PR #9, opened 2017-03-06
Adding `time` to the list of dependencies

yumaikas:
As documented [here](https://www.reddit.com/r/oilshell/comments/5xfyg0/if_you_have_debianubuntu_a_web_server_and_five/), I discovered that the bash builtin `time` on Ubuntu 14.04 does not work like the time binary from apt. This adds `time` as a dependency.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,7 +13,8 @@ set -o errexit
 install-deps() {
   # python-dev: for pylibc
   # gawk: used by spec-runner.sh for the special match() function.
-  sudo apt-get install python3-dev gawk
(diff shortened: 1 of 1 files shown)


andychu:
Great, thanks for fixing this.  I edited your comment a little bit -- the reason we're using it is actually for the --output and --format flags.

BTW I wrote about time in my comment on this blog post:

https://lobste.rs/s/hwgt2h/usr_bin_time_not_command_you_think_you_know

yumaikas:
I saw that shortly after I had it all figured out.

PR #11, opened 2017-03-07
Update spec.sh

D630:
- Make sure IFS is not set to / etc.
- Use mksh for version info itself
- Use grep -E instead egrep
- Use dirname together with readlink and BASH_SOURCE

### DIFF
--- a/spec.sh
+++ b/spec.sh
@@ -7,15 +7,17 @@ set -o nounset
 set -o pipefail
 set -o errexit
 
+IFS=$' \t\n'
+
 readonly DASH=/bin/dash
 readonly BASH=/bin/bash
 readonly MKSH=/bin/mksh
-readonly ZSH=/usr/bin/zsh  # Ubuntu puts it here
+readonly ZSH=/usr/bin/zsh  # Debian/Ubuntu puts it here
 readonly BUSYBOX_ASH=_tmp/shells/ash 
 
 readonly OSH=bin/osh
 
-# ash and dash are similar, so not including ash by default.  zsh is not quite
+# ash and dash are similar, so not including ash by default. zsh is not quite
 # POSIX.
 readonly REF_SHELLS=($DASH $BASH $MKSH)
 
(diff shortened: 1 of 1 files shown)


andychu:
What's the motivation for these changes?  Except for the mksh change, which seems to give more detailed version information, I don't really agree with them.

They might be slight improvements, but they could applied to every script in the repo, and I'd rather concentrate effort on making Oil an example of how to remove these issues with shell.

OSH is basically a vehicle for Oil and no more.

- Oil won't do word splitting with IFS unless explicitly asked for
- Oil will have a simple function to get the location of the script being run.  And it will have a proper module system.

In other words, we could go add "$var" to every single var expansion, and add "--" to every single command (except the ones that don't support it like "echo").  But I feel like this is an exercise in futility, and we should just write a better shell.  I should probably blog about that if the motivation wasn't clear.

Another way of saying it is that the shell scripts basically treat the file system as "trusted" rather than untrusted. I don't think there is any other option in shell.  But in Oil, you should be able to write scripts that don't trust their external input.

I also have this page:

https://github.com/oilshell/oil/wiki/Shell-Security-Problems

If there are any other "pedantic" practices that are good, please list them there so we can subsume them in Oil.

D630:
Hey, I had no real motivation. Before I send you the spec results, I had a quick look at the script and modified it in the habit of a bash shell scripter, who is condemned to quoting. So, just close the pr, if you want to :)

andychu:
Yeah this helped me formulate a design goal.  Right now in shell you have the tradeoff between what I call Naive Style and Pedantic Style.  Note that both of those are slightly negative adjectives.

So most of my scripts are written in Naive Style.  But for Oil we want to get rid of the tradeoff.  I like the "whipupitude" of shell (in Perl terms), but the problem is that you end up with all these tiny bugs all over place.  Death by a thousand cuts.

And honestly I would like to categorize them... there is double quoting and --.  Double quoting does several things though:

- it prevents word splitting
- it prevents globs from being evaluated from inside variables and command output
- it prevents brace expansion

etc.

I had a good link on shell security practices, but I can't find it anymore...

I will probably write a blog post about this to clarify the goals.

PR #13, opened 2017-03-26
Use $(uname -m) for non-x86_64 support in build.sh

FauxFaux:
`build.sh` has a hardcoded `x86_64`. Replacing it with a call to `$(uname -m)` means that at least some of the tests pass on `armv6l`.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,7 +24,7 @@ pylibc() {
 
   # Wildcard to match any Python 3 version.
   shopt -s failglob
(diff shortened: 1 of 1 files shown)


andychu:
Cool thanks.

Linux and Python are very portable across hardware, so I suspect that pretty much all spec test failures are related to shell versions.  There was some talk about this on oil-dev, but right now there's no good solution for getting the right version of every shell unfortunately.

PR #14, opened 2017-07-08
Some portability fixes

myfreeweb:
- include correct regex header on FreeBSD
- use `#!/usr/bin/env` instead of hardcoded Linux paths
- find shells with `which`

`test/spec.sh smoke` works on FreeBSD with this patch, but the `spec-runner.sh` is way too full of linuxisms and GNUisms for me to fix without losing functionality (e.g. `time --output … --format …` — BSD time has `-o` for output file, but no custom format option whatsoever. BSD xargs has no `--verbose`. `nproc` is linux only. etc. etc. etc.)

### DIFF
--- a/test/spec.sh
+++ b/test/spec.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Usage:
 #   ./spec-file.sh <function name>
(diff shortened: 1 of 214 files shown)


test/wild2.sh
```diff
@@ -18,7 +18,7 @@ readonly OSH=~/git/oil/bin/osh
 replace-shebang() {
   local dir=$1
   find $dir -name '*.sh' \
-    | xargs -- sed -i 's|^#!/bin/bash|#!/home/andy/git/oil/bin/osh|'
+    | xargs -- sed -i 's|^#!/usr/bin/env bash|#!/home/andy/git/oil/bin/osh|'
```
andychu: This one should remain /bin/bash because it's actually for changing OTHER repos.  Like the toybox repo -- if I want to test out OSH then I change all the shebang lines to OSH.

test/spec.sh
```diff
@@ -7,18 +7,27 @@ set -o nounset
 set -o pipefail
 set -o errexit
 
-readonly DASH=/bin/dash
-readonly BASH=/bin/bash
-readonly MKSH=/bin/mksh
-readonly ZSH=/usr/bin/zsh  # Ubuntu puts it here
+if hash dash 2>/dev/null; then
```
andychu: Nitpick: it seems to use both `hash` and `which`.

I think something like:

    readonly DASH=$(which dash || echo /bin/sh)

is cleaner.  Apparently you can also use `hash -t` but I never use that.

myfreeweb:
Fixed these.

I can't do the make thing:

```
test -d _build/oil && \
        build/actions.sh app-deps oil ~/git/oil bin.oil
build/actions.sh: line 46: _build/cpython-full/python: No such file or directory
test -d _build/hello && \
        build/actions.sh app-deps hello build/testdata hello
build/actions.sh: line 46: _build/cpython-full/python: No such file or directory
gmake: *** No rule to make target '_build/oil/app-deps-c.txt', needed by '_build/oil/all-deps-c.txt'.  Stop.
```

Is there a step I missed or something?

andychu:
I'm in the middle of working on the release tarball.  I expect that to work on FreeBSD eventually, but I'm sure it doesn't right now.  When I have an initial cut, I will ping this thread again and it would be great if you can try it.

Building from the repo is a different story and is a bit Linux-specific like the test harnesses.  It's a little complicated because I build a CPython interpreter in `_build/cpython-full` and then dynamically discover dependencies so you can freeze them in the tarball.  But this only has to be done on the developer machine, not the end user machine.

BTW I fixed a couple problems with the last commit:

https://github.com/oilshell/oil/commit/655c530ca39acdaa6465715caf2be247c5131b05

https://github.com/oilshell/oil/commit/9f0d8e56f44419e38a2923ccef624f0218a787e0

Thanks for trying it, I'll let you know when I have a tarball.

PR #15, opened 2017-07-18
exec without fork

orent:
Trying to figure out fork_external, I realized there are many more opportunities where forking can be avoided in favor or direct exec. A bit of stracing shows bash does the same.

Also, fixed unit tests.

### DIFF
--- a/test/count-procs.sh
+++ b/test/count-procs.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+#
+# Usage:
+#   ./count-procs.sh <function name>
+
+set -o nounset
+set -o pipefail
+set -o errexit
+
+count-procs() {
+  local sh=$1
+  local code=$2
+
+  # Hm I didn't know -e is like a mini-language.  '-e open' is thes same as '-e
+  # trace=open'.  signal=none turns off the SIGCHLD lines.
+  #
+  # NOTE we grep for [pid and rely on the code itself to echo [pid-of-sh $$].
+
+  code='echo "[pid-of-sh $$]";'" $code"
+  strace -e 'trace=fork,execve' -e 'signal=none' -e 'verbose=none' -ff -- \
+    $sh -c "$code" 2>&1 | fgrep '[pid' || true
+}
+
+test-many() {
+  for code in "$@"; do
+    echo
+    echo
+    echo "--- $code ---"
+    echo
+
+    for sh in dash bash mksh zsh; do
+      echo
+      echo
+      echo "--- $sh ---"
+      echo
+      count-procs $sh "$code"
+    done
+  done
+}
+
+t1() {
+  test-many \
+    'echo hi' \
+    '/bin/echo one; /bin/echo two' \
+    '{ /bin/echo one; /bin/echo two; }' \
+    '{ echo one; echo two; } | wc -l' \
+    '( echo one; echo two ) | wc -l' 
+}
+
+"$@"


andychu:
Thanks for looking into this (and for fixing the tests).

I'm doing some stracing now to see what bash does.  With bash 4.3 on my machine, it doesn't seem to try to save the last process?  If I'm reading the patch correctly, it's observing that something like:

    echo $$; ls ; ls /; 

Can be executed with ONE external process instead of two.  Below I see that bash is PID 27289, the first ls is PID 27290, and then the second one is 27291.  Is this patch making to that the second ls can just reuse 27289?

I also straced a brace group `{ echo $$; ls ; ls /; }` and got the same results.

Not saying we shouldn't do it if bash doesn't do it, but I want to make sure I'm understanding it.

I think there is a related issue right now, the difference between:

    { echo one; echo two; } | wc -l
    ( echo one; echo two ) | wc -l

These should be executed identically, because pipelines force a subshell.  But I just tested it and it appears OSH has an extra fork.  I just checked bash and it seems like it DOES omit the fork that OSH doesn't.

I think what we really need is to have a mechanism to test this (even if it's platform-specific).  `strace -e -ff` and counting PIDs is close.  If you have any suggestions let me know.

```
$ strace -e fork,execve -ff -- bash -c '  echo $$; ls ; ls /;  '
execve("/bin/bash", ["bash", "-c", "  echo $$; ls ; ls /;  "], [/* 72 vars */]) = 0
27289
strace: Process 27290 attached
[pid 27290] execve("/bin/ls", ["ls"], [/* 72 vars */]) = 0
a2.txt        a.txt       bin.oil  config-status-snippet.sh  c.sh        foo.sh  hi.txt       libc.so   Makefile   oil-version.txt  out.txt        README.md  slots.py  test              TODO.txt  yapf-style.cfg
a3.txt        benchmarks  build    configure                 deb-ifs.sh  f.t     __init__.py  LICENSE   native     opy              patches        _release   spec      tests             tools     zsh-osh.txt
ansi2html.sh  bin         _build   core                      deps.sh     f.txt   install      lint2.sh  NOTES.txt  osh              _pybuild       rev.py     s.txt     _tmp              t.sh      zsh-sh.txt
asdl          _bin        _chroot  count.txt                 doc         help    INSTALL      local.sh  oil        osh.log          Python-2.7.13  scripts    tags      TODO-cpython.txt  web
[pid 27290] +++ exited with 0 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=27290, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
strace: Process 27291 attached
[pid 27291] execve("/bin/ls", ["ls", "/"], [/* 72 vars */]) = 0
bin  boot  cdrom  dev  etc  home  initrd.img  initrd.img.old  lib  lib32  lib64  lost+found  media  mnt  opt  proc  root  run  sbin  snap  srv  sys  tmp  usr  var  vmlinuz  vmlinuz.old
[pid 27291] +++ exited with 0 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=27291, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
+++ exited with 0 +++
```

orent:
Check my comment for fork_external. Whan it really means (with negative sense) is that 'sys.exit(status)' and 'return status' would be equivalent at this point. Execing is basically delegatingt that exit() to another program.

The rest is just passing this flag as far as you can to the right, down to the last command in the chain. I don't care if it is exactly what bash does or not, as long as it is simple enough to be "obviously" correct. I intentionally ignored the pipe case for now to keep things simple.

I checked basic stuff like ensuring that a wrapper script that just verifies and modifies some some arguments to a program will not keep a copy of the shell around.

andychu:
But as mentioned, they aren't equivalent because of `_CheckStatus`, which happens in the `errexit` case.  I didn't blog about it yet, but one of the "initial features" of OSH is to give good diagnostics when using `errexit/pipefail/nouonset` ("strict mode").  If you replace the process image then you don't have a chance to print a diagnostic.

This issue is worth thinking about though, and I started a script to test it (on a branch):

https://github.com/oilshell/oil/commit/6cd4e0351ccbb1188f3ef074f9799cc584a45cb2

So thanks for bringing it up.

I won't merge the code changes as is, but I appreciate the test fix.  (Not sure if I can merge it separately).  Also I'm confused that you said "a bit of stracing shows bash does the same" then "I don't care if it is exactly what bash does or not".  As you can probably tell from the blog, I care about these differences :)

orent:
What I meant is that I did not try to reproduce the exact behavior in all cases. I checked simple cases like a wrapper script.

I don't see an issue with errexit. The fork_external is carried along and only triggered in the very last command at the end of a chain that is absolutely going to exit, regardless of its status or errexit - it makes no difference.  Same for pipefail: the parent shell of the pipe is the one that waits on the exit codes of all stages. It does not need any help from the specific subshell that forked the final command of the pipe so that subshell is free to exec instead. 

If this is not obvious enough then perhaps my code needs further work until it is.

orent:
In shell, non zero exit codes are sometimes equivalent to exceptions in other languages - but not always. With set -e enabled, it is tempting to treat unchecked nonzero exit codes the way unhandled exceptions are treated in other languages by printing a diagnostic message before exiting, but sometimes it will not quite be the right thing. 

I sometimes write shell functions like this:

    dosomething() (
         set -e
         step
         for x in y; do
           more steps
        ...
     )

And then call it like this:
   dosomething || somethingelse

In this case, I definitely would not appreciate an error message being printed when the inner function exits. This is not an unhandled exception.

The last command in a scope should never be considered an errexit - it is simply the result of the scope. Only if the enclosing scope also has errexit enabled and fails to check it should it be reported. The tricky part is that you may want to report the command with nonzero status command as well as the code locations that should have checked it. This can become an actual exception with a traceback. If errexit is not enabled, this exception is caught is turned into a regular exit code. Access to nonexistent variables with nounset enabled should always be considered an exception, whether last in scope or not. 

The connection between errexit reporting and the exec/fork optimization is the condition of "last in scope". It's not exactly the same, because it applies to any command, including builtins, while exec/form is only relevant to external commands.

andychu:
As in `bash` and other shells, errexit is disabled in the `||` context, and a handful of other places mentioned here:

https://github.com/oilshell/oil/blob/master/core/state.py#L31

So there won't be any spurious diagnostics there.  Also search for `_PushErrExit` in the code.

I would say I don't see the motivation for this patch right now, because it interacts/complicates the two things I mentioned, and it doesn't have tests:

- errexit diagnostics
- Pipeline / Subshell combo (although honestly probably no real user will notice this either; it's invisible)

It also has some interaction with `shopt -s lastpipe`, which I have spec tests for but haven't implemented.  And that is a feature that someone might conceivably need (`zsh` does it by default, no other shell does.)

I'd be more interested in patches along these lines:

http://www.oilshell.org/blog/2017/06/22.html

Basically I just try random shell scripts that have known output and see what goes wrong.  And then I add spec tests for those features, and then fix the code.

Patches that are focused and have tests should fly through... ones with no tests and where I don't see an actual user need ("would be nice") are less interesting.

andychu:
Also keep in mind that I'm primariliy interested in OSH as far as it relates to the Oil language...  it's a vehicle.  In other words, there's the possibility of going too far in cloning bash / making a better bash-like shell.  I don't think we're close to that point but some minor changes have to be kept in the larger context.

For example `shopt -s lastpipe` is sort of detail, but it is something that trips people up and I haven't decided whether it should be the default in Oil or not.  That is where you do something like

echo foo | read x

This doesn't work in bash but works in zsh, because `read` is run in a subshell and the variable is lost.

orent:
Yes, I am aware of how errexit works and when it does not trigger an immediate exit. Read my message again: the function dosomething() does not check each step. This is not an oversight or an error. It is intentional. If any step fails it expects the posix-compliant shell running it to exit with a nonzero exit code. Since the function is in parentheses this will simply return to the caller with a nonzero exit status. The caller checks the result and takes appropriate action. 

Your shell tries to be "too helpful" - it will assume that this is an unhandled error condition and print a message that may be confusing for the user because the script actually handled the condition correctly. 

I think that one lesson here for the oil language is to think carefully about when nonzero exit codes are equivalent to unhandled exceptions - and when they are not.

PR #28, opened 2017-08-08
Build fixes

lheckemann:
Removes some assumptions about the environment.

### DIFF
--- a/build/actions.sh
+++ b/build/actions.sh
@@ -60,10 +60,10 @@ app-deps() {
   local prefix=_build/$app_name/app-deps
 
   # I need the right relative path for Oil
-  ln -s -f $PWD/build/app_deps.py ~/git/oil/_tmp
+  ln -s -f $PWD/build/app_deps.py _tmp
 
   PYTHONPATH=$pythonpath \
-    $PREPARE_DIR/python -S ~/git/oil/_tmp/app_deps.py $main_module $prefix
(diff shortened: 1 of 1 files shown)


build/actions.sh
```diff
@@ -12,6 +12,7 @@ set -o errexit
 source build/common.sh
 
 write-release-date() {
+  mkdir -p _build
```
andychu: The Makefile is supposed to do this on line 71:

$(shell mkdir -p _bin _release _tmp _build/hello _build/oil)

But I think I am going to use the more consistent style having shell functions write to stdout, and then the Makefile should redirect to `> $@` (i.e. the output file).  Right now _build/release-date.txt is unnecessarily repeated in the shell script and Makefile.

lheckemann: Right, but given the instructions on the Contributing page `make` won't have run the first time `write-release-date` does.

andychu: Yes good point, I added a comment about this:

https://github.com/oilshell/oil/commit/74c0356e162337afa73a80f7fa4735de31d7a126

I'm trying to be a little principled about how I use the `Makefile` and `build/{compile,actions}.sh`. That is, try to use them in a purely functional way, but Make doesn't really work like that so it could be a false hope.  Ideally the build system would make directories for you and you wouldn't have to worry about it.

PR #30, opened 2017-08-09
Make more shebangs compatible

lheckemann:
Several of the tests continue to refer to /usr/bin/python. I don't
know of a way to fix these, as they call python -S and IIRC shebangs
can only include one parameter…

### DIFF
--- a/build/doc.sh
+++ b/build/doc.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Usage:
 #   ./run.sh <function name>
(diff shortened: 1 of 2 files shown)


andychu:
Yeah I ran into the issue of -S.  Well I guess one way is to invoke them from shell with $PYTHON or something like that, rather than using the shebang.  Not sure I really like it but it depends how many places we have to change.

What is failing for you now?

lheckemann:
Nothing's failing, it all seems to be working now :) Thanks again for the help.

andychu:
OK well I think `test/unit.sh all` might fail due to shebangs for you, if you don't have /usr/bin/python, but yeah let's defer that until it's blocking something...

Issue #34, opened 2017-09-11
Problem using OSH as login shell


icarroll:
On an AWS EC2 instance running RedHat 7.3 I set up a user and used chsh to change its login shell to OSH. When I try to log in as that user I get the following:

    Usage: oil MAIN_NAME [ARG]...
           MAIN_NAME [ARG]...
    
    oil behaves like busybox.  If it's invoked through a symlink, e.g. 'osh', then
    it behaves like that binary.  Otherwise the binary name can be passed as the
    first argument, e.g.:
    
        oil osh -c 'echo hi'
    
    
    Invalid main '-osh'

Then the connection is closed.

OSH is installed as /usr/local/bin/oil with a symlink from /usr/local/bin/osh . There is an entry in /etc/shells for /usr/local/bin/osh .

I get the same error if I try `sudo -i -u $USER` or `su - $USER` .

icarroll:
With this patch I am able to use OSH as a login shell, though some expected behavior such as setting shell variables and sourcing profile still needs to be implemented.

PR #35, opened 2017-09-12
support running as -osh for use as login shell

icarroll:
This is a fix for issue #34

### DIFF
--- a/bin/oil.py
+++ b/bin/oil.py
@@ -164,7 +164,7 @@ def _ShowVersion():
   print('Platform: %s' % platform.version())
 
 
-def OshMain(argv):
+def OshMain(argv, login_shell):
   spec = args.FlagsAndOptions()
   spec.ShortFlag('-c', args.Str, quit_parsing_flags=True)  # command string
   spec.ShortFlag('-i')  # interactive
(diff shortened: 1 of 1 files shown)


andychu:
OK great, thanks for fixing this!

I saw your issue but I wasn't quite sure what is different about a login shell.  But I think this is a good start.

andychu:
And BTW I mostly expected people to use it as an interpreter for shell scripts, and to start it from your login shell (e.g. bash).  For example, it will sometimes crash on invalid input right now.  

Please report those bugs if you encounter them.  Although they may be a bit harder to catch if it's a login shell?  If the login shell crashes do you lose the ssh connection?

The completion support is still nascent too.

icarroll:
I ran into the login shell issue when I was trying to test OSH on a complex shell script that configures and runs a web application. The simplest way to test was to change the user's shell to OSH and try to start the application, so it's sort of a yak shaving thing. Now that I have a fix, I will try again and report any problems that I find.

I'll also try the ;s error to see what happens when a login shell crashes.

andychu:
Thanks for testing this out.  Let me know if there are some debugging features that might help.

I've been thinking of those as "carrots" for trying OSH.

1. a better `set -x`, in particular it should show the exit codes, and `set -e` failures.  (This one is always annoying).  Also `set -x` should be properly concurrent and show PIDs perhaps.

2. Being able to dump the state of the shell on a crash.  It's quite easy to dump all variables on the stack, but bash doesn't support this.  You can also dump all the `set/shopt` settings, as well as the file descriptor state, and a few other things.

icarroll:
The main thing that would be useful to me so far would be saying what file and line a specific error was encountered at. On a crash I get plenty of (too much) information about where in the python code the error happened, but almost none about the shell code that triggered it.

andychu:
OK I opened issue #38 about that.

There is infrastructure for providing location information, but sometimes I just raise NotImplementedError and it's not caught.

PR #40, opened 2017-09-22
Tests: portability improvements and additional test

lheckemann:
Hope you don't mind me lumping this all up in one PR, if that's a problem I can split it.

This improves portability to the point that the tests will run on NixOS; there's still a multitude of failures caused by the assumption that ls lives in /bin, but it's a start :)

### DIFF
--- a/spec/show-fd-table.sh
+++ b/spec/show-fd-table.sh
@@ -13,7 +13,7 @@ show-fd-table() {
   echo "   function in pid $$"
   # Make it truly parallel
   sleep 0.5
-  /usr/bin/time --output /tmp/$$.txt -- spec/bin/show_fd_table.py "$@"
+  /usr/bin/env time --output /tmp/$$.txt -- spec/bin/show_fd_table.py "$@"
 }
 
 # Trying to recreate spec-runner problem with file descriptors.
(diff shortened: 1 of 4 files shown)


spec/var-sub.test.sh
```diff
@@ -19,6 +19,11 @@ echo ${foo:-$({ which ls; })}
 # BUG bash stdout-json: ""
 # BUG bash status: 2
 
+### :- with empty alternative
+foo=
+argv.py "${foo:-}"
+# stdout: ['']
+
```
andychu: OK wow I'm surprised this is broken!!  Can you put this test at the very top of spec/var-sub-quote.test.sh?

Then I will pull it.  Thanks.

lheckemann: Done!

andychu: OK can you push it or do I have to pull?  I haven't used this "approve" feature by Github before.  I would assume it should non-committers push one change?

lheckemann: Don't think so, it just marks approval. I think you can set it up to e.g. automatically merge once it has n approvals, but it doesn't allow the PR author to merge it themselves.

andychu: OK but I think you need to push the update you made?  I still see 3 commits and the change in `var-sub` not `var-sub-quote`.  I would like it to allow pushing so it doesn't require another round trip :)

lheckemann: Whoops, sorry, misread — I didn't realise you were referring to the other script. Fixed!

andychu:
OK thanks, I will look at fixing this tonight... pretty surprised that doesn't work!

PR #45, opened 2017-10-09
Implement test -e and test -h

icarroll:
I think I got them right, but it'd be good to double check.

This is to support running wso2server.sh to start wso2is. These are the only changes necessary to run the script.

https://gist.github.com/icarroll/e798a299c86843ad6d03ef684bab1d24


andychu:
OK thanks for trying it.  It would be better if there are unit tests like this, particularly if you aren't sure:

https://github.com/oilshell/oil/commit/89a89be9fd1ed5109e1631b327223e206cfda228

Also I see some noise on the pull request, i.e. commit 5906b6f which was many days ago.  I think you need to bring your fork up to date with the main repo?  Then it won't show those requests.

However since these are 1-line changes I'll just squash and merge and see what happens.

andychu:
OK it looks like Github did a pretty good job merging it.  I think it will be easier if you bring your fork up to date now, since there is probably a merge.

Thanks and let me know what else is not working.  If you have questions about the spec tests let me know too.

PR #52, opened 2017-11-19
Adds error handling to the `pushd` and `popd` builtins.

timetoplatypus:
`pushd` and `popd` were crashing when given malformed input or absent input. I've added some error handling to prevent those crashes. The error messages are the same as given by `pushd` and `popd` in bash.

### DIFF
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -492,6 +492,7 @@ def Pushd(argv, dir_stack):
     return 1
   elif num_args > 1:
     log("pushd: too many arguments")
+    return 1
 
   dest_dir = argv[0]
   if os.path.isdir(dest_dir):
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -484,7 +484,6 @@ def Cd(argv, mem):
 
 
 def Pushd(argv, dir_stack):
-  dir_stack.append(os.getcwd())
   num_args = len(argv)
 
   if num_args <= 0:
@@ -501,6 +500,7 @@ def Pushd(argv, dir_stack):
     util.error("cd: %r: %s", dest_dir, os.strerror(e.errno))
     return 1
 
+  dir_stack.append(os.getcwd())
   return 0
 
 
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -488,17 +488,17 @@ def Pushd(argv, dir_stack):
   num_args = len(argv)
 
   if num_args <= 0:
-    log("pushd: no other directory")
+    util.error('pushd: no other directory')
     return 1
   elif num_args > 1:
-    log("pushd: too many arguments")
+    util.error('pushd: too many arguments')
     return 1
 
   dest_dir = argv[0]
-  if os.path.isdir(dest_dir):
+  try:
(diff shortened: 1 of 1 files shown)


core/builtin.py
```diff
@@ -485,18 +485,37 @@ def Cd(argv, mem):
...
+    return 1
+  elif num_args > 1:
+    log("pushd: too many arguments")
+
   dest_dir = argv[0]
-  os.chdir(dest_dir)  # TODO: error checking
+  if os.path.isdir(dest_dir):
+    os.chdir(dest_dir)
+  else:
+    log("pushd: " + dest_dir + ": No such file or directory")
```
andychu: Please use

    util.error('pushd: %r: message', dest_dir)

The code is not super consistent, but `util.error()` is for fatal errors.

- Using the % style is preferred over +
- Be consistent with the rest of the file and use single quotes everywhere

core/builtin.py
```diff
@@ -485,18 +485,37 @@ def Cd(argv, mem):
...
 
 def Popd(argv, dir_stack):
   try:
     dest_dir = dir_stack.pop()
   except IndexError:
-    log('popd: directory stack is empty')
+    log("popd: directory stack is empty")
+    return 1
+
+  if os.path.isdir(dest_dir):
```
andychu: Please change this test and the one above to try the chdir and catch OSError.

I don't think `isdir()` matches the conditions where `chdir()` fails (symlink to dir, etc.?).  So it's cleaner to only show an error when the actual chdir() call fails.

core/builtin.py
```diff
@@ -485,18 +485,37 @@ def Cd(argv, mem):
 
 def Pushd(argv, dir_stack):
   dir_stack.append(os.getcwd())
+  num_args = len(argv)
+
+  if num_args <= 0:
+    log("pushd: no other directory")
```
andychu: As mentinoed use single quotes and util.error() everywhere

timetoplatypus:
No problem, I can make those changes

core/builtin.py
timetoplatypus: Would you prefer that the error message printed when catching OSError is the same as the error message produced in bash? Or would you prefer that I follow the same error message that you print in the Cd() builtin? (which is: `util.error("cd %r: %s", dest_dir, os.strerror(e.errno))`)

andychu: Yes good question.  The cd style is better, because it prints the actual error from libc.

Actually, bash probably does the same thing -- it doesn't always print "No such file or directory".  It will also print "Permission denied", for example, if that is the `errno` code.

timetoplatypus: Ok, sounds good. I'll stick with the Cd() style

timetoplatypus:
The latest commit (f57fd2e...) should address the inconsistent use of double quotes (now single quotes), and the error handling should now use the `util.error(...)` function with `Cd()` style output where appropriate. Please let me know if there is anything I'm missing.

andychu:
OK great, thanks for doing this!  I merged it.

PR #54, opened 2017-11-23
First draft of `osh` manpage

timetoplatypus:
Adds first draft of an `osh` manpage. It's been added to the doc directory and can be tested by invoking `man ./osh.1`


andychu:
OK thanks for doing this.

Eventually I would like to auto-generate usage strings.  Right now `osh --help` and `help osh-usage` come from `doc/osh-quick-ref-pages.txt`.

But part of that file should be generated from `core/args.py`.  Both the builtins and the `osh` binary itself use core/args.py.  Let me look at that... and let me know if that doesn't make sense.

The easiest way to understand the refs is to run:

    $ build/doc.sh osh-quick-ref

That generates Python code that is included in the binary, text snippets, and HTML.

Issue #42, opened 2017-09-26
spec tests should run in a known environment


andychu:
There have been some issues with assuming /usr/bin/time, /usr/bin/python, /bin/bash, etc.  This seems to be the first issue everybody runs into when developing Oil.

The tests also depend on gawk I think.  And the shells: busybox, dash, bash, ash, etc.

Right now I think the best solution is an Alpine Linux chroot, because it's small and easy to set up.  Should be runnable on any distro quite easily.

FYI @lheckemann 

Also see issue #12

andychu:
Porblem on Ubuntu 17 due to bash/mksh/etc. version mismatches :-(

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=17.10
DISTRIB_CODENAME=artful
DISTRIB_DESCRIPTION="Ubuntu 17.10"
```

```
'_tmp/web' -> '/home/andy/git/oilshell/oil/web'                                          
alias failed with status 1                  
array failed with status 1                  
assoc failed with status 1                  
blog2 failed with status 1                  
builtin-io failed with status 1             
builtins failed with status 1               
builtin-test failed with status 1           
glob failed with status 1                   
sh-options failed with status 1             
var-op-other failed with status 1           
var-op-strip failed with status 1           
var-ref failed with status 1                
var-sub failed with status 1                
xtrace failed with status 1                 

*** 14 tests FAILED 
```

andychu:
@Yorwba Let's continue the discussion here.

I said Alpine might be harder and Ubuntu is probably easier.  The reason I think that is:

1. The tests are already written against Ubuntu
2. Alpine is busybox-based while Ubuntu is coreutils-based.

But maybe this is not as big a deal as I thought.  You know I think the main issue is: what shells do we test against?

1. Binary Ubuntu packages (which we do now essentially)
2. Binary Alpine packages
3. Compile shells from source on Ubuntu or Alpine

I was thinking #3 because it gives us the most control over what shell version.  Although honestly I think "whatever shell version the latest LTS version of Ubuntu uses" is not bad.

I was thinking "the latest version of every shell", but in theory that is more work (though I think it shouldn't be THAT much work.)

[...]

PR #55, opened 2017-12-02
Implement `dirs` builtin

timetoplatypus:
Implements the `dirs` builtin, following the behavior of `dirs` in bash. This includes using the -c, -v, and -p flags. Two flags have not yet been implemented (-l and -N/+N).

I've been hit with some other work so I'll tackle the spec tests later. Wanted to get this merged anyway in case it takes me a while to get back to working on this.


andychu:
Did you try the spec tests at all?  If you're having trouble we can chat about it.  I prefer to have tests with new functionality.  It's honestly hard for me to review otherwise, because I don't know what it's supposed to do.

At the very least, try

    $ test/spec.sh builtins2

If that doesn't work, or you can't get it working within a few minutes, something is wrong :)

timetoplatypus:
No worries, I understand. We can hold off merging until I have the spec tests correct.

Aside from just reading some of those files, I haven't tried to write or execute the spec tests yet. I'll have some free time tomorrow, so I can give that `test/spec.sh builtins2` a shot.

andychu:
OK I started a wiki page here: 

If you hvae problems let me know and we can fill out the wiki page so it's not as hard to get started.

timetoplatypus:
Sounds good, thanks for creating that page.

I think I know what I need to do develop the spec tests. Looking at the spec/builtins.test.sh file helped a lot actually, lots of good examples in there. As someone doing this for the first time, I'll try to make notes of what I thought was useful and the order of things that I'm looking at to develop the test; if I can get that nice and formalized, it may make for a good section in that Wiki.

andychu:
Yeah the overall concept is simple: 

1) Start a test case with `### my description`
2) Write some shell
3) End with some assertions `# stdout: foo`

I think in my e-mail to oil-dev@ I may have overcomplicated things and scared you away :)

It should be easy to copy and paste and run `test/spec.sh builtins2` over again.  Notice that `builtins2` is much smaller and faster to run.  I would add your tests there and we can organize them later.

timetoplatypus:
Haha a little bit. No problem though, I'm making progress with writing the spec tests.

timetoplatypus:
Alright, the spec tests for `dirs` and the three flags that have been implemented are in the recent commit (b319b60...). Do let me know if there's anything wrong with it that I need to fix. I've deferred the implementation and spec tests for the -N/+N flags since, as discussed in oil-dev@, this will suffice for a first pass.

core/builtin.py
```diff
@@ -482,6 +482,15 @@ def Cd(argv, mem):
   state.SetGlobalString(mem, 'PWD', dest_dir)
   return 0
 
+def PrintDirStack(dir_stack):
+  stack_size = len(dir_stack)
```
andychu: Can this be simplified to print(' '.join(dir_stack))  ?

core/builtin.py
```diff
@@ -496,11 +505,12 @@ def Pushd(argv, dir_stack):
   dest_dir = argv[0]
   try:
     os.chdir(dest_dir)
+    dir_stack.append(os.getcwd())
+    PrintDirStack(dir_stack)
```
andychu: Put the things that can't raise OSError outside the except: block.

I think it's better to append `dest_dir` rather than calling os.getcwd() again.

core/builtin.py
```diff
@@ -513,15 +523,37 @@ def Popd(argv, dir_stack):
...
+
 def Dirs(argv, dir_stack):
-  print(dir_stack)
+  arg, i = DIRS_SPEC.Parse(argv)
+  if arg.l:
+    util.warn('*** dirs -l not implemented ***')
+  # Following `bash` behavior for order of operations
+  if arg.c:
+    while len(dir_stack) > 0:
+      dir_stack.pop()
```
andychu: This loop can be replaced with `del dir_stack[:]`

In Python 3 there is `dir_stack.clear()`, but that's the idiomatic way to do it in Python 2.

core/builtin.py
```diff
@@ -513,15 +523,37 @@ def Popd(argv, dir_stack):
...
+  elif arg.v:
+    for i in range(0, len(dir_stack)):
```
andychu: `for i, entry in enumerate(dir_stack)` is more idiomatic.

core/builtin.py
```diff
@@ -543,11 +542,10 @@ def Dirs(argv, dir_stack):
...
   # Following `bash` behavior for order of operations
   if arg.c:
-    while len(dir_stack) > 0:
-      dir_stack.pop()
+    del dir_stack[:]
   elif arg.v:
-    for i in range(0, len(dir_stack)):
-      print(' ' + str(i) + ' ' + dir_stack[i])
+    for i, entry in enumerate(dir_stack):
+      print(' ' + str(i) + ' ' + entry)
```
andychu: Also this can be `print(' %2d %s' % (i, entry))`

That makes the numbers right justified to 2 chars, which more closely matches what bash does.

andychu:
Is this ready to look at again?  I just got some notifications that you pushed, but it wasn't clear if you wanted me to look.

I also sent a comment earlier that I don't think was addressed.  See above:

> Can this be simplified to print(' '.join(dir_stack)) ?

timetoplatypus:
Apparently I hit preview or something instead of post so it looked like (on my end) that I responded to that question when I really hadn't. Whoops!

core/builtin.py
timetoplatypus: In retrospect, I think there may be some ternary operator magic I can use to consolidate this section of code. Will test and get back to you

andychu: Well you can just use an if statement?

BTW `is not`  compares for identity, not equality.  It happens to work on small integers, but shouldn't be used on integers in general.

for example:

```
>>> 5 is (3+2)
True
>>> 50000 is (30000+20000)
False
>>> 50000 == (30000+20000)
True
```

But I think this code should be changed anyway.

core/builtin.py
```diff
@@ -513,15 +522,36 @@ def Popd(argv, dir_stack):
...
-  print(dir_stack)
+  arg, i = DIRS_SPEC.Parse(argv)
+  if arg.l:
+    util.warn('*** dirs -l not implemented ***')
+  # Following `bash` behavior for order of operations
+  if arg.c:
+    del dir_stack[:]
+  elif arg.v:
+    for i, entry in enumerate(dir_stack):
+      print('%2d %s' % (i, entry)) #print(' ' + str(i) + ' ' + entry)
```
andychu: remove the comment here

timetoplatypus: Sorry, should have caught that before pushing. I'll remove it

core/builtin.py
timetoplatypus: I was thinking the same thing, I threw in an `if` statement to just print the directory stack only if it's size is larger than 0. Should be in the latest commit now.

timetoplatypus:
This GitHub "Review" thing is a bit confusing...

Anyway, the comments should be addressed now. Printing the directory stack looks much cleaner now.

andychu:
OK great, I just merged and will pull it and check it out.  Thanks.

andychu:
Hm I just ran the tests and they don't pass again shells other than osh?

Did you run it like this/

    $ test/spec.sh builtins2

I looked at the first case, case #2 for bash, and it fails because the dir stack is NOT empty the first time?  It contains the current directory.

What version of bash are you running?

Mine is:

```
$ bash --version
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)
```

Tests look like this, after I disabled the dash and mksh tests:

```
builtins2.test.sh
  0       3     ok      pass    pass    pass    pass    command -v
  1      16     BUG     BUG     ok      pass    pass    command -v with multiple names
  2      28     N-I     FAIL    N-I     FAIL    pass    dirs builtin
  3      35     N-I     FAIL    N-I     FAIL    pass    dirs -c
  4      45     N-I     FAIL    N-I     FAIL    pass    dirs -v
  5      55     N-I     FAIL    N-I     FAIL    pass    dirs -p
10 passed, 2 ok, 8 known unimplemented, 2 known bugs, 8 failed, 0 skipped
```

andychu:
I just pushed a few tweaks so make sure to pull/merge:

https://github.com/oilshell/oil/commits/master

timetoplatypus:
```bash
bash --version
GNU bash, version 4.4.12(1)-release (x86_64-unknown-linux-gnu)
```
Looks like I'm on a newer version of bash.

Either way, in bash, the directory stack is never empty. It always contains the current directory, even though it's not a poppable element.
```bash
$> cd ~
$> dirs
~
$> cd ..
$> dirs
/home
$> dirs -c
$> dirs
/home
$> popd
bash: popd: directory stack empty
```
OSH doesn't have this behavior. When the stack is empty, it's empty, and printing the directory stack will return nothing. That's why it fails the tests for other shells.

andychu:
OK, so then just adding the current directory in `PrintDirStack()` will fix it?

Do you want to try that?  Generally the behavior of OSH should match at least one shell, unless there's a good reason not to.  I don't see a good reason here.

The tests should all be green or yellow -- no red (unless `--allowed-failures` is adjusted).

timetoplatypus:
Yes, adding the current directory in `PrintDirStack()` will fix this. I can make that change.

Can you tell me more about how to use the `--allowed-failures` option? I'm not sure where to use it or if I'm supposed to write the tests in such a way that they accomodate using that flag.

timetoplatypus:
Just a heads up, it may be a little bit before I can implement adding the current directory in `PrintDirStack()` because the `dirs -l` flag is a prerequisite. Bash defaults to tilde-prefixed paths (relative to the home directory) and in order for my `dirs` implementation to pass the bash-compliance tests, it needs to also print directories in such a manner.

andychu:
Yeah there is a bit of a problem because the working directory that the tests run in could be `/home/andy/git/oilshell/oil` on my machine, and a different one on your machine.  We want the tests to run in the same environment, which is issue #42, but I don't have a concrete plan for that.

Let me think about how we can write the tests for `dirs` without hard-coding someone's home directory.

andychu:
I fixed the spec tests so that they pass on bash and zsh, but fail on OSH.  So now OSH can match that behavior.  Let me know if you have problems doing that.

It shouldn't require `-l` to be implemented.  I added `cd /` at the top of every test, so there are no entries in the dir stack under `$HOME`.

I wrote some more documentation about the tests.  Let me know if you have questions:

Two relevant commits:

https://github.com/oilshell/oil/commit/523ed53403f5bca70b64c64f4f00a202ea884c3e

https://github.com/oilshell/oil/commit/993d4a44f6aeaaed8fe230a111d1d3858c82c41d

It's bad to check in failing spec tests because OSH can't be released until all the failures are "known" / allowed.  If there are red boxes not accounted for in `--allowed-failures`, then that's something to fix.

The point of this is so that we can allow failing tests, but not allow regressions.  If you make a change, and the number of failing tests goes from 3 to 4, that's a regression.

So OSH never regresses beyond what is documented in `test/spec.sh`.

andychu:
To be clear, what's left is:

1. Make the 4 tests pass
2. Change the `--allowed-failures` for `builtins2` back to 0.  Right now it's 4.

I added this FATAL error message to be clearer:

```
$ test/spec.sh builtins2
builtins2.test.sh
  0       3     ok      pass    pass    pass    pass    command -v
  1      16     BUG     BUG     ok      pass    pass    command -v with multiple names
  2      28     N-I     pass    N-I     pass    FAIL    dirs builtin
  3      36     N-I     pass    N-I     pass    FAIL    dirs -c to clear the stack
  4      47     N-I     pass    N-I     ok      FAIL    dirs -v to print numbered stack, one entry per line
  5      60     N-I     pass    N-I     pass    FAIL    dirs -p to print one entry per line
13 passed, 3 ok, 8 known unimplemented, 2 known bugs, 4 failed, 0 skipped

FATAL: 4 tests failed (4 osh failures)
```

timetoplatypus:
I think something is wrong with the spec tests where you've added the `>/dev/null`. Try the following and you'll see what I mean.

Make a file (I called it test.sh)
```
cd /
pushd /tmp >/dev/null
dirs
dirs -c
dirs
```
and run it using `bash test.sh` and `osh test.sh`. It produces different output on my end:
```
$> osh test.sh
/tmp /
/tmp /
/tmp
$> bash test.sh
/tmp /
/tmp
```
Is this only happening for me, or can you reproduce this behavior too?

andychu:
Right, I only fixed the tests so they pass against bash.  I didn't fix OSH.  

bash and OSH do not match now.


You wrote the tests so they pass against OSH but fail against bash.  Do you understand why that is backward?

Here is the output I get:

```
$ bash test.sh
/tmp /
/tmp

$ bin/osh test.sh
/tmp
/tmp
```

Make sure to pull so you have the latest version of osh.  Although I don't think I changed any behavior.  I only changed the tests.

andychu:
In other words, OSH is the thing that has to change, not the tests.

timetoplatypus:
> You wrote the tests so they pass against OSH but fail against bash. Do you understand why that is backward?
I understand why that's backwards, the tests should certainly be written such that other shells pass and OSH has to come up to standard to meet them.

However, on a clean clone and clean build of OSH, I still get
```
$> osh test.sh
/tmp /
/tmp /
/tmp 
$> bash test.sh 
/tmp /
/tmp
```
Something must be wrong with my environment, I'll keep investigating.

andychu:
You probably have an old version in your $PATH?  Try `bin/osh` from the root of the repo rather than `osh`?  And make sure to pull and check `git status`.  (Or clone the repo again.)

andychu:
Also note that I slightly updated the instructions for running from the repo here:

It's `build/dev.sh minimal` instead of `build/dev.sh all`, because as of a few days ago, the latter requires the `re2c` tool to generate C code for the lexer.  If you just do the minimal build it will fall back to the Python version.

timetoplatypus:
I was fetching the latest stuff from oilshell/oil.git and I got some errors when building. To be sure, I did a clean clone and got the same error again.
```bash
$> git clone https://github.com/oilshell/oil.git
...
$> cd oil/
$> build/dev.sh all
Wrote _build/doc/osh-quick-ref.html
'libc.so' -> '_devbuild/py-ext/x86_64/libc.so'
libc.so: broken symbolic link to _devbuild/py-ext/x86_64/libc.so
-- Generating AST, IDs, and lexer in _devbuild/gen
osh/ast_gen.py: line 4: $'\nast_gen.py\n': command not found
import: unable to load module '/usr/lib/ImageMagick-7.0.7/modules-Q16HDRI/coders/ps.la': file not found @ error/module.c/OpenModule/1275.
import: no encode delegate for this image format `PS' @ error/constitute.c/WriteImage/1167.
osh/ast_gen.py: line 8: from: command not found
osh/ast_gen.py: line 9: from: command not found
osh/ast_gen.py: line 10: from: command not found
osh/ast_gen.py: line 12: lex_mode_e: command not found
osh/ast_gen.py: line 15: syntax error near unexpected token `('
osh/ast_gen.py: line 15: `def main(argv):'
```
If I try building with the minimal option, I get a different error.
```
$> build/dev.sh minimal
Wrote _build/doc/osh-quick-ref.html
'libc.so' -> '_devbuild/py-ext/x86_64/libc.so'
libc.so: broken symbolic link to _devbuild/py-ext/x86_64/libc.so
```
Not really sure if it's something wrong on my end or if it's because of a change from recent commits.

andychu:
Here is what I get running it.

What machine are you running on?

```
~/git/oilshell/tmp/oil$ build/dev.sh minimal
Wrote _build/osh-quick-ref/ and _devbuild/osh_help.py
Wrote _build/doc/osh-quick-ref.html
running build
running build_ext
'libc.so' -> '_devbuild/py-ext/x86_64/libc.so'
libc.so: symbolic link to _devbuild/py-ext/x86_64/libc.so
['__doc__', '__file__', '__name__', '__package__', 'fnmatch', 'glob', 'regex_match', 'regex_parse']
.GLOB
['__init__.py']
.Error compiling regex: Invalid use of repetition operators such as using '*' as the first character.
Error compiling regex: Trailing backslash.
Error compiling regex: Invalid use of repetition operators such as using '*' as the first character.
Invalid regex at runtime
-1
..
----------------------------------------------------------------------
Ran 4 tests in 0.000s

OK
```

Output from `build/dev.sh all`

```
OK
-- Generating AST, IDs, and lexer in _devbuild/gen
build/codegen.sh: line 52: _deps/re2c-1.0.3/re2c: No such file or directory
```

This is expected because re2c needs to be installed.  I'm not sure how the output you got is possible.  It looks like you are somehow executing `osh/ast_gen.py` as a shell script, but it's a Python script.  The shebang line of that script is `/usr/bin/python`.

timetoplatypus:
Wow, so I figured it out and I have no idea how it happened, but my python2 binary somehow got wiped. The file existed (/usr/bin/python), but it was empty. I'm guessing my shell tried to execute the ast/gen.py file as a shell script because of the absence of the #!/usr/bin/python that was being referenced.

Anyway, it's building now...

andychu:
OK great, glad you figured it out.

That is a pretty weird error -- `/usr/bin/python` is owned by root, so most processes cannot overwrite it.  Take care to use `sudo` only when necessary!

timetoplatypus:
In retrospect, there was some other stuff I was tinkering around with while root (it required root) that may have messed with Python, so it's unrelated to OSH for sure.

timetoplatypus:
Hey Andy, I think I'm running into an output redirection bug but I want to get someone else to reproduce it too so I know it's not unique to the computers I've been testing on.

If you build OSH from my fork and change the `dirs -c` spec test to the following, you'll (hopefully) reproduce the weird behavior I'm seeing on my end:
```bash
cd /
pushd /tmp >/dev/null
echo "" >/dev/null
dirs
dirs -c
dirs
```

The added `echo` line shouldn't functionally do anything but adding it makes my `dirs` implementation go from failing the test to passing the test.

timetoplatypus:
Looks like throwing in a sys.stdout.flush() resolves that issue actually.

PR #56, opened 2017-12-16
Implement `dirs` builtin flags

timetoplatypus:
This pull request contains the code for the `dirs` builtin flags: -v, -p, and -c. The `bash` implementation also has the -l flag, but that has not yet been written.

### DIFF
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -540,11 +540,17 @@ def Dirs(argv, dir_stack):
   if arg.c:
     del dir_stack[:]
   elif arg.v:
-    for i, entry in enumerate(dir_stack):
-      print('%2d %s' % (i, entry))
+    temp_stack = dir_stack[:]
+    temp_stack.insert(0, os.getcwd())
+    for i, entry in enumerate(temp_stack):
+      print('%2d  %s' % (i, entry))
+    sys.stdout.flush()
   elif arg.p:
-    for entry in dir_stack:
+    temp_stack = dir_stack[:]
+    temp_stack.insert(0, os.getcwd())
+    for entry in temp_stack:
       print(entry)
+    sys.stdout.flush()
   else:
     PrintDirStack(dir_stack)
 
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -484,9 +484,8 @@ def Cd(argv, mem):
 
 
 def PrintDirStack(dir_stack):
-  if dir_stack:
-    print(' '.join(dir_stack))
-
+  print(os.getcwd() + ((' ' + ' '.join(dir_stack)) if dir_stack else ''))
+  sys.stdout.flush()
 
 def Pushd(argv, dir_stack):
   num_args = len(argv)
@@ -498,13 +497,14 @@ def Pushd(argv, dir_stack):
     return 1
 
   dest_dir = argv[0]
+  current_dir = os.getcwd()
   try:
     os.chdir(dest_dir)
   except OSError as e:
     util.error("pushd: %r: %s", dest_dir, os.strerror(e.errno))
     return 1
 
-  dir_stack.append(os.getcwd())
+  dir_stack.append(current_dir)
   PrintDirStack(dir_stack)
   return 0
 
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -483,9 +483,25 @@ def Cd(argv, mem):
   return 0
 
 
-def PrintDirStack(dir_stack):
-  print(os.getcwd() + ((' ' + ' '.join(dir_stack)) if dir_stack else ''))
-  sys.stdout.flush()
+WITH_PREFIX = 1
+WITHOUT_PREFIX = 2
+SINGLE_LINE = 3
+def PrintDirStack(dir_stack, mode = 3):
+  dirs = [os.getcwd()]
+  dirs.extend(dir_stack)
+  if mode == 1:
+    for i, entry in enumerate(dirs):
+      print('%2d  %s' % (i, entry))
+    sys.stdout.flush()
+  elif mode == 2:
+    for entry in dirs:
+      print(entry)
+    sys.stdout.flush()
+  elif mode == 3:
+    if dirs:
+      print(' '.join(dirs))
+      sys.stdout.flush()
+
 
 def Pushd(argv, dir_stack):
   num_args = len(argv)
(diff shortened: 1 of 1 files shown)


core/builtin.py
```diff
@@ -540,11 +540,17 @@ def Dirs(argv, dir_stack):
   if arg.c:
     del dir_stack[:]
   elif arg.v:
-    for i, entry in enumerate(dir_stack):
-      print('%2d %s' % (i, entry))
+    temp_stack = dir_stack[:]
```
andychu: How about unifying -v and -p under `PrintDirStack()` ?  It should just have a second parameter indicating what format to print it in.

Then the os.getcwd() and the sys.stdout.flush() can be shared among all 3 options.

core/builtin.py
```diff
@@ -484,9 +484,8 @@ def Cd(argv, mem):
 
 
 def PrintDirStack(dir_stack):
-  if dir_stack:
-    print(' '.join(dir_stack))
-
+  print(os.getcwd() + ((' ' + ' '.join(dir_stack)) if dir_stack else ''))
```
andychu: This feels a bit awkward.  I think `if dir_stack:` is a lot more straightforward.

But see the other comment about unifying this calculation.

core/builtin.py
```diff
@@ -540,11 +540,17 @@ def Dirs(argv, dir_stack):
...
-      print('%2d %s' % (i, entry))
+    temp_stack = dir_stack[:]
+    temp_stack.insert(0, os.getcwd())
+    for i, entry in enumerate(temp_stack):
+      print('%2d  %s' % (i, entry))
+    sys.stdout.flush()
   elif arg.p:
-    for entry in dir_stack:
+    temp_stack = dir_stack[:]
+    temp_stack.insert(0, os.getcwd())
```
andychu: This could be written as:

```
dirs = [os.getcwd()]
dirs.extend(dir_stack)
```

In general it's more idiomatic to append to lists rather than insert into the middle.  Because insertion is `O(n)` while appending is constant time.

core/builtin.py
timetoplatypus: Sure, I think that can be done. I was keeping it separated for organizational reasons, but it would be more efficient to consolidate that code.

core/builtin.py
timetoplatypus: No problem, I can make that change.

core/builtin.py
timetoplatypus: I added a mode argument (integer) to the PrintDirStack() function but I'd rather not have magic numbers. At the moment, I just have the constants defined right before the function, like so:
```python
WITH_PREFIX = 1
WITHOUT_PREFIX =2
def PrintDirStack(dir_stack, mode):
  if mode == WITH_PREFIX:
    #etc etc etc
```
Is it fine to just define those constants right before the function or is there somewhere else I should be putting them?

core/builtin.py
```diff
@@ -483,9 +483,24 @@ def Cd(argv, mem):
...
-def PrintDirStack(dir_stack):
-  if dir_stack:
-    print(' '.join(dir_stack))
+WITH_PREFIX = 1
+WITHOUT_PREFIX = 2
+SINGLE_LINE = 3
+def PrintDirStack(dir_stack, mode = 3):
+  dirs = [os.getcwd()]
+  dirs.extend(dir_stack)
+  if mode == 1:
```
andychu: Use WITH_PREFIX here, not 1.

core/builtin.py
```diff
@@ -483,9 +483,24 @@ def Cd(argv, mem):
   return 0
 
 
-def PrintDirStack(dir_stack):
-  if dir_stack:
-    print(' '.join(dir_stack))
+WITH_PREFIX = 1
+WITHOUT_PREFIX = 2
+SINGLE_LINE = 3
+def PrintDirStack(dir_stack, mode = 3):
```
andychu: add a blank line after these constants

core/builtin.py
```diff
@@ -483,9 +483,24 @@ def Cd(argv, mem):
...
+  if mode == 1:
+    for i, entry in enumerate(dirs):
+      print('%2d  %s' % (i, entry))
+    sys.stdout.flush()
+  elif mode == 2:
+    for entry in dirs:
+      print(entry)
+    sys.stdout.flush()
+  elif mode == 3:
+    if dirs:
```
andychu: dirs will always be non-empty, so you can remove this.

Also sys.stdout.flush() should go at the end of the function.

core/builtin.py
andychu: Remove the default 3 and just pass SINGLE_LINE in the other few places.

Python has some pitfalls with regard to mixing default values and passing positional args (as opposed to named args).

andychu:
So I ran the spec tests, and they passed, which is good.  However when I poked around manually, there were a couple obvious bugs.  The dir stack was printed in the wrong order, and the wrong entries were printed (you can't print os.getcwd() every time.)

I updated the tests to catch these bugs.  Generally it is good to be test driven, but that requires that the tests are thorough.

See these three commits:

https://github.com/oilshell/oil/commit/91d4bccb7a930c61b8396e5e341fd07c9bfaf545

https://github.com/oilshell/oil/commit/87f1f86dc74a6b137568aaa6ac614268024f868b

https://github.com/oilshell/oil/commit/f17ff5298db23f6775ba6946c8639edca3606533

Thanks for working on this.

timetoplatypus:
I'm going to spend some time and revamp the tests for the pushd/popd/dirs functions so it's more comprehensive. Thinking about it more, there are still other cases that the implementation still fails. It feels like I'm just bandaging or piece-mealing something that should be well-defined before I even try implementing.

For example, it fails the following spec test:
```bash
### dirs test of path aliases
cd /tmp
pushd .. >/dev/null
dirs -p
# stdout-json: "/\n/tmp\n"
# status: 0
```

andychu:
Yes, good point.  I introduced that bug (although the old code wasn't correct either), so that is a good one to add.

timetoplatypus:
Do you have any more thoughts regarding tests that use home directories? One of the critical tests of a well-functioning pushd/popd/dirs is whether it correctly prints the directory stack with tilde-prefix format when the directories in question are in the user's home directory, e.g. /home/timetoplatypus/Documents would be printed by default as ~/Documents

andychu:
It looks like in bash you can set $HOME:

```
$ dirs
~/git/oilshell/oil

~/git/oilshell/oil$ HOME=~/git/oilshell

andy@lisa:~/oil$ dirs
~/oil
```

The environment variable `$TMP` is set to `_tmp/spec-tmp` in the test harness.  You can possibly set `HOME=$TMP/dirs-test` and do `dirs -l`.

Although I think this is mostly orthogonal to the other problem (`..` bug).  I would solve one problem at at time.

timetoplatypus:
Ah, interesting. Didn't know it would stick to the tilde-prefix format if you changed HOME manually. Using that with the test harness should work just fine.

andychu:
FYI I just made a 0.3.0 release and am going to announce it tomorrow. 

http://www.oilshell.org/releases.html

`dirs` is shipped as is, but that's OK since it's definitely better than it was before.

andychu:
FYI I implemented a new multiline test format, and changed `spec/builtins2.test.sh` to use it:

https://github.com/oilshell/oil/commit/a086779f89e2e6e51da67d7c3eea328d22c3a7f2

The `stdout-json:` format was really annoying!!!  Hard to read and hard to write.

I added a bit of documentation here:

As a headers up, if you have any pending changes to that file, this might merge badly with them.  Sometimes I just port the changes manually instead of messing with git.

If anything about the format is unclear, let me know.

timetoplatypus:
Sounds good. Ya, this multi-line format should make things a lot easier. I'll probably just merge it by hand, I do have a bunch of tests I've written but they're in the old format.

PR #57, opened 2017-12-29
Updates `dirs` builtin

timetoplatypus:
- Switches the `dirs` implementation over to the tilde-prefix format when printing the directory stack.
- Fixes a few bugs including when the user invokes Pushd/Popd/Dirs with path aliases such as `.` and `..`
- Adds more spec tests for `dirs`


core/builtin.py
```diff
@@ -604,28 +604,51 @@ def Cd(argv, mem, dir_stack):
...
-WITH_PREFIX = 1
-WITHOUT_PREFIX = 2
+WITH_LINE_NUMBERS = 1
+WITHOUT_LINE_NUMBERS = 2
 SINGLE_LINE = 3
+TILDE_PREFIX_FORM = 1
+LONG_FORM = 2
 
-def _PrintDirStack(dir_stack, style):
+def _PrintDirStack(dir_stack, mem, style, form):
```
andychu: I think the interface to this function can be simplified.  Why not just pass an optional `home_dir=None` arg?  If it's not None, then do the translation.

We don't need `mem` or `form`.  That can be pushed to the outer levels.

core/builtin.py
```diff
@@ -604,28 +604,51 @@ def Cd(argv, mem, dir_stack):
...
   """Helper for 'dirs'."""
 
-  if style == WITH_PREFIX:
+  if style == WITH_LINE_NUMBERS:
     for i, entry in enumerate(dir_stack.Iter()):
-      print('%2d  %s' % (i, entry))
+      if form == TILDE_PREFIX_FORM:
```
andychu: Likewise you can get rid of the `if` here with a function like

`_FormatDir(entry, home_dir=None)`

core/builtin.py
```diff
@@ -604,28 +604,51 @@ def Cd(argv, mem, dir_stack):
...
+      elif form == LONG_FORM:
+        print(entry)
 
   elif style == SINGLE_LINE:
-    print(' '.join(dir_stack.Iter()))
+    if form == TILDE_PREFIX_FORM:
+      for i, entry in enumerate(dir_stack.Iter()):
+        # Don't add a space after printing the path if it's
+        # the last entry being printed.
+        sys.stdout.write(_TranslateHomePrefix(entry, mem) + \
```
andychu: This looks equivalent to `' '.join()`

core/builtin.py
```diff
@@ -604,28 +604,51 @@ def Cd(argv, mem, dir_stack):
...
+      print(' '.join(dir_stack.Iter()))
 
   sys.stdout.flush()
 
 
-def Pushd(argv, dir_stack):
+def _TranslateHomePrefix(dir_name, mem):
+  val = mem.GetVar('HOME')
+  if val.tag == value_e.Str:
+    return dir_name.replace(val.s, '~', 1)
```
andychu: I think you have to check if the value starts with the home_dir.  `replace()` will replace it in the middle, which isn't correct.

core/state.py
```diff
@@ -263,6 +263,8 @@ def Iter(self):
     """Iterate in reverse order."""
     return reversed(self.stack)
 
+  def Size(self):
+    return len(self.stack)
```
andychu: not needed if you follow my suggestions

core/cmd_exec.py
```diff
@@ -275,13 +275,13 @@ def _RunBuiltin(self, builtin_id, argv):
       status = builtin.Jobs(argv, self.job_state)
 
     elif builtin_id == EBuiltin.PUSHD:
-      status = builtin.Pushd(argv, self.dir_stack)
+      status = builtin.Pushd(argv, self.mem, self.dir_stack)
```
andychu: Although it's a little redundant, I think all three functions should just take `self.mem.GetVar('HOME')`

In general it's better to have a narrower interface, and not depend on all of `Mem()` when you only need to read a single variable.

core/builtin.py
timetoplatypus: Good point, I can add a check for that.

core/cmd_exec.py
timetoplatypus: True, it does only need one variable. I can change it to just take the value of $HOME specifically

andychu:
Not sure if you are done with this, but it looks like the `replace()` call still needs to be fixed.

I get a notification from Github when you push.  In the future I will not look unless you explicitly respond with some comments.  Just so the workflow is clear.

timetoplatypus:
Yep, I'm still working on it.

Sounds good, I'll post on whatever thread we're working on when the code is ready for review.

timetoplatypus:
Do you think that situation is worth writing a test for? (Where a path may have "/home/<username>" in the middle of it, so that shouldn't be translated to a tilde)

timetoplatypus:
I fixed the `replace()` issue by checking if the input starts with the home directory path. Let me know if you'd like a spec test for that too. Otherwise, the code should be ready for another review.

core/builtin.py
```diff
@@ -604,28 +604,35 @@ def Cd(argv, mem, dir_stack):
...
-    print(' '.join(dir_stack.Iter()))
+    print(' '.join(_FormatDir(entry, home_dir) \
+      for entry in dir_stack.Iter()))
 
   sys.stdout.flush()
 
 
-def Pushd(argv, dir_stack):
+def _FormatDir(dir_name, home_dir):
+  if home_dir and home_dir.tag == value_e.Str and dir_name.startswith(home_dir.s):
```
andychu: I think there is still a bug here, where

```
HOME=/home/andy
```

and

one of the entries is `/home/andyc`.  In that case it doesn't make sense to print `~c`.

So you have to handle at least two cases, and yes there should be tests for these.

core/builtin.py
```diff
@@ -604,28 +604,35 @@ def Cd(argv, mem, dir_stack):
...
 
-  elif style == WITHOUT_PREFIX:
+  elif style == WITHOUT_LINE_NUMBERS:
     for entry in dir_stack.Iter():
-      print(entry)
+      print(_FormatDir(entry, home_dir))
 
   elif style == SINGLE_LINE:
-    print(' '.join(dir_stack.Iter()))
+    print(' '.join(_FormatDir(entry, home_dir) \
```
andychu: Don't need `\`.  In Python wrapping is assumed between `()`.  Also indent continuation lines by 4.

core/builtin.py
timetoplatypus: Ahh, yes I missed that case. I'll write up some tests and figure out how to handle that

timetoplatypus:
I added spec tests for those aforementioned cases.

For the bug you mentioned where $HOME appears as a substring of a different legitimate path, I changed the code so that when it does the comparison using `.startswith()` the path has to also end with another forward slash in order to match.

Can you take a look and let me know if I missed anything?

andychu:
What about the case where `$PWD` is exactly `$HOME` ?

timetoplatypus:
Good point. I've added a spec test for that case and fixed it in code by checking if the directory name passed in is exactly the home directory passed in. If it is, it'll short circuit and proceed with the tilde replacement. Otherwise, it'll continue to check for whether the directory name passed in starts with the home directory.

andychu:
OK thanks.  I split the dirs tests out into their own file because only bash and zsh implement it.

Generally everything should be green and yellow, and dash and mksh were all red.

There was one case that zsh was red on too.  If there is red, then I mark it "OK" and with the differing stdout.

https://github.com/oilshell/oil/commit/4a12881dfb6d56295001201136c33019217a0c

If anything about the spec tests don't make sense let me know.

PR #71, opened 2018-02-01
Add support for DESTDIR in install script

kbillings:
This adds better support for installing in local directories.

### DIFF
--- a/install
+++ b/install
@@ -26,15 +26,17 @@ main() {
   #local exec_filename=oil.ovm-dbg  # Should this go in configure output?
   local exec_filename=oil.ovm  # Should this go in configure output?
 
-  local dest=$PREFIX/bin/$exec_filename
+  local parent=$DESTDIR$PREFIX
+  local bindir=$parent/bin
+
+  local dest=$bindir/$exec_filename
 
   echo "Installing to $dest"
-  if ! install _bin/$exec_filename $dest; then
+  if ! install -Dm755 _bin/$exec_filename $dest; then
     die "Couldn't install oil binary"
   fi
 
-  cd $PREFIX/bin
-  if ! ln -s -f $exec_filename osh; then  # -f to overwrite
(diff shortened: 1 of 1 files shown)


andychu:
Can you explain the motivation for this change?

You should be able to install into an arbitrary directory with --prefix:

http://www.oilshell.org/release/0.3.0/doc/INSTALL.html

kbillings:
Well you need to rerun configure every time you need to change the installation directory.
Also DESTDIR is a standard variable in installation scripts and is used very often in package creation such as deb.

andychu:
I haven't seen that before.  Do you have a reference?

If this change were to be made, it would need to be commented and documented (in configure --help, in INSTALL.txt, etc.)

However, I'm inclined to just let someone try packaging it for Debian.  I'm pretty sure they can do it without this -- OSH is already packaged for Arch and Nix.

Probably what they would do is simply have the Debian makefile run `./configure --prefix $DESTDIR` or something.

kbillings:
DESTDIR is used in scripts like debuild and dh_auto_install.

Generally DESTDIR is not specified in the configure script but only in the makefile:

https://www.gnu.org/prep/standards/html_node/DESTDIR.html

Since osh has a simple deployment without using any configuration files, you don't have a massive need for DESTDIR and you can use PREFIX for a local install without any problems. But with config files you can't install something to /etc locally because it lays outside of PREFIX.

It's just you have named these installation scripts like they are autotools-generated but they are not, so I attempted to use DESTDIR and it failed so I wrote this patch.

andychu:
OK thanks for the link and background.  For now I'm inclined to leave it out, but @timetoplatypus added a simple man page that has yet to be deployed:

https://github.com/oilshell/oil/blob/master/doc/osh.1

I'm honestly not sure where distros want to put the man page and how it works, but it might be related.

Though I do want to learn toward simplicity, and anything like this that is added should be documented.

I'll close this for now, but I won't be surprised if it comes up again.

PR #72, opened 2018-02-01
Let's run tests on all new code submissions

cclauss:
There seems to be nice test coverage so it would be a good idea to run those tests on all new code submissions.  This configuration will enable Travis CI to run [flake8](http://flake8.pycqa.org) tests on all pull requests before they are reviewed. We could also run the /test directory tests.  This allows contributors and maintainers to make sure no errors are introduced before new code is reviewed.  The owner of this repo would need to go to https://travis-ci.org/profile (log in via GitHub id) and flip the repository switch __on__ to enable free, automated testing of each pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_script:
     # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
     - flake8 . --count --exclude=./Python-2.7.13 --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 script:
-    - true  # pytest --capture=sys  # add other tests here
+    - test/lint.sh
 notifications:
     on_success: change
     on_failure: change  # `always` will be the setting once code changes slow down
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: python
+cache: pip
+python:
+    - 2.7
+    - 3.6
+    #- nightly
+    #- pypy
+    #- pypy3
+matrix:
+    allow_failures:
+        - python: 3.6
+install:
+    #- pip install -r requirements.txt
+    - pip install flake8  # pytest  # add another testing frameworks later
+before_script:
+    # stop the build if there are Python syntax errors or undefined names
+    - flake8 . --count --exclude=./Python-2.7.13 --select=E901,E999,F821,F822,F823 --show-source --statistics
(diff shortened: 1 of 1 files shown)


andychu:
Hi thanks for doing this.  I installed flake8 and it did find some useful errors.

Can Travis run `test/lint.sh` ?  If so, I prefer to put all the logic in that script, so I can run it locally.  How are you testing this -- by copying and pasting the flake8 command?

If Travis can do it, please put flake8 next to pep8 in this style:

https://github.com/oilshell/oil/commit/da69caed3771b95386c15e77b71c92b1381a4360

The dev environment is pretty closely tied to Ubuntu 16.04 at the moment, unfortunately.  But hopefully we can make flake8 run.

I will also likely limit flake8 to the four directories that pep8 is running.  At least until the number of errors gets lower.

I also experimented with "yapf", but I didn't want a lot of huge diffs related to autoformatting.

cclauss:
New commit adds test/lint.sh

cclauss:
flake8 is a superset of pep8.  They come from the same place... https://github.com/PyCQA

flake8 glues together pep8, pyflakes, mccabe, and third-party plugins to check the style and quality of python code.

If you are running flake8 then you don't need pep8 (now called pycodestyle) because it is built in.

Putting flake8 into test/lint.sh is a good way to go...  Just remove the __before_script__ section of the current PR and leave the __script__ section to do the heavy lifting.  I already did __--exclude=./Python-2.7.13__ but more could be done to focus the tool on what you want.  I would highly recommend that you keep __--select=E901,E999,F821,F822,F823__ in the mix because those are the showstopper issues.  Many other issues are merely "style" violations.

andychu:
OK thanks, I'll iterate on it and see how well it works.  The shell tests might be a challenge to get working on Travis, but we'll see.

Do you know if Travis creates an image every time?  Like you have to "cache" the PIP install -- what about installing custom versions of shells like zsh and mksh, and even more obscure ones?

That's the main issue that made me avoid Travis, but it's worth giving it a try now.

cclauss:
The default behavior seems to be that Travis hands you a new Docker container each time.  Then if you have something cached, it pushes that into the container.  And then the rest of your .travis.yml runs.

Alternatively you can start from your own Docker container (zsh, mksh, etc.) https://docs.travis-ci.com/user/docker

PR #73, opened 2018-02-02
import optparse for line 96

cclauss:
https://travis-ci.org/oilshell/oil/jobs/336386811#L471-L473


andychu:
optparse is in an unused function, will probably be a different fix

PR #74, opened 2018-02-02
import zipfile for line 30

cclauss:
https://travis-ci.org/oilshell/oil/jobs/336386811#L474-L476

### DIFF
--- a/build/testdata/hello.py
+++ b/build/testdata/hello.py
@@ -20,7 +20,7 @@ if not os.getenv('_OVM_DEPS'):
 
 import lib
 
-#import zipfile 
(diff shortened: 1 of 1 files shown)


andychu:
The zipfile is behind an `if 0`, which should probably be cleaned up in some other manner.

Thanks for the fixes -- I've merged a few, but I plan to get `test/lint.sh` and Travis in shape and do most of it in a big cleanup pass.

I will also put off doing the `opy/` directory, since that code hasn't been touched in awhile (and it was written by others.)  I plan to get back to it later, and then fixing bugs like these will be a good warmup.

PR #76, opened 2018-02-02
Don't del def_op

cclauss:
https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

https://travis-ci.org/oilshell/oil/jobs/336386811#L492-L500


andychu:
This one is in the `opy/` dir, which should be handled after all the core stuff passes.

Issue #81, opened 2018-02-03
Use Travis for more things


andychu:
Potential things to do:

- [x] tarball build 
- [x] minimal dev build
- [x] full dev build with re2c ?
- [ ] binary build
- [x] Python unit tests
- [x] spec tests -- some spec tests are done
- [ ] gold tests

Or even `scripts/release.sh test-and-build`.

Also, they support OS X, although it's overloaded.  See issue #70 .

One issue I found:

https://docs.travis-ci.com/user/reference/overview/

`sudo: required` means that the machine is a VM, which boots in 20-50 s vs. 1-6 s.

andychu:
I would prefer to use Alpine Linux because I think it will download and start faster.  But I don't think they have an option of providing your own VM.

I like Travis's live console output in JS, might be worth stealing.

andychu:
NOTE: If we want to upload artifacts to a web server or Github, then Travis has encrypted environment variables.

https://docs.travis-ci.com/user/environment-variables/

andychu:
Done with toil!

https://github.com/oilshell/oil/tree/master/services

PR #84, opened 2018-02-20
Make travis perform developer build, run {unit,spec} tests.

dreamtigers:
Spent a shameful amount of time trying to see why it didn't build the shell.
(Please consider I hadn't worked with Travis before)

About VM's vs containers... I don't know if you explicitly prefer one over
the other. I chose to approach the problem using containers since they boot
considerably faster than VMs. If you'd rather use VMs, just
substitute the `addons.apt.packages` and `before_install` lines for:

```yml
sudo: required

before_install:
    - build/dev.sh ubuntu-deps
```
Again, the advantage I see of containers over VMs is that they boot faster. The
disadvantage is that now dependecies must be stated in `build/dev.sh` AND
`.travis.yml`.

I'll look into making the tarball and binary builds.
Any guidance is well-received.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,6 @@ cache: pip
 python:
     - 2.7
 
-# TODO: remove on merge with master.
-branches:
-    only:
-        travis-ci # Where I'm trying to make the build and run the tests.
-
 addons:
     apt:
         packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,42 @@ language: python
 cache: pip
 python:
     - 2.7
+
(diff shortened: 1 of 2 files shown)


andychu:
OK thanks, sorry I replied on the wrong commit.  It looks like a step forward, so I'll merge and try to make it pass.  I'll probably disable some of the spec tests for now.

andychu:
OK thanks, I tweaked it and now it is green, and doing some meaningful stuff:

https://travis-ci.org/oilshell/oil/builds/345164398

More things could be done on issue #81, but this is a good start!

PR #91, opened 2018-03-22
Adds spec test for stdout and stderr redirection appending to a target.

timetoplatypus:
Just adding a spec test for stdout/stderr output redirection appending to a target (file, stream, device, etc). It's basically a copy pasta of the regular spec test for `&>` except that it puts some data into the target file before trying to append (so that it's evident that the subsequent operation appends to the file instead of overwriting it)

### DIFF
--- a/spec/redirect.test.sh
+++ b/spec/redirect.test.sh
@@ -277,3 +277,14 @@ echo second 1>&8
 echo CONTENTS
 cat $TMP/rw.txt
 # stdout-json: "line=first\nCONTENTS\nfirst\nsecond\n"
+
+### &>> appends stdout and stderr
+echo "ok" > $TMP/f.txt
+stdout_stderr.py &>> $TMP/f.txt
+grep ok $TMP/f.txt >/dev/null && echo 'ok'
+grep STDOUT $TMP/f.txt >/dev/null && echo 'ok'
+grep STDERR $TMP/f.txt >/dev/null && echo 'ok'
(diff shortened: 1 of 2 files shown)


spec/redirect.test.sh
```diff
@@ -277,3 +277,14 @@ echo second 1>&8
...
 cat $TMP/rw.txt
 # stdout-json: "line=first\nCONTENTS\nfirst\nsecond\n"
+
+### &>> appends stdout and stderr
+echo "ok" > $TMP/f.txt
+stdout_stderr.py &>> $TMP/f.txt
+grep ok $TMP/f.txt >/dev/null && echo 'ok'
+grep STDOUT $TMP/f.txt >/dev/null && echo 'ok'
+grep STDERR $TMP/f.txt >/dev/null && echo 'ok'
+# stdout-json: "ok\nok\nok\n"
```
andychu: Can you change this to use the new style?  Example here:

https://github.com/oilshell/oil/commit/9a9d3421f06bd801dcee8e03af3984c692dc65ba#diff-403617b9ff0265fda249640593bda3c1

The `stdout-json` is hard to read when there are multiple lines.

timetoplatypus: Sure, I can make that change.

timetoplatypus:
I've incremented the number of allowed failures and tried converting the spec test to the new format. Let me know if the changes look alright.

PR #95, opened 2018-04-08
byterun: Fix check for bound methods on falsy objects.

Yorwba:
The test failures on core/glob_test, core/lexer_gen_test and
core/lex_test were caused by calling a method on a falsy object
and byterun incorrectly treating the `self` parameter as nonexistent.

(At least those tests pass after applying the fix, so I assume that must have been the only issue.)

core/process_test still fails, I will look at it tomorrow if possible.

By the way, I was quite confused by the shell scripts for testing opy, and I ended up manually running opyc on the respective test files. Could you provide some example of the way you typically call those scripts?

### DIFF
--- a/opy/byterun/pyvm2.py
+++ b/opy/byterun/pyvm2.py
@@ -830,7 +830,7 @@ class VirtualMachine(object):
             #debug('im_self %r', (func.im_self,))
             #debug('posargs %r', (posargs,))
 
-            if func.im_self:
+            if func.im_self is not None:
                 posargs.insert(0, func.im_self)
 
             #debug('posargs AFTER %r', (posargs,))


andychu:
Wow amazing!  I was tearing my hear out about this bug last night!  Thanks so much.

I just tested it and it does fix it.

I cloned your repo and I did discover that I unfortunately broke the testing scripts! :-(  I was running out of an old copy of the bytecode that the scripts no longer produce.  (It's `$REPO_ROOT/_tmp/oil-with-opy` now, but some of the scripts assume `$REPO_ROOT/opy/_tmp/oil-opy`.  This is because I moved things to build from the root `Makefile`.)

I will fix that and document it, and merge your request later today.

BTW I think `core/process_test.py` passes -- let me double check.  I temporarily broke that test and fixed it I believe.

Just curious how did you spot it?  I was having a lot of problems.  One reason for that is because byterun confuses host and guest  a lot.  You'll notice that I added the `GuestException`, and used the real stack traces.  Somehow I'm not sure the stack traces were pointing to the right problem.  They are better than before but I am still confused sometimes.

Another problem I ran into was that my debug prints call `repr()`, which somehow interacted badly with the `__repr__` on `Subpattern` instances in `sre_parse.py`.  Still don't know why.

Oh also I have a few interesting things about the metaprogramming idea we were talking about... in particular PyPy does the same thing!  You can program with  full Python at startup, and then the "entry point" concept in PyPy is where it switches to RPython.




(Another point of confusion between host and guest: imports.  Because the `IMPORT` bytecode calls `__import__`, which is written in C and respects `PYTHONPATH`.  It would have been nicer to have an import implementation that respect `BYTERUNPATH` instead.  But I may not need that for OPy because of the "metaprogramming" style.)

andychu:
Ah I see the `core/process_test.py` failure is because no shell scripts make `_tmp`.  Sorry let me fix that too.

This whole thing is more of a learning experience to help me write my own interpreter, not final code, so it's pretty messy.

But it's been worth it!  I didn't really understand the "block stack" within the "call stack" before yesterday, used by `SETUP_{LOOP,EXCEPT,WITH,...}` ...

andychu:
OK I merged this and all the Oil unit tests pass under byterun.  Awesome, thanks a lot!!!

And I fixed the scripts:

https://github.com/oilshell/oil/commit/ac26ef01b35ea39bc26ec4d73916d047c640cd4b

    cd opy/
    ./build.sh oil-repo  # build with OPy
    ./test.sh oil-unit  # run with CPython
    ./test.sh oil-unit-byterun  # run with byterun

Also, another way I test it is like this:
 
    opy$  testdata/regex_compile.py  # run with CPython
    opy$  ../bin/opyc run testdata/regex_compile.py

That was failing before your change, and now works.

Some of this might need to be cleaned up -- I need to add a section to https://github.com/oilshell/oil/wiki/Contributing about OPy.   Thanks for suffering through it -- as mentioned this is a big experiment, but it's been worth it so far.

Also, it's probably worth filing a bug against the original byterun.  Since you fixed the bug I'll let you do that, or if you don't have time let me know and I'll do it.

https://github.com/nedbat/byterun/blob/master/byterun/pyvm2.py

andychu:
FYI I updated the doc here with some shell commands:

https://github.com/oilshell/oil/blob/master/opy/README.md

Yorwba:
I found the stack traces pretty confusing as well. I noticed the `GuestException` in there when I looked over the code, but I didn't know about `more_info`, so I pretty much ignored that.

Instead I grepped directly for the string in the error message and put `import pdb; pdb.set_trace()` in front, so I could look at the live state. Since the error was about `append` being called on a tuple instead of a `SubPattern`, I printed the `func.im_self` and I was quite surprised when it showed `[]` instead of either of those types. But then I realized that `SubPattern` was doing something with its `__repr__` so I suspected that boolean conversion wasn't reliable either. And `bool(func.im_self)` did return `False`.

(Actually I thought that it had a `__bool__` or `__nonzero__` magic method, but just now I checked and it doesn't. Turns out that `bool` will also check the `len` and return `False` if it is 0. Sometimes Python is trying to be a bit too smart for my taste.)

andychu:
OK interesting, yeah I should use pdb more!  I tend to use shell scripts and print statements to debug, but they're not as good for inspecting state.

In retrospect the `__repr__` was probably a consequence of the other bug.  So I was stuck on the symptom and not the real cause.  Thanks!

PR #98, opened 2018-04-30
Fixes typo in umask (symbolic) spec test causing shell and bash to fail.

timetoplatypus:
There's a typo causing the spec test for symbolic umask to fail. This pull requests just fixes that typo.

### DIFF
--- a/spec/builtins.test.sh
+++ b/spec/builtins.test.sh
@@ -181,6 +181,6 @@ umask g-w,o-w
 echo two > $TMP/umask-two
 stat -c '%a' $TMP/umask-one $TMP/umask-two
 # status: 0
-# stdout-json: "664\n644\n"
+# stdout-json: "644\n644\n"
 # stderr-json: ""
 


andychu:
Hm I think this might be because you and I have different umasks?  On my machine, dash, bash, and mksh pass, and osh fails.

On your machine, bash and osh fail?

If so, this is related to issue #42.  I think there is a bug here, but I'm not sure this is the right fix.

Also, good point about `-S`, but I think bash appears to "guess" at what you meant, even if you don't provide `-S`.  I think other shells like dash might not accept `-S`?

andychu:
Actually can you type `umask` in your shell and paste the output?

````
andy@lisa:~/git/oilshell/oil$ umask
0002
```

timetoplatypus:
``` bash
$> umask
0022
$>
```

timetoplatypus:
> On your machine, bash and osh fail?

On my machine, there are three shells tested when I run the spec test: sh, bash, and osh. Originally, all three failed for me.

andychu:
OK thanks, I will look into this after I do the release, along with the completion script.  It does seem related to the different umasks.  Maybe I can reset those at the beginning of the test.

andychu:
Or feel free to propose a patch... I'm pretty sure the one you proposed will fail on my machine.

timetoplatypus:
Ya I think it just comes down to resetting at the start of the test.

andychu:
Yes, can you pull master and make sure this works on your machine?  Thanks for the report.

https://github.com/oilshell/oil/commit/5ce621d2c1f445b78eb40058735fd3ed065746be

timetoplatypus:
Pulled and checked. Your fix works, we can close this.

Issue #100, opened 2018-05-05
OSH crashes while parsing some malformed parameter expansions


Yorwba:
Examples:
```bash
bash$ echo '${a/"' | bin/osh
bash$ echo '${a+"' | bin/osh
bash$ echo '${a%"' | bin/osh
```
What `bash` says:
```bash
bash$ echo '${a/"' | bash
bash: line 1: unexpected EOF while looking for matching `"'
bash: line 2: syntax error: unexpected end of file
```
`bin/osh` crashes at [`osh/word_parse.py`, line 155](https://github.com/oilshell/oil/blob/d9bebcd0001dc98ff6f3ae6885dda5095a7f330b/osh/word_parse.py#L155) because the previous call to `_ReadCompoundWord` returned `None`.

I found this bug using my experimental fork of the RPython type inferencer, so that might be approaching usability (although the likelihood of a bug in the inference is still greater than that of a bug in the program).

PR #101, opened 2018-05-05
Handle parsing failure in _ReadVarOpArg (#100)

Yorwba:
I was expecting the fix to require additional changes further up the call stack, but apparently most of the error handling was already in place and only this one function didn't have the necessary check.

### DIFF
--- a/osh/word_parse.py
+++ b/osh/word_parse.py
@@ -145,6 +145,9 @@ class WordParser(object):
 
     w = self._ReadCompoundWord(
         lex_mode=arg_lex_mode, eof_type=eof_type, empty_ok=empty_ok)
+    if not w:
+      return None
+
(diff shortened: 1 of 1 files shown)


andychu:
Thanks, merged!

Unfortunately the error handling style is left over when I thought I could translate the parser from Python to C++.  (And I am used to writing C++ without exceptions).

Were I to do it again I would just use exceptions everywhere, which would avoid bugs like this.  I've fixed a few missing checks exactly like this.

I hope to get time to change it at some point!  There is a multi-level error message now, which kind of makes sense with return codes, but I think you can do it find with exceptions too.

```
$ echo '${a/"' | bin/osh
Line 2 of '<stdin>'
  ${a/"
      ^
Unexpected EOF reading double-quoted string that began here
---
Line 2 of '<stdin>'
  ${a/"
        ^
Error reading command word
```

PR #102, opened 2018-05-05
Replace leftover use of tdop.ParseError by p_die

Yorwba:
Just an oversight in a previous refactoring, it seems. One input that triggers this code is `$(((a+b)(c)))`.


andychu:
Filed issue #104 for this callgraph bug.  And also issue #103 for my last comment.

Yorwba:
`flake8` only complains about whitespace issues in `osh/arith_parse.py`. `pylint` would have caught the error, but it also reports lots of false positives.

PR #109, opened 2018-05-17
Implement cd -P and -L; respect --

### LINKED ISSUE #86: Bare double dash support
From https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash#11382

> a double dash (--) is used in bash built-in commands and many other commands to signify the end of command options, after which only positional parameters are accepted.

Individual binaries implement this behavior sometimes but OSH as a shell also has to support it too in order for builtins to properly work. For example, using `touch -- -r` while in OSH works, however `cd --` does not ("No such file or directory" error).

I was going to write some spec tests for this but I've only ever written them for specific features. I'm not really sure how to write a spec test for whether the OSH interpreter is parsing user input correctly, if that makes sense. If there isn't a way to do this, I can just write the spec tests for every builtin OSH currently has.

BatmanAoD:
Fixes #86 (at least for `cd`, which is the example given in the description).

Currently the only supported arguments are `-L` and `-P`. These are the only arguments required for POSIX compliance, but others do exist (e.g. my version of Bash has `-e` and, of all things, `-@`).

Adds three spec tests for checking how `cd` interacts with symlinks. The new spec tests all pass.

I do not know if `args.py` supports a way to ensure that "mode" arguments can be overridden. In this case, `cd -P -L` should behave like `cd -L`, but I don't know how to implement that.

### DIFF
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -533,7 +533,7 @@ def Cd(argv, mem, dir_stack):
 
   # Set $PWD.
   if arg.P:
-    state.SetGlobalString(mem, 'PWD', os.readlink(dest_dir))
+    state.SetGlobalString(mem, 'PWD', os.path.realpath(dest_dir))
   else: # `-L` is the default behavior; no need to check it
     # TODO: ensure that if multiple flags are provided, the *last* one overrides
     # the others
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -532,12 +532,12 @@ def Cd(argv, mem, dir_stack):
     return 1
 
   # Set $PWD.
-  if arg.P:
-    state.SetGlobalString(mem, 'PWD', os.path.realpath(dest_dir))
-  else: # `-L` is the default behavior; no need to check it
-    # TODO: ensure that if multiple flags are provided, the *last* one overrides
-    # the others
-    state.SetGlobalString(mem, 'PWD', dest_dir)
+  # `-L` is the default behavior; no need to check it
+  # TODO: ensure that if multiple flags are provided, the *last* one overrides
+  # the others
+  pwd = os.path.realpath(dest_dir) if arg.P else dest_dir
+  state.SetGlobalString(mem, 'PWD', pwd)
+
   dir_stack.Reset()  # for pushd/popd/dirs
 
   return 0
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -322,7 +322,7 @@ def Jobs(argv, job_state):
 # - If there are more words than names, the remaining words and their
 # intervening delimiters are assigned to the last name.
 # - If there are fewer words read from the input stream than names, the
-# remaining names are assigned empty values. 
+# remaining names are assigned empty values.
 # - The characters in the value of the IFS variable are used to split the line
 # into words using the same rules the shell uses for expansion (described
 # above in Word Splitting).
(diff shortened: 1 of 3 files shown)


core/builtin.py
```diff
@@ -1088,7 +1099,7 @@ def Umask(argv):
     # NOTE: dash disables interrupts around the two umask() calls, but that
     # shouldn't be a concern for us.  Signal handlers won't call umask().
     mask = os.umask(0)
-    os.umask(mask)  #
```
BatmanAoD: Sorry about the whitespace diff lines; I auto-indented the whole file because I inadvertently started out using 4-width indents in the new code.

BatmanAoD:
Found a bug, putting this on hold...

BatmanAoD:
There we go--`readlink` didn't do what I thought it did. (I had a stupid idea of what it did, but still, it passed my initial smoketest.)

core/builtin.py
```diff
@@ -526,7 +531,13 @@ def Cd(argv, mem, dir_stack):
     util.error("cd %r: %s", dest_dir, os.strerror(e.errno))
     return 1
 
-  state.SetGlobalString(mem, 'PWD', dest_dir)  # Set $PWD.
+  # Set $PWD.
+  if arg.P:
+    state.SetGlobalString(mem, 'PWD', os.path.realpath(dest_dir))
+  else: # `-L` is the default behavior; no need to check it
```
andychu: nitpick: I use two spaces before the # in an end-of-line comment.  And I don't like backticks unless they are actually rendered in a monospace font :)

Since the lint infrastructure is not in great shape, and I'm travelling now, I'll just merge this since it looks good otherwise.  But for future reference I'm pointing out minor style issues.

spec/builtins.test.sh
```diff
@@ -54,6 +54,47 @@ cd-nonexistent
 status=0
 ## END
 
+### cd permits double bare dash
+cd -- /
+echo $PWD
+# stdout: /
+
+### cd to non-symlink with `-P`
```
andychu: same nitpick about backticks.

Also I think it's safe to say that $TMP is never going to have spaces, since it's controlled by `test/sh_spec.py`.  I tend to leave off quotes but I know that most people are in the habit of using quotes.  That is one of the things that Oil is supposed to fix :)  (i.e. not needing quotes everywhere "just in case")

core/builtin.py
```diff
@@ -526,7 +531,13 @@ def Cd(argv, mem, dir_stack):
...
+    # TODO: ensure that if multiple flags are provided, the *last* one overrides
+    # the others
+    state.SetGlobalString(mem, 'PWD', dest_dir)
```
andychu: I would factor out the dir like:

    pwd = os.path.realpath(dest_dir) if arg.P else dest_dir
    state.Set...

spec/builtins.test.sh
BatmanAoD: Ah, I didn't realize that `$TMP` wouldn't be inherited, if available, from the parent environment. I was just reading [Safe ways to do things in Bash](https://github.com/anordal/shellharden/blob/master/how_to_do_things_safely_in_bash.md), which suggests using quotes even when not necessary. I'll change it to be consistent with the existing tests, though.

BatmanAoD:
Without write access, I can't merge even with the 👍 review. But I've made the changes suggested, so I think it's ready for you to merge.

PR #111, opened 2018-05-18
Implement the pwd builtin

### LINKED ISSUE #110: `pwd` default behavior is `-P`, but other shells use `-L`
While working on #109, I discovered that in dash, bash, and mksh, `$PWD` is equivalent to `$(pwd -L)`, but in `bin/osh`, it's equivalent to `$(pwd -P)`. I believe the default should probably be changed to match the common behavior.

`-P` dereferences symlinks.

With  `/tmp/cd-symtarget` a directory and `/tmp/cd-symlink` a symlink to `cd-symtarget`, this is the behavior in most shells versus `osh`:

With `bash`:
```
bash$ cd /tmp/cd-symlink
bash$ pwd
/tmp/cd-symlink
bash$ pwd -P
/tmp/cd-symtarget
bash$ pwd -L
/tmp/cd-symlink
```

With `osh`:
```
osh$ cd /cd-symlink
osh$ pwd
/tmp/cd-symtarget
osh$ pwd -P
/tmp/cd-symtarget
osh$ pwd -L
/tmp/cd-symlink
```

BatmanAoD:
A simple `pwd` builtin that emulates the other common shells by making `-L` the default (rather than `-P`, the default in the GNU coreutils binary).

Fixes #110.

### DIFF
--- a/core/cmd_exec.py
+++ b/core/cmd_exec.py
@@ -119,7 +119,7 @@ class Executor(object):
     """
     self.mem = mem
     self.fd_state = fd_state
-    self.status_lines = status_lines  
+    self.status_lines = status_lines
     # function space is different than var space.  Not hierarchical.
     self.funcs = funcs
     self.completion = completion
@@ -317,6 +317,9 @@ class Executor(object):
     elif builtin_id == builtin_e.DIRS:
       status = builtin.Dirs(argv, self.mem.GetVar('HOME'), self.dir_stack)
 
+    elif builtin_id == builtin_e.PWD:
+      status = builtin.Pwd(argv, self.mem)
+
     elif builtin_id in (builtin_e.SOURCE, builtin_e.DOT):
       status = self._Source(argv)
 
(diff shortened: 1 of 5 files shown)


core/cmd_exec.py
```diff
@@ -1421,7 +1424,7 @@ class Tracer(object):
     - argv and span ID of the SimpleCommand that corresponds to that
     - then print line number using arena
     - set -x doesn't print line numbers!  OH but you can do that with
-      PS4=$LINENO
```
BatmanAoD: Huh, this time I didn't do an auto-indent intentionally, so I must have an auto-PyLint tool set up on my Vim.

BatmanAoD:
Oops, didn't re-test after rebasing. I should also add a spec test.

BatmanAoD:
Passing spec tests added.

andychu:
Great, thanks for doing this!  I merged both your changes.

I guess one thing I learned is that it would be more orthogonal if shell had something like $(realpath $PWD) instead of $(pwd -P) ?

In other words, it is confusing/odd to have a builtin variable `$PWD` as well as a builtin `pwd`.  I wanted to have some docs / style guide to point out these issues, but it's probably not a big priority now.

BatmanAoD:
You're welcome! Unfortunately I don't know if I'll have time to become a more regularly contributor any time soon, but I'm glad I could add something.

I've always thought the `$PWD`/`pwd` distinction seemed pretty odd, but apparently `readlink` and `realpath` are historically [pretty inconsistently available](https://unix.stackexchange.com/a/136527/38050)

PR #120, opened 2018-05-26
Implement set -o vi/emacs

### LINKED ISSUE #117: Question: how to change `readline` mode interactively?
Does `osh` provide a way to change the `readline` input-mode?

In Bash, I typically do:

```
set -o vi
```

But in `osh`, I get:

```
osh$ set -o vi
usage error: Invalid option name 'vi'
```

BatmanAoD:
Resolves #117.

### DIFF
--- a/bin/oil.py
+++ b/bin/oil.py
@@ -80,8 +80,10 @@ from core import ui
 from core import util
 
 if HAVE_READLINE:
+  import readline
   from core import completion
 else:
+  readline = None
   completion = None
 
 from tools import deps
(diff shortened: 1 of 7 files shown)


andychu:
Thanks, this looks great!

Can you add a trivial test to `spec/sh-options.test.sh` ?

Maybe just like this:

    set -o vi
    echo $?
    set -o emacs
    echo $?

I realize it's not testing much, but it's important to have some coverage for some of the "OPy" transformations I'm doing.

BatmanAoD:
Got it! Looks like there was no test for invalid `-o` options either, so I added one for that as well.

andychu:
Great thanks for doing this!

PR #123, opened 2018-05-28
Fix crash on empty for expression ((;;))

Yorwba:
Before this change
```bash
bin/osh -c 'for ((;;)) do done'
```
crashes with
```python
  ...
  File "core/cmd_exec.py", line 1000, in _Dispatch
    self.arith_ev.Eval(node.init)
  File "core/expr_eval.py", line 252, in Eval
    if node.tag == arith_expr_e.ArithVarRef:  # $(( x ))
AttributeError: 'NoneType' object has no attribute 'tag'
```
because there is no initializer expression.

Additionally (but unrelated to this bug), I'm not sure whether `do done` should be a valid loop body. `bash` doesn't accept it.

### DIFF
--- a/core/cmd_exec.py
+++ b/core/cmd_exec.py
@@ -997,17 +997,20 @@ class Executor(object):
 
     elif node.tag == command_e.ForExpr:
       status = 0
-      self.arith_ev.Eval(node.init)
+      init, cond, body, update = node.init, node.cond, node.body, node.update
+      if init:
+        self.arith_ev.Eval(init)
 
       self.loop_level += 1
       try:
         while True:
-          b = self.arith_ev.Eval(node.cond)
-          if not b:
-            break
+          if cond:
+            b = self.arith_ev.Eval(cond)
+            if not b:
+              break
 
           try:
-            status = self._Execute(node.body)
+            status = self._Execute(body)
           except _ControlFlow as e:
             if e.IsBreak():
               status = 0
@@ -1017,7 +1020,8 @@ class Executor(object):
             else:  # return needs to pop up more
               raise
 
-          self.arith_ev.Eval(node.update)
+          if update:
+            self.arith_ev.Eval(update)
 
       finally:
         self.loop_level -= 1
--- a/spec/for-expr.test.sh
+++ b/spec/for-expr.test.sh
@@ -42,3 +42,18 @@ done  # A construct borrowed from ksh93.
 ## N-I mksh status: 1
 ## N-I mksh stdout-json: ""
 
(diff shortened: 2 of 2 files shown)


andychu:
Thanks for catching and fixing this!  Can you add this case to `spec/for-expr.test.sh` ?

http://www.oilshell.org/release/0.5.alpha3/test/spec.wwz/for-expr.html

Run like this:

    test/spec.sh for-expr

Let me see about the empty loop body issue.

(I think I want to add a NoOp to `arith_expr` like there is one in `command` in `osh/osh.asdl`, so there aren't 3 if statements in the loop evaluation, but that could be done later.  I guess there has to be a NoOp boolean which evaluates to true for the loop condition though)

andychu:
Hm you are right about the empty loop body -- bash doesn't accept it.  And mksh and dash don't implement this `for ((` construct, so we can't consult them.

I don't really see why it shouldn't be allowed.  I guess if there is an easy fix I would accept it; otherwise I don't think it matters too much.  Logically it seems to make sense?

I mean you can have an empty case I think?  But maybe you can't have an empty if statement ?

andychu:
Because I'm obsessed I tested it and:

    case foo in 
    esac

works in all shells.

But

   if foo; then
   fi

does NOT work in any shell except OSH.

I think OSH is the consistent one -- it accepts empty do/done, then/fi, and case/esac!  The other shells seem inconsistent.  I'm inclined to leave it, but if it turns out to be a problem and there is an easy (consistent) change, we could make it more strict.  Logically I don't see a problem.

andychu:
Not that this is super important, but I enshrined the empty bodies as part of OSH here:

https://github.com/oilshell/oil/commit/3315b78f876bd9aef9ef8481f2a857a683eb51a4

As long as it's consistent and stable, I think this behavior is OK.  There would be fewer special cases in the parser as far as I can tell.

```
$ test/spec.sh empty-bodies 
empty-bodies.test.sh
  0       3     ok      ok      pass    pass    Empty do/done
  1      11     pass    pass    pass    pass    Empty case/esac
  2      17     ok      ok      ok      pass    empty then/fi
```

andychu:
Update: zsh behaves like OSH!

```
  0       3     ok      ok      pass    pass    pass    Empty do/done
  1      11     pass    pass    pass    pass    pass    Empty case/esac
  2      17     ok      ok      ok      pass    pass    Empty then/fi
```

Yorwba:
I added a spec test, and I agree that allowing empty bodies is more consistent.

andychu:
Great, thanks!

PR #127, opened 2018-05-28
A release tarball built on Linux now compiles on OS X

aykamko:
- modify detect_readline for Mac OS in configure
- update Python-2.7.13/pyconfig.h: remove some (seemingly) unused feature switches, move platform dependent switches into `configure` detects
- add `--readline-dir` flag to `configure` to enable linking against a different `readline` installation (the `readline` that ships with Mac OS is apparently outdated, so linking against e.g. homebrew `readline` may be necessary)

Addresses https://github.com/oilshell/oil/issues/70


build/common.sh
```diff
@@ -22,6 +22,18 @@ readonly PREPARE_DIR=_devbuild/cpython-full
...
+readonly UNAME=$(uname)
+
+is_macos() {
+  [[ "$UNAME" == "Darwin" ]] || return 1
+}
+
+if is_macos; then
+  readonly NPROC=$(sysctl -n hw.ncpu)
+else  # Linux
+  readonly NPROC=$(nproc)
```
andychu: Where is nproc used?

build/compile.sh
```diff
@@ -171,22 +171,33 @@ build() {
 
   pushd $PY27
 
-  local compile_readline
-  local link_readline
+  if is_macos; then
+    cp pyconfig.macos.h pyconfig.h
+  else
+    cp pyconfig.linux.h pyconfig.h
```
andychu: I wouldn't call this pyconfig.linux.h, because in theory it should work for all platforms.  I am accepting pyconfig.macos.h for expedience, as a hack, but ideally these things would converge and it would all be detected at `configure` time.

So maybe just call it pyconfig.default.h and pyconfig.darwin.h.  I think darwin is better because that's what `uname` says.

Or it would be even better if the headers can be merged, but I don't know how much work that is.

build/compile.sh
```diff
@@ -171,22 +171,33 @@ build() {
...
     # We should create a special no_readline_raw_input().
 
     c_module_src_list=$(cat $abs_c_module_srcs)
+
     # NOTE: pyconfig.h has HAVE_LIBREADLINE but doesn't appear to use it?
-    compile_readline='-D HAVE_READLINE'
-    link_readline='-l readline'
+    readline_flags=(-D HAVE_READLINE)
+
+    if is_macos; then
```
andychu: This logic should use feature detection, not version detection.  See this comment:

https://news.ycombinator.com/item?id=15195370

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Shell.20builtins.20and.20toybox

Alternatively, why not just make a homebrew package for OSH and set `--readline-dir` there, and remove this special logic?

I don't think it's safe to assume that all OS X users use HomeBrew.  Someone could have manually installed readline without HomeBrew, and it would live in a different dir.

build/compile.sh
```diff
@@ -258,18 +268,36 @@ python-sources() {
...
+    $(cat $abs_c_module_srcs)
+}
+
+_headers() {
+  local c_module_srcs=${1:-_tmp/hello/c-module-srcs.txt}
+  local abs_c_module_srcs=$PWD/$c_module_srcs
+
+  cd $PY27
+
+  # HACK: We extract Python headers twice, once with a Linux pyconfig.h and
```
andychu: Hm can you show the difference between these two sets of headers?  I feel like OS X should just use `posixmodule.c` too, and the headers shouldn't be very different, or at all.

configure
```diff
@@ -115,15 +126,30 @@ EOF
...
+
 # Write a shell script to standard out with variables, or fail.
 detect_readline() {
-  if cc_quiet build/detect-readline.c -l readline; then
-    echo 'HAVE_READLINE=1'
-  else
-    if test "$FLAG_with_readline" = 1; then
-      die 'readline was not detected on the system (--with-readline passed).'
+  if is_macos; then
+    if quiet command -v brew && quiet brew --prefix readline; then
```
andychu: Why is it necessary to call out to `brew`?  Is it because the C compiler is installed with Xcode but the libraries are installed with homebrew?

Why not just have people either:

1) Install a homebrew package of OSH, which already depends on readline.  No compilation necessary.
2) compile readline themselves and compile OSH against it.

It seems weird to install readline through `brew`, and then compile OSH manually.  It seems like both should be installed in the same fashion -- either with or without `brew`.

andychu:
Oh also I think a better name is `pyconfig.posix.h` and `pyconfig.darwin.h`.  Darwin is technically POSIX, but if merging the two would hold up the PR, we can do it in multiple steps.

build/common.sh
aykamko: no longer used, took it out
(this was a hold-over from my original approach)

build/compile.sh
aykamko: I decided to take a stab at unifying the headers and adding proper feature detection where needed 😁

build/compile.sh
aykamko: thanks for the advice! feature detection is indeed more robust.

and i think a `--readline-dir` flag is the way to go too

build/compile.sh
aykamko: the difference ended up being fairly minimal as you suspected!

aykamko:
Took another stab at this and managed to unify to a single `pyconfig.h`! Took out a few `#define` statements that don't seem too relevant and also added some additional feature detects in `configure`.

Also added a `--readline-dir` flag which I think makes sense. However: I feel like we should output a warning if `readline` is disabled or isn't recent enough?

build/compile.sh
```diff
@@ -171,22 +171,26 @@ build() {
 
   pushd $PY27
 
-  local compile_readline
-  local link_readline
-  if test "$HAVE_READLINE" = 1; then
+  local readline_flags
```
andychu: Maybe better to be explicit with `local -a readline_flags` ?

build/compile.sh
```diff
@@ -171,22 +171,26 @@ build() {
...
+
     # NOTE: pyconfig.h has HAVE_LIBREADLINE but doesn't appear to use it?
-    compile_readline='-D HAVE_READLINE'
-    link_readline='-l readline'
+    readline_flags=(-D HAVE_READLINE)
+
+    if [[ -n "$READLINE_DIR" ]]; then
+      readline_flags+=(-L "$READLINE_DIR/lib" -I "$READLINE_DIR/include")
+    fi
+    readline_flags+=(-l readline)
```
andychu: Can we put `-D HAVE_READLINE` here instead of at the beginning to slightly simplify the code?

I know that order matters for `-l` flags, but I think `-D` should be OK to move around.

build/detect-readline.c
```diff
@@ -4,8 +4,12 @@
 #include <readline/readline.h>
 #include <readline/history.h>
 
+static int test_event_hook(void) {
+  return 0;
+}
+
 int main(void) {
   char *line = readline("");
+  rl_event_hook = test_event_hook;  // ensure readline version is recent enough
```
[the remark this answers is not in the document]
aykamko: [This line Modules/readline.c](https://github.com/aykamko/oil/blob/babc25c6532e98003d116b07a49477327cdd3bae/Python-2.7.13/Modules/readline.c#L1111) breaks the MacOS build. I didn't dig any further but I suspect that the version of `readline` that you used when you initially generated `pyconfig.h` is more recent than the one that ships with MacOS.

Makefile
```diff
@@ -83,6 +83,8 @@ COMPILE_SH := build/compile.sh
 #BYTECODE_ZIP := bytecode-cpython.zip
 BYTECODE_ZIP := bytecode-opy.zip
 
+UNAME := $(shell uname)
```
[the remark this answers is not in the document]
aykamko: will do

configure
```diff
@@ -63,15 +71,19 @@ while true; do
       FLAG_without_readline=1
       ;;
 
+    --readline-dir)
+      assert_arg readline-dir "$@"
```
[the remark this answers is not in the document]
aykamko: ok, ill revert to the existing style. although `--prefix=` doesn't seem to work with that either.

configure
```diff
@@ -242,6 +279,17 @@ EOF
   else
     echo '#define VA_LIST_IS_ARRAY 1'
   fi
+
+  cc_header_file HAVE_PTY_H 'pty.h'
+  cc_header_file HAVE_LIBUTIL_H 'libutil.h'
+  cc_header_file HAVE_UTIL_H 'util.h'
+
+  cc_statement HAVE_STAT_TV_NSEC \
```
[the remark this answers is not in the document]
aykamko: I'll leave a TODO in here to investigate further for now.

Here's what I can guess:
- `pty.h` and alternatives seem important to OSH since that's related to pseudoterminals.
- [`HAVE_STAT_TV_NSEC`](https://github.com/aykamko/oil/blob/babc25c6532e98003d116b07a49477327cdd3bae/Python-2.7.13/Modules/posixmodule.c#L1467-L1515) is relevant to the `stat()` filesystem function, seems important too but maybe not

andychu:
What problem did you run into with `--prefix=` ?  I just tested and it appears to work fine:

```
andy@andy-VirtualBox:~/src/oil-0.5.alpha3$ ./configure --prefix=bar

andy@andy-VirtualBox:~/src/oil-0.5.alpha3$ cat _build/detected-config.sh 
HAVE_READLINE=1
PREFIX=bar
```

Although it is annoying, we should support `--readline=` for consistency.  (Again I'm open to another solution but just duplicating those few lines of code seems OK.)

-----

Also, can you add a reference to that line in `Modules/readline.c` in detect_readline.c` ?  

-----

I have to prepare a talk tomorrow and then I think I'll have some time to test this out on my Mac!  Thanks again.

aykamko:
Ah, I totally missed the `--prefix=*)` switch! Just added that for the `--readline` option.
Also added that reference in `detect_readline`

andychu:
Thanks for doing this!  I just merged and tested on my Macbook (I hadn't even installed a C compiler on it, and just did so).

It is failing with an error about `readline_flags[@]`.  I think this must be because bash 3.2 doesn't support arrays?  OS X ships with an ancient version of bash from 2007 or so.

I think for now `build/compile.sh` should just use strings as pseudo-arrays.  It's not ideal, but since there are no spaces in any of the arguments, it should be fine.  I guess one case is if `--readline-dir` contains a space, but I think nobody does that because tons of stuff would break.

I guess you must be using a newer version with Homebrew?  I might give this a stab tomorrow, since there are some other things I want to fix.  Like I looked at R and bash `./configure`, and they just support setting `LDFLAGS`.  They don't support the equivalent of `--readline-dir`.

So I think the homebrew recipe could just set `LDFLAGS` and `CFLAGS`, and osh `configure` and `build/compile.sh` would have to support it.  And then no special dir locations for libraries.

```
Andys-MacBook-Air:oil-0.5.alpha4 andy$ make
build/compile.sh build-opt _build/oil/ovm-opt _build/oil/module_init.c _build/oil/main_name.c _build/oil/c-module-srcs.txt
~/src/oil-0.5.alpha4/Python-2.7.13 ~/src/oil-0.5.alpha4
build/compile.sh: line 199: readline_flags[@]: unbound variable
make: *** [_build/oil/ovm-opt] Error 1

Andys-MacBook-Air:oil-0.5.alpha4 andy$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin15)
Copyright (C) 2007 Free Software Foundation, Inc.
```

PR #128, opened 2018-05-29
Replace leftover lex.VAR_NAME_RE.match by match.IsValidVarName

Yorwba:
This caused a few spec tests to fail. Maybe the CI should run them to prevent similar mistakes in the future.


andychu:
Thanks, sorry for the build failures!  That was my fault :-/

Fixing now.

PR #129, opened 2018-05-29
Use argv.py instead of argv in spec tests

Yorwba:
Maybe my dev environment just isn't set up correctly, but `argv` alone doesn't work for me. Since some of the other spec tests use `argv.py` as well, I assume that's the correct command.

### DIFF
--- a/spec/append.test.sh
+++ b/spec/append.test.sh
@@ -91,13 +91,13 @@ f
 f() {
   # NOTE: mksh doesn't like a=() after keyword.  Doesn't allow local arrays!
   local x+=(a b)
-  argv "${x[@]}"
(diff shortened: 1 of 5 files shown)


andychu:
This is correct, thank you!  `spec/bin/argv.py` is set to be in the `$PATH` by `sh_spec.py`.

Unfortunately I have an `argv` script in my own $PATH which means bugs like this sneak through.

Issue #42 addresses this.  I'm reluctant to enshrine Travis CI as the definitive environment, although that would probably work since it's Ubuntu and this similar to mine.  I wanted to make a chroot image but maybe that is overkill and the Ubuntu hack is more effective.

PR #130, opened 2018-06-01
fix build/compile.sh on bash 3.2

aykamko:


andychu:
As mentioned on Zulip, I fixed this a different way.  Thanks!

PR #134, opened 2018-06-03
parse glob to AST

aykamko:
Zulip thread: https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss
Github issue: https://github.com/oilshell/oil/issues/125

- [x] Use `osh/glob.asdl` to generate Python code that can represent a glob expression as an AST
- [x] Write a parser by hand to parse a glob expression directly into an AST
    - [ ] consider using `re2c` to generate a parser
- [ ] Convert the glob AST to a POSIX extended regex

### DIFF
--- a/core/glob_.py
+++ b/core/glob_.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
   from benchmarks import fake_libc as libc
 
+from osh.meta import glob as glob_ast
 from core import util
 log = util.log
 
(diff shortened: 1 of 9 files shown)


aykamko:
@andychu this is a rough draft, let me know if i'm heading in the right direction.

A couple of questions:
1. [This line](https://github.com/oilshell/oil/pull/134/files#diff-337eb6aa80f03d8bcd2af2d790107b2eR16) breaks my tests when running `test/unit.sh core/glob_test.py`. For now, I've disabled fastlex, but how should I go about this?
1. Do we really need an [`EscapedChar` class](https://github.com/oilshell/oil/pull/134/files#diff-94d11b2c2ffb76f7f6c63b649921434aR25)? Or can we just treat escaped chars (e.g. `\*`) as `Literals`?

andychu:
@aykamko Can you show me the unit test error?  The tests are passing for me and in the Travis build, but yes I have noticed the issue where the presence of `fastlex` changes the behavior of the unit tests.  (Travis runs without fastlex.)  

There may be something to change here, but I don't see why putting `fastlex = None` at the top level should change things.

Oh maybe I didn't check in something?  Let me try a clean clone.

------

I'm pretty sure we need `EscapedChar` because the representation of the value is different?

In other words, `Literal 'a'` is just `a`, but `EscapedChar \*` is `*`.  Does that make sense?

In regex form, I guess `\*` is `\*`.  But I guess that would be taking advantage of the fact that all glob metacharacters are regex metacharacters?

It appears to be true for `*`, `?`, `[`, and `\`.  They all have different meanings in regex, but I guess they are still metacharacters.  Hm so on the other hand, maybe you can get away with it.  If the tests pass I guess it's OK, although it seems a little weird.  I guess neither glob or regex syntax is going to change much in the next 20 years :-/

andychu:
Hm I made a fresh clone, and followed instructions here:

Then

    $ test/unit.sh all

finishes either with both `build/dev.sh minimal` (no fastlex) and `build/dev.sh all`.  (as mentioned, I could probably get rid of this distinction to simplify things.  Just vendor `re2c`.)

```
real    0m4.636s
user    0m3.691s
sys     0m1.002s

All unit tests passed.
```

aykamko:
Tried testing from a fresh clone and everything worked, thanks! Must have broken something along that way

I'll keep `EscapedChar`, seems correct

andychu:
Yeah I would leave `EscapedChar` in until everything is working, i.e. tests are passing.  It might become evident that we don't need it, but I think it makes sense to start with.

andychu:
Let me know you want more feedback on this... I was assuming it's still in progress?

No rush since I have a bunch of other pull requests to review :)

aykamko:
Still in progress! Got busy this work week, but hoping to finish this up this weekend!

aykamko:
@andychu I think this is mostly ready but I'm not sure if my tests are robust enough. Let me know what you think.

Also, turns out the `EscapedChar` class wasn't necessary. When we convert from glob AST to regex, I just check if `<literal> in '.|^$()+*?[]{}\\'` and that does the trick.

aykamko:
@andychu I think this is mostly ready but I'm not sure if my tests are robust enough. Let me know what you think.

Also, turns out the EscapedChar class wasn't necessary. When we convert from glob AST to regex, I just check if <literal> in '.|^$()+*?[]{}\\' and that does the trick.

core/glob_.py
```diff
@@ -67,67 +70,157 @@ def GlobEscape(s):
...
+      RuntimeError: If error during parsing the character class.
+    """
+    i = start_i
+    if glob_pat[i] != '[':
+      raise RuntimeError('invalid CharClassExpr start!')
 
     i += 1
+    # NOTE: Both ! and ^ work for negation in globs
+    # https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html#Pattern-Matching
+    negated = glob_pat[i] in '!^'
```
andychu: I was able to crash with this case:

```
~/git/oilshell/oil-aykamko (aleks/glob_parse)$ bin/osh -c 'echo ${a//[/}'                                                                                    
  File "bin/osh", line 623, in <module>
    main(sys.argv)
....
  File "/home/andy/git/oilshell/oil-aykamko/bin/../core/glob_.py", line 92, in Parse
    return self._ParseUnsafe(glob_pat)
  File "/home/andy/git/oilshell/oil-aykamko/bin/../core/glob_.py", line 124, in _ParseUnsafe
    char_class_expr, i = self.ParseCharClassExpr(glob_pat, i)
  File "/home/andy/git/oilshell/oil-aykamko/bin/../core/glob_.py", line 153, in ParseCharClassExpr
    negated = glob_pat[i] in '!^'
```

Issue #116, opened 2018-05-24
`FUNCNAME` highest-index element is missing in `osh`


BatmanAoD:
Currently, `osh`'s `FUNCNAME` variable doesn't include the last element.

From the [Bash manual](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html) description of `FUNCNAME`:

 > The bottom-most element (the one with the highest index) is "main".

Here's a demonstration:

```
> bash <(echo echo \${FUNCNAME[0]})
main
```

However, on my machine, the behavior seems to be different in an interactive or sourced context.

Interactively, the `FUNCNAME` array starts empty (just as in `osh`):

```
> echo ${FUNCNAME[0]}

```

In a sourced context, the word `source` is added upon entry into the sourced script:

```
> bash <(echo '. "$1"') <(echo echo \${FUNCNAME[0]})
source
```

andychu:
OK interesting, yeah it only seems to apply when you have a file.  The test harness uses `stdin` mostly, so there is a slightly different strategy to test this.

However since this is all bash-specific, I also want to revisit what it will take to make a good shell debugger.  bash falls short of this.

andychu:
Hm actually this was fixed a long time ago with the `source` argv implementation.  I am refactoring some of this code, but there is a test to prevent it from breaking again.

```
$ bash <(echo '. "$1"') <(echo echo \${FUNCNAME[0]})
source

$ bin/osh <(echo '. "$1"') <(echo echo \${FUNCNAME[0]})
source
```

BatmanAoD:
I've just tested this; it works with `source`, but Bash also creates a `main` stack-frame when starting:

    $ osh <(echo echo \${FUNCNAME[0]})

    $ bash <(echo echo \${FUNCNAME[0]})
    main

...I'm guessing this should be pretty simple?

andychu:
Yes thanks, I'll fix this.

andychu:
@BatmanAoD I'm looking at this now.  It's not super hard to implement, but it's not trivial because FUNCNAME, BASH_SOURCE, and BASH_LINENO are supposed to line up, and then this extra entry only happens when you pass file input, not in the `-c` or `stdin` case.  IMO this is a weird behavior.

Right now I don't distinguish heavily between file input, `-c`, and `stdin` so it causes a cascade of if statements.

Are you relying on this behavior of `${FUNCNAME[@]}` in your scripts ?

[...]

PR #138, opened 2018-06-06
Set up arguments appropriately before 'source'-ing script

### LINKED ISSUE #137: Arguments are ignored when sourcing a script
Bash accepts arguments when sourcing a script:
```
$ . <(echo 'echo $@') foo bar
foo bar
```
Osh does not:
```
osh$ . <(echo 'echo $@') foo bar

```

This may be a bit of a weird thing to do, but it turns out it's something I've done in a script I wrote.

BatmanAoD:
Fixes #137; addresses part of #116.

It looks like my `builtins` test for `set -o vi/emacs` somehow didn't make it in as part of #120 (??), so that is included as well.


andychu:
Hm this looks promising, but I feel like function calls might be different than source in other ways.  Let me think about it.

I moved your tests here, so I think you can drop that part of the PR:

https://github.com/oilshell/oil/commit/c1ad118a3beba52580e3f55527ab0913676cdb41

BatmanAoD:
Ah, you're right; experimenting a bit, it seems that Bash treats 'locals' in the scope of the sourced file as local to the scope *from which* the file was sourced. Accordingly, I've created a new Push/Pop pair.

andychu:
Hm I still have to look at this, but yes I was thinking of mentioning that bash has **dynamic scope**, unlike almost any other language (except early lisps).

It means it looks up the call stack (which is easy to implement, but not useful), rather than looking in the lexical context (which is hard to implement, but useful in Scheme/JS/etc.).

I'm not sure exactly if that is related to the change, but it is worth understanding for stuff like this :)

andychu:
I should also mention:

- it would be interesting to know what other shells do, with regard to both arguments to `source` and the scope of variables.
  - "bash does it and I use it" is a very strong argument, and many features are there for that reason.  But I try to also look at other shells.
  - Here is a minor counterexample: https://github.com/oilshell/oil/pull/123 . @Yorwba brought up
    a parsing issue and I tested all shells, and came to the conclusion that bash is "wrong".  But in this case the issue is so minor that it doesn't matter.
  - I think we should accept arguments to source, but I have to examine the scope issue.

- I am also thinking about the Oil semantics.  OSH should be convertable to Oil, so I wonder if this is desired Oil behavior or it needs a flag to turn it off.  (Not in this commit, but something to make note of for the future.)

BatmanAoD:
For arguments to `source`, check out the test I added: `mksh` lets `source` take extra arguments, but `dash` does not. (I haven't tested other shells.)

As a matter of consistency, I like the idea of treating the source code of a sourced file as the lexical scope of a function named `source`, or even `source.filename` (or something like that). This would have the advantage of making file-scope `local` variables actually `local` to that file, which I believe is otherwise impossible in Bash. (I could of course be wrong.) It's up to you, though.

BatmanAoD:
Further testing: `ksh` supports arguments, `sh` (on my system, so probably a Bash variant) does not.

core/state.py
```diff
@@ -475,6 +475,18 @@ def PopTemp(self):
     self.var_stack.pop()
     #util.log('**** PopTemp()')
 
+  def PushSource(self, argv):
+    """For the scope of a file read with 'source' or '.'."""
+    self.func_name_stack.append("source")
+    self.argv_stack.append(_ArgFrame(argv))
+    # Bash treats 'locals' in the scope of the sourced file as local to the
```
andychu: Hm I'm confused by this wording.  Are you just referring to top-level variables in the sourced file?

I think this is better illustrated by a spec test.  You can refer to it from a comment in the code.

spec/builtins.test.sh
```diff
@@ -156,6 +156,13 @@ echo status=$?
 # OK bash stdout: status=2
 # OK dash stdout: status=127
 
+### Source with extra arguments
+lib=$TMP/to-be-sourced
+echo 'echo $@' > $lib
+. $lib foo bar  # dash doesn't have source
```
andychu: This is a good test, but how about some more test cases?

What happens when you source a file inside a shell function?

How about testing multiple argv arrays?  If you need to make temp files you can do it with

```
tmp=$TMP/test-case-dir
mkdir -p $tmp
cd $tmp
echo lib > lib.sh
...

```

core/state.py
BatmanAoD: Yes, just top-level variables.

spec/builtins.test.sh
BatmanAoD: Yeah, sourcing a file inside a shell function is the only way I know of to demonstrate the "locals are exposed" issue I mentioned in the comment. That would be a good one.

I'm not sure what you mean about multiple `argv` arrays. There's only one `argv` array per scope, right?

BatmanAoD:
Sorry for a few days of silence. I currently don't have a Linux dev environment available (I'm trying to switch from VirtualBox to Hyper-V and ran into some issues trying to convert my Linux guest), so I won't be able to work on this for a while.

I feel fairly confident in the implementation, so I believe the only thing left to do is add the test cases you mentioned and maybe refer to them from the comment in the code. I think that would be a pretty small task if someone wanted to take this over.

BatmanAoD:
Hm, I was wrong -- Bash and `mksh` actually continue to expose any existing arg array if no new array is created when a script is sourced.

This is also true of nested functions, actually:

    bash$ f () { echo $@; }
    bash$ g () { f $@; }
    bash$ g foo bar
    foo bar

`osh` currently has this wrong. I think in order to fix this cleanly, one of the following needs to happen:

 * When looking up `$*` or `$@`, if the top of the `argv` stack is empty, try to use the next entry, and continue doing this recursively until the bottom of the stack is reached or a non-empty array is found. (I'm guessing this is essentially the dynamic scoping you mentioned?)
 * The `Push`/`Pop` functions should be turned into *objects* implementing `__enter__` and `__exit__`. When `argv` is empty, the object would record this and neither push nor pop a new `argv` entry.

These aren't mutually exclusive solutions, and the `__enter__`/`__exit__` refactoring would probably be valuable in any case.

BatmanAoD:
I...guess I didn't think about what I was doing when I wrote `g () { f $@; }`! Sorry. So it seems that the odd argument array behavior may only apply to `source`d scripts. I'll double check I didn't make a stupid error when checking the behavior there, too...

andychu:
OK I did a lot of testing like this manually, and it is annoying to do without mistakes.

The spec tests are useful in this respect, and you can put data files in `spec/testdata`.  Relative paths should work to reference those files.

Also, I put some semi-automated demos in `demo/`, e.g. for the recent readlink change.

It's better if things are fully automated in `spec/`, but sometimes I start with `demo/` and then move stuff over to `spec/` once I figure out what the relevant test cases are.

andychu:
Hm sorry that was so hidden.  I added these hints to the wiki page:

And I linked that page more prominently on the wiki.

BatmanAoD:
One other issue that confused me: I tried to leave a blank line in a `STDOUT:` section, but it wasn't honored as I expected:

    ## STDOUT:
    foo bar
    
    ## END

became `foo bar\n` rather than `foo bar\n\n`, causing spurious failures. Are blank lines intentionally trimmed?

spec/builtins.test.sh
```diff
@@ -1,160 +1,188 @@
...
+f() { . $lib; . $lib args to src; echo $foo; }
+f args to func
 ## STDOUT:
-cd-nonexistent
-status=0
+args to func
+args to src
+foo_val
 ## END
+## N-I dash
```
BatmanAoD: @andychu `bash` and `mksh` pass this test as-is; note that `args to func` is printed even though the script is sourced without `$@`. (Side-note: to get the "N-I" status for Dash, do I need to add a `stdout-json` to the `N-I` line, or can I specify a multi-line `STDOUT` for a `N-I` entry?)

So sourcing, at least, *does* need to create some kind of "optional" `argv` entry. I still think the `__enter__`/`__exit__` pattern would be the best way to handle scope push/pop logic; would you be amenable to that change?

andychu:
- Hm the empty line issue seems to be caused by the `LineIter()` tokenizer in `test/sh_spec.py` ignoring ALL empty lines, regardless of context.  That does seem like a bug... let me think about the best way to fix it.  For this change can we just work around it by adding some common prefix?  (And feel free to file a bug about this.)

- `N-I dash STDOUT:` is valid.  I usually use `stdout-json` for values like the empty string, which is a little clearer in the format.

- I avoided `__enter__/__exit__` in OSH because it's one more thing the OPy compiler has to understand.  I did some research on it, and it appears that `with` is more or less defined in terms of `try/finally`.  I think I am not understanding the benefit of enter/exit?  Is it mostly syntactic sugar or is there something fundamental?

- I think I see what you mean by the optional argv array.  But I will have to play with and test the semantics before I give a concrete suggestion.  It does seem pretty weird, but if both bash and mksh agree, I suppose we should implement that.

(I don't think `source` will be a recommended part of the Oil language; I would like something more static, and something that doesn't rely as much on global state.)

andychu:
Hm I guess I would have to see an example.  I thought that the `Mem()` object could hold the state?

`with` is definitely nicer in that it's 2 lines instead of 4.  But other than that I have been reasonably satisfied with try/finally.

Is this ready to look at again?  Last night I think I saw a bunch of commented out tests which made the diff big.

BatmanAoD:
I've uncommented those tests, but I don'can't think of any clean way for the `Mem()` object to hold the state. It can't just have a boolean indicating whether the most recent `source` had any arguments or not, since `source`d files can `source` files recursively, and so on, and sometimes the `source` statement may have arguments, sometimes not:

In script A:

    . B

In script B:

    . C foo bar

...so there needs to be some indication in `Mem` *per instance* of `source` whether or not a new `argv` array was pushed. This would either mean a parallel stack, or a different way of storing the stack of `argv` entries. Neither seems very clean to me.

If each stack entry were its own object, each object would either have an `argv` array or not.

BatmanAoD:
It doesn't pass all the tests yet, but here's (more or less) what it would look like if we used `with`: https://github.com/BatmanAoD/oil/compare/master...BatmanAoD:FramesUseWithSyntax?expand=1

BatmanAoD:
Let me know if you're okay with me continuing in that direction, and if not, what alternative you think would be preferable.

andychu:
Thanks for writing the test cases.  I think I need to pull this and play with it on my machine to fully understand what you're saying.   I think I'll have time for that this weekend.  In the meantime, let me throw out some things that may or may not make sense:

> so there needs to be some indication in Mem per instance of source whether or not a new argv array was pushed. This would either mean a parallel stack, or a different way of storing the stack of argv entries. Neither seems very clean to me.

I don't quite follow this.  Can we extend `_ArgFrame`, which is a tuple of (argv, num_shifted) ? 

How about one of these two strategies:

1. If source is passed arguments, push a new _ArgFrame() with those args.

2a. If it's not passed arguments, push an _ArgFrame with the current args.  That is, make a copy of the top of the stack.

2b. If it's not passed arguments, push an "_DelegatingArgFrame" that delegates all its methods to the one below.

Whether you use 2a or 2b depends on the behavior of "shift" when you source.  Can a "shift" in the sourced module affect the parent module?

-----

I'm not 100% sure if what I'm saying makes sense, but I'm throwing it out there until I have a chance to play with the code myself.

Also, I should add that I don't see anything "unclean" about parallel stacks, as that already exists within the `Mem()` class and it seems to work fine.  The stacks are hidden behinds methods so that invariants are preserved.

BatmanAoD:
Yes, you're right, we could extend `_ArgFrame`.

That's an interesting question about `shift`! I'll investigate.

BatmanAoD:
Okay, it looks like in both `bash` and `mksh`, `shift` inside a sourced script *can* affect the outer `argv` array if no new arguments are passed when sourcing the script.

andychu:
Hm interesting.  Can you push the commits  with the tests so I can try them on my machine?  I should have time tis weekend.

It seems like github is saying it's on the master branch of BatmanAoD/oil but the last commit I see there is 3 days ago?  I could be misunderstanding github.

https://github.com/BatmanAoD/oil

BatmanAoD:
Sorry for the confusion; I put the new test code in the `FramesUseWithSyntax` branch linked above.

andychu:
OK, I finally have some time to play with this!  Sorry for the delay.

I did a minor cleanup of your tests and then pushed them:

https://github.com/oilshell/oil/commit/1d6bbeb408bfbe5af7b76f5d232d75059460ec69

And then I made them pass here:

https://github.com/oilshell/oil/commit/cb9850fe12528ee6887bfd565aecc1d216b5140a

I get why you wanted to use `with`.  The duplicate conditional is slightly ugly.  But it's simple and straightforward.  It's many fewer lines of code.

Do you see any other cases that this doesn't handle?  I'd appreciate it if you can run your original script with this change, and try to come up with some things that don't work.  I basically did the simplest thing that makes the tests pass.

andychu:
OK and I also moved the tests to their own file:

https://github.com/oilshell/oil/commit/094636c1101659be2a237b952bb9512ed808535a

And surprisingly zsh does the exact same thing!  I'm always surprised when I see this weird behavior replicated by many shells.  But I guess the authors must have done similar regression testing.

BatmanAoD:
Hm, one issue is that the func-name is not dependent on whether or not args are provided. So `source` must be pushed onto and popped from `mem.func_name_stack` unconditionally.

BatmanAoD:
I'm also not quite convinced that the non-`with` version is "many" fewer lines of code, since all of the call-sites are significantly simpler (and in the case of `Frame_Temp` could be made even simpler).

andychu:
Yes good point!  I just fixed that:

https://github.com/oilshell/oil/commit/1ad997dc72468ff4be6128ef8f070c0269a8f85f

Regardless of whether it's shorter or not, I want to use as few constructs in the code as possible (to a reasonable degree).  If "with" is truly useful it could be added, but for this case I don't see any problem at all with try/finally.

andychu:
Thanks for all the help on this!  Hopefully I can get a release out soon :)  There are some obstacles because I used to do benchmarks on 2 very specific computers which I don't have access to when I'm travelling :-(

Issue #133, opened 2018-06-03
Implement ${BASH_REMATCH[@]}


andychu:
For the construct `[[ foo123 =~ ([a-z]+)([0-9]+) ]]`, `BASH_REMATCH` is filled with the captured groups.

```
$ [[ foo123 =~ ([a-z]+)([0-9]+) ]]
$ spec/bin/argv.py "${BASH_REMATCH[@]}"
['foo123', 'foo', '123']
```

OSH doesn't currently implement this.  This is a small task that involves working with the Python/C API.

PR #139, opened 2018-06-06
implement BASH_REMATCH (#133)

okayzed:
This diff adds simple support for BASH_REMATCH environment variable that exports a regex's capture groups

### DIFF
--- a/native/libc.c
+++ b/native/libc.c
@@ -224,7 +224,6 @@ func_regex_match(PyObject *self, PyObject *args) {
   }
 
   regex_t pat;
-  regmatch_t m[2];
 
   // Should have been checked by regex_parse
   if (regcomp(&pat, pattern, REG_EXTENDED) != 0) {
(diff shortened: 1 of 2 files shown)


native/libc.c
```diff
@@ -224,48 +224,51 @@ func_regex_match(PyObject *self, PyObject *args) {
...
-    if (pat.re_nsub>0) {
-      ret->s = xmprintf("%.*s", m[1].rm_eo-m[1].rm_so, target+m[1].rm_so);
-      if (TT.refree) free(TT.refree);
-      TT.refree = ret->s;
-    } else assign_int(ret, m[0].rm_eo);
-    */
-    ret = 1;
 
+
+  PyObject *ret = PyList_New(pat.re_nsub+1);
```
andychu: Looks like we can move outlen above this statement.

native/libc.c
```diff
@@ -224,48 +224,51 @@ func_regex_match(PyObject *self, PyObject *args) {
...
+
+  if (ret == NULL) {
+    regfree(&pat);
+    return NULL;
+  }
+
+  int outlen = pat.re_nsub + 1;
+  int nomatch = 0;
```
andychu: nitpick: can we make it `match = 0` and then set it inside the if, and reverse the last test?    I don't like bools with negatives in their names; it leads to double negatives like `!nofoo`.

okayzed:
> In any case we should be consistent with the prevailing CPython style (though it is annoying)

not sure if you want anything changed here (if you mean something aside from the 2 comments you left)

andychu:
OK great!  I was just talking about the C89 style of not doing `for (int i...`.

Hm I just realized we should have added a spec test to `test/spec.sh regex` for BASH_REMATCH.  We can do that in another commit.  The case in the issue can just be copied in there with the output assertion.

PR #141, opened 2018-06-06
Allow for multiple possible outcomes of spec tests

Yorwba:
This doesn't address issue #42 directly, but does fix the mentioned failures on Ubuntu 17. In cases where different versions of the same shell behave differently, you can now just specify all known possibilities. Then the spec tests won't break when one of the other shells lands a bugfix.

The new test format assumes that status and stdout specifications are adjacent when they belong together, which was fortunately already the case for the existing tests. Specifying a different combination of shells or setting the same value multiple times creates a new separate outcome. I think it's an intuitive extension of the old test format.

### DIFF
--- a/spec/alias.test.sh
+++ b/spec/alias.test.sh
@@ -128,3 +128,5 @@ shopt -s expand_aliases  # bash requires this
 alias e_.~x='echo'
 e_.~x X
 ## stdout: X
+## OK mksh status: 127
+## mksh R56 2017/08/08 only allows [][A-Za-z0-9_!%,.@:-] in aliases
--- a/spec/array.test.sh
+++ b/spec/array.test.sh
@@ -313,8 +313,8 @@ a=('123' '456')
 argv.py "${a[0]}" "${a[0][0]}"
 # stdout-json: ""
 # status: 2
-# OK mksh status: 1
-# bash is bad -- it IGNORES the bad subscript.
+# OK bash/mksh status: 1
+# old bash version is bad -- it IGNORES the bad subscript.
 # BUG bash status: 0
 # BUG bash stdout: ['123', '123']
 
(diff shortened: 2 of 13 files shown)


Yorwba:
Having a controlled environment for the tests is important because it allows separating failures due to breakage in OSH from failure due to unrelated changes in the environment. So that should eventually be the default way to test. I might have tackled that first, but I'm a bit fuzzy on how chroots even work, so I didn't.

But it should also be possible to test with just about any environment to see whether anything breaks (see e.g. the issues with gcc 8). In that case, you really only care about OSH failures, so failures of other shells shouldn't be counted as errors for that purpose.

But you do want to know when other shells decide to change their behavior, so if anything unexpected happens in someone's personal environment, it should be possible for them to tell what happened, e.g. whether it's a known bug or a new bugfix. So the spec tests should document the behavior of at least a few recent versions.

I realize that this is susceptible to bloat if you only ever add to the tests and never remove cases only relevant to shell versions you really don't care about anymore. Maybe they could be tagged with the version number they apply to, then it's clearer when the case can be deleted?

andychu:
Quickly about chroots, there is a demo here of how to do it:

https://github.com/oilshell/oil/blob/master/test/alpine.sh

It is very easy -- Alpine provides a base image that you just untar into a directory.  And then you use the chroot command to run `/bin/sh` in that image (busybox sh) with `_tmp/alpine-foo` as `/`.

So you need to set up a few things and then you can run shell commands or enter an interactive shell.

The only downside is you need root.  That is why I was mentioning the setuid binary that could chroot.

-----

I think I agree with everything you say, but I don't see why having "a few versions" is important.  If the user's environment doesn't match the "golden" test environment, then they will get some warning about `mksh` results not matching for say 17 test cases.

If they really want to drill down, they should just use the golden image.

I think that is OK and anything more complicated is just more work ?

-----

The main practical issue is whether the golden environment should be something lightweight like Alpine, or Ubuntu 16.04.  I was thinking the former, but I think it's too much work for now, and should be done in steps.

So the first step is to just use Ubuntu 16.04 (which is relatively straighforward with debootstrap), and then the second step would be to reduce the test dependencies (e.g. maybe they could just use busybox rather than coreutils.)

Yorwba:
Hm, if it's so easy with Alpine I don't understand why it would also be too much work compared to Ubuntu.

-----

My thinking with the multiple versions was that it's better to troubleshoot each version only once and then embed that knowledge in the tests, rather than everyone having to repeat that process when they need to drill down. I guess the utility depends on how often that is really necessary, so maybe the benefit isn't that great. I'll close this PR for now.

Issue #142, opened 2018-06-06
Implement utf-8 string operations


andychu:
- [x] length of string - `${#s}`
- [x] Get rid of dependency on `str.decode()` (`unicodeobject.c`)
- [x] slice of string (start index, length) - `${s:5:3}`

NOTE: `${s[5]}` is only for array indexing; it does NOT work for strings!

What about patsub?  That uses the libc regex engine, so does it respect global locale settings?

(NOTE: ping if starting this, because I may use it for RC pairing)

andychu:
note: string length passes tests, but it would be nicer to remove the dependency on `s.decode()` in Python (i.e. unicode objects).

andychu:
Done with 0.6.pre2

PR #145, opened 2018-06-08
Fix unicode string length

caroline-lin:
The length operator ${#s} returns the number of code points now instead
of the number of bytes. Also added a test for this. Addresses #142.


andychu:
Please remove tab

PR #154, opened 2018-06-27
Script to convert .pyc to C definitions

darius:
… for the experimental interpreter. (Easier than loading a code file at runtime.)

I wasn't sure where to put this, so it's just in opy/

### DIFF
--- a/opy/cstruct_from_code.py
+++ b/opy/cstruct_from_code.py
@@ -1,0 +1,67 @@
+"""
+Convert a Python2 code object (actually a subset of what's
+possible) to a sequence of C definitions suitable for the
+stripped-down experimental OPy interpreter.
+Usage like: python cstruct_from_code.py _tmp/fibonacci.pyc
+"""
+
+import marshal
+import json
+import sys
+
+def dump_code(code):
+    print """\
+typedef unsigned char uint8;
+typedef long long int64;
+
+typedef struct {
+    uint8 *co_code;
+    int64 *co_consts;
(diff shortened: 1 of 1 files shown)


opy/cstruct_from_code.py
```diff
@@ -0,0 +1,67 @@
...
+    print '  co_consts,'
+    print '  co_names,'
+    print '  %d,  // stacksize' % code.co_stacksize
+    print '};'
+
+def dump_constant(x):
+    if isinstance(x, int):
+        return '%d' % x
+    else:
+        return '0'   # XXX super hack
```
andychu: This issue and the JSON issue above gave me an idea in the shower (where else?)

I want to try using this oheap format I designed to serialize constants.  There are already 3 different types of constants in even the fibonacci example: int, None, and tuple of string.

In ASDL that can be represented like:

    const = None | Int(int i) | String(string s) 

And then on the C++ side you get a typed API.  It should be reasonably efficient, although perhaps not optimal.

Blog post: http://www.oilshell.org/blog/2017/01/09.html

There is a demo like this:

```
$ asdl/run.sh arith-encode-oheap
$ asdl/run.sh osh-demo
```

However they don't work for me right now because the shell scripts pointed to a version of Clang I had on my home machine... let me fix that.

andychu:
I'm closing this since we ended up going in a different direction (that I still have to get back to now that RC is over!)

Thanks for working with me on this!

PR #156, opened 2018-07-02
Spec Test Update

granttrec:
Spec tests will now follow the new convention.


andychu:
Thanks a lot for doing this!  I tested this on my machine and merged it.

PR #157, opened 2018-07-03
Update code comments to match new spec format

granttrec:
Missed the comments, so this just fixes that error.

### DIFF
--- a/spec/builtins.test.sh
+++ b/spec/builtins.test.sh
@@ -209,7 +209,7 @@ umask | grep '[0-9]\+'  # check for digits
 ## status: 0
 
 #### set umask in octal
-rm $TMP/umask-one $TMP/umask-two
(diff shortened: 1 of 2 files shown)


spec/builtins.test.sh
```diff
@@ -209,7 +209,7 @@ umask | grep '[0-9]\+'  # check for digits
 ## status: 0
 
 #### set umask in octal
-rm -f $TMP/umask-one $TMP/umask-two
+rm $TMP/umask-one $TMP/umask-two
```
andychu: This looks like you didn'dt merge https://github.com/oilshell/oil/commit/71cb657d78307938609ae87a742707815c9ca899

?

granttrec: There was a merge conflict but I picked the wrong one, will fix.

PR #165, opened 2018-08-06
Remove trailiing whitespace from *.py and *.c files

granttrec:
This is the first pass to format the project code. Potential idea for travis, for issue 148
```sh
if git --no-pager grep -P '\s+$' -- *.{c,py}; then
  printf "Bad commit: trailing white space ...\n"
fi
```

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -26,6 +26,7 @@
 #   $0 git-changelog-$VERSION
 #   $0 announcement-$VERSION
 #   MAYBE: ./local.sh test-release-tree if you want to preview it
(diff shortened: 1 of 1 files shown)


andychu:
Hm can you revert all the stuff in `Python-2.7.13/` and `opy/_regtest` ?

- We will barely ever change the Python dir.  I don't want to make it more complicated to do a `diff -r` to the orginal.
- The `_regtest` dir is golden source data, so it also won't be edited.

The Travis hook might be useful, but we can discuss separately.  I will probably have to reconfigure my editor ... otherwise this will cause a lot of annoying whitespace commits :-/

PR #166, opened 2018-08-06
Remove trailiing whitespace from *.py and *.c files

granttrec:
I managed to make a mess out of the previous pull request so here it is again.
The first pass for formatting the project code.
The travis idea for dealing with whitespace, also good to check the gitignore...
```sh
if git --no-pager grep -P '\s+$' -- *.{c,py}; then
  # do something
fi

### DIFF
--- a/asdl/asdl_.py
+++ b/asdl/asdl_.py
@@ -218,7 +218,7 @@ class Type(AST):
 
 class _CompoundType(AST):
     """Either a Product or Constructor.
-    
+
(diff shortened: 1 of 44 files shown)


andychu:
Sorry for the delay on this, I've been traveling and babysitting big rendering jobs (blog post may be forthcoming :-)  )

Thanks for doing this!

I will keep the `git grep` thing in mind, although I wonder why not just use `grep`?  We can discuss on issue #81.

PR #172, opened 2018-08-23
Fix comment errors

granttrec:
There are some small mistakes in some of the comments that have
been overlooked from the commit that reformatted the spec tests.
This commit amends those errors for consistent documentation.

### DIFF
--- a/test/sh_spec.py
+++ b/test/sh_spec.py
@@ -80,8 +80,8 @@ def log(msg, *args):
 
 
 # EXAMPLES:
-## stdout: foo
-## stdout-json: ""
+# stdout: foo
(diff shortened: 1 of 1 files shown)


test/sh_spec.py
```diff
@@ -80,8 +80,8 @@ def log(msg, *args):
 
 
 # EXAMPLES:
-## stdout: foo
-## stdout-json: ""
+# stdout: foo
+# stdout-json: ""
```
andychu: This change looks wrong?

The one below is right.

granttrec: yes my mistake

PR #174, opened 2018-08-26
Add compgen function to osh

granttrec:
This merge adds `compgen -A function` to osh. Not sure if `compgen` should be in `cmd_exec.py` or `builtins` so the previous is commented out.

### DIFF
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -801,12 +801,14 @@ COMPGEN_SPEC.ShortFlag('-A', args.Str)
 def CompGen(argv, funcs):
   arg, i = COMPGEN_SPEC.Parse(argv)
   status = 0
+
   if arg.A:
-    for func_name in funcs:
+    for func_name in sorted(funcs.iteritems()):
       print(func_name)
   else:
     util.warn('*** command without -A not implemented ***')
     status = 1
+
   return status
 
 
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -803,10 +803,14 @@ def CompGen(argv, funcs):
   status = 0
 
   if arg.A:
-    for func_name in sorted(funcs.iteritems()):
-      print(func_name)
+    if arg.A != 'function':
+        status = 1
+        raise args.UsageError('compgen: {}: invalid action name'.format(arg.A))
+    else:
+      for func_name,func_def in sorted(funcs.iteritems()):
+        print('{}'.format(func_name))
   else:
-    util.warn('*** command without -A not implemented ***')
+    util.warn('*** not implemented ***')
     status = 1
 
   return status
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -877,7 +877,7 @@ def Command(argv, funcs, path_val):
         # This is for -v, -V is more detailed.
         print(arg)
   else:
-    util.warn('*** command without -v not not implemented ***')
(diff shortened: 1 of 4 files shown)


core/builtin.py
```diff
@@ -794,6 +794,24 @@ def Shopt(argv, exec_opts):
...
+COMPGEN_SPEC = _Register('compgen')
+COMPGEN_SPEC.ShortFlag('-A', args.Str)
+
+
+def CompGen(argv, funcs):
+  arg, i = COMPGEN_SPEC.Parse(argv)
+  status = 0
+
+  if arg.A:
+    for func_name in sorted(funcs.iteritems()):
```
andychu: The value of arg.A should be tested for `function` here.  The other actions can print a warning for "not implemented".

core/cmd_exec.py
```diff
@@ -180,8 +180,8 @@ def _Complete(self, argv):
       util.error('Oil was not built with readline/completion.')
     return 0
 
-  def _CompGen(self, argv):
-    raise NotImplementedError
+  #def _CompGen(self, argv):
```
andychu: Can remove these lines

spec/compgen.test.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+#### Print function list
+env -i /bin/bash --norc --noprofile -c 'add () { expr 4 + 4; }; \
```
andychu: OSH isn't being tested here, because you're invoking bash.

You can probably do something with the value of `$SH`.  Or maybe we can add something to `sh_spec.py` that starts bash with a clean environment?

That is probably good for test isolation in general.

andychu:
I tried running `compgen -A function` inside the spec test environment with bash, and I get no functions -- it looks like it doesn't leak my `.bashrc` ?

Why do you have the `--norc --noprofile` stuff?

granttrec:
@andychu I just wanted to be sure, there could be other global definitions that could be sourced but if it not a problem then I will remove it

andychu:
OK, well you can just do the minimum to make it pass on your machine.  If there is some leakage of the environment, we can handle it separately.  It isn't testing OSH as written -- it's only testing bash.

andychu:
I looked at this when seeing the push notifications, but it looks like I caught it in an intermediate state.  Let me know when it's ready to look at.

Make sure to merge master so that the [osh2oil] change doesn't show up in the diffs.

granttrec:
@andychu I think it is ready to merge.

core/builtin.py
```diff
@@ -794,6 +794,28 @@ def Shopt(argv, exec_opts):
...
+
+  if arg.A:
+    if arg.A != 'function':
+        status = 1
+        raise args.UsageError('compgen: {}: invalid action name'.format(arg.A))
+    else:
+      for func_name,func_def in sorted(funcs.iteritems()):
+        print('{}'.format(func_name))
+  else:
+    util.warn('*** not implemented ***')
```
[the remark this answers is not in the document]
granttrec: I changed this, if you see it when you review again I may have messed up my push

granttrec:
@andychu Made the changes.

andychu:
OK, I pulled it, ran the tests, and manually poked at it.  Seems to work well.  Thanks!

PR #193, opened 2018-10-09
[osh] adding PS1 parsing and evaluation

okayzed:
This is a continuation of #183 pull request, but this time from a branch in oilshell (instead of raisjn)


osh/word_parse.py
```diff
@@ -269,6 +269,8 @@ def _ParseVarExpr(self, arg_lex_mode):
     if self.token_type == Id.Right_VarSub:
       return part  # no ops
 
+    log('%s', self.token_type)
```
okayzed: remove?

osh/lex.py
```diff
@@ -445,6 +445,15 @@ def IsKeyword(name):
   R(r'[^\\\0]+', Id.Char_Literals),
 ]
 
+# https://www.gnu.org/software/bash/manual/html_node/Controlling-the-Prompt.html#Controlling-the-Prompt
+PS1_DEF = _C_STRING_COMMON + [
```
okayzed: is this okay? you mentioned refactoring multiple groups into one common group

core/word_eval.py
```diff
@@ -552,7 +552,16 @@ def _EvalBracedVarSub(self, part, part_vals, quoted):
 
     elif part.suffix_op:
       op = part.suffix_op
-      if op.tag == suffix_op_e.StringUnary:
+      if op.tag == suffix_op_e.StringNullary:
+        if op.op_id == Id.VOp0_P:
+          # TODO: Use dependency injection
```
okayzed: what's this comment?

osh/word_parse.py
andychu: Hm not sure when that snuck in, I pushed a change to remove it.

osh/lex.py
andychu: Discussed on Zulip, we should not use _C_STRING_COMMON

core/word_eval.py
andychu: This is to possibly replace the global ui.PROMPT with dependency injection.  We can leave it in for this PR, but I probably want to refactor it later to follow the prevailing style.

andychu:
Merged!

Issue #182, opened 2018-09-19
`cd` changes PWD to entered path


okayzed:
```
osh$ pwd
/home/okay/tonka/apps/oil
osh$ cd ..
osh$ pwd
..
osh$
```

andychu:
Hm, I looked into this, but couldn't find an easy fix.

There is some interaction with `cd -P` that I don't quite understand.  (Somebody contributed `cd -P` and `pwd -P`; I never really use it.  Though it is implemented by all shells, even dash.)

Somewhat surprisingly, I think we might need to write some custom code to iterate over path segments and match what bash does:

https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#Bourne-Shell-Builtins

I looked at dash and it has some nontrivial code to update PWD that I don't quite understand either (in `src/cd.c`).

andychu:
Since this is a really localized problem that doesn't require understanding the rest of the code, I'm labelling it a good first issue.

The task is basically to fix this:

```
test/spec.sh builtins --range 6 --verbose
```

```
test/spec.sh builtins --range 6 --verbose
```

while not breaking the rest of the tests:


```
test/spec.sh builtins 
```

BatmanAoD:
This seems like it should be pretty easy, but I'm getting confused by how the Python path-manipulation functions behave around `..`. The current implementation (mine 😄) doesn't apply *any* path-normalization, so it should be as simple as applying `abspath` or `normpath`.

....but it looks like the current implementation *also* doesn't actually prepend the old working directory...? So I'm not sure if `cd` to a relative path *ever* resulted in a valid `PWD` value.

andychu:
@BatmanAoD I think our general strategy is to avoid `os.path` in Python.  I discovered that while implementing `readlink -f`.  We use `libc.realpath` (our own wrapper) instead of `os.path.realpath()`.

The Python functions don't really have anything to do with what shells like bash do.  So the best bet seems to be to write C in Python, e.g. use `os.symlink()` since that exists in libc.  `strace` helps.

okayzed:
the issue here is that we are using os.chdir() to move the CWD of osh around but we are not updating pwd to reflect that same dir. i think the line we want is `pwd = os.getcwd()`

okayzed:
notice the line: `pwd = libc.realpath(dest_dir) if arg.P else dest_dir` comes after the line `os.chdir(dest_dir)`

[...]

PR #195, opened 2018-10-16
[builtins] add relative path joining for cd

okayzed:
this is related to issue #182. After looking at line 215 of [cd.c in dash](https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/cd.c#n215), I 
decided to try their first piece of logic: join if relative, else replace and now tests pass. 

TESTPLAN: tested with test/spec.sh builtins, all pass except 'set umask symbolically'

### DIFF
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -544,7 +544,14 @@ def Cd(argv, mem, dir_stack):
   # '-L' is the default behavior; no need to check it
   # TODO: ensure that if multiple flags are provided, the *last* one overrides
   # the others
-  pwd = libc.realpath(dest_dir) if arg.P else dest_dir
+
+  # if we are cd'ing relatively: we do a join and fix any '..'s
+  # else: do a replace
+  if dest_dir[0] != '/':
+    pwd = os.path.normpath(os.path.join(pwd.s, dest_dir))
+  else:
+    pwd = libc.realpath(dest_dir) if arg.P else dest_dir
+
   state.SetGlobalString(mem, 'PWD', pwd)
 
   dir_stack.Reset()  # for pushd/popd/dirs


andychu:
Thanks for doing this!  I verified it fixes the test so I merged.

https://github.com/oilshell/oil/commit/426e5a9ddd03e2235263b73cdf1353bb182e0326

But I also came up with a test where it fails.  I think explicitly `help cd` might help us, I will try it:

```
        -L      force symbolic links to be followed: resolve symbolic links in
        DIR after processing instances of `..'
        -P      use the physical directory structure without following symbolic
        links: resolve symbolic links in DIR before processing instances
        of `..'
```

andychu:
OK I fixed the test I came up with, and I think the code is cleaner too:

https://github.com/oilshell/oil/commit/e91822260db66db5649603f5daec521fec1305f7

Also cleaned up the spec tests.  They were a bit confusing.

okayzed:
this doesn't make sense to me: 

```
 abspath = os.path.join(pwd.s, dest_dir)  # make it absolute, for cd ..
  if arg.P:
    # -P means resolve symbolic links, then process '..'
    pwd = libc.realpath(abspath)
  else:
    # -L means process '..' first.  This just does string manipulation.  (But
    # realpath afterward isn't correct?)
    pwd = os.path.normpath(abspath)```

why do you join the path unconditionally? in dash, they join the path if the leading character is not a '/'. are we relying on some behavior of normpath to fix that?

**Comment by andychu** [comment]
Python's `os.path.join()` already handles that.  If `dest_dir` begins with `/`, it gives you dest_dir back:

```
>>> os.path.join('/home/', '/')
'/'
```

**Comment by okayzed** [comment]
I was spending all my time reading about normpath (instead of join). Ok, this makes sense!

**Comment by andychu** [comment]
BTW the clue for me that something was off was that you had a branch for `args.P` in one case but not the other.  So I managed to construct that test case to show the bug, and then fixed it with this code.

**Comment by andychu** [comment]
I wouldn't be totally surprised if something is still wrong.  But can you make a spec test case out of that and show a difference between bash/dash, and osh?

Or just provide a standalone shell script I can run with both shells.

If so I should probably be able to fix it.

I don't think the last "this implies" statement is right but let's try it and see.

**Comment by okayzed** [comment]
```
diff --git a/spec/builtins.test.sh b/spec/builtins.test.sh
index 19615804..347ed014 100644
--- a/spec/builtins.test.sh
+++ b/spec/builtins.test.sh
@@ -47,6 +47,20 @@ dir-two
 dir-one
 ## END
 
+#### pwd after cd .. involving symlinks
+mkdir -p symlinktest
+cd symlinktest
+mkdir -p a/b/c
+mkdir -p a/b/d
+ln -s a/b/c c > /dev/null
+cd c
+cd ..
+ls # expecting a c/ (since we are in symlinktest) but osh gives c d (thinks we are in b/)
+## STDOUT:
+a
+c
+## END
+
 #### pwd -P
 tmp=$TMP/builtins-pwd-1
 mkdir -p $tmp
```

```
okay@chalk:~/tonka/apps/oil$ test/spec.sh builtins --verbose --range 7
builtins.test.sh
  7      50     pass    pass    pass    FAIL    pwd after cd .. involving symlinks
case: 7
[osh stdout] Expected 'a\nc\n', got 'c\nd\n'
osh stdout:
c
d

osh stderr:
ln: failed to create symbolic link 'c/c': File exists


3 passed, 0 ok, 0 known unimplemented, 0 known bugs, 1 failed, 26 skipped

FATAL: Got 1 failures (1 osh failures), but 2 are allowed
```

andychu:
Thanks a lot!  I just fixed this:

https://github.com/oilshell/oil/commit/867933d3ba754fc892ab1f5b0002085ef7c9cfe5

Keep the reports coming :)

PR #199, opened 2018-10-20
Use #!/usr/bin/env for shebangs

bb010g:
This allows easier development with tools like nix-shell that put alternative Pythons on the PATH.

### DIFF
--- a/Python-2.7.13/Tools/gdb/libpython.py
+++ b/Python-2.7.13/Tools/gdb/libpython.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 '''
 From gdb 7 onwards, gdb's build can be configured --with-python, allowing gdb
 to be extended with Python code e.g. for library-specific data visualizations,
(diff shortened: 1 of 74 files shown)


bb010g:
Apparently this option is new in [GNU coreutils 8.30](https://github.com/coreutils/coreutils/blob/571f63f5010b047a8a3250304053f05949faded4/NEWS#L79) (released in July), so that's a problem for Travis (running Trusty) and anyone else on boxes without that option. No current Ubuntu releases have coreutils 8.30 or higher. On the bright side, it looks like FreeBSD's env got this further back.

I'll leave closing this up to you, I guess. This could be resolved by finding a way to avoid `python -S` in shebangs.

andychu:
OK, I see the issue with -S.  If you're able to develop without changing this I'd be inclined to leave it.  It isn't ideal but I feel like it's never going to be right on all distros :-(

Linux Standard Base was supposed to say something about this but it doesn't seem like it was successful...

Issue #153, opened 2018-06-27
Find broken spec tests and fix them


andychu:
This is a "meta issue" for people looking for something to do on the project.

There are still a bunch of failing spec tests.  Almost all of them are things that should be fixed, although they vary in importance, and some of them may not be great first issues.

Example:

`$ test/spec.sh array`

has a bunch of failing test cases.  (Although I'm not saying this is the most important or interesting thing to fix!)

Note the `--verbose` and `--range` flags here:

https://github.com/oilshell/oil/wiki/Spec-Tests

They help narrow down what the exact failure is.

-----

If you find something interesting, ping this bug before starting so we're on the same page about it :)

andychu:
Down to 58 failures, most of them are not trivial, so closing this

PR #201, opened 2018-10-20
Implement array index arrays with ! (#153)

bb010g:
Resolves spec test "Retrieve all indices with !" in array and introduces spec test "Retrieve all sparse indices with !".


andychu:
Great thanks for doing this!  I merged and made some minor changes:

https://github.com/oilshell/oil/commit/d5150d7ab15e715e6c76ea4f426c30e2612ce864

The most salient one is changing it to 3 allowed failures rather than 4.  I do that so we can have a few red tests but not regress.

PR #205, opened 2018-10-30
[spec/builtins] Tuck a couple of stray directories under tmpdir.

gnprice:
This leaves `git status` clean after `test/spec.sh all`.


andychu:
Great thanks for fixing this!

(looks like the Travis failure is due to my recent change, will fix)

Issue #208, opened 2018-11-05
Run the bash_completion project


andychu:
See also #175  (Run git-completion.bash)

Things that don't work:

- [x] `printf -v var %q foo` -- should be changed to `var=${foo@Q}`
- [x] `${!x}` where `x=a[i]`.  PR by Greg Price
- [ ] `${!x#pattern}` -- var ref then prefix/suffix op.  Reported by Greg Price.
- [ ] Make `local "$2"` work and revert the `eval` patches (I think).  This is related to static vs. dynamic parsing, mentioned in #207.
  - This change is somewhat invasive because parsing assignments is somewhat subtle.  There is some inconsistency between `export` and `local/readonly/declare` that I would like to resolve.  Right now `export` is only dynamically parsed, but `local/readonly/declare` are statically parsed.

TODO: What else?

andychu:
https://github.com/oilshell/oil/commit/752111843df24bd354540049717ac36e809eb53f

gnprice:
I just made a quick skim through the patches in
https://github.com/scop/bash-completion/compare/master...oilshell:dev/osh-compat
to find things that that patched out that might be reasonable to implement.

* **`printf -v array[i]`** -- `printf -v "$2[i]"` appears in one place in the tip commit. Other `printf -v "$foo"` could also turn out to be array refs.
* **`$array`** meaning the same as `${array[0]}` -- This accounts for two of the patches, I think. Seems like basically any script relying on this behavior would be improved by a patch like these; but also shouldn't be too hard to implement, and my suspicion is a lot of scripts do this.
* **`local "$2"`** -- I don't totally understand what's happening here. There are a couple of iterations on this patch; seems like the first workaround caused some other issue?
* `BASH_VERSINFO` -- OK, patching this is definitely the only reasonable solution :slightly_smiling_face:

andychu:
Part of the goal of OSH is also to be saner, so if there is a trivial patch like $array -> ${array[0]} then I'm inclined to keep it.  Possibly later there could be a flag to enable "not sane" behavior if we run into some scripts that are really hard to change.

I think it's probably time to contact the maintainers of bash-completion and see how amenable they are to patches that will work in both bash and OSH.  This will obviously be a lot more convincing if we can run the whole project :)

Whether we have to maintain a fork will probably influence some decisions.  But since all the patches are one-liners, I think it will be easy to maintain a fork if we have to (hopefully not for too long).

[...]

PR #209, opened 2018-11-05
Crude first version of printf builtin

gnprice:
Passes one spec test!

And good enough to run bash_completion, and now without the message

    -vprintf: warning: ignoring excess arguments, starting with ‘quoted’

we were seeing, which is from /usr/bin/printf not understanding `-v`.

This shadows /usr/bin/printf, which has more functionality but also
differs from the shell builtin.  So some scripts that worked may now
fail, in which case the fix is to add the missing functionality here.

Fixes an item in #208.

### DIFF
--- a/core/builtin.py
+++ b/core/builtin.py
@@ -82,6 +82,7 @@ _SPECIAL_BUILTINS = {
 _NORMAL_BUILTINS = {
     "read": builtin_e.READ,
     "echo": builtin_e.ECHO,
+    "printf": builtin_e.PRINTF,
     "cd": builtin_e.CD,
     "pushd": builtin_e.PUSHD,
     "popd": builtin_e.POPD,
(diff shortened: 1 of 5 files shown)


core/builtin.py
```diff
@@ -235,6 +236,84 @@ def Echo(argv):
...
+    state.SetLocalString(mem, arg.v, result)
+  else:
+    sys.stdout.write(result)
+    sys.stdout.flush()
+  return 0
+
+
+def _PrintfQuote(val):
+  # TODO match specifics
+  return "'" + val.replace("'", "'\\''") + "'"
```
andychu: I don't think this is right, because `\` isn't special within single quotes.

core/builtin.py
```diff
@@ -235,6 +236,84 @@ def Echo(argv):
...
+  result = ''
+  f = 0
+  v = 0
+  # Loop invariant: vals[v:] and fmt[f:] remain, to accumulate into result.
+  while True:
+    f_next = fmt.find('%', f)
+    if f_next < 0:
+      f_next = len(fmt)
+
+    result += fmt[f:f_next]  # TODO backslash-escapes, at least \n
```
andychu: I think the requirement to lex `\` in addition to `%` means we should use re2c to lex.  See `osh/lex.py PS1_DEF` for a lexer definition for `$PS1`.  It's just a list of regex -> token mappings.

This change could be OK without `\` if it gets us past some practical problems... but long-term I see it being rewritten to use the regex-style lexers.  They are fast because it compiles to C code.

It also tends to help with corner cases, i.e. this style of string processing tends to end up with a lot of IndexError (handled or not).   Using a sentinel (i.e. NUL) tends to be cleaner in my experience (and that's the style that OSH uses throughout).

core/builtin.py
```diff
@@ -235,6 +236,84 @@ def Echo(argv):
...
+        break
+      else:
+        # (handy!) bash printf quirk: re-use fmt to consume remaining vals.
+        f = 0
+        continue
+
+    c = fmt[f+1]
+    if c == '%':
+      f += 2
+      result += '%'
```
andychu: Please use the `parts.append` and `''.join(parts)` idiom here.  I know CPython optimizes it not to be quadratic, but I already started writing the OVM interpreter loop and it won't be special-cased :)

core/builtin.py
```diff
@@ -235,6 +236,84 @@ def Echo(argv):
...
+      continue
+    elif c == 'd':
+      f += 2
+      val = vals[v] if v < len(vals) else '0'
+      v += 1
+      try:
+        num = int(val)
+      except ValueError:
+        # TODO should print message but carry on as if 0
+        util.error('printf: %s: invalid number' % (val,))
```
andychu: Can use `util.error(' %s ', val)` .  It does the % for you.

core/builtin.py
gnprice: Ah, it's a bit subtle, but I think it really is a correct implementation of "quote this for the shell" -- in fact it's the simplest one to write down explicitly that I know :-)

Consider a value like `FOO'BAR`. This function encodes that as `'FOO'\''BAR'` -- i.e. the single-quote gets encoded as a `\'` located *outside* any quoting, and everything else gets put into single quotes.

In other words, in that replacement string `"'\\''"` (i.e. the four characters `'\''`), the quote at the beginning *ends* a single-quoted string, and the one at the end starts a new one; they're there precisely to let the `\'` be written outside of any quotes.

Now, that said, this logic probably isn't the final version of this helper. It meets the spec in bash's `help printf`:
```
      %q	quote the argument in a way that can be reused as shell input
```
but it could e.g. skip quoting where none is needed. Bash's version does that, and also has a preference for backslashes:
```
bash$ printf %q$'\n' '"\$"'
\"\\\$\"

osh$ printf %q$'\n' '"\$"'
'"\$"'
```
I don't so much like the backslashes, but it's possible even those will turn out to be worth matching for compatibility.

core/builtin.py
gnprice: OK :)

core/builtin.py
gnprice: Oh neat! Definitely better, thanks.

core/builtin.py
gnprice: Cool, thanks for the pointer! Took a quick look, and yeah I definitely agree that that should be a better way for this to ultimately be implemented.

As it is, this code is basically C (in my head) translated to Python. I figured a real lexer would be better, but wasn't sure how to get my hands on that in the context of this codebase, and this seemed like a quicker way to get something working in order to see what bash-completion would need just next. C generated from regexes would be much better than Python translated from C :slightly_smiling_face:

andychu:
I don't see the comment change requested for `_PrintfQuote` ?  Maybe it got lost in the torrent of github e-mails :)  In any case I will merge and please add that in your WIP.

gnprice:
Ah, indeed I hadn't seen that :-) -- thanks for highlighting it. I'll send a trivial PR for it momentarily.

PR #213, opened 2018-11-16
Fix compiler warning; `pos` was unused

tekknolagi:


andychu:
Thanks for fixing this!

PR #217, opened 2018-12-07
Create _tmp dir automatically

### LINKED ISSUE #216: `build/dev.sh minimal` doesn't work without manually creating `_tmp`
I'm not sure why this is, but today when I tried to build Oil I got this error:

    build/dev.sh: line 80: _tmp/types_asdl.py: No such file or directory

I was able to fix this by manually creating `_tmp` *before* running the build; the instructions, however, seem to indicate that it can be created later.

BatmanAoD:
Fixes #216 (and removes a manual step from the "contribution" instructions)

### DIFF
--- a/build/dev.sh
+++ b/build/dev.sh
@@ -68,6 +68,8 @@ gen-help() {
   build/doc.sh osh-quick-ref
 }
 
+mkdir -p _tmp
+
 # Helper
 gen-asdl-py-pickle() {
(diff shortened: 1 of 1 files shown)


andychu:
Thanks for catching this and sorry about the wrong instructions!

I moved it inside the minimal function because I don't like side effects at the top level :)  If it needs to be at the top level, I would move it right above the last line which is `"$@"`.

Issue #92, opened 2018-03-25
Try out OSH on Void Linux scripts


andychu:
Thread that got me interested here: https://lobste.rs/s/8fd1pd/into_void_void_linux_description_review

Code:  https://github.com/voidlinux/xbps

Since OSH 0.4 can run Alpine, Aboriginal, and Debian scripts, there's a good chance it can run Void scripts, or is within spitting distance of it.

andychu:
Actually I'm not sure where the source is, xbps has mostly C code, and this one is obsolete:

https://github.com/voidlinux/xbps-src

maxice8:
xbps-src is only 800 lines but in the process of building it sources various files in `common/`

```
$ tokei common
-------------------------------------------------------------------------------
 Language            Files        Lines         Code     Comments       Blanks
-------------------------------------------------------------------------------
 Shell                 154         6945         5311          751          883
-------------------------------------------------------------------------------
 Total                 154         6945         5311          751          883
-------------------------------------------------------------------------------
```

olamiko:
Hi, I can possibly do this if you point out the right way to me

andychu:
Are you a Void Linux user?  Presumably it uses bash, and you can rebuild Void Linux packages with bash.

If that's the case, then in theory you should be able to build Void Linux packages with OSH, and that's what I'm looking to test.

Somebody already built OSH on Void Linux, and apparently it works.  (There is a conflict with the packaging system that prevents it from being packaged, but that shouldn't stop you from building and running it.)

andychu:
Do you have any shell scripts you've written yourself or that you use often and know well?  I would start by building OSH and running those scripts with it.  Then let me know if it succeeds or fails.

You can post a message on #oil-discuss on oilshell.zulipchat.com, or file a new issue if there is a clear bug.

Thanks!

andychu:
Closing this since I haven't encountered any Void Linux users and it seemed to be harder than I expected to bootstrap the distro.

PR #222, opened 2019-01-25
Create Dockerfile

afunsten:
get started faster with a docker sandbox
Tested OK with Docker Desktop Version 2.0.0.2 (30215) on MacOS 10.14.1 (18B75)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+#BUILD EXAMPLE: docker build . -t osh
+#Tested with Docker Desktop Version 2.0.0.2 (30215) on MacOS 10.14.1 (18B75) 
+
+#FROM alpine
+#RUN apk add --no-cache gcc musl-dev make file sed
+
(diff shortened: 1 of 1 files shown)


andychu:
Thanks for the PR.  This might be useful in the future, but I don't see the need to have it in the repo right now.  I would just keep it for private use, or you could post a message on oilshell.zulipchat.com and see if anyone else wants to use it.

If there are multiple contributors who want to use it then it would probably be worth it,.

FWIW the process may change in the next 1 to 3 months, and I probably won't update this since I don't use Docker myself.  It's likely that I'll port it to run on Python 2 and 3 simultaneously to take advantage of Python 3 type annotations.

Right now the most useful low-hanging fruit is to run OSH on existing shell scripts and tell me what happens (success or failure).  For example, issue #92 is along those lines.

afunsten:
No worries cancel it.

PR #229, opened 2019-02-10
add support for \W in $PS1

tyxieblub:
Hello,

Since I've seen no import of `os.path` in core, I switched pylib's os_path...

### DIFF
--- a/core/ui.py
+++ b/core/ui.py
@@ -13,7 +13,7 @@ from __future__ import print_function
 import posix
 import pwd
 import sys
-import os.path
+from pylib import os_path
 
 from asdl import const
 from asdl import format as fmt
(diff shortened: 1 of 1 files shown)


andychu:
Thanks a lot, and yes switching to `os_path` is right.  It's so we can minimize dependencies in Oil (discussed in [Dev Log #7](http://www.oilshell.org/blog/2018/11/15.html) )

Also, I added a quick spec test case for this.  Let me know if it doesn't make sense!

https://github.com/oilshell/oil/commit/33f826895d64837174da86c7cb1837e8f5157cdf

tyxieblub:
Thanks for the link! The test seems fine.

PR #233, opened 2019-02-15
$PS1: \h and \H 

tyxieblub:
The current `\h` should be `\H`, and `\h` should stop at the first '.' as described [here](https://www.gnu.org/software/bash/manual/html_node/Controlling-the-Prompt.html).


andychu:
Great thanks for fixing this!

PR #235, opened 2019-02-19
Fixes libedit based tab completion.

sethwoodworth:
According to this [article](https://pewpewthespells.com/blog/osx_readline.html) libedit based readline requires `readline.parse_and_bind("bind ^I rl_complete")` rather than `readline.parse_and_bind("tab: complete")`.  This is backed up by the note in the [python2.7 readline](https://docs.python.org/2.7/library/readline.html) docs.
AFAICT: otherwise completion is implemented the same except that using the readline's `set_completion_display_matches_hook` is going to be ignored on OSX.  OSH isn't using that hook, nor `Cmd` which seems to call it.

I can't test this until I get to my office tomorrow and try it on an OSX machine.

### DIFF
--- a/core/completion.py
+++ b/core/completion.py
@@ -1074,7 +1074,10 @@ class ReadlineCallback(object):
 if __name__ == '__main__':
   # This does basic filename copmletion
   import readline
-  readline.parse_and_bind('tab: complete')
+  if 'libedit' in readline.__doc__:
+    readline.parse_and_bind("bind ^I rl_complete")
+  else:
+    readline.parse_and_bind("tab: complete")
   while True:
     x = raw_input('$ ')
     print(x)


andychu:
OK interesting!  I think this is definitely part of the puzzle.

The other part to it is that I made some patches to readline.c that I think broke OS X.  There's all this __APPLE__ stuff and libedit stuff in that file, but I didn't test it.

https://github.com/oilshell/oil/blob/master/Python-2.7.13/Modules/readline.c

Another complication is that it's only possible to make a release tarball  on Linux now.  The tarball itself builds on OS X, but making the tarball is nontrivial because of the "OPy" stuff.

BUT I have this standalone demo that I'm going to integrate into OSH.  It might be worthwhile to try this demoish script on OS X, as it will be easier to run than a new build of OSH.

https://github.com/oilshell/blog-code/blob/master/interactive-shell/demoish.py#L979

-----

FWIW the tarball build is described at the end of this page:

I'm not sure anyone besides me has done it though.  There was also a pending PR to make the dev env work on OS X, but it was only partially done and didn't land.

andychu:
BTW I forked the Python readline wrapper to `line_input.c`.  So a lot of the stuff I said above should be less complicated (sorry it was a lot of internal detail).

It would be useful if you patch demoish.py and then let me know if it works on OS X!  It should be easy to get running since it's a single Python script with no dependencies.

I am going to integrate it into OSH.  It's a better UI, with some screencasts here:

https://asciinema.org/~andyc

https://github.com/oilshell/blog-code/blob/master/interactive-shell/demoish.py#L979

PR #242, opened 2019-03-04
Fix hostname test for $PS1

tyxieblub:
`\H` in bash is not equal to `hostname -f` as the latter one gets seems to get the FQDN from `getaddrinfo()` in the [Debian sources](https://salsa.debian.org/meskes/hostname/blob/master/hostname.c#L336) and not only `gethostname()`

So this is what I get on bash:

```bash
joris ~ $ cat /etc/hostname
xps15
joris ~ $ hostname -f
xps15.home
joris ~ $ PS1="\H ~ "
xps15 ~ PS1="\h ~ "
xps15 ~
```
I set `/etc/hostname` to xps15.local (and rebooted) and then there was a difference between `\h` and `\H`, as the first one was `xps15` and the second one was `xps15.local`. The [bash documentation on prompt](https://www.gnu.org/software/bash/manual/html_node/Controlling-the-Prompt.html#Controlling-the-Prompt) is quite explicit in this regard as there no mention of FQDN, so I didn't check how it is done there.

I took the liberty to remove the NOTE, feel free to add it again if I missed something.

### DIFF
--- a/spec/prompt.test.sh
+++ b/spec/prompt.test.sh
@@ -135,14 +135,11 @@ $
 
 #### hostname
 
-# NOTE: This test is not hermetic.  On my machine the short and long host name
-# are the same.
-
 PS1='\h '
 test "${PS1@P}" = "$(hostname -s) "  # short name
 echo status=$?
 PS1='\H '
-test "${PS1@P}" = "$(hostname -f) "  # fully qualified
+test "${PS1@P}" = "$(hostname) "  # fully qualified
 echo status=$?
 ## STDOUT:
 status=0


andychu:
Thanks!  I committed this here:

https://github.com/oilshell/oil/commit/5a3ce35d69eb190d42a0314793b1641af54bccdc

When I wrote "hermetic" I mean that whether the test passes or fails depends on what machine you run it on, which is bad!

Really we should do some kind of container-based testing ... then you could do the /etc/hostname change without booting your machine.

We already do a little bit with `test/alpine.sh`, where the tarball is built inside Alpine without readline and with musl libc on each release, but it could be extended to the spec tests.

tyxieblub:
Thanks for the answer!
I fiddled a bit with `test/alpine.sh` to run the spec tests but ran into some trouble:
  - Alpine ships with bash4.3 which, as far as I could see does not support the `@P` expansion syntax, which is somewhat the main component of `prompt.test.sh` which is the only one I played with.
  - Moreover, changing the hostname within a chroot will still tamper the host's hostname, so on the case in hand it would still be problematic.

andychu:
Yeah that's unfortunate about bash 4.3.  I imagine it's not too hard to build bash 4.4 or 5.0 in Alpine, or make an APKBUILD for them?  Or maybe there is a new Alpine image we should use.

OK I guess you can change /etc/hostname in the chroot, but you're saying the kernel will still be consulted?  If we really wanted we could do a `LD_PRELOAD` hook, but it's probably not worth it now.

I would like to eventually get all the tests running in some kind of container.  Or maybe getting them to run on Travis is a prerequisite (issue #81).  Travis can be annoying but it does have a pretty stable Ubuntu environment and has reliably run all the unit tests.

tyxieblub:
Indeed, bumping Alpine version from 3.6.2 to 3.9.2 is enough to get bash 4.4.x.

I looked a bit more about modifying the hostname in a chroot without tampering the host, and namespaces with [unshare -u](https://linux.die.net/man/1/unshare) seems to work fine; however as bash only sets the `$HOSTNAME` at startup, separating hostname test from other prompt tests was needed.

I have done a quick implementation in c092af3f7a479bf53ffbd2fe5017842e30dc3689 which gets the same results as `sudo ./test/spec.sh prompt` on my Debian host. I did not run any other spec tests than prompt yet. Do you think using unshare would fit the bill?

andychu:
OK this looks like an awesome start!  `unshare -u` looks like the right thing, although right now the host name isn't being changed?

Right now the silly thing about the spec tests is that I mostly run them on my machine.  Although over the summer I removed most of the quirks, so I believe they should run on any Ubuntu box.

For example, it used to depend on all the system bash; now it depends on `_tmp/spec-bin/{bash,dash,...}` which is better.

I like Alpine because it's smaller and simpler.  But Ubuntu has more cloud support, e.g. Travis gives you an Ubuntu environment but I don't think you can use Alpine or even upload your own custom image.  I could be wrong about that.

I think it would be useful to try running different spec tests with this harness and seeing if there are any differences with Ubuntu?  I guess that depends on you being able to run:

`test/spec.sh all`

OUTSIDE to compare what happens inside.

I do that all the time and it should pass almost all the time.  I tried to remove most of the quirks but there's no guarantee.

This is what it looks like for me now:

```
$ test/spec.sh all
...
test/spec-runner.sh test-to-html sh-options 
test/spec-runner.sh test-to-html smoke 
test/spec-runner.sh test-to-html special-vars 
test/spec-runner.sh test-to-html strict-options 
test/spec-runner.sh test-to-html subshell 
test/spec-runner.sh test-to-html tilde 
test/spec-runner.sh test-to-html type-compat 
test/spec-runner.sh test-to-html var-num 
test/spec-runner.sh test-to-html var-op-len 
test/spec-runner.sh test-to-html var-op-other 
test/spec-runner.sh test-to-html var-op-patsub 
test/spec-runner.sh test-to-html var-op-strip 
test/spec-runner.sh test-to-html var-op-test 
test/spec-runner.sh test-to-html var-ref 
test/spec-runner.sh test-to-html var-sub-quote 
test/spec-runner.sh test-to-html var-sub 
test/spec-runner.sh test-to-html word-eval 
test/spec-runner.sh test-to-html word-split 
test/spec-runner.sh test-to-html xtrace 
'_tmp/web' -> '/home/andy/git/oilshell/oil/web'

*** All 76 tests PASSED

Results: file:///home/andy/git/oilshell/oil/_tmp/spec/index.html

real    0m13.987s
user    0m59.517s
sys     0m17.083s
```

andychu:
To be more concise, if a spec test behaves differently in the Alpine chroot vs. Ubuntu (where it's tested), then that's probably something to be tightened up.

Although as mentioned, it's ALSO a worthy task to simply get spec tests running on an Ubuntu cloud machine (and the delta might be smaller than porting to Alpine).

I want other people to be develop Oil easily, and that requires running the spec tests.  But right now there's no "source of truth" for the pass/fail status of spec tests (other than my machine, although I do try to make it work in general).

I guess you can tell me if the spec test suite is a barrier to developing Oil :)  It's essential for correctness but it's a bit fiddly.

tyxieblub:
The hostname is set in `prompt.test.sh`  (line 141 & 143) just before spawning the new shell. 

I certainly don't get the same results with my Debian sid host (I get 16 test failed), so yeah having some results to base our expectations on is a good idea. I'll start looking at what fails on my host.

I don't think it is a barrier as for what I have done so far (messing with $PS1). Maybe it will be when I dig further :D

andychu:
Ah OK great, feel free to send a pull request.  If a given test works both outside a chroot on Ubuntu and inside a chroot, that is a big improvement!

andychu:
And updating Alpine is a good idea too!

PR #244, opened 2019-03-12
[cleanup] Fix location of script in README.md

rhencke:
scripts/count.sh was moved to metrics/source-code.sh previously in
commit ce71e0c3d216a4963a41f04de31b18e0be76500f

(also this project is seriously cool. :)


andychu:
Thank you, good call!

PR #251, opened 2019-03-18
add 'builtin' builtin

jyn514:
add spec tests for
- builtin x when a function called x exists
- builtin for an invalid command
- builtin without arguments

see https://github.com/oilshell/oil/issues/146 for details

### DIFF
--- a/osh/cmd_exec.py
+++ b/osh/cmd_exec.py
@@ -19,8 +19,6 @@ import posix
 import resource
 import time
 
-from sys import stderr
-
 from _devbuild.gen.id_kind_asdl import Id
 from _devbuild.gen.syntax_asdl import (
     command_e, redir_e, lhs_expr_e, lhs_expr_t, assign_op_e, word_part, token
--- a/osh/cmd_exec.py
+++ b/osh/cmd_exec.py
@@ -391,10 +391,7 @@ class Executor(object):
           if to_run == builtin_e.NONE:
               to_run = builtin.ResolveSpecial(argv[0])
           if to_run == builtin_e.NONE:
-              print("{}: {}builtin: {}: not a shell builtin".format(
-                  self.mem.GetArgNum(0).s,
-                  self.mem.line_num.s + ': ' if not self.exec_opts.interactive else '' ,
-                  argv[0]), file=stderr)
+              util.error("builtin: {}: not a shell builtin".format(argv[0]))
               status = 1
           else:
               status = self._RunBuiltin(to_run, argv, span_id)
--- a/osh/cmd_exec.py
+++ b/osh/cmd_exec.py
@@ -19,6 +19,8 @@ import posix
 import resource
 import time
 
+from sys import stderr
+
 from _devbuild.gen.id_kind_asdl import Id
 from _devbuild.gen.syntax_asdl import (
     command_e, redir_e, lhs_expr_e, lhs_expr_t, assign_op_e, word_part, token
@@ -378,6 +380,24 @@ class Executor(object):
 
     elif builtin_id == builtin_e.REPR:
       status = builtin.Repr(argv, self.mem)
+
+    elif builtin_id == builtin_e.BUILTIN:
+      if not argv:
+          status = 0
+      else:
+          to_run = builtin.Resolve(argv[0])
+          # check special builtins now, this is a strange thing to ask
+          # since they can't be overwritten but it is valid
+          if to_run == builtin_e.NONE:
+              to_run = builtin.ResolveSpecial(argv[0])
+          if to_run == builtin_e.NONE:
+              print("{}: {}builtin: {}: not a shell builtin".format(
+                  self.mem.GetArgNum(0).s,
+                  self.mem.line_num.s + ': ' if not self.exec_opts.interactive else '' ,
+                  argv[0]), file=stderr)
+              status = 1
+          else:
+              status = self._RunBuiltin(to_run, argv, span_id)
 
     else:
       raise AssertionError('Unhandled builtin: %s' % builtin_id)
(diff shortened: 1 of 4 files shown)


osh/cmd_exec.py
```diff
@@ -379,6 +381,24 @@ def _RunBuiltin(self, builtin_id, argv, span_id):
...
+      if not argv:
+          status = 0
+      else:
+          to_run = builtin.Resolve(argv[0])
+          # check special builtins now, this is a strange thing to ask
+          # since they can't be overwritten but it is valid
+          if to_run == builtin_e.NONE:
+              to_run = builtin.ResolveSpecial(argv[0])
+          if to_run == builtin_e.NONE:
+              print("{}: {}builtin: {}: not a shell builtin".format(
```
andychu: Please use `util.error()` here, following some of the examples in `osh/builtin.py`.  The error handling probably needs another pass, but having them all go through that function will make it easier to change at once.

Then you don't need to worry about the $0 issue... we can change that all at once if necessary.

I probably do want to change the interactive errors vs. batch errors as you've done, but it should be done globally.

spec/builtins2.test.sh
```diff
@@ -136,3 +136,38 @@ status=127
...
+/bin/mksh: <stdin>[1]: builtin: ls: not found
+## N-I dash status: 127
+## N-I dash STDERR:
+/bin/dash: 1: builtin: not found
+## OK osh STDERR:
+osh: 1: builtin: ls: not a shell builtin
+
+#### builtin no args
+builtin
+## status 0
```
andychu: this line looks like it's missing a colon

status: 0

spec/builtins2.test.sh
```diff
@@ -136,3 +136,38 @@ status=127
...
+## N-I dash STDOUT:
+hi
+## N-I dash STDERR:
+/bin/dash: 3: builtin: not found
+
+#### builtin ls not found
+builtin ls
+## status: 1
+## OK bash STDERR:
+/bin/bash: line 1: builtin: ls: not a shell builtin
```
andychu: It looks like there need to be a bunch of `## END` here:

Is the rest green or yellow for all shells?   Mine looks like this:

```
andy@lisa:~/git/oilshell/oil$ test/spec.sh builtins2
builtins2.test.sh
  0       3     ok      pass    pass    pass    pass    command -v
  1      31     BUG     BUG     ok      pass    pass    command -v with multiple names
  2      60     pass    pass    pass    pass    pass    command skips function lookup
  3      80     pass    pass    pass    N-I     pass    command command seq 3
  4      90     pass    pass    pass    N-I     pass    command command -v seq
  5      99     N-I     ok      BUG     BUG     pass    history usage

```

jyn514: Yes, they are all `pass`, `ok`, or `N-I`. I don't mind adding `## END` but all the tests were working without it.

Is there a way to have the tests output `pass` for an output specific to a shell? When I add the following I get a parse error:

```
 ## status: 1
-## OK bash STDERR:
+## PASS bash STDERR:
```

```
  File "test/sh_spec.py", line 996, in <module>
    sys.exit(main(sys.argv))
  File "test/sh_spec.py", line 915, in main
    cases = ParseTestFile(tokens)
  File "test/sh_spec.py", line 310, in ParseTestFile
    test_case = ParseTestCase(tokens)
  File "test/sh_spec.py", line 279, in ParseTestCase
    assert kind == TEST_CASE_BEGIN, (line_num, kind, item)  # Invariant
AssertionError: (157, 4, '/bin/bash: line 1: builtin: ls: not a shell builtin')
```

osh/cmd_exec.py
jyn514: I can look at implementing this in `util.error`, I think having line numbers in error messages is easy and makes scripting much friendlier for the user.

spec/builtins2.test.sh
andychu: It's not supposed to be green/ PASS unless the output is exactly what's expected.  It can be yellow as BUG or N-I though; that's the intended usage.

I don't think it could have ever worked without ## END ?  In any case it looks right now.

osh/cmd_exec.py
jyn514: I'm not sure this can be done from `util`, those are all static methods and don't have information about the runtime. We could move `error` into somewhere osh-specific where it has access to the running interpreter, but I imagine that would require a pretty substantial rewrite.

andychu:
Thanks for implementing this!

I merged and noticed that the way I handle `builtin command ls` is wrong... so I'm fixing that.

I also changed `"{}".format` to `util.error("%s", argv[0])`.  OSH uses the older style of formatting to be consistent everywhere, and because it's probably easier to port to C.  (in progress)

jyn514:
No problem! I really like the idea behind this project, bash globbing has bitten me quite a few times in the past. I might be adding a few more thing here and there :)

andychu:
OK great thanks for testing bats too.  I have a private wiki page with a bunch of things to test and I will surface that in an issue.

Yeah unfortunately globs are parsed and executed dynamically -- AFTER variable/command/arith substitution.  I think that is the major problem with it.

I want to fix that with the Oil language, but that's still pretty far off :-(

andychu:
@jyn514 FYI I removed some of the stderr assertions because they don't pass with the "hermetic" binaries made by `test/spec-bin.sh`.  [1]

https://github.com/oilshell/oil/commit/8fc93e43e9215425beab0c22047d9379830db42b

In an ideal world we would test all the errors, but I generally haven't been doing that, and I think the exact error text will be polished anyway.

[1] Briefly mentioned here: https://github.com/oilshell/oil/wiki/Spec-Tests

For example, some spec tests fail on bash 4.3 but pass on bash 4.4, so we have these hermetic binaries.

Issue #146, opened 2018-06-11
Implement remaining builtins


andychu:
This is a catch-all issue for people who are looking for small tasks.

There are various builtin commands that are unimplemented here, usually with a red X.  (note: Oil language stuff is mostly not ready to tackle; contact me if you're not sure what that means.)

http://www.oilshell.org/release/0.6.pre16/doc/osh-quick-ref.html

It's best if you have a bit of experience using the builtin!  I don't necessarily want to copy all of bash's builtins -- just the ones people use.

~~http://www.oilshell.org/release/0.5.alpha3/doc/osh-quick-ref.html~~

andychu:
Example: We found that `compgen -A functions` and `declare -F` are similar and missing by running this script:

https://github.com/dylanaraps/pure-bash-bible/blob/master/test.sh

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Tests

jyn514:
I would like to see `type` show the definition of a function if possible.
osh:
```$ type cd
cd is a function
```

bash:
```
$ type cd
cd is a function
cd () 
{ 
    if [ -z "$1" ]; then
        builtin cd && ls;
    else
        builtin cd "$1" && shift 2> /dev/null && ls "$@";
    fi
}

```

jyn514:
I'm not sure how to work on this. I tracked it down to `builtin_e` in `osh/builtin.py` but it looks like that gets imported from `_devbuild.gen.runteim_asdl` and I don't know where that is or how to edit it.

jyn514:
The `builtin` builtin is not supported, I can work on this. The other bit looks hard to implement.

andychu:
OK great.  I implemented `command` awhile ago, but I didn't see anyone using `builtin`.  It looks like your function is using it, so OSH should have it.

The way I would tackle this is to copy your function into a case here:

https://github.com/oilshell/oil/blob/master/spec/builtins2.test.sh

Notes about spec tests:

https://github.com/oilshell/oil/wiki/Spec-Tests

Then run something like:

`test/spec.sh builtins2 --verbose --range 6`

And make sure the test is failing.

Then this change may give a clue (although I'm not entirely sure)

https://github.com/oilshell/oil/commit/874d1f8d166647a52b79ebececd430c9e60e28b4

jyn514:
[...]

Issue #232, opened 2019-02-13
Implement globstar


andychu:
This is from bash 4.x, borrowed from zsh.

- I will be surprised if nobody brings it up
- It's a reason we may want our own glob engine, in addition to extended globs.

Although, I would like to avoid writing our own glob engine if possible.  Oil could have a strict front end to validate the language, but delegate to `libc` at runtime if possible.

So I wonder if globstar can be implemented using fnmatch and some logic?  Ditto for extended glob.

glyh:
How a about a more generalized version, like regex?

andychu:
I think we should start with globstar first, then talk about other things later

extended globs also give globs the power of regexes, so probably not.  They are already implemented in OSH

We need  `configure`  detection for extended globs in glibc though - @Melkor333  had a related PR  with Android pwent() and so forth

PR #253, opened 2019-03-19
Begin implementing $RANDOM

sethwoodworth:
As requested in #146, I tried my hand at implementing $RANDOM.  Tonight I was able to figure out the asdl generator configuration, where the lexer connects regexes to the Id kinds, the implementation of special cases in `_EvalWordPart` and how to use `pdb` and `util.log` functions to interact with `ovm`.
Clearly, this will need a lot of work.  This still needs, at a minimum:
+ to return a 32b random integer
+ allow setting the variable to seed the psudo-random number generator
+ a better name for the `VSub` kind than `DollarSpecialName`
+ a `C` regex that captures `RANDOM` and `SECONDS` 
+ implement bracket expansion in addition to `DollarName` expansion
+ tests and specs
But I thought it would be worthwhile to start the PR to begin getting feedback.

A few questions that I still have:
+ Am I returning the correct type by using `value.Str`?
For the `_EvalWordPart` function, would you be open to a dict of cases where the keys are `part.tag`?
```
part_tag_actions = {
    word_part_e.LiteralPart: lambda part, part_vals, quoted:  part_vals.append(part_value.String(part.token.val, not quoted),
    word_part_e.EscapedLiteralPart: None,
    word_part_e.SingleQuotedPart: None,
    word_part_e.DoubleQuotedPart: None,
    word_part_e.CommandSubPart: None,
}
value_function = part_tag_actions[part.tag](part, part_vals, quoted)
```
Alternately, 
```

def eval_word_part_LiteralPart(part, part_vals, quoted):
    part_vals.append(part_value.String(part.token.val, not quoted)
part_tag_actions {
    word_part_e.LiteralPart: eval_word_part_LiteralPart,
}
word_part_e.LiteralPart(part, part_vals, quoted)
```
rather than:
```
if part.tag == word_part_e.ArrayLiteralPart:
...
elif part.tag == word_part_e.LiteralPart:
...
elif part.tag == word_part_e.EscapedLiteralPart:
...
elif part.tag == word_part_e.SingleQuotedPart:
...
elif part.tag == word_part_e.DoubleQuotedPart:
...
elif part.tag == word_part_e.CommandSubPart:
...
```
I realize you're likely doing speed optimization here that I am not immediately understanding, and that a function call is costly. But if you're amenable to the proposal at all, I would be happy to send you a PR with a full example that we could discuss.

Thanks again for this wonderful project.  I look forward to using and contributing to `ovm` and `oil`.

### DIFF
--- a/core/id_kind.py
+++ b/core/id_kind.py
@@ -281,6 +281,7 @@ def AddKinds(spec):
   # But then you have AS_STAR, or Id.Arith_Star maybe
 
   spec.AddKind('VSub', [
+      'DollarSpecialName', # $RANDOM or ${RANDOM}
       'DollarName',  # $foo
       'Name',        # 'foo' in ${foo}
       'Number',      # $0 .. $9
--- a/frontend/lex.py
+++ b/frontend/lex.py
@@ -133,6 +133,7 @@ VAR_NAME_RE = r'[a-zA-Z_][a-zA-Z0-9_]*'
 # All Kind.VSub
 _VARS = [
   # Unbraced variables
+  C(r'$RANDOM', Id.VSub_DollarSpecialName),
   R(r'\$' + VAR_NAME_RE, Id.VSub_DollarName),
   R(r'\$[0-9]', Id.VSub_Number),
   C(r'$!', Id.VSub_Bang),
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -828,7 +828,13 @@ class _WordEvaluator(object):
     elif part.tag == word_part_e.SimpleVarSub:
       maybe_decay_array = False
       # 1. Evaluate from (var_name, var_num, token) -> defined, value
-      if part.token.id == Id.VSub_DollarName:
+      if part.token.id == Id.VSub_DollarSpecialName:
+          SPECIAL_TOKENS = ['RANDOM', 'SECONDS']
+          if not part.token.val[1:] in SPECIAL_TOKENS:
+              raise AssertionError("DollarSpecialName lookup failed")
+          var_name = part.token.val[1:]
+          val = value.Str(str(42))
+      elif part.token.id == Id.VSub_DollarName:
         var_name = part.token.val[1:]
         val = self.mem.GetVar(var_name)
       elif part.token.id == Id.VSub_Number:


andychu:
Thanks for looking at this!

- Yes, numbers should be `value.Str()`, because shell doesn't really have a number type.  I hope to add a proper number type in the Oil language (still far in the future).
- Please just use the if/elif style for now.  The code is definitely too slow, but a dict won't speed it up enough.  I'm working right now on a translation to C++, and if/elif is better for that.  (Even though I've been using Python for 15+ years, this project is the first one where I've *really* wanted a switch statement!)
- I don't think you should use the lexer to recognize `$RANDOM`.  I thought about this, because it is indeed faster, but the problem is statements like this:

    echo $RANDOM ${RANDOM} ${RANDOM/42/-}

Instead there is yet another if/elif chain in `GetVar()` in `osh/state.py`.  This is not ideal, but I think it makes sense to focus first on the functionality, and then change that style later if necessary.

Just curious: do any shells besides bash implement $RANDOM?  I recall a blog post about random in shell ... let me see if I can dig it up.

andychu:
Oh yeah, this is the post I remember.

https://nullprogram.com/blog/2018/12/25/

Honestly I felt it was overkill / overthinking it.  My guess it that $RANDOM can just use libc `rand()` and `srand()`, so we don't have to bother writing and maintaining our own random number generator.  We can "punt" it to libc.

If a user wants good randomness, they should consult the **kernel** via /dev/{random,urandom} or Linux's new `getrandom()`.

I don't think any shells will give secure randomness, and I don't think any shell scripts should rely on secure randomness!  It is sort of a low quality best-effort thing.

sethwoodworth:
If you'd prefer, I'll implement SECONDS instead. This is more an exercise in understanding the codebase than a feature I currently need.

andychu:
Well I think it's important to prioritize by features that real shell programs use.  There's a very long tail of bash features that we could get caught up in.  I won't be surprised if people use RANDOM or SECONDS, but I'd like to know where!

Running programs in the wild usually reveals bugs pretty quickly.  I put some suggestions here:

https://github.com/oilshell/oil/wiki/Shell-Programs-That-Run-Under-OSH

Although it is getting to the point where many common scripts run.

What happens when you try to `source ~/.bashrc` in OSH?  (Maybe you hit https://github.com/oilshell/oil/issues/236 like I do.)

My main goal right now is for people to use OSH.  I think that is mostly blocked on interactive features, but I'm sure there are some language/builtin issues left as well.

andychu:
In other words I guess I'm curious what is preventing people from switching to OSH right now.  My guess is:

1) Some interactive bugs
2) OSH is slower but doesn't have that many more features.  Is there a killer feature that would make you switch?  (I can think of a few, but I'm curious what other people say.  Some of them will also take a long time to implement unfortunately.)

jyn514:
This is what I get when I `source ~/.bashrc`:

```
osh$ . /home/joshua/.bashrc 
osh 'shopt' usage error: Invalid option 'checkwinsize'
osh 'shopt' usage error: Invalid option 'globstar'
osh 'shopt' usage error: Invalid option 'checkhash'
Line 330 of '/usr/share/bash-completion/bash_completion'
      local "$2" "$3" "$4" && _upvars -a${#words[@]} $2 "${words[@]}" \
            ^
Variable names must be unquoted constants
```

I think bash_completion is a known issue, but the `shopt`s might be new. From the manpage:

```
              checkhash
                      If  set,  bash  checks that a command found in the hash table exists
                      before trying to execute it.  If a hashed command no longer  exists,
                      a normal path search is performed.
...
              checkwinsize
                      If  set, bash checks the window size after each command and, if nec‐
                      essary, updates the values of LINES and COLUMNS.
...
              globstar
                      If  set,  the  pattern  ** used in a pathname expansion context will
                      match all files and zero or more directories and subdirectories.  If
                      the  pattern is followed by a /, only directories and subdirectories
                      match.

```

sethwoodworth:
In that case, I'll close this PR and look into shopts

andychu:
@sethwoodworth Thanks, I've already stubbed out a bunch of 'shopt' as no-ops, and that might be OK here.

- checkhash: doesn't apply to OSH, maybe it's OK to stub out.
- checkwinsize: I think this could be useful, but I wonder who uses it?
- globstar: This is #232, probably shouldn't be stubbed out because that would be misleading.  I expect people will want this, but it's an open question on how to do it.  We don't implement our own glob parser, so I'm not sure how to implement `**`.  Some research in this area would be valuable.

Honestly I have a hard time understanding how this feature even works... I put some experiments in `demo/globstar.sh` if anyone wants to play with them.

andychu:
And yes @jyn514 the "unquoted constants" error is because we don't have dynamic assignment, https://github.com/oilshell/oil/issues/236 .  That is not hard, but it's probably going to be a large-ish change.

jyn514:
I think implementing `**` when it's the only part of a path fragment is fine, I don't know how much it's used besides that. So we stay consistent with bash's treatment of `$root/**` and `$root/**/` but ignore all the other cases.

BTW, `$root/**.py` doesn't match *anything*.

Aversiste:
@andychu I use RANDOM on my [rr](https://github.com/Aversiste/rr) project but only if no better tool is available for obtaining random numbers. It is neither safe to use nor serious.

Aversiste:
I spotted one legitimate usage of RANDOM: ensuring all cron jobs are executed daily but not at the exact same time. An example of such use can be seen in [acme-client(1)](https://man.openbsd.org/acme-client#EXAMPLES) man page.

PR #256, opened 2019-03-19
Fix readline problems

jyn514:
Closes https://github.com/oilshell/oil/issues/252

### DIFF
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -30,6 +30,20 @@ PROMPT_RESET = '\x01%s\x02' % _RESET
 PROMPT_UNDERLINE = '\x01%s\x02' % _UNDERLINE
 PROMPT_REVERSE = '\x01%s\x02' % _REVERSE
 
+def _PromptLen(prompt_str):
+  """Ignore all characters between \x01 and \x02."""
+  escaped = False
+  length = 0
+  for c in prompt_str:
+    if c == '\x01':
+      escaped = True
+    elif c == '\x02':
+      escaped = False
+    elif not escaped:
+      length += 1
+  return length
+
+
 class PromptState(object):
   """For the InteractiveLineReader to communicate with the Display callback."""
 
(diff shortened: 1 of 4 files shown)


andychu:
Two things:

- We can't use `re` directly in OSH; we have to port regexes to `fastlex`.  In `frontend/match.py`, there is a Python version and a "fast" version of every function that uses regex.  This is because OSH no longer depends on the Python regex module.
- So as a result I think it's better to NOT try to detect ANSI escapes in the prompt.  Instead we just iterate through the characters and count them, except we look for `\[` and \]` and do NOT count anything inside them.

ANSI escapes are expected to be inside `\[` and `\]`, so we don't have to detect them ourselves.  Bash doesn't detect them.

jyn514:
Sounds good. The osh code should only affect completion in any case.

core/comp_ui.py
```diff
@@ -30,18 +30,29 @@
...
 
-
 class PromptState(object):
   """For the InteractiveLineReader to communicate with the Display callback."""
 
   def __init__(self):
     self.last_prompt_str = None
     self.last_prompt_len = -1
 
+  def _PromptLen(self, prompt_str):
```
andychu: Please make this a free function, since it doesn't use `self`.  (pet peeve of mine)

Also even though it's simple, I would write a few unit tests ... i.e. testing the case where they're adjacent, mismatched, etc.

The `elif escaped` makes me skip a beat although it might happen to be correct.


As far as unmatched \[ \], that's a good point.  Maybe in prompt.py we should render an error string like we do when `\W` can't be determined.

osh/prompt.py
```diff
@@ -117,9 +117,11 @@ def _ReplaceBackslashCodes(self, tokens):
 
       elif id_ == Id.PS_LBrace:
         non_printing += 1
+        ret.append('\x01')
 
       elif id_ == Id.PS_RBrace:
         non_printing -= 1
```
andychu: Yeah I think it makes sense to check if non_printing is not 0 at the end of the function and render an error?

What does bash do?  (If I had to guess, it causes problems without producing a useful error...)

jyn514: This is interesting: if PS1 is `> \[\[ >`, then when you go off the line, *it wraps properly*, but then a character later goes to the start of the second line and starts overwriting the line. I would have expected one or the other, seeing both is strange.

jyn514: So you think we should catch the error here instead of in `comp_ui.py`? That sounds like a plan, I can move my invalid inputs tests over to that test file

andychu: Yeah I think validating in prompt.py makes sense.  Then just return an error string.  And then the length calculation will work fine on the error string -- it won't have unbalanced `\x01` and `\x02`, because you will have caught the unbalanced `\[` and `\]`.

jyn514: I added this, it will show `<error parsing prompt>` but still let you use the shell without throwing an exception.

image

andychu:
@jyn514 Thanks for doing this!  I merged and made a followup commit:

https://github.com/oilshell/oil/commit/4c28743be8328c4e0a7c287c3ee9bdaabb163a93

The spec tests caught a subtle issue!  I had to remove the special codes in `${x@P}`.  readline normally removes them, so `@P` in OSH has to remove them too.

Issue #259, opened 2019-03-20
type builtin doesn't handle -p/P/a


pepa65:
In bash I often use the pattern `type -p $command` to find the path of a command if it exists; this doesn't work in OSH.

jyn514:
I wrote a patch for this, I haven't added unit tests yet

andychu:
OK awesome, the way I find the tests is

```
$ grep -w type spec/*.sh
spec/12-command-sub.sh:  type sleep  # external command
spec/builtin-bash.test.sh:# Bash implements type -t.
spec/builtin-bash.test.sh:#### type -t builtin -> function
spec/builtin-bash.test.sh:type -t f
spec/builtin-bash.test.sh:#### type -t builtin -> builtin
...
```

So it looks like they should go in builtin-bash.

jyn514:
from `help type` in bash:

```
type: type [-afptP] name [name ...]
    Display information about command type.
    
    For each NAME, indicate how it would be interpreted if used as a
    command name.
    
    Options:
      -a        display all locations containing an executable named NAME;
                includes aliases, builtins, and functions, if and only if
                the `-p' option is not also used
      -f        suppress shell function lookup
      -P        force a PATH search for each NAME, even if it is an alias,
                builtin, or function, and returns the name of the disk file
                that would be executed
      -p        returns either the name of the disk file that would be executed,
                or nothing if `type -t NAME' would not return `file'
      -t        output a single word which is one of `alias', `keyword',
                `function', `builtin', `file' or `', if NAME is an alias,
                shell reserved word, shell function, shell builtin, disk file,
                or not found, respectively
    
    Arguments:
      NAME      Command name to be interpreted.
    
    Exit Status:
    Returns success if all of the NAMEs are found; fails if any are not found.
```

jyn514:
@andychu It looks like there's no mechanism to do a file lookup without going through `_ResolveNames` and check if it's a function first, I'm going to factor it out into a new function.

andychu:
OK `-p` will be out with the next release!

andychu:
This is released.  Latest version of OSH: http://www.oilshell.org/release/0.6.pre20/

I will leave this open until the other flags are implemented.  It's not on my near-term TODO list now so any contributions are appreciated :)

andychu:
`-a` is not implemented, although I'm not sure if scripts use it

jyn514:
`-a` is useful if you have a function that uses `command`:
```
$ type -a kubectl
[...]

PR #267, opened 2019-03-22
add -p option to type builtin

jyn514:
addresses https://github.com/oilshell/oil/issues/259

if found:
  if file, print the filename
  otherwise do nothing

otherwise, set status to 1

add unit tests for type -p builtin

### DIFF
--- a/spec/builtin-bash.test.sh
+++ b/spec/builtin-bash.test.sh
@@ -84,3 +84,21 @@ type -P mv tar grep cd builtin command type
 /bin/tar
 /bin/grep
 ## END
+
+#### type -f builtin -> not found
+type -f FOO BAR NOT FOUND
+## status: 1
+
+#### type -f builtin -> function and file exists
+mv () { ls; }
+tar () { ls; }
+grep () { ls; }
+type -f mv tar grep
+## STDOUT:
+/bin/mv is a file
+/bin/tar is a file
+/bin/grep is a file
+## OK bash STDOUT:
+mv is /bin/mv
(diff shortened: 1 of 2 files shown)


spec/builtin-bash.test.sh
```diff
@@ -37,3 +37,16 @@ echo "help=$?"
 cat $TMP/err.txt | grep -i 'no help topics' >/dev/null
 echo "grep=$?"
 ## stdout-json: "help=1\ngrep=0\n"
+
+#### type -p builtin -> file
+type -p mv tar grep
+## stdout-json: "/bin/mv\n/bin/tar\n/bin/grep\n"
```
andychu: Please use the newer format here:

```
## STDOUT:
/bin/mv
...
## END
```

For awhile, we didn't have anything but the JSON format, but multiline is more readable for many cases.  I tend to change those when touching older files.

andychu:
Oh sorry I didn't realize this was still open.  It looks ready to merge?

I think I only get a notification when you respond with a comment, so the commits where you addressed the last comment are silent.  Sorry about that!

jyn514:
I think the only thing I haven't address is you wanted the initial commit message changed, I'd have to figure out how `git rebase` works. Other than that I think it's ready to merge, not sure how well it can be extended to `-a` and I haven't tested how the options interact though.

andychu:
OK great I merged it and ran the tests on my machine.  Looks good, thanks!

I forgot to add #259 on the edited change description; oh well.

PR #271, opened 2019-03-25
print newline after process is interrupted

jyn514:
closes https://github.com/oilshell/oil/issues/258

this turned out to be simpler than I expected, since posix has a way to get how a process exited from its status.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -864,7 +864,8 @@ class Waiter(object):
 
     # TODO: change status in more cases.
     if posix.WIFSIGNALED(status):
-      pass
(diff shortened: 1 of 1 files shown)


jyn514:
Not sure how to test this, the normal testing framework doesn't allow you to interrupt a command that I know of. I've been testing locally with

```
$ cat
^C
```

andychu:
Thanks!  Yeah this kind of thing is hard to test, so simply describing how you tested it is useful.

Actually it could be in the commit description, but it's too late ... no big deal.

PR #281, opened 2019-04-21
don't crash if we can't fork a process

jyn514:
Addresses https://github.com/oilshell/oil/issues/279

Retries up to MAX_TRIES times to fork the processes. If we still can't fork, print an error and continue.

TODO:
- set status to an error code
- add spec test
- don't throw recursive errors. if we have a pipeline, then multiple children of osh will try to fork, resulting in exponentionally more errors. example:

```
$ prlimit --nproc=$(($(ps -u $USER -L | wc -l) + 1)) --pid $$
$ ls | ls | ls | ls
osh: fork: retry: Resource temporarily unavailable
osh: fork: retry: Resource temporarily unavailable
osh: fork: retry: Resource temporarily unavailable
osh: fork: retry: Resource temporarily unavailable
osh: fork: retry: Resource temporarily unavailable
```

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -16,6 +16,7 @@ import posix
 import pwd
 import signal
 import sys
+from time import sleep
 
 from _devbuild.gen.id_kind_asdl import Id
 from _devbuild.gen.runtime_asdl import redirect_e, process_state_e
@@ -25,6 +26,7 @@ from pylib import os_
 
 from typing import Optional
 
+MAX_TRIES = 10
 
 def GetHomeDir():
   # type: () -> Optional[str]
@@ -550,7 +552,7 @@ class Process(Job):
     # a single process group.
     #
     # - posix.setpgid()
-    # - posix.setpgrp() 
+    # - posix.setpgrp()
     # - posix.tcsetpgrp()
     #
     # NOTE: posix.setsid() isn't called by the shell; it's should be called by the
@@ -558,10 +560,19 @@ class Process(Job):
     #
     # The whole job control mechanism is complicated and hacky.
 
-    pid = posix.fork()
-    if pid < 0:
-      # When does this happen?
-      raise RuntimeError('Fatal error in posix.fork()')
+    pid = None
+    for attempt in range(MAX_TRIES):
+        try:
+            pid = posix.fork()
+            break
+        except OSError:
+            print("osh: fork: retry: Resource temporarily unavailable",
+                  file=sys.stderr)
+            sleep(1)
+
+    if pid is None:
+        print("osh: fork: could not fork child process", file=sys.stderr)
+        return 0
 
     elif pid == 0:  # child
       # Respond to Ctrl-\ (core dump)


andychu:
The `prlimit` invocation is really useful.  I got bash in some odd state where it is retrying periodically?  It seemed a bit flaky, I'm not sure why.

But dash seems to behave more predictably.

```
$ prlimit --nproc=$(( $(ps -u $USER -L | wc -l)  )) --pid $$
$ echo hi
hi
$ echo $?        
0
$ echo $(echo hi)
dash: 22: Cannot fork
$ echo $?
2
```

I'm not sure there's a good reason to retry?

Also, I think returning a PID of 0 is bad.  I think we might want to catch this exception in either the several invocations of `p.Start()` in `Executor` in `osh/cmd_exec.py`, or maybe just in `main_loop.py` at the top level.

I think that will make it easier to set the status as well.

We might want to turn `OSError` with this `errno` into some other exception, like CannotFork or something.

I think checking in a small script with the `prlimit` command and some comments in `demo/` or `test/` is pretty useful.

One confusing thing about exceptions in Oil is that they can be thrown **after** fork as well... so we probably would have to catch it in two places, which is not a problem.  It's the `SubProgramThunk` in `core/process.py`, which is what executes the AST in a forked process.  So if we can't fork there, then we can print an error message.  The exit code probably needs testing, I don't remember how that works ... :-/

andychu:
Hm yeah on second thought, it might be OK to do something really blunt and catch **all** `OSError` and `IOError` in `core/main_loop.py` and in `SubProgramThunk`.

I mean the shell should never crash, so that is better than nothing.

Let me experiment with it and see if the behavior is decent... It seems blunt but this is what exceptions were designed for :)

jyn514:
The use case for retrying is when we try to spawn multiple processes. If we have a pipeline, then we can run the first and retry on the rest, which will start working when the first process exits.

andychu:
I think that only works in very limited cases.  `ls | ls` probably works, but the more common case is that you are transfering more than `PIPE_SiZE` bytes across the pipe, and the first process can't complete until the second one starts reading from the pipe.

That is, the two processes must run concurrently for the shell to have fulfilled its "contract".

All things being equal I prefer simpler code ... in my mind retries are fuzzy, non-deterministic, harder to debug, etc. and thus less simple.  But I could be convinced if a lot of shells do things this way for some important reason.

I haven't encountered this error myself in practice, but it's probably an issue in certain container environments.  The shell definitely shouldn't crash so regardless of whether we try, returning an error is a huge improvement.

andychu:
I just landed a very simple fix for this:

https://github.com/oilshell/oil/commit/bd969417a29d1f46202006355a90869b6f52ee44

I tested dash, bash, and zsh interactively with `prlimit --nproc=1 --pid=$$`.  bash and zsh both fail deterministically for me, but bash does the retrying, which actually seems to make the shell unresponsive?

This might not be the last word on this -- let me know if there are more corner cases this doesn't handle, or if it behaves oddly in any situations.  Thanks for testing the corner cases!

jyn514:
osh still crashes if PS1 has commands in it.

```
(osh) /.../src/oil $ echo $PS1
(osh) $(display_pwd) $
(osh) /.../src/oil $ prlimit --nproc=1 --pid=$$
  File "/usr/local/src/oil/bin/oil.py", line 778, in <module>
    main(sys.argv)
  File "/usr/local/src/oil/bin/oil.py", line 747, in main
    sys.exit(AppBundleMain(argv))
  File "/usr/local/src/oil/bin/oil.py", line 721, in AppBundleMain
    status = ShellMain('osh', argv0, main_argv, login_shell)
  File "/usr/local/src/oil/bin/oil.py", line 534, in ShellMain
    return main_loop.Interactive(opts, ex, c_parser, display, arena)
  File "/usr/local/src/oil/core/main_loop.py", line 102, in Interactive
    c_parser.ResetInputObjects()
  File "/usr/local/src/oil/osh/cmd_parse.py", line 436, in ResetInputObjects
    self.line_reader.Reset()
  File "/usr/local/src/oil/frontend/reader.py", line 110, in Reset
    self.prompt_str = self.prompt_ev.FirstPromptEvaluator()
  File "/usr/local/src/oil/osh/prompt.py", line 209, in FirstPromptEvaluator
    return self.EvalPrompt(val)
  File "/usr/local/src/oil/osh/prompt.py", line 203, in EvalPrompt
    val2 = self.ex.word_ev.EvalForPlugin(ps1_word)
  File "/usr/local/src/oil/osh/word_eval.py", line 970, in EvalForPlugin
    val = self.EvalWordToString(w)
  File "/usr/local/src/oil/osh/word_eval.py", line 928, in EvalWordToString
    self._EvalWordPart(p, part_vals, quoted=False)
  File "/usr/local/src/oil/osh/word_eval.py", line 821, in _EvalWordPart
    v = self._EvalCommandSub(part.command_list, quoted)
  File "/usr/local/src/oil/osh/word_eval.py", line 1139, in _EvalCommandSub
    stdout = self.ex.RunCommandSub(node)
  File "/usr/local/src/oil/osh/cmd_exec.py", line 1274, in RunCommandSub
    pid = p.Start()
  File "/usr/local/src/oil/core/process.py", line 561, in Start
    pid = posix.fork()
OSError: [Errno 11] Resource temporarily unavailable
```

andychu:
This was fixed awhile ago so I'm closing.  Thanks for filing the bug and keep them coming!

PR #289, opened 2019-04-25
build/compile.sh: fix bug in $READLINE_DIR flags

ariabuckles:
Hi! I was trying to compile on Mac OS X with readline and ran into some small trouble with
build/compile.sh. Making this change solved it. I haven't tested this on linux as the change is rather trivial, but I can if that would be helpful.

Summary:

If $READLINE_DIR is specified, $readline_flags becomes something like:

`"-L /usr/local/opt/readline/lib -I /usr/local/opt/readline/include-l readline -D HAVE_READLINE"`

And the lack of space between the include and the `-l` trips up clang on mac.

Changes:

This commit adds an extra space to separate the `-I` and `-l` flags to `cc`.

Test plan:

Was able to build osh on Mac OS X with
`./configure --prefix ~ --with-readline --readline /usr/local/opt/readline`
where I was not previously.

### DIFF
--- a/build/compile.sh
+++ b/build/compile.sh
@@ -195,7 +195,7 @@ build() {
     c_module_src_list=$(cat $abs_c_module_srcs)
 
     if [[ -n "$READLINE_DIR" ]]; then
-      readline_flags+="-L $READLINE_DIR/lib -I $READLINE_DIR/include"
+      readline_flags+="-L $READLINE_DIR/lib -I $READLINE_DIR/include "
     fi
 
     # NOTE: pyconfig.h has HAVE_LIBREADLINE but doesn't appear to use it?


andychu:
Thanks for the fix!  I guess nobody tested that code path before :-/

BTW I have not tested tab completion on OS X:

https://github.com/oilshell/oil/labels/portability

I think Python uses `libedit` by default (for licensing reasons?  not sure), but for a number of reasons GNU `readline` might  be better... that is a long story related to how Oil is more principled about text and parsing.  We use GNU readline in a different way than most programs.

andychu:
@ariabuckles I just made a pre-release of OSH with this change:

http://www.oilshell.org/release/0.6.pre19/changelog.html

I also announced it on https://oilshell.zulipchat.com/ -- feel free to sign up if you're interested Oil development!

I noted that there's a performance bug in pre19 unfortunately, so I plan to make a pre20 release right away.

PR #336, opened 2019-06-15
build/dev.sh: When no arguments, print usage

mrshu:
* Print usage when no arguments are passed to the `build/dev.sh` script.

### DIFF
--- a/build/dev.sh
+++ b/build/dev.sh
@@ -91,7 +91,7 @@ gen-asdl-py() {
 
   # abbrev module is optional
   core/asdl_gen.py mypy "$@" > $tmp
-  
+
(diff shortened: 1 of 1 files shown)


build/dev.sh
```diff
@@ -219,4 +219,9 @@ all() {
   fastlex
 }
 
+if [ $# -eq 0 ]; then
+        echo "usage: $0 <function name>"
```
andychu: Thanks, there a lot of scripts that need something like this, but we might as well do build/dev.sh first since that's the first thing people will run into.

But please follow the existing style and use two spaces everywhere.

mrshu: Roger that, sorry bout this -- I'll happily change that!

PR #337, opened 2019-06-15
Update man page to use (semantic) mdoc macros

msingle:
Following mdoc/man documentation ( [intro/synopsis_util](http://mandoc.bsd.lv/mdoc/intro/synopsis_util.html), [man(7)](http://mandoc.bsd.lv/man/mdoc.7.html) )

TODO: Update info from other pieces of existing Osh documentation

### DIFF
--- a/doc/osh.1
+++ b/doc/osh.1
@@ -1,42 +1,82 @@
-.TH OSH "1" "November 2017" "OSH version 0.2.0" "User Commands"
-.SH NAME
-osh \- A bash-compatible UNIX shell
-.SH SYNOPSIS
-.B osh
-[\fI\,options\/\fR] \fI\,script \/\fR[\fI\,args\/\fR]
-.br
-.B osh
-[\fI\,options\/\fR] -c [\fI\,command_string\/\fR] [\fI\,args\/\fR]
-.SH DESCRIPTION
-.PP
-OSH is a command-line shell. It behaves like busybox and `osh` is a symlink to it.
-.PP
-.SH OPTIONS
-OSH accepts POSIX \fBsh\fR flags, with the following differences:
-.TP
-\fB\-n\fR
-only validate the syntax.  Also prints the AST.
-.TP
-\fB\-\-show\-ast\fR
-print the AST in addition to executing.
-.TP
-\fB\-\-ast\-format\fR
-what format the AST should be in
-.SH AUTHOR
-Written by Andy Chu.
-.SH REPORTING BUGS
-Please report bugs to <https://github.com/oilshell/oil/issues>
-.SH COPYRIGHT
-Copyright © 2016 Andy Chu. All rights reserved.
-.br
-Licensed under the Apache License, Version 2.0 (the "License");
-.br
-You may not use this file except in compliance with the License. You may obtain a copy of the License at
-.IP
-http://www.apache.org/licenses/LICENSE-2.0
-.SH SEE ALSO
-Developer documentation for
-.B OSH
-may be found online at
-.IP
-http://www.oilshell.org/release/0.2.0/doc/osh-quick-ref.html
+.\" Copyright © 2016 Andy Chu. All rights reserved.
+.\"
+.\" Licensed under the Apache License, Version 2.0 (the "License");
+.\" You may not use this file except in compliance with the License.
+.\"
+.\" You may obtain a copy of the License at:
+.\" http://www.apache.org/licenses/LICENSE-2.0
+.Dd June 28, 2019
+.Dt OSH 1
+.Os
+.Sh NAME
+.Nm osh
+.Nd command language interpreter
+.Sh SYNOPSIS
+.Nm
+.Op Ar options
+.Ar script
+.Op Ar args
+.Nm
+.Op Ar options
+.Fl c
+.Op command_string
+.Op args
+.Sh DESCRIPTION
+The
+.Nm
+utility is a
+.Xr bash 1 Ns -compatible
+command language intepreter:
+it reads commands either from standard input, a
+.Ar script
+file or command-line arguments and executes them.
+.Pp
+.Nm
+accepts POSIX
+.Xr sh 1
+arguments, with the following differences:
+.Bl -tag -width Ds
+.It Fl n
+Only validate the syntax.
+Also prints the AST.
+.It Fl -show-ast
+Print the AST in addition to executing.
+.It Fl -ast-format
+What format the AST should be in.
+.El
+.Sh ENVIRONMENT
+.Bl -tag -width "OSH_CRASH_DUMP_DIR"
+.It Ev OSH_HIJACK_SHEBANG
+.It Ev OSH_CRASH_DUMP_DIR
+.El
+.Sh FILES
+The interactive shell only sources
+.Pa ~/.config/oil/oshrc
+on startup.
+.Sh EXIT STATUS
+The
+.Nm
+utility exits with one of the following values:
+.Bl -tag -width 3n
+.It 1
+One or more runtime errors occurred.
+.It 2
+One or more parsing errors occurred,
+either due to incorrect script syntax or improper invocation of a shell
+builtin.
+.It 126
+Permission was denied when running a command.
+.It 127
+The referenced command or script could not be found.
+.El
+.Sh SEE ALSO
+.Xr bash 1 ,
+.Xr busybox 1 ,
+.Xr sh 1
+.Pp
+.Lk http://www.oilshell.org/release/0.6.pre22/doc/osh-quick-ref.html OSH Quick Reference
+.Sh AUTHORS
+The
+.Nm
+utility was written by
+.An Andy Chu Aq Mt andy@oilshell.org
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,10 @@
+# Documentation
+
+## Man pages
+Osh.1 is a ["man" page](https://en.wikipedia.org/wiki/Man_page) written using the ['mdoc'](http://mandoc.bsd.lv/man/mdoc.7.html) macro language for the troff (groff) formatter.
+Osh.1 is viewable with the `man` or `mandoc` commands by typing (in the doc directory) `man ./osh`. 
+
+### TODO:
+Install oil to the $MANPATH
+
+


doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
-.Br
-.Nm 
+.Nm
 .Op Ar options
 .Fl c
 .Op command_string
 .Op args
 .Sh DESCRIPTION
-OSH is a command-line shell. It behaves like busybox and `osh` is a symlink to it.
+.Nm OSH
```
The-King-of-Toasters: This should instead start like so:

```mdoc
The
.Nm
utility is a command-line shell.
```

`Nm` automatically inserts the name of the program, no need to put it here.

doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
+is a command-line shell.
+It behaves like busybox and `osh` is a symlink to it.
```
The-King-of-Toasters: This is an implementation detail and not really needed.

doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
 .Sh OPTIONS
```
The-King-of-Toasters: No need for a separate heading, just create a new paragraph:

```mdoc
.Pp
```

doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
 OSH accepts POSIX \fBsh\fR flags, with the following differences:
```
The-King-of-Toasters:
```mdoc
.Nm
accepts POSIX
.Xr sh 1
flags, with the following differences:
```

This is good for a first start, but I recommend in the future listing all of these flags.

doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
 .Bl -tag -width Ds
 .It Fl n
-only validate the syntax.  Also prints the AST.
+only validate the syntax.
+Also prints the AST.
 .It Fl Fl show-ast
 print the AST in addition to executing.
```
The-King-of-Toasters: Capital letters when starting sentences.

The-King-of-Toasters: ditto

doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
 .It Fl Fl ast-format
 what format the AST should be in
 .El
-.Sh AUTHOR
+.Sh SEE ALSO
```
The-King-of-Toasters: Too long, I recommend something like this instead:

```mdoc
.Xr busybox 1
.Xr sh 1

.Lk http://www.oilshell.org/release/0.6.pre22/doc/osh-quick-ref.html OSH Quick Reference
```

doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
+Developer documentation for
+.Nm OSH
+may be found online at
+.Lk http://www.oilshell.org/release/0.6.pre22/doc/osh-quick-ref.html
+.Sh AUTHORS
 Written by
```
The-King-of-Toasters: The convention for the authors section is:

```mdoc
The
.Nm
utility was written by
.An Andy Chu Aq Mt andy@oilshell.org
```

doc/osh.1
```diff
@@ -5,45 +5,47 @@
...
 .An Andy Chu .
 .Sh REPORTING BUGS
 Please report bugs to
 .Aq Lk https://github.com/oilshell/oil/issues
 .Sh COPYRIGHT
```
The-King-of-Toasters: Don't put license info in the manpage, it's not needed. You could stick it up in the top of the file as a series of comments like:

```mdoc
.\" Copyright © 2016 Andy Chu. All rights reserved.
.\"
.\" Licensed under the Apache License, Version 2.0 (the "License");
.\" You may not use this file except in compliance with the License.
.\"
.\" You may obtain a copy of the License at:
.\" http://www.apache.org/licenses/LICENSE-2.0
```

andychu:
Thanks @The-King-of-Toasters for the great feedback.

One issue is that this documentation will likely get very long, and most of it doesn't exist yet.  For example, bash is bigger than say OpenBSD's ksh.  And Oil is going to grow over time (and there is going to be a difference between `osh` and `oil`, and some other commands may get installed too).

I'll likely want to generate flag specs for the builtins from the code, at the very least.  Because there a lot of builtins and each one has many flags.  And there is an online `help` command.

So I think the first pass should concentrate on something with the right basic structure, but has mostly pointers to other pieces of documentation.  Any list of signals, etc. will get out of date.

That said this page has some info about exit codes: https://github.com/oilshell/oil/blob/master/doc/osh-manual.md

msingle:
Thanks @The-King-of-Toasters ! The newest changes should address the in-line comments.
I figured I would address missing sections/more content in the next pull request.

doc/osh.1
```diff
@@ -1,42 +1,84 @@
...
+.Bl -tag -width Ds
+.It Fl n
+Only validate the syntax.
+Also prints the AST.
+.It Fl Fl show-ast
+Print the AST in addition to executing.
+.It Fl Fl ast-format
+What format the AST should be in
+.El
+.Sh EXIT STATUS
```
The-King-of-Toasters: The convention is to have a definition list of exit codes. How about something like:

```mdoc
The
.Nm
utility exits with one of the following values
.Bl -tag -width 3n
.It 1
A runtime error occured.
.It 2
An parser error occured,
either due to incorrect script syntax or improper invocation of a shell
builtin.
.It 126
A command was found but was not executable
.It 127
A command or script could not be found.
.El
```

doc/osh.1
```diff
@@ -1,42 +1,84 @@
...
+.Pp
+.Nm
+accepts POSIX
+.Xr sh 1
+flags, with the following differences:
+.Bl -tag -width Ds
+.It Fl n
+Only validate the syntax.
+Also prints the AST.
+.It Fl Fl show-ast
```
The-King-of-Toasters: I prefer using the form:

```mdoc
.Fl -long-opt
```

as mandoc generates proper hyperlinks and `less(1)` tags

doc/osh.1
```diff
@@ -37,37 +37,25 @@ flags, with the following differences:
...
-POSIX mentions these:
-.Bl -bullet
-.It
-126 for permission denied when running a command (errno EACCES)
-.It
-127 for command not found
-.El
+The
+.Nm
+utility exits with one of the following values
```
The-King-of-Toasters: I forgot to put a colon at the end of `values`, whoops

doc/osh.1
msingle: done

doc/osh.1
msingle: done

doc/osh.1
msingle: done

The-King-of-Toasters:
I made a couple more changes to the file:

```patch
--- osh.1	2019-06-28 13:06:56.129014570 +1000
+++ osh-new.1	2019-06-28 17:20:51.626877439 +1000
@@ -4,13 +4,13 @@
 .\" You may not use this file except in compliance with the License.
 .\"
 .\" You may obtain a copy of the License at:
-.\" http://www.apache.org/licenses/LICENSE-2.0d June 14, 2019
-.Dd June 14, 2019
+.\" http://www.apache.org/licenses/LICENSE-2.0
+.Dd June 28, 2019
 .Dt OSH 1
 .Os
 .Sh NAME
 .Nm osh
-.Nd A bash-compatible UNIX shell
+.Nd command language interpreter
 .Sh SYNOPSIS
 .Nm
 .Op Ar options
@@ -24,15 +24,17 @@
 .Sh DESCRIPTION
 The
 .Nm
-utility is a command-line shell.
-It behaves like busybox and
-.Nm
-is a symlink to it.
+utility is a
+.Xr bash 1 Ns -compatible
+command language intepreter:
+it reads commands either from standard input, a
+.Ar script
+file or command-line arguments and executes them.
 .Pp
 .Nm
 accepts POSIX
 .Xr sh 1
-flags, with the following differences:
+arguments, with the following differences:
 .Bl -tag -width Ds
 .It Fl n
 Only validate the syntax.
@@ -48,16 +50,18 @@
 utility exits with one of the following values:
 .Bl -tag -width 3n
 .It 1
-A runtime errors occured.
+One or more runtime errors occurred.
 .It 2
-A parse errors occured,
-either due to incorrect script syntax or improper invocation of a shell builtin.
+One or more parsing errors occurred,
+either due to incorrect script syntax or improper invocation of a shell
+builtin.
 .It 126
-Permission denied when running a command (errno EACCES)
+Permission was denied when running a command.
 .It 127
-A command or script could not be found
+The referenced command or script could not be found.
 .El
 .Sh SEE ALSO
+.Xr bash 1 ,
 .Xr busybox 1 ,
 .Xr sh 1
 .Pp
@@ -65,8 +69,5 @@
 .Sh AUTHORS
 The
 .Nm
-utility is written by
+utility was written by
 .An Andy Chu Aq Mt andy@oilshell.org
-.Sh REPORTING BUGS
-Please report bugs to
-.Aq Lk https://github.com/oilshell/oil/issues
```

You can download this patch [here](https://www.sgregoratto.me/paste/osh.1.patch).

msingle:
Sounds good! Applied.

msingle:
@The-King-of-Toasters any other issues?

doc/osh.1
```diff
@@ -1,42 +1,85 @@
...
+.Bl -tag -width Ds
+.It Fl n
+Only validate the syntax.
+Also prints the AST.
+.It Fl -show-ast
+Print the AST in addition to executing.
+.It Fl -ast-format
+What format the AST should be in.
+.El
+.Sh ENVIRONMENT
```
The-King-of-Toasters: What do these do?

doc/osh.1
```diff
@@ -1,42 +1,85 @@
...
+OSH_HIJACK_SHEBANG
+OSH_CRASH_DUMP_DIR
+.Sh FILES
```
The-King-of-Toasters: This could be rewritten as:

```mdoc
.Sh FILES
.Bl -tag -width "~/.config/oil/oshrc" -compact
.It Pa ~/.config/oil/oshrc
User configuration file.
Sourced on startup.
.El
```

`.Pa` is what is used to reference system paths/files

doc/osh.1
```diff
@@ -1,42 +1,85 @@
...
+On startup,
+the interactive shell sources
+.Em only
+.Ql ~/.config/oil/oshrc .
+The
+.Nm
+utility intends to avoid
```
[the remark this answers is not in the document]
msingle: not sure where that came from, removed

doc/osh.1
msingle: This pull request is just to get the structure (see @andychu 's first comment)

doc/osh.1
msingle: updated with path link. I'm leaving the wording - yours implies that there's multiple configs, and I believe (in addition to the specific path), that multiple configs are being avoided. @andychu does that sound right?

doc/osh.1
```diff
@@ -51,11 +51,9 @@ OSH_CRASH_DUMP_DIR
 On startup,
```
The-King-of-Toasters: Since there is only one file, how about:

```mdoc
The interactive shell sources only
.Pa ~/.config/oil/oshrc
on startup.
```

doc/osh.1
The-King-of-Toasters: In that case, they should be in a list for now. Something like:

```mdoc
.Bl -tag -width "OSH_CRASH_DUMP_DIR"
.It Ev OSH_HIJACK_SHEBANG
.It Ev OSH_CRASH_DUMP_DIR
.El
```

doc/osh.1
msingle: Done.

doc/osh.1
msingle: Done.

andychu:
Can you add doc/README.md with links to mdoc, how to use it, how to preview the man page, etc.?

Sorry I should have been looking at this more carefully, but I don't know much about man pages.

I did read a chapter in The Art of Unix Programming which mentions mdoc and troff, so I have a bit more context for it now.

The-King-of-Toasters:
Hey @msingle, perhaps you could give me commit access to your repo so I can add what @andychu wants.

The-King-of-Toasters:
@andychu Would you like readme.md to include descriptions about all files in `doc/`, or just `osh.1`?

andychu:
For now just describing `osh.1` is OK.  I will add descriptions for the rest of them.  Thanks!

The-King-of-Toasters:
I think we're almost done then. Some remaining things todo:

- [ ] rebase all `osh.1` commits into one
- [ ] add osh.1 to the install script
- [ ] rewrite the readme to reference the installed manpage

The-King-of-Toasters:
Rebase complete, prepare for breakage!

andychu:
Should I merge this first?  And then we can do the install stuff in another PR?

andychu:
OK I checked it out and previewed on my machine.  Looks pretty good as a first pass.

As of a week ago, the 0.6.0 release, there are some new docs with every release, e.g.

https://www.oilshell.org/release/0.6.0/

https://www.oilshell.org/release/0.6.0/doc/osh-manual.html

But we can add those in subsequent PRs.  I'll merge this tomorrow morning unless there's anything else you want to change.  Normally the review protocol I use is like this:

https://github.com/oilshell/oil/wiki/Contributing#pull-requests

The-King-of-Toasters:
I think it's better if the install stuff is added first, but I'm a bit confused by your install scripts. Would this be added to `build/doc.sh`?

andychu:
`install` is included in the tarball, and `make install` is an alias for that, for people who are used to that autoconf-like way of doing things.

`build/doc.sh` isn't included in the tarball, only the repo.  

FWIW here is some background on the build system: https://github.com/oilshell/oil/wiki/Contributing#three-kinds-of-builds

OSH doesn't use autoconf.  But like autoconf, the repo is different than the tarball.  And in Oil the repo is even *more* different than the tarball! 

Let me know if you have more questions.

andychu:
I guess I don't know what the normal convention is for installing man pages on Unix systems.   If so I would try it myself, since the `install` script is currently tiny and simple.

Looking at Debian sort of gives me a clue, but not entirely.  Ideally it wouldn't be dependent on a particular Linux distro or the variant of BSD.

```
$ dpkg --listfiles bash |grep man
/usr/share/doc/bash/README.commands.gz
/usr/share/man
/usr/share/man/man1
/usr/share/man/man1/bash.1.gz
/usr/share/man/man1/clear_console.1.gz
/usr/share/man/man1/bashbug.1.gz
/usr/share/man/man1/rbash.1.gz
/usr/share/man/man7
/usr/share/man/man7/bash-builtins.7.gz

$ dpkg --listfiles dash |grep man
/usr/share/man
/usr/share/man/man1
/usr/share/man/man1/dash.1.gz
/usr/share/man/man1/sh.1.gz
package diverts others to: /usr/share/man/man1/sh.distrib.1.gz
```

andychu:
This page gives me a clue for Linux.  I guess `/usr/share/man/man1` is the place to put it?  Since it's in section 1 of the manual?

https://linux.die.net/man/7/hier

I think that's probably OK to start?  I would rather do it in a separate PR though, in case it needs to be rolled back.  The manual contents itself looks good.

andychu:
Actually, looking at dash, it's pretty simple.  I think it's just `$prefix/share/man/man1`.  We already have prefix, which is usually `/usr/local`, and can also be `/usr/`.

Although I don't know who gzips the man page.

```
configure:mandir
configure:mandir='${datarootdir}/man'
```

```
andy@lisa:~/src/languages/dash-0.5.8$ ./configure --help|grep datarootdir
  --datarootdir=DIR       read-only arch.-independent data root [PREFIX/share]

```

The-King-of-Toasters:
In general, `/usr/share/man` is the directory used for system/base programs and `/usr/local/man` is used for external/packaged programs. In my projects, I use two variables:

- `PREFIX` - the root prefix to install to (e.g. `/usr` or `/usr/local`).
- `MANPREFIX` - the root prefix to install manpages (e.g. `$PREFIX/share/man` or `$PREFIX/man`).

These would default to `/usr/local` and `/man`, but they should be able to be redefined.

andychu:
OK, I will merge this PR then, and take a stab at the install.

Another question though: who uses the semantic markup?  It seems like the `man` program doesn't care about the semantics?

Is there a program that does cross-referencing and such or generates a different presentation like HTML?  Maybe we should also generate HTML to make sure the semantic markup makes sense?

The-King-of-Toasters:
>... Who uses the semantic markup? It seems like the `man` program doesn't care about the semantics?

It does, but the most of the benefits depend on the output mode.
Text/terminal output mostly remains the same, but groff PDFs are better formatted and mandoc generates high quality HTML5 (hyperlinks to websites/email/other manual pages, tables, lists etc).

> Is there a program that does cross-referencing and such or generates a different presentation like HTML? Maybe we should also generate HTML to make sure the semantic markup makes sense?

In general you don't need to do this. Manual viewers can output to different formats at the user's request. If you wish to upload the HTML version of the manual to oilshell.org, then you could add a configure/build command to generate it with `mandoc` (using `mandoc -Thtml doc/osh.1`) and rsync/scp it to the server.

PR #348, opened 2019-06-19
expr_eval: Implement `[[ -o val ]]` option

### LINKED ISSUE #338: Implement test -o to test if an option is set
From #322

mrshu:
* Implement the `-o` option for the test command (`[[ -o val ]]`).

* Fix  #338


andychu:
Technically this could use some spec tests [1] , but I want to talk about fzf and then maybe circle back and do that!

[1] https://github.com/oilshell/oil/wiki/Spec-Tests

PR #349, opened 2019-06-19
buildtin: Add first version of bind

### LINKED ISSUE #339: Implement `bind` builtin
From #322 

I don't know how this works ... hopefully it isn't a big time sink ...

mrshu:
* Add a simple, introductory version of the buildtin `bind` command that
  just passes configuration options to `readline`.

* Fixes #339.

### DIFF
--- a/osh/builtin.py
+++ b/osh/builtin.py
@@ -122,6 +122,8 @@ _NORMAL_BUILTINS = {
     "alias": builtin_e.ALIAS,
     "unalias": builtin_e.UNALIAS,
 
+    "bind": builtin_e.BIND,
+
     # OSH only
     "repr": builtin_e.REPR,
 }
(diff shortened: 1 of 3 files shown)


osh/builtin.py
```diff
@@ -1213,6 +1215,25 @@ def __call__(self, arg_vec):
...
+class Bind(object):
+  def __init__(self, readline_mod):
+    self.readline_mod = readline_mod
+
+  def __call__(self, arg_vec):
+    if len(arg_vec.strs) == 1:
+      raise args.UsageError('bind keyseq:function-name')
+
+    status = 0
+    for i in xrange(1, len(arg_vec.strs)):
```
andychu: Does bash process all positional arguments?

When I do `help bind`, it doesn't have the customary ... syntax for the last arg:

[keyseq:readline-function or readline-command]

I've never used `bind` honestly.  Can you maybe write some manual test cases in the comments?

Ideally tests should be automatic, but interactive readline stuff is hard to test of course... I maintain some manual test cases here, and those would be useful for `bind`

https://github.com/oilshell/oil/wiki/Testing-the-Interactive-Shell

PR #350, opened 2019-06-19
fix: TravisCI linter by not importing unused sys

mrshu:
* Do not import the `sys` module in `word_eval.py` as it is not used
  there.

* Fix TravisCI build as a result.


andychu:
Thank you!

Issue #355, opened 2019-06-20
Implement hash / hash -r for speed


andychu:
Didn't realize this was POSIX

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/hash.html

andychu:
Related to #306

andychu:
Does this help with the speed of Python's configure?   #306

andychu:
Released with 0.7.pre1: https://www.oilshell.org/release/0.7.pre1/

PR #365, opened 2019-06-23
Run files with /bin/sh if shebang is missing

jyn514:
Resolve the full filepath since it might be a command in PATH, not the current directory.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -17,7 +17,7 @@ import sys
 
 from _devbuild.gen.id_kind_asdl import Id
 from _devbuild.gen.runtime_asdl import redirect_e, job_state_e
-from asdl import pretty
+from asdl import const, pretty
 from core import ui
 from core.util import log
 from pylib import os_
@@ -496,13 +496,14 @@ class ExternalProgram(object):
       os_.execvpe(argv[0], argv, environ)
     except OSError as e:
       # Run with /bin/sh when ENOEXEC error (no shebang). Because all shells do it.
-      if e.errno == errno.ENOEXEC:
+      if e.errno == errno.ENOEXEC and argv[0] != "/bin/sh":  # avoid infinite loops
         _, realpath = _ResolveFile(argv[0], environ["PATH"].split(':'))
         # Check if file was deleted in the meantime
         if realpath is None:
           e.errno = errno.ENOENT
         else:
           arg_vec.strs = ["/bin/sh", realpath] + argv[1:]
+          arg_vec.spids = [const.NO_INTEGER] + arg_vec.spids
           # Never returns
           self.Exec(arg_vec, environ)
 
--- a/core/process.py
+++ b/core/process.py
@@ -21,6 +21,7 @@ from asdl import pretty
 from core import ui
 from core.util import log
 from pylib import os_
+from osh.builtin import _ResolveFile
 
 import posix_ as posix
 
@@ -494,8 +495,16 @@ class ExternalProgram(object):
     try:
       os_.execvpe(argv[0], argv, environ)
     except OSError as e:
-      # TODO: Run with /bin/sh when ENOEXEC error (noshebang).  Because all
-      # shells do it.
+      # Run with /bin/sh when ENOEXEC error (no shebang). Because all shells do it.
+      if e.errno == errno.ENOEXEC:
+        _, realpath = _ResolveFile(argv[0], environ["PATH"].split(':'))
+        # Check if file was deleted in the meantime
+        if realpath is None:
+          e.errno = errno.ENOENT
+        else:
+          arg_vec.strs = ["/bin/sh", realpath] + argv[1:]
+          # Never returns
+          self.Exec(arg_vec, environ)
 
       # Would be nice: when the path is relative and ENOENT: print PWD and do
       # spelling correction?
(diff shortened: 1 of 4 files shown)


core/process.py
```diff
@@ -494,8 +495,12 @@ def Exec(self, arg_vec, environ):
     try:
       os_.execvpe(argv[0], argv, environ)
     except OSError as e:
-      # TODO: Run with /bin/sh when ENOEXEC error (noshebang).  Because all
-      # shells do it.
+      # Run with /bin/sh when ENOEXEC error (no shebang). Because all shells do it.
+      if e.errno == errno.ENOEXEC:
+          arg_vec.strs = (["/bin/sh"]
+            + [_ResolveFile(arg_vec.strs[0], environ["PATH"].split(':'))[1]]
```
andychu: This looks like the right idea, but it's a bit hard to read because of the [1] and so forth.
_ResolveFile can return (None, None).  It probably won't happen, but there's no guarantee because files can disappear at any time.

It can just be argv[0] too.  It might be clearer if we make a private version of this function that's like

_Exec(self, argv, argv0_spid):
  ...

and then it can call itself in the ENOEXEC case.

jyn514: Good catch that _ResolveFile can return `None`! That found another bug - _ResolveFile wasn't handling relative paths properly, which I fixed and added to the spec tests.

I cleaned up the code a little, hopefully it's a bit easier to read now.

spec/command_.test.sh
```diff
@@ -34,3 +34,10 @@ chmod +x $TMP/no-shebang
 $SH -c '$TMP/no-shebang'
 ## stdout: hi
 ## status: 0
+
+#### File with relative path and no shebang is executed
+echo 'echo hi' > no-shebang
```
andychu: I'm pretty sure this will litter the repo.  `cd $TMP` first (TMP is always set in the test env)

osh/builtin.py
```diff
@@ -1031,6 +1031,12 @@ def _ParsePath(path_val):
 
 def _ResolveFile(name, path_list):
   # Now look for files.
+  # Relative path, won't be found by looking in PATH
+  if name.startswith('./') or name.startswith('../'):
```
andychu: What happens when we run `subdir/myprogram` without `./` ?

I think there are probably 2 bugs.  One is a bug in `type` and the other is a bug in path lookup for execution.  We might need two different functions.

- $PATH shouldn't behave like the current directory is in it
- only programs directly under $PATH are run

core/process.py
```diff
@@ -494,8 +495,16 @@ def Exec(self, arg_vec, environ):
...
       os_.execvpe(argv[0], argv, environ)
     except OSError as e:
-      # TODO: Run with /bin/sh when ENOEXEC error (noshebang).  Because all
-      # shells do it.
+      # Run with /bin/sh when ENOEXEC error (no shebang). Because all shells do it.
+      if e.errno == errno.ENOEXEC:
+        _, realpath = _ResolveFile(argv[0], environ["PATH"].split(':'))
+        # Check if file was deleted in the meantime
+        if realpath is None:
+          e.errno = errno.ENOENT
```
andychu: I think it's clearer to do sys.exit(127) here than to mutate the exception object, if that's the intention.

core/process.py
```diff
@@ -494,8 +495,16 @@ def Exec(self, arg_vec, environ):
...
+        else:
+          arg_vec.strs = ["/bin/sh", realpath] + argv[1:]
```
andychu: I guess this is technically correct, but I don't like the style of mutating arg_vec either... it breaks the invariant that arg_vec.strs and arg_vec.spids are parallel (and the same length).

core/process.py
jyn514: If we exit immediately, we don't print the error message that the file was missing. Especially if we add suggestions for misspelled files, it would be nice not to duplicate that logic.

core/process.py
jyn514: How would I get an spid to add to arg_vec? `Exec` uses it on errors.

I guess that brings up a second question of if the original fails _and_ `/bin/sh` fails whether we want to print '/bin/sh' or the original, that's kind of a weird edge case.

osh/builtin.py
jyn514: Hmm.. I want to keep this in the same function so that we don't have to duplicate the logic and we can make sure they're the same (`command -v` shouldn't print something different than what will be executed!). What if I changed the check to `name.contains('/')`? That won't look in the current directory unless you prefix it with `./` but it will still check subdirectories

core/process.py
andychu: One way is to use const.NO_INTEGER, e.g.

arg_vector(["/bin/sh" ... , [const.NO_INTEGER] + arg_vec.spids)

Or another way is factoring out a private function like _Exec(self, argv, argv0_spid), so there's no arg_vector any more.

-----

I think this code also will enter an infinite loop if `/bin/sh` returns ENOEXEC.  That is a corner case but we should probably avoid these things even in theory, since shells are often used to fix broken systems, etc.


------------

Note this code may change in the near future due to issue #355, which caches the PATH lookups.

jyn514:
This is ready for review, I think I addressed everything you mentioned. It still mutates various objects, if you want me to change that style I can.

andychu:
Yeah I don't like the mutating style, but let's get the related `source` change merged first, and then I will merge this.

And maybe change a few things around afterward.  I think I would rather use `mem.GetVar("PATH")` rather than `environ["PATH"]` too, but I'm not sure how to plumb that around yet.

But since it passes tests I don't want to get hung up on the style that much.

Issue #269, opened 2019-03-24
comp_ui.NiceDisplay should take into account unicode character width


andychu:
@jyn514 I'm pretty sure this is the issue you just mentioned  in #257.  There are two separate bugs:

- If you keep typing past the end of the terminal, OSH has no idea (but zsh and fish do).  This is #257.
- If your prompt contains a unicode character that is say 4 bytes long in utf-8, the code will currently move right FOUR cursor positions instead of ONE.  I think that explains the screenshot, since you have a unicode character there.

I think we can just call the POSIX function wcwidth(), although I guess that involves some conversion to `wchar_t`, which is annoying:

http://man7.org/linux/man-pages/man3/wcwidth.3.html

There is a pure Python version of it, but I'm not sure if we should use it.

https://pypi.org/project/wcwidth/#files

andychu:
It looks like `wchar_t` might introduce some portability problems?  Not sure how much of an issue this is in practice.

https://en.wikipedia.org/wiki/Wide_character#C/C++

Is there a version wcwidth() that works on utf-8 data?

Note: this function might be useful for short-circuiting?

http://man7.org/linux/man-pages/man3/iswprint.3.html

andychu:
Fairly related, since OSH essentially does this!

https://www.openbsd.org/papers/eurobsdcon2016-utf8.pdf

The OpenBSD base system supportsexactly two LC_CTYPE locales:

1.    UTF-8
2.    C=POSIX = US-ASCII

We  don’tevensupport ISO-LATIN-1anylonger in the base system.

andychu:
Here's an interesting alternative:

https://unix.stackexchange.com/questions/245013/get-the-display-width-of-a-string-of-characters

We could just print the prompt, and query cursor positions before and after!  Then we know exactly how many displayable characters it is.  Downsides?

jyn514:
Why do we need multithreading? Shouldn't we have at most one main thread at a time?

andychu:
Yeah we don't really need multi-threading, but functions that use globals are a "smell" to me ... it's not what I'm used to.

Some people have asked for Oil to be a Lua-like library, in which case it should use no globals.  This would be nice in theory, but it's way down on the priority list now.

----

Although querying the cursor position might seem like a hack, I think it might be the right thing for the case of the prompt, if not in general.

jyn514:
Since this is the last blocker from me using osh as my fulltime shell, I think I'll take a look at it.

[...]

Issue #366, opened 2019-06-23
Consider using a utf-8 library in C


andychu:
This one now maintained by the Julia developers looks like a good bet:

https://github.com/JuliaStrings/utf8proc

formerly: https://github.com/JeffBezanson/cutef8

related to #269

andychu:
At the very least, we could use their uppercase / lowercase implementation, e.g. for `${x^}`  and so forth.

andychu:
I sketched out what we need in https://github.com/oilshell/oil/blob/master/doc/unicode.md

and I believe we don't need this

We will use

- Bjoern DFA -- and create a BACKWARD version of it too?
- fallbacks to libc, for case folding

TODO: `cpp/unicode_test.cc` can test how LANG and LC_COLLATE behave etc.

PR #368, opened 2019-06-23
Take unicode into account when calculating display width

jyn514:
Addresses #269. Since the release build of python doesn't support unicode, we have to use C libraries to decode to and from it.

### DIFF
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -43,10 +43,10 @@ def _PromptLen(prompt_str):
     elif not escaped:
       display_str += c
   try:
-      return libc.wcswidth(display_str)
+    return libc.wcswidth(display_str)
   # en_US.UTF-8 locale missing, just return the number of bytes
-  except SystemError:
-      return len(display_str)
+  except (SystemError, UnicodeError):
+    return len(display_str)
 
 
 class PromptState(object):
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -6,7 +6,6 @@ from __future__ import print_function
 import sys
 
 import libc
-import wcwidth
 
 _RESET = '\033[0;0m'
 _BOLD = '\033[1m'
@@ -36,14 +35,17 @@ def _PromptLen(prompt_str):
   number of bytes or the number of unicode characters."""
   escaped = False
   length = 0
-  for c in unicode(prompt_str, 'utf-8'):
+  chunk = ""
+  for c in prompt_str:
     if c == '\x01':
       escaped = True
+      length += libc.wcswidth(chunk)
+      chunk = ""
     elif c == '\x02':
       escaped = False
     elif not escaped:
-        length += wcwidth.wcwidth(c)
-  return length
+      chunk += c
+  return length + libc.wcswidth(chunk)
 
 
 class PromptState(object):
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -34,18 +34,19 @@ def _PromptLen(prompt_str):
   In particular, the display width of a string may be different from either the
   number of bytes or the number of unicode characters."""
   escaped = False
-  length = 0
-  chunk = ""
+  display_str = ""
   for c in prompt_str:
     if c == '\x01':
       escaped = True
-      length += libc.wcswidth(chunk)
-      chunk = ""
     elif c == '\x02':
       escaped = False
     elif not escaped:
-      chunk += c
-  return length + libc.wcswidth(chunk)
+      display_str += c
+  try:
+      return libc.wcswidth(display_str)
+  # en_US.UTF-8 locale missing, just return the number of bytes
+  except SystemError:
+      return len(display_str)
 
 
 class PromptState(object):
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 import sys
 
 import libc
-
+import wcwidth
 
 _RESET = '\033[0;0m'
 _BOLD = '\033[1m'
@@ -31,16 +31,18 @@ PROMPT_REVERSE = '\x01%s\x02' % _REVERSE
 
 
 def _PromptLen(prompt_str):
-  """Ignore all characters between \x01 and \x02."""
+  """Ignore all characters between \x01 and \x02 and handle unicode characters.
+  In particular, the display width of a string may be different from either the
+  number of bytes or the number of unicode characters."""
   escaped = False
   length = 0
-  for c in prompt_str:
+  for c in unicode(prompt_str, 'utf-8'):
     if c == '\x01':
       escaped = True
     elif c == '\x02':
       escaped = False
     elif not escaped:
-      length += 1
+        length += wcwidth.wcwidth(c)
   return length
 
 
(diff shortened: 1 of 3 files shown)


native/libc.c
```diff
@@ -5,8 +5,11 @@
 #include <stdarg.h>  // va_list, etc.
 #include <stdio.h>  // printf
 #include <limits.h>
+#define _XOPEN_SOURCE
```
andychu: comment on why this is needed?

native/libc.c
```diff
@@ -383,6 +386,30 @@ func_get_terminal_width(PyObject *self, PyObject *unused) {
...
 }
 
+static PyObject *
+func_wcswidth(PyObject *self, PyObject *args){
+    char *string;
+    if (!PyArg_ParseTuple(args, "s", &string)) {
+        return NULL;
+    }
+
+    const char *old_locale = setlocale(LC_CTYPE, "en_US.UTF-8");
```
andychu: I'm pretty sure this does a file system operation, so the prompt length calculation will do more than one in general.

I guess the long term solution is to use our own library as in #366.

We might not be able to set the locale globally, because GNU readline is in the same process and also uses libc.  Stupid global variables...

core/comp_ui.py
```diff
@@ -31,17 +30,22 @@
...
+  """Ignore all characters between \x01 and \x02 and handle unicode characters.
+  In particular, the display width of a string may be different from either the
+  number of bytes or the number of unicode characters."""
   escaped = False
   length = 0
+  chunk = ""
   for c in prompt_str:
     if c == '\x01':
       escaped = True
+      length += libc.wcswidth(chunk)
```
andychu: RuntimeError isn't handled.  Maybe we should fall back on the old calculation in that case?

core/comp_ui.py
```diff
@@ -31,17 +30,22 @@
...
+      chunk = ""
     elif c == '\x02':
       escaped = False
     elif not escaped:
-      length += 1
-  return length
+      chunk += c
+  return length + libc.wcswidth(chunk)
```
andychu: It seems like it would be simpler to just collect all the characters first, and make a single wcswidth() call at the end.  That seems like the same thing.

jyn514: It would do the same thing, yeah. My reasoning is that it would use less memory not to copy the prompt string, but if `setlocale` is expensive it probably makes sense to do it all at once.

native/libc.c
jyn514: It's mentioned in the man page for wcswidth, I'm not sure if it's actually necessary.

```
WCSWIDTH(3)                      Linux Programmer's Manual                     WCSWIDTH(3)

NAME
       wcswidth - determine columns needed for a fixed-size wide-character string

SYNOPSIS
       #define _XOPEN_SOURCE             /* See feature_test_macros(7) */
       #include <wchar.h>

       int wcswidth(const wchar_t *s, size_t n);
```

native/libc.c
jyn514: I just ran this locally and I got this wrong - setlocale returns the locale you _called_ it with, not the existing one. The way you get current locale is `setlocale(LC_CTYPE, NULL)`. Changing this real quick.

jyn514: And you're right, this mmaps /usr/lib/locale/locale-archive

core/comp_ui.py
```diff
@@ -31,17 +30,23 @@
...
     if c == '\x01':
       escaped = True
     elif c == '\x02':
       escaped = False
     elif not escaped:
-      length += 1
-  return length
+      display_str += c
+  try:
+      return libc.wcswidth(display_str)
```
andychu: Please use the existing 2 space indents

native/libc.c
andychu: It looks like a libc versioning thing.  I would be inclined to leave it out (since you probably get a compile time error on an ancient system), or at least add a comment that it's for wcswidth.

native/libc.c
```diff
@@ -383,6 +386,30 @@ func_get_terminal_width(PyObject *self, PyObject *unused) {
...
+    }
+
+    const char *old_locale = setlocale(LC_CTYPE, NULL);
+    if (setlocale(LC_CTYPE, "en_US.UTF-8") == NULL) {
+        PyErr_SetString(PyExc_SystemError, "en_US.UTF-8 is not a valid locale");
+        return NULL;
+    }
+    int len = mbstowcs(NULL, string, 0);
+    if (len == -1) {
+        PyErr_SetString(PyExc_UnicodeError, "Invalid UTF-8 string");
```
andychu: Maybe change this to "mbstowcs error" so we don't get confused, i.e. think it's from CPython

andychu: The locale should be restored in this error case

native/libc_test.py
```diff
@@ -189,6 +190,9 @@ def testGetTerminalWidth(self):
     else:
       print('width % d' % width)
 
+  def testWcsWidth(self):
+    self.assertEqual(1, libc.wcswidth("▶️"))
+    self.assertEqual(28, libc.wcswidth("(osh) ~/.../unchanged/oil ▶️ "))
```
andychu: Please add coverage for the UnicodeError case

core/comp_ui.py
```diff
@@ -31,17 +30,23 @@
...
+  # en_US.UTF-8 locale missing, just return the number of bytes
+  except SystemError:
```
andychu: UnicodeError needs to be handled as well, i.e. this is arbitrary user input

andychu:
Should I look at these PRs again?  I see you pushed something, but I assume it's still in progress unless you reply.  I know not everyone has the same workflow which can be confusing.

jyn514:
Yes please! In the future I'll say something.

andychu:
I just merged this, but when I tested it against some random unicode prompt, libc.wcswidth() is returning -1?

I reduced it to this, which works in bash:

```
PS1=$'\xe2\x86\x92'
```

Can you reproduce this on your machine?  Either way, we should handle the -1 return value.

But I'm not sure why this is happening.  Maybe the `'\x01` stripping isn't right?  That would be an unprintable character.


https://unix.stackexchange.com/questions/25903/awesome-symbols-and-characters-in-a-bash-prompt

andychu:
Hm the `\x01` stripping seems to be working OK.  It shouldn't matter since we are changing it, but my locale settings are:

```
$ locale
LANG=en_US.UTF-8
LANGUAGE=en_US
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
```

jyn514:
I can't replicate this on my machine. Here's my locale:

```
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```

I noticed you have `LANGUAGE=en_US`, could you try changing that and seeing what happens? Currently we only change `LC_CTYPE` because that's what the linux man pages document, maybe it's different on MacOS.

PR #393, opened 2019-07-04
Lookup files to be sourced in $PATH

jyn514:
Gives precendence to $PATH before sourcing files in current directory
TIL: sourcing files in the current directory is a bash-ism, not part of POSIX,
see http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot

Addresses https://github.com/oilshell/oil/issues/389

### DIFF
--- a/spec/builtin-eval-source.test.sh
+++ b/spec/builtin-eval-source.test.sh
@@ -135,3 +135,23 @@ FALSE
 FALSE
 ## END
 
+#### source works for files in current directory
+echo "echo current dir" > cmd
+. cmd
+rm cmd
+## STDOUT:
+current dir
+## N-I dash stdout-json: ""
+## N-I dash status: 2
+## N-I mksh stdout-json: ""
+## N-I mksh status: 1
+
+#### source gives precendence to PATH
+mkdir -p dir
+echo "echo path" > dir/cmd
+echo "echo current dir" > cmd
+PATH="dir:$PATH"
+. cmd
+rm -r dir cmd
+## STDOUT:
+path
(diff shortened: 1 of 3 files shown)


spec/posix.test.sh
```diff
@@ -142,3 +142,41 @@ three
 EOF2
 ## stdout-json: "one\ntwo\nthree\n"
 
+#### source works for files in subdirectory
```
andychu: Since  you mention this  isn't a POSIX feature, please move the tests to spec/builtin-source-eval.

The way I find tests for a particular feature:

`grep source spec/*.sh`

jyn514: I moved the tests, could you take another look at this?

spec/builtin-eval-source.test.sh
```diff
@@ -135,3 +135,23 @@ FALSE
...
+## N-I mksh stdout-json: ""
+## N-I mksh status: 1
+
+#### source gives precendence to PATH
+mkdir -p dir
+echo "echo path" > dir/cmd
+echo "echo current dir" > cmd
+PATH="dir:$PATH"
+. cmd
+rm -r dir cmd
```
andychu: I think `cd $TMP` is the better thing to do, but maybe I should just change the test framework to automatically do it.

Filed #396

spec/posix.test.sh
```diff
@@ -142,3 +142,20 @@ three
...
+mkdir -p dir
+echo "echo path" > dir/cmd
+. dir/cmd
+rm dir/cmd
+## STDOUT:
+path
```
andychu: It looks like this is missing an `# END`.  Although now that I check the parser in `test/sh_spec.py`, END is optional!  That is weird.

PR #402, opened 2019-07-06
Allow prompts to contain newline

jyn514:
Addresses https://github.com/oilshell/oil/issues/394#issuecomment-508942298

### DIFF
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -32,10 +32,12 @@ PROMPT_REVERSE = '\x01%s\x02' % _REVERSE
 def _PromptLen(prompt_str):
   """Ignore all characters between \x01 and \x02 and handle unicode characters.
   In particular, the display width of a string may be different from either the
-  number of bytes or the number of unicode characters."""
+  number of bytes or the number of unicode characters.
+  Additionally, if there are multiple lines in the prompt, only give the length of the last line."""
   escaped = False
   display_str = ""
-  for c in prompt_str:
+  last_line = prompt_str.split('\n')[-1]
+  for c in last_line:
     if c == '\x01':
       escaped = True
     elif c == '\x02':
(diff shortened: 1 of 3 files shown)


core/comp_ui.py
```diff
@@ -32,18 +32,23 @@
...
       escaped = True
     elif c == '\x02':
       escaped = False
     elif not escaped:
       display_str += c
   try:
-    return libc.wcswidth(display_str)
+    width = libc.wcswidth(display_str)
+    if width == -1:
+        return len(display_str)
```
andychu: Please move this outside the try block and use 2-space indent.

IIRC PEP 8 says somewhere to minimize the scope of try blocks.

(It would be nice to set up `test/lint.sh` to check indents)

core/comp_ui.py
```diff
@@ -32,18 +32,23 @@
 def _PromptLen(prompt_str):
   """Ignore all characters between \x01 and \x02 and handle unicode characters.
   In particular, the display width of a string may be different from either the
-  number of bytes or the number of unicode characters."""
+  number of bytes or the number of unicode characters.
+  Additionally, if there are multiple lines in the prompt, only give the length of the last line."""
   escaped = False
   display_str = ""
-  for c in prompt_str:
+  last_line = prompt_str.split('\n')[-1]
```
andychu: OK now I get your question more... Although I was sort of agnostic about this, it does seem better to strip `\x01` first, and then take the last line of `display_str`?  

Because otherwise you will count stuff that the user intends for you to ignore.   My interpretation of `x01` is "ignore everything inside here when calculating the length".

jyn514: I guess if `\x01` is wrong, that's a user error ... It might nice to give a warning though. Say you have `PS1=$'abc\x01def\n\033[00m\x02> '`, then the prompt will look like it has 2 characters (`> `) but we'll calculate it as having 5 (`abc> `). It will be wrong either way we calculate it (splitting the line first won't ignore the ANSI escapes)

jyn514:
Could you take another look at this?

andychu:
Great, thanks for fixing this!  I merged and verified it manually.

Issue #135, opened 2018-06-04
Consider "vendoring" re2c and getting rid of build/dev.sh minimal


andychu:
This reduces the possibility of test breakages with and without `fastlex`.

Although this would mean we need to install re2c on Travis?  Or maybe just provide a pre-built binary somewhere?

andychu:
Related to issue #159 

Instead of the source, we can publish binaries (along with scripts to reproduce them)

andychu:
#187 (continuous build) and #513 (Nix) make this less interesting.  It's not a special case

PR #411, opened 2019-07-07
Add POSIX compatibility for release scripts

jyn514:
These are the issues that popped up when following the instructions for creating the release tarball (from https://github.com/oilshell/oil/wiki/Contributing#three-kinds-of-builds). More will probably crop up at some point.

### DIFF
--- a/Python-2.7.13/Parser/asdl_c.py
+++ b/Python-2.7.13/Parser/asdl_c.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python2
 """Generate C code from an ASDL description."""
 
 # TO DO
--- a/build/actions.sh
+++ b/build/actions.sh
@@ -192,7 +192,7 @@ join-modules() {
 clean-pyc() {
   # skip _chroot, _tmp, etc.  But delete __init__.pyc
   find . \( -type d -a -name '_*' -a -prune \) -o -name '*.pyc' -a -print |
-    xargs --no-run-if-empty -- rm --verbose
+    xargs -- rm -f -v
 }
 
 # NOTE: Not deleting _devbuild, including cpython-full.  Maybe we should, or
--- a/build/codegen.sh
+++ b/build/codegen.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # For textual code generation.
 #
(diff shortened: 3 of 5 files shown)


jyn514:
Note that the last commit changing `./configure` is in the release tarball. All other changes are to the scripts which create the tarball.

andychu:
I'm not sure how I feel about this strategy.  It doesn't really "scale", in the sense that there about 10K lines of other "devtools" shell scripts in the repo that would require similar changes.

That's more or less why I added the first line here:

I think the better strategy would be to write some kind of "sh-env" akin to virtualenv, which modifies $PATH so that bash and coreutils are in it, in a sandbox.  And then I guess change shebangs to `/usr/bin/env bash`.  Or it probably already exists?

Related to #135 -- re2c could go in the sh-env

Although maybe we should do some feature detection of bash and coreutils up front to prevent errors.

andychu:
Also courtesy of @drwilly , we have our own `tools/find/xargs.py` that we can test here instead of changing the xargs invocation.  i.e. it accepts `--no-run-if-empty`

So yeah I think it would be better to port EVERYTHING at once via some kind of `sh-env`.

c.f. "It's easier to port a shell than a shell script" -- Larry Wall

jyn514:
I guess since this is for the development environment, that makes sense. It makes it difficult to test on other platforms though :/ I like having at least the release scripts compatible so I can make changes on the platform I'm testing out. If we limit the changes to scripts in `./build`, there's really not that much that has to change.

---

Note that coreutils may not exist on every platform, at least not by the same names as on linux. See https://www.freebsdsoftware.org/sysutils/coreutils.html, most of them will be called `gexpr` instead of `expr`. Not sure how you could change the environment in an automated way to smooth that out.

jyn514:
Additionally the changes to `./configure` and `build/actions.sh` do go in the tarball, so I think it makes sense to worry about compatibility there.

andychu:
The `expr` change is good -- I would make a new PR with just that change, and add a pointer to:

https://www.freebsd.org/cgi/man.cgi?query=expr&sektion=&n=1

Let me think about what to do with the rest... 

`build/actions.sh` is in the tarball, but `clean-repo` doesn't make sense there.  The Makefile is intentionally used in both places (dev machine and user machine), which does make it a bit confusing.

andychu:
Thanks for investigating this!  Although as mentioned I don't want to go backward to POSIX for all the dev scripts.  The release tarball can be made on Linux and built on other OSes.

PR #412, opened 2019-07-07
Remove dependency on bash in tarball

jyn514:
**NOT READY FOR MERGE**

This removes `set -o pipefail` without supplying a replacement, I
don't see a way around this using POSIX sh (without writing a monster library, see https://stackoverflow.com/a/54931544)

This leaves the OIL_SYMLINKS and OPY_SYMLINKS arrays unchanged
even though they're incompatible because I'm not sure if they should be
turned into string variables (like INCLUDE_PATHS) or moved to a different
file, since they're not used in the tarball. @andychu which would you prefer?

### DIFF
--- a/build/actions.sh
+++ b/build/actions.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 #
 # Build actions used in the Makefile.
 #
@@ -6,10 +6,9 @@
 #   ./actions.sh <function name>
 
 set -o nounset
-set -o pipefail
 set -o errexit
 
-source build/common.sh
+. build/common.sh
 
 write-release-date() {
   mkdir -p _build  # Makefile makes this, but scripts/release.sh needs it too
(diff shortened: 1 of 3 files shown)


andychu:
This one makes more sense since it's in the release tarball.  However my idea for moving to POSIX sh for end users is to finish the `mycpp` translation.  Then build/{actions,compile} would largely go away, and there would be some simple `build` script like:

cc -o oil.ovm oil_file{1,2,3}.cc

i.e. just a single line to compile a bunch of files, not Makefile or bash.

build/{actions,compile} are mixed between dev time and end user now.  If we were going to remove the deps, I'd rather do that separation more cleanly.  I like using bash at dev time, and OSH also works.

For developers, we should really just have everyone use OSH.  That will solve the `pipefail` problem.

jyn514:
I'm not sure what's used at dev time and compile time or both for the scripts, I might let you work that out.

I have no problem with requiring bash at development time, this is exclusively for the release tarball.

jyn514:
I'm not planning to follow up on this.

Issue #167, opened 2018-08-14
Release automation: run spec tests with just OSH on different platforms


andychu:
For OS X, BSDs, etc. where we might not want to bother building with `spec-bin.sh`

And maybe ARM Linux would be a good one too.  Although it would be nice to build binaries there.

andychu:
This is is more significant now that we have a `__GLIBC__` test in `native/libc.c` for `extglob`.

In other words, some things don't work on Alpine now, and it would be nice to see that in the tests.

andychu:
TODO: What are dependencies of spec tests?

- Python 2
- bash
- coreutils (including nproc #415)


Anything else?

I guess since they are confined to `test/spec.sh` and `test/spec-runner.sh`, we could make these more like POSIX scripts.  I guess it depends what we do with Python 2, which is needed to run `test/sh_spec.py`.

If we can use the same portability solution for all 3 deps, that would be nice.

Note `test/spec-bin.sh` builds binaries of other shells.  We don't technically need it for this task.  However we could use the same solution to build Python 2, bash, coreutils (which is admittedly onerous, but I guess BSDs are source-based anyway)

andychu:
@jyn514 I realized the spec tests have a fairly significant gawk dependency.  The `match()` function is a useful extension -- see `test/spec-runner.sh`

So ideally we could package up these 4 dependencies somehow.   I don't think they are particularly weird or exotic.

I think putting `gawk` in $PATH is not too hard, as well as accounting for name differences with a tiny shell wrapper.

andychu:
I think having something like:

```
$REPO/
  sh-env/
    darwin/
      nproc
      gawk
      rm
    freebsd/
      nproc
      gawk
      rm
```

Would work.  Some of these can be symlinks.

It's similar to the philosophy of avoiding #ifdefs.  The Go/Plan 9 guys advocate that -- they avoid #ifdefs and instead create portable interfaces.

andychu:
This is kind of the opposite of #42 -- because we want to run at least some tests in a DIFFERENT environment!

andychu:
Updated dependencies:

- Python 2 for `sh_spec.py` 
  - this was in Python 3 at some point, but I dropped support for it, woudln't be too hard to port again.  I think we mostly need `b''` everywhere.
- bash or known working version of Oil (might have arrays, associative arrays, etc. so not just POSIX)
- POSIX `rm`, etc.

OPTIONAL:

[...]

PR #413, opened 2019-07-07
Don't crash when parsing regexes on FreeBSD

jyn514:
Before:
```
osh$ [[ 1 =~ [0-9] ]]
Segmentation fault (core dumped)
```

After:
```
osh$ [[ 1 =~ [0-9] ]]
osh$ echo $?
0
```

This happened because libc.c used the GNU regex library headers but
used the BSD library when dynamically linked. I'm not sure why it was
using different headers, maybe Python was originally statically linked?

### DIFF
--- a/native/libc.c
+++ b/native/libc.c
@@ -16,11 +16,7 @@
 
 #include <fnmatch.h>
 #include <glob.h>
-#ifdef __FreeBSD__
-#include <gnu/posix/regex.h>
-#else
 #include <regex.h>
-#endif
 
 #include <Python.h>
 


andychu:
FWIW someone explicitly added the #ifdef like 2 years ago, and I'm not sure why... but I'm inclined to merge this change, since I never understood the point.  FreeBSD should have a POSIX regex in its libc.

https://github.com/oilshell/oil/commit/7ad8ebe3ab5c22b67419120a430abf957de547cc

I think #167 is more important now.  There is a significant difference between the dynamic "dev build" and the static "release build", and the spec tests often find bugs there.

jyn514:
I think that makes a good argument for making the spec scripts portable. I started working on that but you said that development is only intended to work on Linux so I never finished it.

andychu:
Yeah I'm still thinking it's easier to install coreutils and such, and modify $PATH to point to those.  Someone did that here on OS X, with `brew`:

https://github.com/oilshell/oil/pull/194/files

My problem with the other way is that it's a constant tax on development.  There's basically no way to test on Linux if a script works on OS X or BSD, rather than by memorizing a bunch of POSIX arcana.  The whole point of the Oil project is to improve over POSIX :)

It's the same "porting a shell vs. porting a shell script" argument.

Practically speaking, we could port the scripts, but they would probably keep breaking over and over again.  It would be nicer to have an environment that's the same on all OSes (which I don't think we're that far away from, given the small number of dependencies)

Practically speaking it looks like I should merge the `nproc` changes, since everyone keeps running into that, and spec tests depend on that I think.

It would be interesting if we could run with the xargs from @drwilly , because spec tests use that too.

Issue #159, opened 2018-07-27
Publish developer binaries built with test/spec-bin.sh


andychu:
This scripts builds stable version of bash/dash/mksh/zsh/busybox that we run spec tests against.  This is the mitigation for issue #42.

It would be nice to make some portable Linux binaries so that everyone doesn't have to build them (or run Ubuntu 16.04).

Binaries to build:

- [ ] shells
- [ ] re2c -- to get rid of the pure Python path?
- [ ] A modified version of CPython that we run with?  Not sure if this is necessary.

NOTE: We also use Clang, but their binary releases seem OK.

andychu:
Several people have been able use `test/spec-bin.sh`, and we also have Nix in progress, so closing this

PR #414, opened 2019-07-07
Take a stab at PROMPT_COMMAND

jyn514:
Location information is wrong - PROMPT_COMMAND=';' gives a traceback
pointing to ~/.local/oil/oshrc instead of [ interactive ] or
[ PROMPT_COMMAND ]

Other than that, this seems to be working pretty well with my patched version of bash-git-prompt.

Addresses https://github.com/oilshell/oil/issues/400

### DIFF
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -18,8 +18,10 @@ from _devbuild.gen.syntax_asdl import (
     command_t, command,
     parse_result__EmptyLine, parse_result__Eof, parse_result__Node
 )
+from _devbuild.gen.runtime_asdl import value__Undef, arg_vector
 from core import ui
 from core import util
+from frontend import reader
 #from core.util import log
 
 from typing import Any, Optional, List, TYPE_CHECKING
@@ -43,6 +45,10 @@ def Interactive(opts, ex, c_parser, display, errfmt):
 
     while True:  # ONLY EXECUTES ONCE
       try:
+        prompt_command = ex.mem.GetVar('PROMPT_COMMAND')
+        if not isinstance(prompt_command, value__Undef):
+            arg_vec = arg_vector(['', prompt_command.s], [0, 0])
+            ex._Eval(arg_vec)
         # may raise HistoryError or ParseError
         result = c_parser.ParseInteractiveLine()
         if isinstance(result, parse_result__EmptyLine):
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -40,10 +40,12 @@ def _ExecutePromptCommand(ex):
   command = prompt_var.s
 
   # save this so PROMPT_COMMAND can't set $?
-  old_status = ex.LastStatus()
+  ex.mem.PushStatusFrame()
   arg_vec = arg_vector(['PROMPT_COMMAND', command], [0, 0])
-  ex._Eval(arg_vec)
-  ex.mem.SetLastStatus(old_status)
+  try:
+    ex._Eval(arg_vec)
+  finally:
+    ex.mem.PopStatusFrame()
 
 
 def Interactive(opts, ex, c_parser, display, errfmt):
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -33,16 +33,11 @@ if TYPE_CHECKING:
 
 def _ExecutePromptCommand(ex):
   # type: (Any) -> None
-  command = None
   prompt_var = ex.mem.GetVar('PROMPT_COMMAND')
-  if prompt_var.tag == value_e.Undef:
+  # Undefined, Array, or AssociativeArray
+  if prompt_var.tag != value_e.Str:
     return
-  elif prompt_var.tag == value_e.AssocArray:
-    command = prompt_var.d['0']
-  elif prompt_var.tag == value_e.StrArray:
-    command = prompt_var.strs[0]
-  elif prompt_var.tag == value_e.Str:
-    command = prompt_var.s
+  command = prompt_var.s
 
   # save this so PROMPT_COMMAND can't set $?
   old_status = ex.LastStatus()
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -18,11 +18,9 @@ from _devbuild.gen.syntax_asdl import (
     command_t, command,
     parse_result__EmptyLine, parse_result__Eof, parse_result__Node
 )
-from _devbuild.gen.runtime_asdl import value__Undef, arg_vector
+from _devbuild.gen.runtime_asdl import value_e, arg_vector
 from core import ui
 from core import util
-from frontend import reader
-#from core.util import log
 
 from typing import Any, Optional, List, TYPE_CHECKING
 if TYPE_CHECKING:
@@ -31,6 +29,25 @@ if TYPE_CHECKING:
   from osh.cmd_parse import CommandParser
   # commented out so --strict doesn't follow all
   #from osh.cmd_exec import Executor
+
+
+def _ExecutePromptCommand(ex):
+  command = None
+  prompt_var = ex.mem.GetVar('PROMPT_COMMAND')
+  if prompt_var.tag == value_e.Undef:
+    return
+  elif prompt_var.tag == value_e.AssocArray:
+    command = prompt_var.d['0']
+  elif prompt_var.tag == value_e.StrArray:
+    command = prompt_var.strs[0]
+  elif prompt_var.tag == value_e.Str:
+    command = prompt_var.s
+
+  # save this so PROMPT_COMMAND can't set $?
+  old_status = ex.LastStatus()
+  arg_vec = arg_vector(['PROMPT_COMMAND', command], [0, 0])
+  ex._Eval(arg_vec)
+  ex.mem.SetLastStatus(old_status)
 
 
 def Interactive(opts, ex, c_parser, display, errfmt):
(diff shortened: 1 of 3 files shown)


andychu:
I think this needs to be similar in spirit to WordParser.EvalForPlugin() and Tracer in core/dev.py.

PS4 is evaluating something for every line for `set -x`.  And `PROMPT_COMMAND` is for every line of the interactive shell.

For example I believe it should avoid changing `$?` -- I think you reported that bug with $PS1 or something.

That said, if you can add a test case in `spec/interactive.test.sh`, and make it pass for bash and osh, then we can merge that and figure out all the details in a subsequent change.  (e.g. followups for location info.)

jyn514:
My local `.profile` is messing up the tests for bash. Would you be willing to consider an ignored `--noprofile` argument for osh? I didn't open an issue since it would basically only be useful for tests.

andychu:
I usually do something like:

```
case $SH in *bash) flag='--noprofile` ;; esac
$SH $flag -i ...
```

Does that work?

jyn514:
Hmm, that worked which makes me confused why the tests aren't passing. I'm running `test/spec.sh interactive` and bash fails all the tests. The weird thing is if I run the command manually, without going through `test/spec.sh`, it works fine. Here's the test case I'm looking at:

```
#### fatal errors continue

# NOTE: tried here doc, but sys.stdin.isatty() fails.  Could we fake it?
$SH -i -c '
echo $(( 1 / 0 ))
echo one
exit 42
'
## status: 42
## STDOUT:
one
## END
```

Running the command through spec.sh:
```
test/spec.sh interactive -v
interactive.test.sh
  0       4     N-I     FAIL    N-I     pass    'exit' in oshrc (regression)
[bash stdout] Expected 'one\n', got 'To run a command as administrator (user "root"), use "sudo <command>".\nSee "man sudo_root" for details.\n\none\n'
bash stdout:
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

one
```
Running manually:
```
$ /bin/bash --noprofile --rcfile /home/joshua/Documents/Programming/oil/_tmp/spec-tmp/interactive.test.sh/oshrc -i -c 'echo hello'
one
```

andychu:
OK weird, I reproduced this on another machine.  Let me see what's going on!

andychu:
Weirdly it works on Travis:

https://travis-ci.org/oilshell/oil/builds/555543685

andychu:
I don't understand why `/etc/bash.bashrc` is being run... but somehow it is... still looking.

https://askubuntu.com/questions/22607/remove-note-about-sudo-that-appears-when-opening-the-terminal

andychu:
OK I think this is system bash vs. hermetic bash from `test/spec-bin.sh`.  Although Travis maybe has a newer version of Ubuntu.

I think Ubuntu must have patched bash to run `/etc/bash.bashrc` even if `--rcfile` is passed??  That is annoying.  And maybe they got rid of it in a later version of Ubuntu.



```
andy@lisa ~/git/oilshell/oil (master)$ bash --version
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)
```

```
andy@lisa ~/git/oilshell/oil (master)$ _tmp/spec-bin/bash --version
GNU bash, version 4.4.0(1)-release (x86_64-unknown-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

Here's my Ubuntu:

```
andy@lisa ~/git/oilshell/oil (master)$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.2 LTS"
```

andychu:
I mention the `spec-bin` stuff here, along with issue #159 

I guess you just have to compile bash once, and it should go away.  And it's better to do that anyway so all the rest of the tests pass.  There are definitely differences between bash 4.3 and bash 4.4, so if your system has 4.3 it's worth the time.

jyn514:
I set `HOME` instead and that seems to have fixed it.

jyn514:
I added a test case, could you take another look?

core/main_loop.py
```diff
@@ -31,6 +31,26 @@
...
 
 
+def _ExecutePromptCommand(ex):
+  # type: (Any) -> None
+  command = None
+  prompt_var = ex.mem.GetVar('PROMPT_COMMAND')
+  if prompt_var.tag == value_e.Undef:
+    return
+  elif prompt_var.tag == value_e.AssocArray:
+    command = prompt_var.d['0']
```
andychu: This is what bash does, but OSH explicitly disallows it now.  That is `$a` is not equivalent to `${a[0]}` (I made this an error in the last few days, but it always behaved differently than bash.)

I would just `return` unless it's a string.

core/main_loop.py
```diff
@@ -31,6 +31,26 @@
...
+  elif prompt_var.tag == value_e.StrArray:
+    command = prompt_var.strs[0]
+  elif prompt_var.tag == value_e.Str:
+    command = prompt_var.s
+
+  # save this so PROMPT_COMMAND can't set $?
+  old_status = ex.LastStatus()
+  arg_vec = arg_vector(['PROMPT_COMMAND', command], [0, 0])
+  ex._Eval(arg_vec)
```
andychu: Please use PushStatusFrame() / PopStatusFrame(), and use a try/finally so that it gets restored in the case of exceptions (which happen a lot in OSH!)

It would be nice to test that in the spec tests, though I'm not sure if it's possible to tickle.

spec/interactive.test.sh
```diff
@@ -51,3 +55,13 @@ RCFILE
 ## N-I dash status: 2
 ## N-I mksh status: 1
 
+#### interactive shell runs PROMPT_COMMAND after each command
+TEST_CASE='PROMPT_COMMAND="echo hi"'
+case $SH in
+	*bash) echo "$TEST_CASE" | $SH --noprofile --norc -i;;
+	*osh) $SH --rcfile /dev/null -i -c "$TEST_CASE";;
+	*) $SH -i -c "$TEST_CASE";;
```
andychu: Hm also this seems to be doing a ton of stuff that's different for each shell.

Are you trying to avoid the spurious stdout from bash?

I think if doing the `test/spec-bin.sh` stuff in your environment fixes it, then we should go with that instead.  I added `test/spec.sh interactive` on Travis, and it doesn't have that problem.

For now that is sort of the "golden environment".

jyn514: Yes that should have been commented. This isn't an issue with spurious output, this is because osh and bash have different ideas of what it means to be interactive.

`bash -i -c 'PROMPT_COMMAND="echo hi"'` does not give any output, but `echo 'PROMPT_COMMAND="echo hi"'' | bash -i` does.

osh goes the other way: `echo 'PROMPT_COMMAND="echo hi"' | osh -i` prints the prompt: `osh$ osh$ ^D`

I agree not having the same test for both isn't ideal but I didn't see another way to check this.

jyn514:
Speaking of exceptions, this patch crashes osh if PROMPT_COMMAND throws an exception:

```
(osh) ~/.../Programming/oil ▶️ echo 'PROMPT_COMMAND=""; :' | bin/^C
(osh) ~/.../Programming/oil ▶️ PROMPT_COMMAND='which which'
/usr/bin/which
(osh) ~/.../Programming/oil ▶️ prlimit --nproc=1 --pid=$$
osh I/O error: Resource temporarily unavailable
(bash) ~/src/oil ▶️
```

I'm looking into how `EvalForPlugin` works.

jyn514:
I'm not sure how to treat the word returned by EvalForPlugin as a command substitution, I have the following code but it just prints the word instead of running a command.

```
  w_parser = parser.parse_ctx.MakeWordParserForPlugin(command)
  try:
    ps1_word = w_parser.ReadForPlugin()
  except util.ParseError as e:
    ps1_word = word.ErrorWord("<ERROR: Can't parse PS1: %s>", e)
    _parse_cache[ps1_str] = ps1_word

  # Evaluate, e.g. "${debian_chroot}\u" -> '\u'
  print(ex.word_ev.EvalForPlugin(ps1_word).s)
```

andychu:
I think that behavior is OK for now.

I will probably do some refactoring, i.e. instead of _Eval, add something like RunFuncForCompletion.  But if some new tests pass, then I won't worry too much about the details.

OK thanks for letting me know about the bash/osh difference.  Please add a comment about that in the test.

My main issue now is setting $HOME because your version of bash is reading a file in `~`.  That is sort of regressing against issue $42.

We could set $HOME to something `_tmp/spec-tmp/dummy-home`, but it shouldn't be the actual HOME dir, which can have arbitrary files.

I think it will pass on Travis without that, so please remove it for now, and then we can figure out what to do if it's still failing on your machine.

jyn514:
Ok I took HOME out, let me know if there's any more changes I should make. The tests are failing locally, I'm going to try building bash locally finally.

andychu:
Great, yes feel free to update the wiki!

andychu:
BTW I got the tests working roughly the same under bash/OSH:

https://github.com/oilshell/oil/commit/82d22cc0c24a6b76db752eff0b247fb259e9a0bb

Thanks for adding this!

PR #421, opened 2019-07-11
Add changes to ./configure to build on FreeBSD

jyn514:
See https://github.com/oilshell/oil/pull/411#issuecomment-509020340 and https://github.com/oilshell/oil/issues/415#issuecomment-509854207 for discussion.

### DIFF
--- a/configure
+++ b/configure
@@ -12,7 +12,7 @@
 # Other settings: LTO, PGO?  Consider moving prefix, LTO, PGO to build and
 # install steps.
 
-TMP=${TMP:-/tmp}  # Assume that any system has $TMP set or /tmp exists
+TMP=${TMPDIR:-/tmp}  # Assume that any system has $TMP set or /tmp exists
 readonly TMP  # POSIX sh supports 'readonly'
 
 log() {
--- a/configure
+++ b/configure
@@ -66,7 +66,7 @@ while true; do
       ;;
 
     --readline=*)
-      FLAG_readline=$(expr "$1" : '--readline=\(.*\)')
+      FLAG_readline=$(expr \( "$1" : '--readline=\(.*\)' \) )
       ;;
     --readline)
       if test $# -eq 1; then
--- a/configure
+++ b/configure
@@ -12,7 +12,7 @@
 # Other settings: LTO, PGO?  Consider moving prefix, LTO, PGO to build and
 # install steps.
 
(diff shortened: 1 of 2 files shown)


jyn514:
I'll do one better, this builds on Android as well :)

configure
```diff
@@ -12,7 +12,7 @@
 # Other settings: LTO, PGO?  Consider moving prefix, LTO, PGO to build and
 # install steps.
 
-TMP=${TMP:-/tmp}  # Assume that any system has $TMP set or /tmp exists
+TMP=${TMPDIR:-/tmp}  # Assume that any system has $TMP set or /tmp exists
```
andychu: The comment is now out of date.  Does FreeBSD not have either TMP or /tmp ?

configure
```diff
@@ -66,7 +66,7 @@ while true; do
       ;;
 
     --readline=*)
-      FLAG_readline=$(expr "$1" : '--readline=\(.*\)')
+      FLAG_readline=$(expr \( "$1" : '--readline=\(.*\)' \) )
```
andychu: Please add a comment saying that `\(` is necessary for FreeBSD (which is honestly a great example of sucky language design in shell tools).

And link to the manual  too:

https://www.freebsd.org/cgi/man.cgi?query=expr&sektion=&n=1

configure
jyn514: Android does not have /tmp or $TMP, but it has $TMPDIR. FreeBSD has /tmp but not $TMP or $TMPDIr.

jyn514:
I'm not using the NDK, I have [termux](https://github.com/termux/termux-app) installed. It has a mostly POSIX environment, minus a few quirks like missing /tmp and needing -landroid-glob for `glob` and `globfree`.

configure
```diff
@@ -12,7 +12,7 @@
 # Other settings: LTO, PGO?  Consider moving prefix, LTO, PGO to build and
 # install steps.
 
-TMP=${TMPDIR:-/tmp}  # Assume that any system has $TMP set or /tmp exists
+TMP=${TMPDIR:-/tmp}  # Assume that any system has $TMPDIR set or /tmp exists
```
andychu: Hm I was wondering why you removed TMP.  But this answer seems to say that TMPDIR or /tmp should exist.

https://unix.stackexchange.com/questions/174817/finding-the-correct-tmp-dir-on-multiple-platforms

andychu: Ah OK I see you mentioned that in the commit description.  Thanks.

Issue #430, opened 2019-07-16
exit OSH with 'Ctrl-D' tries to render one last time, which can have side effects


jyn514:
Found this from https://github.com/oilshell/oil/issues/427 - when you exit OSH with 'Ctrl-D', the prompt reader tries to render one last time.

Current behavior:
```
osh$ PS1='$(time sleep 1)'
real    1.013
user    0.006
sys     0.000
^D
real    1.010
user    0.000
sys     0.005
```

Requested behavior (same as bash):
```
osh$ PS1='$(time sleep 1)'
real    0m1.005s
user    0m0.002s
sys     0m0.003s
^D
```
Strangely, this only happens for Ctrl-D, using `exit` behaves as expected.

jyn514:
Tracked this down to a hacky looking double while loop in core/main_loop.py. Going to take a stab at it.

andychu:
Released with 0.7.pre1: https://www.oilshell.org/release/0.7.pre1/

PR #431, opened 2019-07-16
Don't run c_parser.Reset when exiting

jyn514:
Remove while loop with one iteration in favor of `return`
- `is_done` mistakenly ran c_parser.Reset, so I replaced it with a
`return`
- The end of the while loop was only executed when parse_result was a
Node, so I moved it to that branch instead of keeping it at the bottom
- `display.EraseLines` was called before each condition in the `try`
clause, so I moved it out of the conditions to just after parsing
- The funny-looking `pass` at the end of the exception handling is
because Python expects code after an indented comment for some reason

Addresses https://github.com/oilshell/oil/issues/430

The detailed comments are because the change in indentation makes it really hard to read the diff.

### DIFF
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -34,67 +34,53 @@ if TYPE_CHECKING:
 def Interactive(opts, ex, c_parser, display, errfmt):
   # type: (Any, Any, CommandParser, Any, ErrorFormatter) -> Any
   status = 0
-  done = False
-  while not done:
-    # - This loop has a an odd structure because we want to do cleanup after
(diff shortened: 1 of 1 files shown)


jyn514:
Tested like this:
```
osh$ PS1='$(time sleep 1)'
real    1.013                                                                                
user    0.006                                                                                
sys     0.000                                                                                
^D
```

andychu:
I fixed #430 a different way... this loops is tricky and hard to test, so I don't think it's worth disturbing unless there's something seriously wrong.

PR #433, opened 2019-07-17
[devtools] Add make-all helper function to spec-bin.sh

jyn514:
Instead of typing 5 similar commands in a row, let the shell do it for me.

### DIFF
--- a/test/spec-bin.sh
+++ b/test/spec-bin.sh
@@ -13,6 +13,9 @@
 #   test/spec-bin.sh copy-all     # Put them in _tmp/spec-bin
 #   test/spec-bin.sh test-all     # Run a small smoke test
 #
+# To run all commands at once (with minimal error-checking):
+#   test/spec-bin.sh all-steps     # Download, extract, compile, and test all shells
+#
 # Could also build these:
 # - coreutils
 # - re2c for the OSH build (now in build/codegen.sh)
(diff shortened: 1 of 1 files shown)


test/spec-bin.sh
```diff
@@ -176,4 +172,12 @@ publish-tmp() {
   scp _deps/re2c-1.0.3/re2c ${name}@${name}.org:$dest
 }
 
+make-all() {
+  download     # Get the right version of every tarball
+  extract-all  # Extract source
+  build-all    # Compile
+  copy-all     # Put them in _tmp/spec-bin
+  test/spec-bin.sh test-all     # Run a small smoke test
```
andychu: don't need the script name here

test/spec-bin.sh
```diff
@@ -7,11 +7,7 @@
 #   ./spec-bin.sh <function name>
 #
 # Instructions:
-#   test/spec-bin.sh download     # Get the right version of every tarball
-#   test/spec-bin.sh extract-all  # Extract source
-#   test/spec-bin.sh build-all    # Compile
-#   test/spec-bin.sh copy-all     # Put them in _tmp/spec-bin
-#   test/spec-bin.sh test-all     # Run a small smoke test
+#   test/spec-bin.sh make-all     # Download, extract, compile, and test all shells
```
andychu: I like the comment as is because it tells people to run steps separately... but you can add a note about the "all-in-one" function too

jyn514:
I changed this to use `set -e`, not sure if that will catch all (or even most errors).

andychu:
We already have `set -o errexit` at the top, which is the same thing... But I will just remove it myself.

PR #437, opened 2019-07-18
Read in unbuffered mode for `read -n`

jyn514:
Puts TTY into 'noncanonical' mode for duration of read
Calls posix.read n times, since each time will only return 1 byte
Restore original mode even if an exception occurs

Addresses https://github.com/oilshell/oil/issues/274

### DIFF
--- a/demo/read-n-demo.sh
+++ b/demo/read-n-demo.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+read-one () {
+	# try typing a letter without hitting newline
+	# the shell should respond immediately
+	read -n 1
+	echo $'\n'"$REPLY"
+}
+
+read-multiple () {
+	read -n 5
+	echo $'\n'"$REPLY"
+}
+
+all () {
+	echo "one char:"
+	read-one
+	echo "multiple chars:"
+	read-multiple
+}
+
+"$@"
--- a/demo/read-n-demo.sh
+++ b/demo/read-n-demo.sh
@@ -2,22 +2,24 @@
 set -euo pipefail
 
 read-one () {
-	# try typing a letter without hitting newline
-	# the shell should respond immediately
-	read -n 1
-	echo $'\n'"$REPLY"
+  # try typing a letter without hitting newline
+  # the shell should respond immediately
+  read -n 1
+  echo
+  echo "$REPLY"
 }
 
 read-multiple () {
-	read -n 5
-	echo $'\n'"$REPLY"
+  read -n 5
+  echo
+  echo "$REPLY"
 }
 
 all () {
-	echo "one char:"
-	read-one
-	echo "multiple chars:"
-	read-multiple
+  echo "one char:"
+  read-one
+  echo "multiple chars:"
+  read-multiple
 }
 
 "$@"
--- a/osh/builtin.py
+++ b/osh/builtin.py
@@ -518,13 +518,13 @@ class Read(object):
         name = names[0]
       except IndexError:
         name = 'REPLY'  # default variable name
+      s = ""
       if sys.stdin.isatty():  # set stdin to read in unbuffered mode
         orig_attrs = termios.tcgetattr(stdin)
         attrs = termios.tcgetattr(stdin)
         # disable canonical (buffered) mode
         # see `man termios` for an extended discussion
         attrs[3] &= ~termios.ICANON
-        s = ""
         try:
           termios.tcsetattr(stdin, termios.TCSANOW, attrs)
           # posix.read always returns a single character in unbuffered mode
@@ -534,7 +534,14 @@ class Read(object):
         finally:
           termios.tcsetattr(stdin, termios.TCSANOW, orig_attrs)
       else:
-        s = posix.read(stdin, arg.n)
+        s_len = 0
+        while arg.n > 0:
+          buf = posix.read(stdin, arg.n)
+          # EOF
+          if buf == '':
+            break
+          arg.n -= len(buf)
+          s += buf
 
       state.SetLocalString(self.mem, name, s)
       # NOTE: Even if we don't get n bytes back, there is no error?
--- a/osh/builtin.py
+++ b/osh/builtin.py
@@ -26,6 +26,7 @@ http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_
 """
 from __future__ import print_function
 
+import termios # for read -n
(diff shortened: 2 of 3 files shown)


osh/builtin.py
```diff
@@ -512,12 +513,25 @@ def __call__(self, arg_vec):
...
       try:
         name = names[0]
       except IndexError:
         name = 'REPLY'  # default variable name
-      s = posix.read(sys.stdin.fileno(), arg.n)
-      #log('read -n: %s = %s', name, s)
+      orig_attrs = termios.tcgetattr(stdin)
+      attrs = termios.tcgetattr(stdin)
+      # disable canonical (buffered) mode
+      # see `man termios` for an extended discussion
```
andychu: My understanding  is a little fuzzy here, but I don't think this makes sense because stdin isn't always a terminal.

Well first I'm not clear exactly what problem this is solving.  Do we have a test that fails with the old code and passes with this code (whether manual or automated)?

Is it that you want `read -n 1` to return after a keypress?  OK I see that doesn't work in OSH.

Why doesn't it work?  I don't think Our posix.read(0, 1) call is supposed to buffer anything?  If I look at the source to `native/posixmodule.c`, it calls `read()` directly.  So is the source of buffering in libc?

But what happens when stdin is not a terminal?  e.g. a file?   `read -n 1 < myfile`

I looked at the source to bash and dash.  It looks like they both call `read(1)` (a single byte) ALL the time.   They don't appear to do anything with the terminal settings ?

Anyway I guess I'm interested in an analysis of what doesn't work about the current code, and what other shells do.  It does seem like something is wrong, but I don't know what.

andychu: Hm Ok I missed the strace... let me look at that.  I was looking at the bash source but didn't see it doing anything like that.

andychu: OK now I see in `builtins/read.def`, bash does `ttfd_onechar()`, which presumably does something just like this.

dash doesn't implement `read -n`, so it doesn't apply.

OK I will look at this in more detail when I get back.  Thanks for the patch.

jyn514: That's a good point about non-interactive stdin, I'll add some error handling for that case.

jyn514: The buffering I think is happening below libc in the kernel drivers for the terminal. I don't have an automated test case, but if you do `read -n 2` it should read two characters _before_ you hit newline, which it wasn't doing before.

jyn514:
Note that the test case I just added for `read -n` from input that never ends will fail until osh gives status 124 on SIGPIPE.

jyn514:
I was wrong about the tests IN https://github.com/oilshell/oil/issues/438 - the issue is that this only works when stdin is interactive! I don't think there's a way to test that non-interactively. Is there somewhere I can put tests that need user input? I found https://github.com/oilshell/oil/wiki/Testing-the-Interactive-Shell but I didn't want to edit the wiki for a feature that hasn't been merged yet

andychu:
- The `demo/` dir is sort of a graveyard of one-off tests.  But adding something there is better than nothing!  It's useful for code reviews and for `git blame`.  
- Yes that wiki page is for stuff I do manually.  Feel free to edit it.
- I  filed #439 to come up with a better solution for tests that involve a terminal.  Probably a lot of the stuff in `demo/` could be automated.

jyn514:
I added a demo, could you take another look at this?

---

You mentioned a few times that you don't know why all shells read one character at a time. See http://pubs.opengroup.org/onlinepubs/9699919799.2016edition/utilities/sh.html under STDIN:
> When the shell is using standard input and it invokes a command that also uses standard input, ... [i]t shall not read ahead in such a manner that any characters intended to be read by the invoked command are consumed by the shell (whether interpreted by the shell or not) or that characters that are not read by the invoked command are not seen by the shell.

I haven't found a test case yet, if I think of one I'll let you know.

demo/read-n-demo.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+read-one () {
+	# try typing a letter without hitting newline
```
andychu: Please be consistent with 2 space indents everywhere.

Which editor do you use?  If vim, maybe we can add "modelines" so things automatically work.  I use vim.

osh/builtin.py
```diff
@@ -512,12 +513,35 @@ def __call__(self, arg_vec):
...
     if arg.n is not None:  # read a certain number of bytes
+      stdin = sys.stdin.fileno()
       try:
         name = names[0]
       except IndexError:
         name = 'REPLY'  # default variable name
-      s = posix.read(sys.stdin.fileno(), arg.n)
-      #log('read -n: %s = %s', name, s)
+      s = ""
+      if sys.stdin.isatty():  # set stdin to read in unbuffered mode
```
andychu: I guess you can use stdin.isatty() instead of sys.stdin

demo/read-n-demo.sh
```diff
@@ -0,0 +1,23 @@
...
+	# the shell should respond immediately
+	read -n 1
+	echo $'\n'"$REPLY"
```
andychu: very minor, but I think

```
echo
echo $REPLY
```

is more conventional and more readable

andychu:
Oh also, in zsh `read -n` apparently means something else.  `run-help read` shows that (which I just learned about a couple days ago).

So I would just do `case $SH in *zsh) return ;; esac` with a comment, and make the test run 1 second faster.

demo/read-n-demo.sh
jyn514: I use vim. Sorry this keeps being an issue, I just fixed it by putting configuration for this directory in my global vimrc (https://stackoverflow.com/a/18933951/7669110). Shouldn't happen again.

osh/builtin.py
jyn514: `stdin` is a number here, not an object. Calling `isatty` would through a reference error.

jyn514:
I didn't realize `read -n` wasn't standardized! The flag is different for `mksh` too, I changed it and now it passes the tests (other than looping indefinitely).

andychu:
@jyn514 `spec/builtins2` was failing for me after this commit.

I pushed a commit to fix it, and changed a few other things:

https://github.com/oilshell/oil/commit/0bfc953ae43ca3ef83cf3b8cffc2abbbc916814a

- exit 124 was from the `timeout` builtin
- The timeout should be avoided anyway because it makes the tests slow
- OSH apparently has a SIGPIPE problem we need to fix.  (zsh might too but it doesn't matter.)

PR #458, opened 2019-08-14
./configure and $DESTDIR

drwilly:
Hi Andy,
first draft. I added the `$DESTDIR` stuff, ignored a good bunch of common `configure` arguments and cleaned up various other stuff I found. As usual feel free to cherry-pick and/or comment.

### DIFF
--- a/configure
+++ b/configure
@@ -19,6 +19,10 @@ log() {
   echo "$0: $@" 1>&2
 }
 
+info() {
+  echo "$0 INFO: $@" 1>&2
+}
+
 die() {
   echo "$0 ERROR: $@" 1>&2
   exit 1
@@ -39,9 +43,17 @@ Detects system settings before a build of Oil.
 EOF
 }
 
-# Default installation is /usr/local/bin/oil, but this can be changed with
-# --prefix.  We roughly follow GNU:
+# This script roughly follows the GNU Standards
+# https://www.gnu.org/prep/standards/html_node/Configuration.html
 # https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
+#
+# Default installation is /usr/local/bin/oil, but this can be changed with
+# --prefix.
+#
+# While this script only uses a handful of the standard directory variables
+# listed on the above documents, it accepts most of them in the --arg=value
+# form as noops. This helps automated build-systems passing preconfigured
+# sets of arguments to configure oil.
 FLAG_prefix='/usr/local'
 FLAG_datarootdir=''  # default initialized after processing flags
 FLAG_with_readline=''  # Fail if it's not available.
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -68,42 +70,100 @@ while true; do
...
       shift
-      FLAG_datarootdir=$1
+      FLAG_datarootdir="$1"
+      ;;
+
+    --with-*|--enable-*)
+      info "Argument ignored '$1'"
+      ;;
+
+    --build=*)
```
andychu: Since we're not using $1, please collapse these flags for readability:
```
--build|--build=*)
  info "Ignored argument '$1'"
  ;;
```

Or I think you can even do all of them in a single column while still keeping the 80 char limit.

configure
```diff
@@ -68,42 +70,100 @@ while true; do
...
+      info "Argument ignored '$1'"
+      ;;
+    --sysconfdir)
+      info "Argument ignored '$1'"
+      shift
+      ;;
+    --localstatedir=*)
+      info "Argument ignored '$1'"
+      ;;
+    --localstatedir)
```
andychu: Please document where all these dirs come from, e.g. point to Gentoo or the GNU standards, etc.

configure
drwilly: That will be a bug if someone does `--ignored notignored`

configure
drwilly: https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables
Right above the parse-loop, where you added it. It is still valid.

configure
drwilly: Is there an argument against using `getopt` (or was it `getopts`? I keep forgetting which one it was)

drwilly:
> Eh this is too big for one change :-/ Let's get all the flag stuff in -- the minimal change to make it work -- and then talk about the rest.
> 
> (This script is fragile because there's no way of testing it -- it has to work on all platforms. Although I want OSH to have some sort of light test framework to help solve this problem.)
> 
> The `${1%=}` refactoring is probably good though, but I don't see why you changed everything else.

If I see something I (think I) can improve, I do. I could of course ask you beforehand, what the exact thought-process behind using `expr` was, but I think it's easier to talk about something you can look at. Also, you can always be like "no"; no hard feelings.

WRT testing shouldnt that be doable on TravisCI? We can at least cover Linux and OSX (and Windows?)
You could do a staged install and check the `find` output - that would at least catch the very dumb mistakes.

configure
andychu: OK I see, well you could have exactly two cases then:

1. --build, --host, etc.
2. --build=, --host=, etc.

There is a POSIX builtin getopt, but it doesn't take long flags, only short flags.

configure
andychu: OK well it should have a comment about why we are accepting and ignoring certain flags.

I know you told me but it should be in the code too, for the next person who goes in there.

andychu:
Adding continuous testing somehow would be nice, but that presumes that there are tests for this script in the first place, which there aren't.  Adding those would change my attitude towards the diffs, but I doubt it's worth it right now.

For OSH in general, right now we only have Linux testing.  There is no automated OS X testing.  If you know how to add that, it would be appreciated.

configure
drwilly: To be honest I'm not 100% sure we need the `--arg opt` cases anyways. I'm currently trying to figure out if that is standard for configure scripts at all.
After all we're targeting automated build-processes with this change and they can be expected to produce standard-conforming input.

drwilly:
2nd attempt. now with less stuff and more shiny.

Issue #456, opened 2019-08-13
./install should respect $DESTDIR


drwilly:
https://www.gnu.org/prep/standards/html_node/DESTDIR.html

andychu:
I think @timetoplatypus might have run into this for the Arch Linux package?

Maybe this is a good thing to add although I'm not sure of the details now.

> If your installation step would normally install /usr/local/bin/foo and /usr/local/lib/libfoo.a, then an installation invoked as in the example above would install /tmp/stage/usr/local/bin/foo and /tmp/stage/usr/local/lib/libfoo.a instead. 


Also, this might be related to #136 .

I thought the underlying cause of that is our weird "app bundle" system which I wanted to go away by now, but probably won't for awhile ... I will have to look in more detail.

andychu:
from Zulip:

> but is it OK if the binary does NOT run in the temp dir??
> I don't see how it can work because under Unix paths are hard-coded.

@drwilly :

> nothing is ever run in that image area
> stuff just gets copied there and then either tared or moved to /

andychu:
@drwilly  I consolidated usages of $DESTDIR as mentioned.  It should behave the same way still.  Let me know if it doesn't :)

https://github.com/oilshell/oil/commit/7e7c1496b2aa88b35db1893ee274450bd6b6add6

andychu:
Or if you think it's ready for a release, let me know.  There are a few other changes that could also go out.

drwilly:
```
#  tree /var/tmp/portage/app-shells/oil-0.7_pre3/image/
/var/tmp/portage/app-shells/oil-0.7_pre3/image/
└── usr
    ├── bin
    │   ├── oil.ovm
    │   ├── osh -> oil.ovm
    │   └── oshc -> oil.ovm
    └── share
        └── man
            └── man1
                └── osh.1.xz

5 directories, 4 files
```
looks good

PR #461, opened 2019-08-16
#456

drwilly:


install
```diff
@@ -21,18 +21,22 @@ main() {
   #local exec_filename=oil.ovm-dbg
   local exec_filename=oil.ovm  # Should this go in configure output?
 
-  local dest="$PREFIX/bin/$exec_filename"
+  local BINDIR="${PREFIX}/bin"
```
andychu: Local variables should be lower case.  Please revert all the renamings of local variables.

Document DESTDIR at the top of the file, since it's part of the user interface.

install
```diff
@@ -21,18 +21,22 @@ main() {
...
 
-  echo "Installing executables in $dest"
-  if ! install -v _bin/$exec_filename $dest; then
+  echo "Installing executables in ${DESTDIR}/${BINDIR}"
+  if ! install -v -d "${DESTDIR}/${BINDIR}/"; then
```
andychu: trailing slash here is inconsistent

install
```diff
@@ -21,18 +21,22 @@ main() {
   #local exec_filename=oil.ovm-dbg
   local exec_filename=oil.ovm  # Should this go in configure output?
 
-  local dest="$PREFIX/bin/$exec_filename"
+  local BINDIR="${PREFIX}/bin"
 
-  echo "Installing executables in $dest"
-  if ! install -v _bin/$exec_filename $dest; then
+  echo "Installing executables in ${DESTDIR}/${BINDIR}"
```
andychu: You don't need `${DESTDIR}` as opposed to `$DESTDIR` in any of these places.  Please remove it and revert all the changes to that add `${}`.

There are some unnecessary double quotes too but I guess those are OK for consistency, and do change the meaning.  The ${} doesn't change anything.

andychu:
Also, please write a proper commit description.  Assume that the repo is going to outlast Github.

andychu:
Also when you document DESTDIR, please link to the docs for it as well as explaining what it's used for (e.g. in Gentoo).

install
drwilly: There are very few unnecessary double-quotes in shell scripts and especially under the aspect of robustness I'd stick to quoting everything.

install
```diff
@@ -21,18 +29,22 @@ main() {
   #local exec_filename=oil.ovm-dbg
   local exec_filename=oil.ovm  # Should this go in configure output?
 
-  local dest="$PREFIX/bin/$exec_filename"
+  local bindir="$PREFIX/bin"
+
+  echo "Installing executables in $DESTDIR/$bindir/"
+  if ! install -v -d "$DESTDIR/$bindir/"; then
+    die "Couldn't create $DESTDIR/$bindir/"
```
andychu: Why not ust make $DESTDIR part of $bindir, rather than repeat it 4 times?

For expediency I will merge this and fix that ... if I break anything let me know :)

PR #466, opened 2019-08-18
[configure] better usage message

drwilly:
Make configure a bit more specific about what options expect arguments and how they relate.


andychu:
OK thanks, looks good!

PR #480, opened 2019-09-11
Fix a typo and capitalization of section titles

waldyrious:
(Also some punctuation tweaks.)


andychu:
OK thanks.  If you're curious about the architecture, the best place is to start a new thread on #oil-discuss or #oil-dev on:

https://oilshell.zulipchat.com/

I will respond, and most likely copy some responses into that doc.  There are no silly questions since most things are undocumented :)

I'm not updating this doc actively at the moment.  It was more something I wrote for myself.

Surfacing some more information about the general architecture to help contributors would be a great thing.

PR #509, opened 2019-10-20
Add Nix based development environment

adisbladis:
This commit also changes shebangs from `#!/bin/bash` to `#!/usr/bin/env bash` to work on NixOS and other non-FHS distros, and to ensure that `bash` is also used from nixpkgs.


andychu:
Thanks for getting this started!

Before changing all the shebang lines, I would want to do a small demo and test it out on 2 distros with Nix on top.

Here's a command that only takes a few seconds, depends on the versions of specific shells (and runs on Travis, not just my machine):

```
$ test/spec.sh smoke
smoke.test.sh
  0       5     pass    pass    pass    pass    builtin
  1       9     pass    pass    pass    pass    command sub
  2      13     pass    pass    pass    pass    command sub with builtin
  3      17     pass    pass    pass    pass    pipeline
  4      21     pass    pass    pass    pass    pipeline with builtin
  5      25     pass    pass    pass    pass    and-or chains
  6      42     pass    pass    pass    pass    here doc with var
  7      50     pass    pass    pass    pass    here doc without var
...
```

If specific versions of the shells are in `_tmp/spec-bin`, then we use those.

https://github.com/oilshell/oil/blob/master/test/spec-bin.sh#L34

andychu:
So here is my question: can I replace `_tmp/spec-bin/{bash,dash}` with `_tmp/nix-bin/{bash,dash}` or something?

I just want to define a set of binaries and then download them with a single command, and have it be reproducible on the machines for multiple developers.  It could be a shell script that wraps Nix.  Or maybe nix-shell is required?

FRidh:
Seeing the thread I also gave it a go. Like @adisbladis, I patched the shebangs. Additionally, I added a Nix expression that has the test dependencies needed by the Travis scripts. Have a look at https://github.com/FRidh/oil/commit/5bcc7fa540f1a91b61174863f9326984565a2dc3. Note a segmentation fault occurs in one of the tests and the build currently fails.

andychu:
@FRidh Thanks for trying it!  I will fix the build, which looks like there are some unrelated type checking errors.

Can we try out the `test/spec.sh smoke` command above before changing the shebangs?  That script already has `/usr/bin/env bash` as the shebang.

It's a very small script that serves as a good demo.   The other good demo is `build/dev.sh fastlex` which depends on the `re2c` code generator.

andychu:
Basically the dependencies I want versioned are:

- bash, dash, mksh, zsh, busybox -- for comparing Oil against in the "spec tests"
- re2c -- for generating the lexer

Actually `build/dev.sh ubuntu-deps` also lists:

- Python 2 headers
- gawk
- time
- libreadline-dev

But I would say the shell issue is the most pressing.  The other dependencies are stable, but the shells actually change significantly from version to version!  Hence the need to pin them.

FRidh:
The commit https://github.com/oilshell/oil/commit/e47ae87e5966afc335165be7398a0476e4511c57 has all the dependencies. With it, I get
```sh
$ test/spec.sh smoke
smoke.test.sh
  0       5     pass    pass    pass    FAIL    builtin
  1       9     FAIL    FAIL    FAIL    FAIL    command sub
  2      13     pass    pass    pass    FAIL    command sub with builtin
  3      17     FAIL    FAIL    FAIL    FAIL    pipeline
  4      21     FAIL    FAIL    FAIL    FAIL    pipeline with builtin
  5      25     pass    pass    pass    FAIL    and-or chains
  6      42     FAIL    FAIL    FAIL    FAIL    here doc with var
  7      50     FAIL    FAIL    FAIL    FAIL    here doc without var
  8      57     pass    pass    pass    FAIL    here doc with builtin
  9      64     FAIL    FAIL    FAIL    FAIL    Redirect external command
 10      70     FAIL    FAIL    pass    FAIL    Redirect with builtin
 11      75     FAIL    FAIL    FAIL    FAIL    Here doc with redirect
 12      83     FAIL    FAIL    FAIL    FAIL    "$@" "$*"
 13      90     FAIL    FAIL    FAIL    FAIL    $@ $*
 14      97     FAIL    FAIL    FAIL    FAIL    failed command
 15     101     pass    pass    pass    FAIL    subshell
 16     109     pass    pass    pass    FAIL    for loop
 17     121     pass    pass    pass    FAIL    vars

                dash    bash    mksh    osh
        pass    7       7       8       0
        FAIL    11      11      10      18
        total   18      18      18      18
```
and `nix-build` fails at
```
building
build flags: SHELL=/nix/store/0qbl8rbx866x6985l0gj0fq33ssm5m4a-bash-4.4-p23/bin/bash
test -d _build/opy && \
  build/actions.sh app-deps opy /build/oil:/build/oil/vendor bin.opy_
build/actions.sh: line 69: _devbuild/cpython-full/python: No such file or directory
test -d _build/oil && \
  build/actions.sh app-deps oil /build/oil:/build/oil/vendor bin.oil
build/actions.sh: line 69: _devbuild/cpython-full/python: No such file or directory
test -d _build/hello && \
  build/actions.sh app-deps hello /build/oil/build/testdata:/build/oil/vendor hello
build/actions.sh: line 69: _devbuild/cpython-full/python: No such file or directory
make: *** No rule to make target '_build/oil/app-deps-c.txt', needed by '_build/oil/all-deps-c.txt'.  Stop.
builder for '/nix/store/gbzc1gr6n408sya4xdwgzly166lz3gzq-oil-0.7.pre5.drv' failed with exit code 2
error: build of '/nix/store/gbzc1gr6n408sya4xdwgzly166lz3gzq-oil-0.7.pre5.drv' failed
```

By the way, there are several that have contributed to the Oil expression in `nixpkgs` which I assume are either users or people interested in Oil. https://github.com/NixOS/nixpkgs/commits/master/pkgs/shells/oil/default.nix

andychu:
Right so my goal is to get that to pass with "one command" (which could be our own wrappers around Nix) on any Linux distro.

The thing that's different about Oil is that the **dev build** is very distinct from the **tarball build**.  I describe that a bit here:

The nix expressions that exist are for the tarball build, which is very easy.  It's just `configure; make ; install` essentially.

However the dev build is very complicated.  I do that so end users don't have to install Python basically.  The parts of Python we need are bundled with Oil.

That's why you're getting those errors about `_devbuild/cpython-full/python` -- that is necessary to bootstrap the dev build and automatically find dependencies.

FRidh:
I see. That command would be:
```sh
nix-shell --run "test/spec.sh smoke"
```

FRidh:
Note also, that when opening the `nix-shell` and executing `build/dev.sh minimal` the produced `bin/osh` functions.

andychu:
I think I need some more background on Nix... my mental model might be wrong or I'm trying to use it in some non-standard way.

When you run:

> nix-shell --run "test/spec.sh smoke"

Where is the environment defined?  I didn't see the Nix files in the PR?

Are you using the tarball build Nix expression implicitly somehow?

andychu:
I have to go to lunch now but I would be interested in chatting about it later tonight (I'm on US pacific time).

Or if you have some pointers to docs or projects doing similar things I'd be interested.

I'm not sure I want to change the "world" to Nix.  I don't want to REQUIRE nix to develop Oil.   I just want to use Nix as an easy way to get stable versions of binaries, if that makes sense.  Do people use it like that?

andychu:
We can chat on https://oilshell.zulipchat.com/ (requires login) , or I am willing to look at docs/examples first to get up to speed.  Thanks again for trying it!

FRidh:
> Where is the environment defined? I didn't see the Nix files in the PR?

In @adisbladis PR there is a `shell.nix` file. That one points to Nixpkgs and then reuses its Oil expression and its dependencies. 

In my commit there is a `default.nix`, which also points to Nixpkgs to obtain its dependencies but does not use its Oil expression. You would likely then point to a new NixOS version every half a year, at which point you would have to account for potentially different versions of dependencies. At the same time, you could have expressions for those in here as well, but you likely don't want that (at least not of the transitive dependencies as that would get big and complicated).

New commit https://github.com/oilshell/oil/commit/810dc541470286d09bee12eed3bce2b6d5a15e18

> I have to go to lunch now but I would be interested in chatting about it later tonight (I'm on US pacific time).

My time zone is CEST so I am turning off soon. Maybe @adisbladis is in your time zone?

> I just want to use Nix as an easy way to get stable versions of binaries, if that makes sense. Do people use it like that?

Yes, sometimes it's easier to just do that and perform the build "impure". 

The project `status.im` uses Nix, but I think they use it also for reproducible builds as well
https://discuss.status.im/t/the-road-to-nix-a-functional-package-manager-to-rule-them-all/1049

andychu:
OK thanks I will test this out later today.  I read the Nix paper and tried it a long time ago but I'm otherwise very out of date on it.

andychu:
@FRidh I installed Nix and tried to `nix-build` the `default.nix` in your PR.  I also want to ignore every directory that starts with `_` in addition to the `.git` dir.  I couldn't figure out how to do that after some Googling and trial and error. 

```
error: file '/home/andy/git/oilshell/oil/_chroot/a3/dev/null' has an unsupported type
(use '--show-trace' to show detailed location information)
```

I put some notes here:

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Nix.20Notes

@adisbladis I didn't understand the binary blob in your Nix expression.  One thing I'm looking for is for the build/binary to be transparent from the sources, and the other one seemed closer.

Although note I'm not trying to build Oil with Nix.  I want to fetch/build Oil's dev dependencies with Nix!  Oil itself will change in the git repo.

And someone else has solved the "tarball build" problem, which is very simple.  This is the separate "dev dependencies" problem.

akavel:
I can't help more now, but for a sample project using Nix to fetch build deps with nix-shell, see QMK - check the Nix subsection in the [installation guide]( https://docs.qmk.fm/#/getting_started_build_tools ) and the corresponding [shell.nix]( https://github.com/qmk/qmk_firmware/blob/master/shell.nix ) script ("expression" in Nix parlance) which makes it possible. I can try answering your further questions, but not sure I'll find time.

adisbladis:
Closing since https://github.com/oilshell/oil/pull/599 was merged

PR #511, opened 2019-10-21
Nix development environment

FRidh:
This commit introduces a Nix expression for Oil. It includes test
dependencies needed for running the Travis scripts.

See https://github.com/oilshell/oil/pull/509 for discussion.


andychu:
In case you didn't see, I had a problem using Nix with temp dirs starting with `_`, described here:

https://github.com/oilshell/oil/pull/509#issuecomment-544370143

Do you know how to exclude those dirs?

FRidh:
@andychu in case you want to `nix-build`, you point in your recipe to the source using the `src` attribute. You could have e.g. `src = builtins.fetchGit ./.;` in which case it will only consider files that are version-controlled or staged. That should solve your issue. In this PR I switched to just a `shell.nix` as you seem to wanted that.

andychu:
Hm not sure I understand.  Can you add something to the README.md about how to use this?

Ideally it would be like:

1. Install nix
2. Run nix-shell
3. build/dev.sh minimal
4. Run bin/osh

Does that work now?

Also it looks like you did a broad search and replace that will break a bunch of things.  Please do the minimal to run `build/dev.sh minimal`, and then the rest of it can be changed later once we have confidence that this works and is desirable.

FRidh:
Closing because #599 got merged

Issue #435, opened 2019-07-18
implement printf %d \'c to print the numeric value of char c


Crestwave:
Basically, `printf %d \'a` should print 97. It's a pretty obscure feature, but it's used in BF interpreters and such.

andychu:
Wow that is a truly bizarre feature!  It took me awhile to find in POSIX.  All shells appear to implement it.

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html#tag_20_94

> If the leading character is a single-quote or double-quote, the value shall be the numeric value in the underlying codeset of the character following the single-quote or double-quote

Crestwave:
This should actually be pending-release, not closed, no?

andychu:
Released with 0.7.pre6 : https://www.oilshell.org/release/0.7.pre6/

PR #517, opened 2019-10-29
[spec/builtin-printf] Implement printf %d \'c to print the numeric value of char c

asokoloski:
Closes: #435

### DIFF
--- a/osh/builtin_printf.py
+++ b/osh/builtin_printf.py
@@ -207,15 +207,21 @@ class Printf(object):
             try:
               d = int(s)
             except ValueError:
-              # This works around the fact that in the arg recycling case, you have no spid.
-              if word_spid == runtime.NO_SPID:
-                self.errfmt.Print("printf got invalid number %r for this substitution", s,
-                                  span_id=part.type.span_id)
+              if s[0] in '\'"':
+                try:
+                  d = ord(s[1])
+                except IndexError:
+                  d = 0
               else:
-                self.errfmt.Print("printf got invalid number %r", s,
-                                  span_id=word_spid)
+                # This works around the fact that in the arg recycling case, you have no spid.
+                if word_spid == runtime.NO_SPID:
+                  self.errfmt.Print("printf got invalid number %r for this substitution", s,
+                                    span_id=part.type.span_id)
+                else:
+                  self.errfmt.Print("printf got invalid number %r", s,
+                                    span_id=word_spid)
 
-              return 1
+                return 1
 
             if typ in 'di':
               s = str(d)
--- a/osh/builtin_printf.py
+++ b/osh/builtin_printf.py
@@ -208,9 +208,9 @@ class Printf(object):
               d = int(s)
             except ValueError:
               if s[0] in '\'"':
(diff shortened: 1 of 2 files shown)


osh/builtin_printf.py
```diff
@@ -207,15 +207,21 @@ def __call__(self, arg_vec):
...
             except ValueError:
-              # This works around the fact that in the arg recycling case, you have no spid.
-              if word_spid == runtime.NO_SPID:
-                self.errfmt.Print("printf got invalid number %r for this substitution", s,
-                                  span_id=part.type.span_id)
+              if s[0] in '\'"':
+                try:
+                  d = ord(s[1])
+                except IndexError:
+                  d = 0
```
andychu: I pulled this and ran the tests, and it works!  Comments:

* How about adding test coverage for this `d = 0` case too?  When you just pass a single or double quote?
* The test confused me because you used `\*`.  Maybe choose `a=97` instead?

I'm probably going to change the exceptions to if statements because of the "mycpp" translation, but that is only a little related to this PR.

asokoloski: Sure, will do.  I actually started out with `a=97` but tried to be cute with `*=42` -- but clarity trumps cuteness :)

asokoloski: * 114e57d6d1109835da06ebceffad58a1065f7c78 -- switch to using "a" character
* c874e2badadf297213961e4905bda1853e78f8d6 -- add spec test for d = 0 case
* a3d47f87f7a10ead084dcc4b73b736a6cdccbead -- avoided adding an instance of exception handling in this MR, at least.

andychu:
I merged this, thanks a lot!  Not many people have made a spec test change so I'm happy to see you got through it :)

I did find one problem:

https://github.com/oilshell/oil/commit/42458b8a965c63c48b843ea627a7e0c4ad0ab875

I also tightened up some corner cases.  Most shells are kind of sloppy about corner cases, but OSH is tring to be a little stricter. 

I made it stricter like `ash` in the empty string case, and stricter like `mksh` in the single quote case.

I debated whether to error out if there is more than one byte, but I didn't change that.

Oh another corner case is what to do in the case of a utf-8 encoded unicode char.  Let me at least test that -- I probably won't implement it for now.

andychu:
OK, bash and zsh support unicode, but the other shells don't.

https://github.com/oilshell/oil/commit/78e089f37d39356b8f16ff7f8ba4301f289a2980

OSH should support Unicode in this case, but I don't think it's that important.  And as mentioned it depends on #366.  Because we have some pretty cheesy Python utf-8 code in `osh/string_ops.py` :)  That code helped us write good tests, but probably needs to be replaced.

Thanks for your help on this!

asokoloski:
Glad to help!  Sorry about being a bit sloppy; thanks for tidying up after me.

andychu:
No problem!  I'm still maintaining these issue labels if you're looking for something to hack on:

https://github.com/oilshell/oil/labels/good%20first%20issue

Or ask me on Zulip if you have any questions!

Issue #173, opened 2018-08-25
${s#?} and ${s%?} should be respect Unicode


andychu:
I thought this was all done with issue #142, but I realized that the `# ## % %%` operators need to support Unicode too.

The glob cases loop over bytes, not code points.  Should be easy to fix with the same utility functions in `core/libstr` (although these 4 operators are somewhat subtle and tedious).

andychu:
Notes about them, and pointers to implementation: https://github.com/oilshell/oil/commit/7a1947fa24a10718b2b34c260013a4a21ed3f1bb

andychu:
I marked this as "good first issue" because it's a highly localized algorithm that doesn't require understanding a lot of the code.

But it is probably only a good first issue for someone who is really good with strings and detail-oriented.  It's fiddly.

In short: `test/spec.sh var-op-strip --range 10` needs to pass

Add `--verbose` flag to see details of failures.

PR #520, opened 2019-10-30
Fix handling of unicode in affix removal (${s#?} and ${s%?})

asokoloski:
Fixes: #173 

Hi @andychu, this one turned out to be a bit tricker than I thought.  I needed to change func_fnmatch in libc.c to set the locale to "" (making it refer to the environment variables) -- otherwise it assumed LANG=C, which meant that fnmatch did not properly handle multibyte characters.  I'm not sure if this was the best way to deal with this.

Also, it turns out that iterating backwards over a UTF-8 string is not substantially harder than going forwards, so I did that, rather than iterating forwards and building up a list of indexes as was suggested by your comment above DoUnarySuffixOp.

I hope this is alright.  I spent more time checking edge cases so as not to embarrass myself again :)

### DIFF
--- a/osh/string_ops.py
+++ b/osh/string_ops.py
@@ -99,16 +99,41 @@ def _NextUtf8Char(s, i):
 def _PreviousUtf8Char(s, i):
   """
   Given a string and a byte offset, returns the position of the
-  character before that offset.
+  character before that offset.  To start (find the first byte of the
+  last character), pass len(s) for the initial value of i.
 
   Validates UTF-8.
   """
-  following_i = i
+  # All bytes in a valid UTF-8 string have one of the following formats:
+  #
+  #   0xxxxxxx (1-byte char)
+  #   110xxxxx (start of 2-byte char)
+  #   1110xxxx (start of 3-byte char)
+  #   11110xxx (start of 4-byte char)
+  #   10xxxxxx (continuation byte)
+  #
+  # Any byte that starts with 10... MUST be a continuation byte,
+  # otherwise it must be the start of a character (or just invalid
+  # data).
+  #
+  # Walking backward, we stop at the first non-continuaton byte
+  # found.  We try to interpret it as a valid UTF-8 character starting
+  # byte, and check that it indicates the correct length, based on how
+  # far we've moved from the original byte.  Possible problems:
+  #   * byte we stopped on does not have a valid value (e.g., 11111111)
+  #   * start byte indicates more or fewer continuation bytes than we've seen
+  #   * no start byte at beginning of array
+  #
+  # Note that because we are going backward, on malformed input, we
+  # won't error out in the same place as when parsing the string
+  # forwards as normal.
+  orig_i = i
+
   while i > 0:
     i -= 1
     byte_as_int = ord(s[i])
     if (byte_as_int >> 6) != 0b10:
-      offset = following_i - i
+      offset = orig_i - i
       if offset != _Utf8CharLen(byte_as_int):
         # Leaving a generic error for now, but if we want to, it's not
         # hard to calculate the position where things go wrong.  Note
--- a/osh/string_ops.py
+++ b/osh/string_ops.py
@@ -61,6 +61,19 @@ def _CheckContinuationByte(byte):
     raise util.InvalidUtf8(INVALID_CONT)
 
 
+def _Utf8CharLen(starting_byte):
+  if (starting_byte >> 7) == 0b0:
+    return 1
+  elif (starting_byte >> 5) == 0b110:
+    return 2
+  elif (starting_byte >> 4) == 0b1110:
+    return 3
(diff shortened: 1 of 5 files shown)


andychu:
Cool!  I cloned this and ran the tests, and it worked.

I will look at the code in more detail later.  But does this make any assumptions about the locale environment variables?

My thinking is that OSH should only support utf-8, because it's a lot simpler to get all the edge cases right.  Unfortunately that is totally at odds with how libc string handling works.

I need to document that more.  I put some more notes in `doc/unicode.sh` just now.

I think we need some more testing with different env variable settings because there were some problems with the `wcswidth()` wrapper in a Docker image.  That function also sets the locale.

----

Long story short: since this works I will likely merge it.  But it's making me think about the general unicode strategy and #366 again.  Glob and fnmatch are tricky because it's something we don't want to reimplement ourselves.  So this is likely the right strategy for those operations.

andychu:
OK thinking about it again, I think this is is the right strategy except that we should always change the locale to utf-8 for this operation.  That is, all our invocations of fnmatch() should be insensitive to the environment and respect utf-8.

This is how wcswidth() works right now.

I think I can rig up the spec framework to expose this.  Right now it only runs with LANG as utf-8 -- see `MakeTestEnv` in `sh_spec.py`.  Yeah let me play with it.

----

At least, this is the external behavior I think makes sense... I'm open to suggestions.

Basically Oil is following newer languages like Go, Rust, Julia, Swift, etc.  -- utf-8 is primary.   Rather than Java, JavaScript, and Python, which have "multibyte characters".

asokoloski:
Yeah, I was not sure -- I think you're right, just copying what wcswidth does is better than what I did -- it doesn't make sense to take LANG into account when the code all expects utf-8 and won't work otherwise.  I can change it, unless you meant that you want to do it -- just let me know.

asokoloski:
Oh, and about using utf-8 everywhere -- is there a good way to set this around where osh starts, rather than on every call to fnmatch and wcswidth?  We don't want to override someone's preferred date/time representation, etc, of course.  Would it be reasonable to look at the user-provided locale, and just replace the character encoding with utf-8, but keep the language the same?  I don't have enough experience with locales to know if I'm just spouting nonsense here.

andychu:
Yeah that's a good question.  I'm not sure either :)  Unicode in libc is confusing.

Let me look at the code in more detail.  I think I will merge it without the change, and then use that to improve the test framework to handle different LANG.  And what we learn from that will probably inform the code style.

asokoloski:
Ok, sounds good, thanks!

andychu:
Also based on the list of operations I have here,

https://github.com/oilshell/oil/blob/master/doc/unicode.md

I would expect that some tests in

```
test/spec.sh var-op-patsub
test/spec.sh glob
test/spec.sh extglob-match
```

would be failing.  I don't remember if that's because they actually work or because we didn't test them :)  

For patsub, we translate a glob pattern to a regex, then use `regexec()` to get the position of the match.  Because `fnmatch()` doesn't give you that.  I learned that there is a weird mismatch in shell because historically ksh did not use libc APIs!  It used its own stuff before there was even POSIX I think.

osh/string_ops.py
```diff
@@ -61,36 +61,64 @@ def _CheckContinuationByte(byte):
...
+    i += length
   except IndexError:
     raise util.InvalidUtf8(INCOMPLETE_CHAR)
 
   return i
 
 
+def _PreviousUtf8Char(s, i):
+  """
+  Given a string and a byte offset, returns the position of the
```
andychu: OK this looks like a short and useful piece of code.

1) Can you write unit tests for it in `osh/string_ops_test.py` ?  I don't think the error cases are covered by the spec tests.

2) Is there a URL to explain this algorithm?  I think I get it but it's not obvious upon reading.

Here's a link I found through Googling, but the explanation is unclear and some people are quibbling with it:

https://stackoverflow.com/questions/22257486/iterate-backwards-through-a-utf8-multibyte-string

asokoloski: 1. Sure, sorry, didn't even consciously notice the test module.  Still learning my way around the code base :)

2. I can try to find another one, but when I was working on it I only referred to that stackoverflow post and the description on wikipedia.  I'll try to come up with a clear explanation in the code.

asokoloski: 0b9445adea807545bb1bc8f49ea5424db78c0c26 -- added tests for this function, as well as _NextUtf8Char, which lacked them.
1e611ed4514fec8e0afd4fe8d480e91bbcea2063 -- added comment to explain algorithm in _PreviousUtf8Char

andychu:
OK excellent, that makes sense now!

For most changes, I rely on spec tests more than unit tests, since external behavior is more important.  But this one has some corner cases that made a unit test worthwhile.

asokoloski:
:D  no worries, I agree about the testing.

PR #526, opened 2019-11-08
[test/spec] fix var-op-patsub "Substitute one unicode character" test

asokoloski:
Using the environment's locale instead of the default "C" locale
allows regcomp to understand UTF-8.

(finally got around to this tiny fix...)


andychu:
Great thanks for doing this!  I merged and it worked.

I have some more thoughts on Unicode which I will post to Zulip.

PR #534, opened 2019-11-12
[types] Convert type checking to use config file for mypy, remove assert-one-error hack

asokoloski:
@andychu I think I got all the instances where we run mypy, except for in mycpp/examples.py -- but I couldn't find an entry point that I could make work.  I think converting it just like the others should work ok, perhaps with some additions to mypy.ini's per-module config sections.

Please feel free to just kick this back to me if I broke or missed anything.  Working through issues is a good way to learn :)

### DIFF
--- a/types/mypy.ini
+++ b/types/mypy.ini
@@ -18,6 +18,6 @@ no_implicit_reexport = True
 
 strict_optional = True
 
-[mypy-core.alloc,core.main_loop,core.pyutil,core.ui,_devbuild.gen/syntax_asdl,frontend.match,frontend.parse_lib,frontend.tdop,oil_lang.expr_parse,oil_lang.expr_to_ast,osh.bool_parse,osh.braces,osh.cmd_parse,osh.word_,osh.word_parse,pgen2.parse,pylib.os_path]
+[mypy-core.alloc,core.main_loop,core.pyutil,core.ui,_devbuild.gen/syntax_asdl,frontend.match,frontend.parse_lib,frontend.tdop,oil_lang.expr_parse,oil_lang.expr_to_ast,osh.bool_parse,osh.braces,osh.cmd_parse,osh.glob_,osh.word_,osh.word_parse,pgen2.parse,pylib.os_path]
 
 strict_optional = False
--- a/types/mypy.ini
+++ b/types/mypy.ini
@@ -16,63 +16,13 @@ warn_unused_ignores = True
 warn_return_any = True
 no_implicit_reexport = True
 
+strict_optional = False
 
+[mypy-asdl.pretty]
 strict_optional = True
 
-# --START-LAX-CONFIG--
-# Everything below this gets stripped out in order to generate a stricter version of this config.
-# see types/osh-parse.sh, typecheck-all function
-
-[mypy-core.alloc]
+[mypy-_devbuild.gen.*]
+strict_optional = True
+[mypy-_devbuild.gen.syntax_asdl]
 strict_optional = False
 
-[mypy-core.main_loop]
-strict_optional = False
-
-[mypy-core.pyutil]
-strict_optional = False
-
-[mypy-core.ui]
-strict_optional = False
-
-[mypy-_devbuild.gen/syntax_asdl]
-strict_optional = False
-
-[mypy-frontend.match]
-strict_optional = False
-
-[mypy-frontend.parse_lib]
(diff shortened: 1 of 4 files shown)


types/mypy.ini
```diff
@@ -0,0 +1,78 @@
...
+strict_optional = False
+
+[mypy-osh.word_parse]
+strict_optional = False
+
+[mypy-pgen2.parse]
+strict_optional = False
+
+[mypy-pylib.os_path]
+strict_optional = False
```
andychu: Hm I kind of feel like this isn't worth it... because we're going to add 100% type checking coverage in the near future, and then every time we'll have to edit this file.

That is, will every new file need to be entered here?

It seems like it's easier to pass --no-strict-optional once.  If I'm understanding the patch correctly.

It might be nice to turn on strict optional in the future, but it's much more important  to translate Oil to C++ first.

types/run.sh
```diff
@@ -36,7 +36,7 @@ pyann-patched() {
 
 iter-demo-asdl() {
   asdl/run.sh gen-typed-demo-asdl
-  typecheck --strict \
+  typecheck $MYPY_FLAGS \
```
andychu: I think the intention was that ASDL code gen is strict, but the rest of Oil is not strict.

That will make it easier to turn on `--strict-optional later`.  That is ASDL won't regress while we work on the OIl codebase in general.

andychu:
I cleaned up the spew:

https://github.com/oilshell/oil/commit/41c6e012b209442210967361e1a201ab80d8b68c

And commented the intention here:

https://github.com/oilshell/oil/commit/95b4932c9b7f7a156bee78b4f73fe17d99cd201f

It might make sense to use a MyPy config at some point.  But I think it's more important to increase the coverage.

The translation seems to be working.  It requires some work for each file, but is tractable.  And translation depends on static types.  So that's why I think increasing type checking coverage is important now.

If there are any other annoyances / problems with the scripts let me know!

types/mypy.ini
asokoloski: Ah, ok.  I have heard of people going for a "ratchet effect" where you keep making things stricter.  That's what I was trying to do -- but no worries if it's overkill.  I guess the manifest file is serving a similar purpose, by enabling type checking one module at a time?

Anyway, main point is it's not necessary to enter every new file here.  You can specify modules as glob patterns - that's what I did originally, and I only needed a few.  Shall I switch back to that approach?

asokoloski: @andychu maybe c1f70139087153c68692dcf28db9c48b6613ec82 is an improvement?  I believe this addresses your comment about ASDL being strict as well, below.  Note that syntax_asdl.py seems to need --no-strict-optional, in contrast to the rest of _devbuild/gen.

If you still think it's not worth the change after this, just say so, I won't be offended :)  Last thing I want to do is to take your attention away from the big stuff :)

andychu: Yeah I'm not convinced we need a MyPy config right now.  The shell function is a hack, but it's a short one.  That's why I use shell -- to make problems go away :)

And generally I prefer flags over config files because it's less syntax to learn.  I'd rather use shell globs and a text file of files than globs in N different config syntaxes.

It's possible we might a config in the future, in which case we can dig this PR up again.  But I haven't needed to adjust MyPy flags in a long time.  I think we can complete the experiment with translating the parser without MyPy config changes, only code changes.

asokoloski: Ok, no worries.  I think I disagree in one sense, but agree in another.

Disagree in the sense that it the mypy command line opts aren't as expressive as the config file.  You can specify just a few files to type check with particular options, but since mypy follows imports, you're pulling in a dependency graph, which will probably overlap with the dependency graph of other modules you want to type check with different options.  That's what the caused the `asdl/pretty.py` problem, I think.  It's getting checked with `strict_optional` set to both true and false in different situations, and so you either get an error about the None or an error about the unnecessary "ignore this None" comment.  

Using a config file to specify options on a module level lets you use the same options for `asdl.pretty` in both cases.  Of course, it would be nice if there were a way to get around this on the command line, or if mypy just wasn't so nitpicky about the unnecessary comment.

But I totally agree in the sense that it's a minor issue that won't hold up translating the parser.  And I'm sympathetic to the desire to avoid more syntaxes.

I'll close the MR, and get on with more useful stuff :)

PR #535, opened 2019-11-13
[types] get frontend/location.py passing with mypy

asokoloski:
Also add it to travis checks to prevent regressions.

@andychu did I do it right with "UP_node"?  What does UP stand for, anyway?  Union... Parameter?

### DIFF
--- a/frontend/location.py
+++ b/frontend/location.py
@@ -8,33 +8,49 @@ TODO: Move some of osh/word_ here.
 """
 from __future__ import print_function
 
-from _devbuild.gen.syntax_asdl import command_e, command_t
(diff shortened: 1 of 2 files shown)


andychu:
OK awesome!  Thanks for doing this.

UP is the opposite of down :)  Because we're downcasting from the base type to the subtype.

And I'm glad you followed this subtlety that only exists in Oil because of the Python -> C++ conversion!

andychu:
In an ideal world we wouldn't need the UP_node thing, so I made it look a bit ugly.  It's a tiny-ish hack to gradually translate the code to C++.

PR #542, opened 2019-11-27
Add annotations and enable typechecking for osh/word_compile.py and osh/builtin_bracket.py

asokoloski:
Also:
* create a manifest for typecheck-more-oil files instead of specifying all the filenames inline.
* make typecheck-more-oil stop returning errors for dependencies of specified modules, using the `--follow-imports=silent` option to mypy.

### DIFF
--- a/osh/bool_parse.py
+++ b/osh/bool_parse.py
@@ -37,9 +37,10 @@ from core.util import p_die
 from frontend import lookup
 from osh import word_
 
-from typing import List, Optional, TYPE_CHECKING
+from typing import List, Optional, Union, TYPE_CHECKING
 if TYPE_CHECKING:
   from osh.word_parse import WordParser
+  from osh.builtin_bracket import _StringWordEmitter
 
 # import libc  # for regex_parse
 
@@ -48,7 +49,7 @@ class BoolParser(object):
   """Parses [[ at compile time and [ at runtime."""
 
   def __init__(self, w_parser):
-    # type: (WordParser) -> None
+    # type: (Union[WordParser, _StringWordEmitter]) -> None
     """
     Args:
       w_parser: WordParser
--- a/types/run.sh
+++ b/types/run.sh
@@ -77,7 +77,7 @@ typecheck-more-oil() {
   typecheck $MYPY_FLAGS \
     osh/glob_.py osh/string_ops.py frontend/location.py \
     osh/history.py core/comp_ui.py frontend/lookup.py \
-    frontend/py_reader.py core/error.py \
+    frontend/py_reader.py core/error.py osh/word_compile.py \
     > $log
 
   assert-one-error $log
--- a/types/run.sh
+++ b/types/run.sh
@@ -68,17 +68,13 @@ typed-arith-asdl() {
   echo
 }
 
-typecheck-more-oil() {
-  #typecheck $flags osh/word_compile.py
 
+typecheck-more-oil() {
+  local manifest=types/more-oil-manifest.txt
   local log=_tmp/typecheck-more-oil.txt
 
   set +o errexit
-  typecheck $MYPY_FLAGS \
-    osh/glob_.py osh/string_ops.py frontend/location.py \
-    osh/history.py core/comp_ui.py frontend/lookup.py \
-    frontend/py_reader.py core/error.py osh/word_compile.py \
-    > $log
+  cat $manifest | xargs -- $0 typecheck $MYPY_FLAGS > $log
 
   assert-one-error $log
 }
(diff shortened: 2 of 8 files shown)


types/run.sh
```diff
@@ -87,12 +87,12 @@ need-typechecking() {
...
-  set +o errexit
-  cat $MORE_OIL_MANIFEST | xargs -- $0 typecheck $MYPY_FLAGS > $log
-
-  assert-one-error $log
+  # The --follow-imports=silent option allows adding type annotations
+  # in smaller steps without worrying about triggering a bunch of
+  # errors from imports.  In the end, we may want to remove it, since
+  # everything will be annotated anyway.  (that would require
+  # re-adding assert-one-error and its associated cruft, though).
+  cat $MORE_OIL_MANIFEST | xargs -- $0 typecheck --follow-imports=silent $MYPY_FLAGS
```
asokoloski: @andychu I decided to just take out the call to assert-one-error here, because it's still used by `typecheck-all` in `types/osh-parse.sh`.  We could also use `--follow-imports=silent` there too, though, and that would allow dropping the function entirely.

I love deleting code, but on a project I'm not so familiar with, I think the parable of "Chesterton's Fence" usually applies :)

osh/bool_parse.py
```diff
@@ -48,7 +49,7 @@ class BoolParser(object):
   """Parses [[ at compile time and [ at runtime."""
 
   def __init__(self, w_parser):
-    # type: (WordParser) -> None
+    # type: (Union[WordParser, _StringWordEmitter]) -> None
```
andychu: FYI: I noticed this Union issue too... it will require change when we translate (since it's not translatable) but of course the first step is to get it to pass MyPy in any way!

types/run.sh
```diff
@@ -68,18 +68,31 @@ typed-arith-asdl() {
...
-    > $log
+need-typechecking() {
+    # This command is useful to find files to annotate and add to
+    # $MORE_OIL_MANIFEST.
+    # It shows all the files that are not included in
+    # $MORE_OIL_MANIFEST or $OSH_PARSE_MANIFEST, and thus are not yet
+    # typechecked by typecheck-more-oil here or
+    # `types/osh_parse.sh travis`.
+    comm -2 -3 \
+         <(metrics/source-code.sh osh-files | grep '.py$' | sed 's@^@./@') \
```
andychu: Doesn't need to be changed for this commit, but when doing my own audits like this, I noticed we were leaving out `oil_lang/` (and there is no `oil-files` -- it's a bit approximate)

For example `oil_lang/regex_translate.py` is the Eggex to POSIX ERE translator that's probably very easy to type check (and translate), since it's all string manipulation.

types/run.sh
andychu: Yeah that makes sense, it looks like it is still used, but I'm sure we can get rid of it eventually.

andychu:
FWIW I got a bunch of errors in `types/osh-parse travis` without this flag:

https://github.com/oilshell/oil/commit/1cab347438c50ca36417eab2b9da70ddcd26960a

It probably makes sense to unify the `osh-parse` and `typecheck-more-oil` suites at some point, but I like the progress we're making now so I don't want to shave that yak yet!

As you may have noticed the difference is that I created `bin/osh_parse.py` as a separate binary that only does string manipulation.  And its dependencies are automatically crawled to form the manifest.

Thanks again for doing this!

andychu:
I changed the `Union` thing since it came up during translation.

https://github.com/oilshell/oil/commit/bd99c93d84cf1340454780707489dd38c5b581bb

I should add translation to Travis at some point, but it looks like we're making progress  with the current setup :)  It depends on the MyPy repo and is pretty slow.

andychu:
Oops, I was running `types/osh-parse.sh travis` to verify my changes, but really `types/run.sh travis` is what I should have run.  So the build has been since this morning.

Fixing...

asokoloski:
> FWIW I got a bunch of errors in `types/osh-parse travis` without this flag:
> 
> [1cab347](https://github.com/oilshell/oil/commit/1cab347438c50ca36417eab2b9da70ddcd26960a)
> 

Hmm, did one of the `if TYPE_CHECKING:` blocks I added/altered cause mypy to pull in some modules it didn't previously?  Whoops...

> It probably makes sense to unify the `osh-parse` and `typecheck-more-oil` suites at some point, but I like the progress we're making now so I don't want to shave that yak yet!

:+1: 
 
> As you may have noticed the difference is that I created `bin/osh_parse.py` as a separate binary that only does string manipulation. And its dependencies are automatically crawled to form the manifest.

I had not noticed.  Thanks for explaining :)
 
> Thanks again for doing this!

:vulcan_salute:

PR #548, opened 2019-11-28
[doc] clarify docstring of frontend/syntax_abbrev.py

asokoloski:
This was a bit confusing to me at first.


andychu:
OK thanks for clarifying it! 

I'm trying to remember the reason for the textual include rather than import.  I think it was related to not needing to repeat the imports at the top.  I guess it's a tiny shortcut.

PR #549, opened 2019-11-28
[fix] fix osh/state_test.py incorrectly assuming where env binary will *not* be

asokoloski:
This makes the test pass on my system (Arch linux).  I tried not to weaken the test too much.


andychu:
Thanks for fixing this!

asokoloski:
No worries :)

PR #550, opened 2019-11-28
Get osh/split.py type checked

asokoloski:
one more!

### DIFF
--- a/osh/split.py
+++ b/osh/split.py
@@ -27,11 +27,16 @@ with SPLIT_REGEX = / digit+ / {
 """
 
 from _devbuild.gen import runtime_asdl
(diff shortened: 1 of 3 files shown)


andychu:
Great thanks!

I fixed an unused import lint error caught by Travis:

https://github.com/oilshell/oil/commit/f3e7935ee192d20828b6a6282431beda26f9a2fc

I know it's kind of like whack-a-mole, and the lint spews a lot whether it's red or green.  At some point I will make a shell function that does both more cleanly but for now it seems like things are working.

asokoloski:
Glad to help!  I'll try to pay more attention to that (and the indentation).

PR #556, opened 2019-12-06
Add typechecking for 2 more files

asokoloski:
oil_lang/regex_translate.py needed substantial annotation changes, while core/main_loop.py was already passing mypy.


andychu:
Thanks a lot, looks good!

PR #557, opened 2019-12-06
[types] Add type annotations for oil_lang/objects.py [tentative]

asokoloski:
I'm unsure about some of these.  I used `Any` in several places, some
of which can probably be more specific.  It'll probably need
revisiting, but it's a step in the right direction, I hope.

### DIFF
--- a/oil_lang/objects.py
+++ b/oil_lang/objects.py
@@ -9,10 +9,20 @@ from __future__ import print_function
 from core.util import log
 from oil_lang import regex_translate
 
-from typing import Union, TYPE_CHECKING, Type, List, Dict, Any, Optional
+from typing import Union, TYPE_CHECKING, List, Dict, Any, Optional
 if TYPE_CHECKING:
+  from typing import Type
+  BoolList = List[bool]
+  IntList = List[int]
+  FloatList = List[float]
+  StrList = List[str]
+  TableDict = Dict[Any, List[Any]]
+  AssocArrayDict = Dict[Any, Any]
   from _devbuild.gen.syntax_asdl import re_t, command__Proc, command__Func, expr__Lambda
   from osh.cmd_exec import Executor
+else:
+  BoolList = IntList = FloatList = StrList = list
+  AssocArrayDict = TableDict = dict
 
 _ = log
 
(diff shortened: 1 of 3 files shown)


oil_lang/objects.py
```diff
@@ -9,9 +9,10 @@
 from core.util import log
 from oil_lang import regex_translate
 
-from typing import TYPE_CHECKING
+from typing import Union, TYPE_CHECKING, Type, List, Dict, Any, Optional
```
andychu: The Travis build is failing because `Type` is not defined in `vendor/typing.py`.

https://travis-ci.org/oilshell/oil/builds/621465709?utm_source=github_status&utm_medium=notification

The reason we have `vendor/typing` is because we don't need to ship the whole `typing.py` file with the production build of Oil.  It has some other dependencies on the Python interpreter that are stripped down version doesn't have.

oil_lang/objects.py
```diff
@@ -33,34 +35,35 @@ def __getitem__(self, typ):
...
   def __call__(self):
+    # type: () -> None
     # Array(1 2 3)
     raise AssertionError("Arrays need a parameterized type")
 
 
 # These are for data frames?
 
-class BoolArray(list):
+class BoolArray(List[bool]):
```
andychu: I'm not positive but I think this will cause a runtime error with `vendor/typing.py`?  i.e. I think `bin/osh` imports that.

Note the PYTHONPATH in `bin/osh`:

```
PYTHONPATH=$REPO_ROOT:$REPO_ROOT/vendor exec $REPO_ROOT/bin/oil.py osh "$@"
```

asokoloski:
@andychu whoops, yeah.  `Type` not being found in `typing` is an easy fix, but the other problem may be a blocker.  I'll have a quick look tomorrow, but I'm leaning toward just abandoning this.

asokoloski:
@andychu Ok, I figured out a relatively simple way to get `oil_lang/objects.py` work both with type checking and at runtime: 02b1a4fa1162d1a18007acb62d5ee5d05b114f3f.  I figure that even if all this code is python-only, having type annotations here may make mypy happier with other code that calls this stuff, maybe?  What do you think?

oil_lang/objects.py
asokoloski: Fixed by: 02b1a4fa1162d1a18007acb62d5ee5d05b114f3f

oil_lang/objects.py
asokoloski: Yep, that was broken too.  Fixed by: 02b1a4fa1162d1a18007acb62d5ee5d05b114f3f

asokoloski:
@andychu Or just let me know if you'd prefer to scrap the change, and ignore this file instead.  I'm not married to it -- I just wanted try it and see if it was possible :)

andychu:
OK this looks good.  There were some significant corrections and although I don't expect this file to translate with mycpp, it's good documentation.  Thanks!

PR #561, opened 2019-12-17
Add typechecking to osh/cmd_exec.py

asokoloski:
I found at least one bug in the process -- fix to follow.


asokoloski:
@andychu ok, that's enough for me for tonight, I think :)

asokoloski:
@andychu sorry, I should clarify -- this MR is ready to go, I think.  I'll put the bugfix stuff into another MR.

andychu:
OK cool, taking a look ...

andychu:
Beautiful, thank you!  And thanks for separating out the stuff that requires rewrites, which makes it easier to review.

asokoloski:
@andychu :D  you're welcome!

Heh, I was planning to go more granular with the commits (splitting out variable renaming, etc.) but I decided it was probably not worth the trouble, in the end.  Maybe next time I can plan things a bit better, but it's tough to tell in advance what you'll need to do to fix the errors.

PR #565, opened 2019-12-20
Assorted fixes, etc.   Aftermath of type-checking osh/cmd_exec.py

asokoloski:
@andychu let me know what you think of these.

### DIFF
--- a/spec/special-vars.test.sh
+++ b/spec/special-vars.test.sh
@@ -225,6 +225,17 @@ f
 ['4']
 ## END
 
+#### $LINENO and redirections (bug regression)
+>_tmp/lineno_regression$LINENO
+ls _tmp/lineno_regression1
+echo $LINENO
+## STDOUT: 
+_tmp/lineno_regression1
+3
+## END
+## BUG zsh STDOUT: 
+## END
+
 #### $LINENO for [[
 echo one
 [[ $LINENO -eq 2 ]] && echo OK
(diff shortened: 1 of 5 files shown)


spec/special-vars.test.sh
```diff
@@ -225,6 +225,17 @@ f
 ['4']
 ## END
 
+#### $LINENO and redirections (bug regression)
+>_tmp/lineno_regression$LINENO
+ls _tmp/lineno_regression1
```
andychu: The spec test frame work sets `$TMP` and that's preferrred, because some tests in `test/spec.sh` pass `--cd-tmp`, which makes the test more isolated from the file system.

asokoloski: Gotcha.  But I do see some tests that have hardcoded references to _tmp/ -- are those just older and haven't been updated?

asokoloski:
@andychu thanks for the merge!

andychu:
Yes I started using $TMP, but some tests haven't been updated!  I fix them if I'm in there for some other reason.  It's just more likely to make the tests flaky if there's garbage from other tests, or a previous run of the same test.

andychu:
Also it's probably a good idea to fix #42 so everyone's on the same page.

I think it would be as simple as adding them to Travis?  We already run a few tests.  Though for some reason I am reluctant to make the Travis build take longer than it does now.

Ideally we would have a continuous build that would let us run the current Travis script and all the spec tests in parallel.  That would force a cleaning up of all the spec tests (which may or may not involve `$TMP`)

Issue #441, opened 2019-07-20
Implement `times` builtin


timetoplatypus:
I'm working on a pull request to implement this feature. Things have changed a lot since the last time I worked on OSH, so I'm encountering errors I'm not familiar with, etc.

Planning to use the following module in the implementation:
https://docs.python.org/2.7/library/resource.html

andychu:
OK great!  What are the errors?  You can post them here or on #oil-dev on Zulip.

`resource.c` appears here, so we have the resource module:

https://www.oilshell.org/release/0.7.pre1/metrics.wwz/line-counts/nativedeps.txt

But I would still accept a PR if we didn't.  "Make it work" and "make it small" are separate steps, etc.

timetoplatypus:
```bash
$> build/dev.sh all
removed '_devbuild/gen/find_asdl.py'
removed '_devbuild/gen/find.marshal'
removed '_devbuild/gen/find_nt.py'
removed '_devbuild/gen/grammar.marshal'
removed '_devbuild/gen/grammar_nt.py'
removed '_devbuild/gen/id.h'
removed '_devbuild/gen/id_kind_asdl.py'
removed '_devbuild/gen/id_kind_asdl.pyc'
removed '_devbuild/gen/__init__.py'
removed '_devbuild/gen/__init__.pyc'
removed '_devbuild/gen/osh_help.py'
removed '_devbuild/gen/osh-lex.re2c.h'
removed '_devbuild/gen/osh-types.h'
removed '_devbuild/gen/runtime_asdl.py'
removed '_devbuild/gen/syntax_asdl.py'
removed '_devbuild/gen/syntax_asdl.pyc'
removed '_devbuild/gen/typed_arith_asdl.py'
removed '_devbuild/gen/typed_demo_asdl.py'
removed '_devbuild/gen/types_asdl.py'
removed '_devbuild/gen/types_asdl.pyc'
Wrote _devbuild/osh-quick-ref/ and _devbuild/gen/osh_help.py
Wrote _release/VERSION/doc/osh-quick-ref.html
renamed '_tmp/types_asdl.py' -> '_devbuild/gen/types_asdl.py'
Wrote _devbuild/gen/types_asdl.py
renamed '_tmp/syntax_asdl.py' -> '_devbuild/gen/syntax_asdl.py'
Wrote _devbuild/gen/syntax_asdl.py
renamed '_tmp/runtime_asdl.py' -> '_devbuild/gen/runtime_asdl.py'
[...]

PR #580, opened 2020-01-15
WIP: Implements `times` builtin

timetoplatypus:
Implements #441

### DIFF
--- a/osh/builtin.py
+++ b/osh/builtin.py
@@ -42,7 +42,6 @@ import libc
 import posix_ as posix
 
 import os
-import resource
 
 from mycpp import mylib
 if mylib.PYTHON:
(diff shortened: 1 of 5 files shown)


timetoplatypus:
Not ready for merge yet, just creating the PR so progress can be tracked.

osh/builtin.py
```diff
@@ -277,6 +280,28 @@ def _AppendParts(s, spans, max_results, join_next, parts):
...
 
 
+TIMES_SPEC = _Register('times')
+
+class Times(object):
+  def __call__(self, arg_vec):
+    if "SC_CLK_TCK" in os.sysconf_names:
+      clock_ticks_per_second = os.sysconf("SC_CLK_TCK")
+      # Convert resource usage (clock ticks) to seconds before pretty printing
+      print("%dm%1.3fs %dm%1.3fs" %
```
andychu: Hm I don't know this API / builtin very well, but can it be done with os.times() ?

https://docs.python.org/2/library/os.html#os.times

timetoplatypus: Ya, that looks like that'll work better. I'll test it out

timetoplatypus:
Looks like `os.times()` works, and it's a good bit cleaner, too.

andychu:
OK excellent!  Thanks for doing this.

Although this reminds me that we have no floating point in the stripped down CPython build !  

https://github.com/oilshell/oil/blob/master/native/libc.c#L358

However maybe this will still work with `mycpp` translation and we won't have to do that.  (Because `oil-native` is going to be the new Oil, as mentioned on the blog)

Either way this is great and gets us further.  Let me know on Zulip if you want to pick up any other tasks.

PR #584, opened 2020-01-21
[types] add scripts/function to help maintain correct imports when adding type annotation

asokoloski:
These are very hacky and intended to be temporary, just until the type annotation is completed.

@andychu I'm working on adding type annotations to osh/word_eval.py, and I decided that I've been spending too much time manually updating imports, so I wrote these helpers.  I don't know if anyone else will want to use them, but I thought it was better to commit them than not.

### DIFF
--- a/types/run.sh
+++ b/types/run.sh
@@ -78,16 +78,16 @@ checkable-files() {
 }
 
 need-typechecking() {
-    # This command is useful to find files to annotate and add to
-    # $MORE_OIL_MANIFEST.
-    # It shows all the files that are not included in
-    # $MORE_OIL_MANIFEST or $OSH_PARSE_MANIFEST, and thus are not yet
-    # typechecked by typecheck-more-oil here or
-    # `types/osh_parse.sh travis`.
-    comm -2 -3 \
-         <(checkable-files | sort | grep '.py$' | sed 's@^@./@') \
-         <(cat $MORE_OIL_MANIFEST $OSH_PARSE_MANIFEST | sort) \
-        | xargs wc -l | sort -n
+  # This command is useful to find files to annotate and add to
+  # $MORE_OIL_MANIFEST.
+  # It shows all the files that are not included in
+  # $MORE_OIL_MANIFEST or $OSH_PARSE_MANIFEST, and thus are not yet
+  # typechecked by typecheck-more-oil here or
+  # `types/osh_parse.sh travis`.
+  comm -2 -3 \
+    <(checkable-files | sort | grep '.py$' | sed 's@^@./@') \
+    <(cat $MORE_OIL_MANIFEST $OSH_PARSE_MANIFEST | sort) \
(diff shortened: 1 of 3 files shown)


asokoloski:
Also, of course, let me know if you want them brought up to a higher standard, or just don't want this sort of junk in the repo.

types/run.sh
```diff
@@ -94,6 +94,37 @@ typecheck-files() {
...
+add-imports() {
+	# Temporary helper to add missing class imports to the 'if
+	# TYPE_CHECKING:' block of a single module, if the relevant
+	# classes are found in one of COMMON_TYPE_MODULES
+
+	# Also, this prints out the typechecking output, to avoid having
+	# to run two redundant (and slow) typechecking commands.
+	local module=$1
+	export PYTHONPATH=.
+	module_tmp=$(mktemp)
```
andychu: - I find it easier to read and debug to use a hard-coded well-named path in `_tmp`.  Then if something goes wrong you can look at those files manually.
- And I wouldn't bother with `trap` because of this.
- please follow the existing style of using two spaces for indents here

asokoloski: @andychu sorry about the indentation, I didn't notice!  I've installed an emacs mode that I expect to make this never happen again (for shell, at least -- it doesn't do python).  Also fixed the need-typechecking function I added.  dde3ffb02ae0c08101bb47a712b1f8578ab3fe18

Temporary file paths have been sorted as well.  92d8805ba3649e6464f999f4b3a57941cc7e165c

asokoloski: Oh, and one more change -- f48392b8f85596ae1db9a356ba0af76e590df079.  I thought it would be simpler to just save the mypy output and let the caller can decide whether to print it or not.

types/run.sh
```diff
@@ -78,22 +78,54 @@ checkable-files() {
...
+  local module=$1
+  export PYTHONPATH=.
+  readonly module_tmp=_tmp/add-imports-module.tmp
+  readonly typecheck_out=_tmp/add-imports-typecheck-output
+  set +o pipefail
+  # unbuffer is just to preserve colorization (it tricks the command
+  # into thinking it's writing to a pty instead of a pipe)
+  unbuffer types/run.sh typecheck-files "$module" | tee "$typecheck_out" | \
+    grep 'Name.*is not defined' | sed -r 's/.*'\''(\w+)'\''.*/\1/' | \
+    sort -u | python devtools/findclassdefs.py "${COMMON_TYPE_MODULES[@]}" | \
```
andychu: Since these are just private scripts it's not worth messing with more, but I changed to use a shebang of `#!/usr/bin/env python2` and executing it directly (no `python`).  This will get Python 3 on some systems.

asokoloski: Ooh! good point.  I've been using a python2 virtualenv for my oilshell work and didn't think of that.  Thanks.

PR #585, opened 2020-01-21
[types] add type annotations to osh/word_eval.py

asokoloski:
This went ok.  I'm getting the hang of it, I think.  I used "Any" in a few places that seemed appropriate.  Also had to add some `assert isinstance(<var>, <type>)` statements to make mypy happy.

### DIFF
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -8,13 +8,19 @@ from _devbuild.gen.id_kind_asdl import Id, Kind
 from _devbuild.gen.syntax_asdl import (
     braced_var_sub, Token,
     word, word_e, word_t, compound_word,
-    bracket_op_e, suffix_op_e, word_part_e
+    word_part__AssocArrayLiteral,
+    bracket_op_e, suffix_op_e, word_part_e,
+    bracket_op__ArrayIndex, bracket_op__WholeArray,
+    suffix_op__Nullary, suffix_op__PatSub, suffix_op__Slice,
+    suffix_op__Unary, sh_array_literal,
 )
 from _devbuild.gen.runtime_asdl import (
     builtin_e, effect_e,
-    part_value, part_value_e, part_value_t, part_value__String,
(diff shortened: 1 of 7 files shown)


osh/word_eval.py
```diff
@@ -1488,7 +1666,7 @@ def EvalWordSequence2(self, words, allow_assign=False):
       argv: list of string arguments, or None if there was an eval error
     """
     if self.exec_opts.simple_word_eval:
-      return self.StaticEvalWordSequence2(words, allow_assign)
+      return self.StaticEvalWordSequence2(words, allow_assign) # follow
```
andychu: What is this comment for?

Also I use 2 spaces before #, although I'm won't  block on anything on that until we have an autoformatter...

asokoloski: Whoops!  That was just me trying to keep track while manually tracing through some calls.  I removed it (actually in 2 places): 37dcecc8c886c3ee8c0841d0cea86348e46dd5ff

asokoloski: For any future work I can jump right to using tagswitch.  Do you prefer that?

asokoloski:
For any future work I can jump right to using tagswitch.  Should I do that?

andychu:
Yeah if you think it's less total work then `with tagswitch()` is better.  It's already being used and translated in the parser so I assume it will work.

There's also this issue where I use `tag = node.tag_()`  in some places instead of `node.tag`.  That is also because of translation.  In C++  `tag_()` is an overloaded function, since you can have an "overloaded field".

I think I started off doing it in 2 steps because the first step is more automatable than the second (with some `sed` scripts).  But if neither step is automatable then it may be less work to combine them into one.  But use your discretion.

Thanks again!

andychu:
To clarify, `tag_()` is used for some if statements.  `with tagswitch()` is translated directly to a switch statement, and does the check internally so you don't need `tag_()`.

asokoloski:
Ok, thanks for clarifying!  And as always, glad to help :)

PR #590, opened 2020-01-24
Spec test: `popd` empty directory stack

timetoplatypus:
`popd` should return status code 1


andychu:
Thanks!  

(There is an unrelated lint failure I'll fix)

timetoplatypus:
Just a note so it's written somewhere. This spec test only checks if the return code is correct. It does not check if `popd` is also printing the error message to stderr

PR #591, opened 2020-01-25
Add '-c' and '-d' arguments to history builtin

lbonanomi:
This PR attempts to address https://github.com/oilshell/oil/issues/399 by adding support for deleting-from shell history and support for clearing shell history.

### DIFF
--- a/osh/builtin.py
+++ b/osh/builtin.py
@@ -739,8 +739,16 @@ class History(object):
 
     # Delete history entry by id number
     if arg.d:
-      cmd_index = int(rest[0]) - 1  # readline is indexed from 1
-      readline_mod.remove_history_item(cmd_index)
+      try:
+        cmd_index = int(rest[0]) - 1  # readline is indexed from 1
+      except IndexError:
+        raise args.UsageError('Requires a numeric argument')
+
+      try:
+        readline_mod.remove_history_item(cmd_index)
+      except ValueError:
+        raise args.UsageError('%r : history position out of range' % arg0)
+
       return 0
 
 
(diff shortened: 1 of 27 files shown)


andychu:
Thanks for looking at this!

It generally looks good although I didn't run it.  I'm thinking about how to test `history`.  There is one test in `spec/builtins2.test.sh` but it's only testing the syntax of the command because the module is hard to test!

Let me look a little more.

andychu:
I just added some unit tests since it's hard to test `readline` "from the outside"

https://github.com/oilshell/oil/commit/69032626528b01334fb319570cfd72d25e9cadaf

It works pretty well, you can now call `_TestHistory(['history', '-c'])` and then assert the output.

Please give it a try and let me know if you run into problems.  Thanks for looking into this!

lbonanomi:
Added some tests to try and reflect interactive use of the history command.

osh/builtin.py
```diff
@@ -731,6 +732,27 @@ def __call__(self, cmd_val):
...
+    # Clear all history
+    if arg.c:
+      readline_mod.clear_history()
+      return 0
+
+
+    # Delete history entry by id number
+    if arg.d:
+      try:
+        cmd_index = int(rest[0]) - 1  # readline is indexed from 1
```
andychu: This has the issue I mentioned on Zulip -- `int()` can raise ValueError.

The idiomatic way to do it is:

HISTORY_SPEC.ShortFlag('-d', args.Int)

And then arg.d is the integer itself.

This also makes `-d4` (no space) and `-d 4` (space) work, since the arg parser understands that.

lbonanomi: `-d4` and `-d 4` working is worth losing the more-specific error message.

andychu:
Great thank you for doing this!  I can definitely use more help with readline and anything else :)

PR #607, opened 2020-02-18
fix NoneType attribute error

abathur:
I'm not sure this is the "right" fix, but after introspecting for a while I did find a fix that solves the AttributeError and does appear returns the right list of attributes here:
```
...
    self.VisitChildren(node)
  File "/Users/abathur/work/oil/tools/deps.py", line 32, in VisitChildren
    for name, _ in node.ASDL_TYPE.GetFields():
AttributeError: 'NoneType' object has no attribute 'GetFields'
```

### DIFF
--- a/tools/deps.py
+++ b/tools/deps.py
@@ -29,7 +29,7 @@ class Visitor(object):
     """
     #print 'CHILD', node.ASDL_TYPE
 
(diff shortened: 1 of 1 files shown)


andychu:
Hm this wasn't an intended API, but hey if it works now, why not.

I don't have any plans for this code in the next year or so, so feel free to hack on it ...

PR #613, opened 2020-02-25
skip libc tests that fail on darwin/macOS

### LINKED ISSUE #611: native/libc_test.py fails on OS X
I think these were legit differences because we were testing out a lot of fnmatch(), and the libc on OS X behaves differently

Thread on Zulip

e.g. this one https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/building.20dev.20env.20in.20Nix.20.2B.20macOS/near/187785631

abathur:
Fixes #611. I went ahead and guarded just the specific test cases that fail.

### DIFF
--- a/native/libc_test.py
+++ b/native/libc_test.py
@@ -10,9 +10,12 @@
 libc_test.py: Tests for libc.py
 """
 import unittest
+import sys
 
 import libc  # module under test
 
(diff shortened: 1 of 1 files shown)


andychu:
OK I don't really like the `None or self.assert` style, but it's fine to get past these errors for now.

Issue #612, opened 2020-02-25
OSH behaves differently on libc's without extended glob


andychu:
- We need extended glob to run `bash-completion`
- We use GNU libc's extended glob implementation
- OS X doesn't have it

So some unit tests fail

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/oil-dev.20as.20a.20dependency.2C.20packaging/near/188993488

andychu:
Similar to #611 in that the underlying cause is  libc differences

andychu:
Related to #192

PR #614, opened 2020-02-26
skip compl. tests that fail on darwin/macOS

abathur:
This is per the tests in #612, but the issue sounds like it's cut a little more broadly, so I won't assert that this fixes it.


andychu:
Cool thanks

Issue #616, opened 2020-02-26
allow users to write obscure shell operators with function call syntax


andychu:
For users who care about readability and not compatibility.  Exact syntax TBD.

- `${x#/}` => `stripPrefix(x, '/')`
- `${x^^}` => `uppercase(x)`
- `${x@P}` => `promptFormat(x)`
- `${x//pat/replace}` => `replace(x, 'pat', 'replace', 'all')` (although I wanted this to use the left-to-right "pass" syntax, which I'm not sure if we'll get to)

Could also add an option to not strip trailing newline, e.g. `commandSub(cmd, newline=true)`

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Trailing.20Newline.20in.20Command.20Sub

Also:

* `glob()` for dynamic globbing
* `split()` for IFS splitting and other kinds of splitting
* `maybe()` for empty elision

shamrin:
What is the advantage of new syntax:
```
replace(x, 'pat', 'replace', 'all')
```

When this exits (it's even shorter):
```
$(echo $x | sed 's/pat/replace/g')
```

Similarly, why support `uppercase(x)` when there's `$(echo $x | tr '[a-z]' '[A-Z]')`?

Contrast with [fish shell approach][1]. Fish does not have function calls with round brackets, only commands.

In other words, what's Oil philosophy regarding function calls versus commands, substitutions and pipes?

P.S. Having said all that, I could never remember all those `${x...}` forms. And I do use some of them in my scripts! E.g. ,`${x-}` and `${x:-}`. In this example I cannot even tell what's the difference without consulting the docs.

[1]: https://fishshell.com/docs/current/design.html

shamrin:
*Brainstorm mode on.*

I wonder if this would be better:

```
${x | sed s/pat/replace/g}
```

And this:

```
${x | tr [a-z] [A-Z]}
```

andychu:
Yes good question.. this is open to debate, but the rationale is:

- `$(basename $x)` implies that you're forking a process, and so does `$(echo x)`, where `echo` is a builtin.  That is, a command substitution is a process invocation.   Something like `$basename(x)` is a function call. 
  - It might be possible to magically optimize things, but in general I would like to avoid magic.  One of the principles of Oil is that syntax and semantics should correspond.  Similar semantics should use similar syntax; different semantics should use different syntax.  So I don't want to use the same syntax for two different things.
[...]

PR #617, opened 2020-02-29
help.md: add lost back ticks

shamrin:


andychu:
Ah thanks, I forgot I had even written this text!  That is related to #616

PR #634, opened 2020-03-07
[WIP] more portable root user expansion

abathur:
Opening for discussion. Needs changes before merge. I suspect we can delete parts or break them out into other tests, and you may have ideas on another approach. 

Thoughts/notes:

1. `[[ ~ == $(eval printf \'%s\\\n\' \~$LOGNAME) ]]` is actually just part of an earlier attempt at fixing the test, but I kept the line for discussion since I haven't seen this narrow slice of behavior covered in another test. If you think it's worthwhile, I could imagine folding it into how _must_exist checks the first expansion, or creating a new test in tilde.test.sh?
2. Because using grep to test/filter the root user's path `/root` weakens the stdout assertion a little, I tried to compensate by also testing that the directories exist. If you think testing the end of the root user's path is enough, I can cut `_must_exist` and some related changes:
    - I pared `{/src,root}` back to `{,root}` since ~/src is less-likely to exist than ~/
    - delayed overriding HOME to test that the user home exists before testing output

### DIFF
--- a/spec/brace-expansion.test.sh
+++ b/spec/brace-expansion.test.sh
@@ -169,10 +169,29 @@ echo {foo~,~}/bar
 #### Two kinds of tilde expansion
 # NOTE: osh matches mksh.  Is that OK?
 # ~/foo and ~bar
+# necessary to establish ~ == ~current_user?
+# (could be a separate test now?)
+[[ ~ == $(eval printf \'%s\\\n\' \~$LOGNAME) ]]
+_must_exist(){ test -d "$1" && test -d "$2"; }
+_check_paths(){
+  echo $1
+  if test -d "$2"; then
+    # hope root's homedir ends in root
+    echo $2 | grep -E -o "/root$"
+  else
+    echo $2
+  fi
+}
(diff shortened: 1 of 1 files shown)


andychu:
Hm I get that the underlying problem is that `~root` isn't isolated... I think the "real" solution is to compare against what's in /etc/passwd then?

However I'm also kind of wary of making the tests less readable in the name of isolation.  Like if we need to go back and debug brace expansion, it's nice to have dumb simple tests.  I have a hard time reading this even if it's correct

I guess this is why I wanted to get an "overview" of what's wrong... i.e. so we can see if there are patterns that can let us fix a ton of tests at once, rather than rewriting tests one-by-one

andychu:
So I am making good progress on #167 but we haven't audited all the failures, either in Alpine or in Nix, and I think that will help with this specific issue

abathur:
Agh. This is partly/temporarily moot anyways. I convinced myself this would work but I was forgetting that, on NixOS, the build sandbox sets root's homedir to `/build`; I'll close for now.

PR #635, opened 2020-03-07
add diffable sh_spec output format

abathur:
Here's the diffable output formatter I've been picking at.

This doesn't address how to integrate it with spec.sh and spec-runner.sh.

Also, I noticed while scanning over the diff that I've smuggled a bugfix in here (lines 794-795). Let me know if you'd like me to separate it. I noticed it while running some tests in verbose mode (including `test/spec.sh var-op-patsub -v`), which produces a stack like:

```
  File "test/sh_spec.py", line 1362, in <module>
    sys.exit(main(sys.argv))
  File "test/sh_spec.py", line 1326, in main
    stats = RunCases(cases, case_predicate, shell_pairs, env, out, opts)
  File "test/sh_spec.py", line 666, in RunCases
    out.WriteRow(i, line_num, result_row, desc)
  File "test/sh_spec.py", line 777, in WriteRow
    self._WriteDetailsAsText(self.details)
  File "test/sh_spec.py", line 787, in _WriteDetailsAsText
    print(stdout.decode('utf-8'), file=self.f)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u03bc' in position 1: ordinal not in range(128)
```

### DIFF
--- a/test/sh_spec.py
+++ b/test/sh_spec.py
@@ -646,7 +646,10 @@ def RunCases(cases, case_predicate, shells, env, out, opts):
           d = (i, sh_label, actual['stdout'], actual['stderr'], messages)
           out.AddDetails(d)
 
-      result_row.append(cell_result)
+      if opts.format == 'diffable':
+        result_row.append((sh_label, cell_result))
+      else:
+        result_row.append(cell_result)
 
       stats.ReportCell(cell_result, sh_label)
 
@@ -788,7 +791,8 @@ class ColorOutput(object):
       print('%s stderr:' % shell, file=self.f)
       try:
         print(stderr.decode('utf-8'), file=self.f)
-      except UnicodeDecodeError:
+      # I got errors here without adding encode; was decode correct?
+      except (UnicodeDecodeError, UnicodeEncodeError):
         print(stderr, file=self.f)
       print('', file=self.f)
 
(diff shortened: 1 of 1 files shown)


abathur:
We discussed this on zulip, but I suppose for posterity I should better describe what this does. 

It outputs spec test results line by line, with the spec name, case name, shell, and result, to make it easier to diff different runs (at least across platforms, hopefully across branches/time/etc.)

An example diff showing a single difference between my local macos build and a macos build on travis-ci:
```diff
--- /Users/abathur/work/oil/diffable_macos_travis
+++ /Users/abathur/work/oil/diffable_macos_local
@@ -2588,7 +2588,7 @@
 builtin-printf.test.sh Runtime error for invalid integer (  zsh): known bug
 builtin-printf.test.sh Runtime error for invalid integer (  osh): ok (not great)
 builtin-printf.test.sh %(strftime format)T ( dash): not implemented
-builtin-printf.test.sh %(strftime format)T ( bash): failed
+builtin-printf.test.sh %(strftime format)T ( bash): passed
 builtin-printf.test.sh %(strftime format)T ( mksh): not implemented
 builtin-printf.test.sh %(strftime format)T (  zsh): not implemented
 builtin-printf.test.sh %(strftime format)T (  osh): not implemented
```

andychu:
I'm going to bikeshed a bit first before looking at the code :)  Why not:

```
builtin-printf [bash] FAIL %(strftime format)T
```

Basically have 3 fields with no spaces, and then the description as arbitrary text...

Also I think we should use the same FAIL N-I strings as we do in the ANSI output.  It will diff fine and it's more consistent, since people will use all formats

andychu:
Or maybe:

```
bash builtin-printf FAIL description of test case
dash builtin-printf FAIL description of test case
```

andychu:
One reason to put the shell first is that sometimes we only want to diff OSH, so you would do something like:

```
diff osh-only.txt <(grep ^osh other-run.txt)
```

And it's nice to have the first 3 fields as structured data, so you can do

```
awk '$1 == "bash" { print }'  run.txt
awk '$2 == "builtin-printf" { print }'  run.txt
awk '$3 == "FAIL" { print }'  run.txt
```

etc.

andychu:
I fixed the Unicode issues here.  I think it was messed up because this script used to be Python 3, and I ported it back to Python 2!

https://github.com/oilshell/oil/commit/b0647a46dcbbc3730a8ba6fa15269f40ac89377c

Sorry for the merge.  Let me know if you see any problems with that approach.

----

Also I think this change is on the right track.  If we have the parseable 4-column format, then we can read it back into `sh_spec.py` and generate something like this with say ORANGE for "unexpected failures" for Nix/Alpine:

http://www.oilshell.org/spec-results/2020-03-07__09-38-00__lisa.wwz/_tmp/spec/osh.html

That is, I want to ignore all the red failures, and concentrate on the orange ones.  Does that make sense?

PR #642, opened 2020-03-08
[spec/builtin-dirs] swap /lib for more-portable /dev

abathur:
Both cases try to use /lib, which neither macos nor nixos have. Instead of
making a longer, more abstract fix, I'm just swapping this for /dev. I checked
several docsets and didn't see a *nix without it, so it should at least be
"more" portable.

[fail before nixos](https://github.com/oilshell/oil/files/4303992/nixos_before_fail.txt)
[fail before macos](https://travis-ci.org/abathur/oil/jobs/659895964#L3349) (forgot to run case 6 here, but it fails the same way)

[pass after nixos](https://github.com/oilshell/oil/files/4303993/nixos_after_pass.txt)
[pass after macos](https://travis-ci.org/abathur/oil/jobs/659922765#L3346)

And, as before:
[pass before linux](https://travis-ci.org/abathur/oil/jobs/659895963#L3274) (forgot to run case 6 here, but it also passes)
[pass after linux](https://travis-ci.org/abathur/oil/jobs/659922764#L3274)

### DIFF
--- a/spec/builtin-dirs.test.sh
+++ b/spec/builtin-dirs.test.sh
@@ -72,7 +72,7 @@ cd /
 pushd /tmp >/dev/null
 echo --
 dirs -v
-pushd /lib >/dev/null
+pushd /dev >/dev/null
 echo --
 dirs -v
 ## status: 0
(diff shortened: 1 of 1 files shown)


andychu:
Cool sounds good!

PR #643, opened 2020-03-09
[spec/brace-expansion] improve ~root test portability

abathur:
To dodge assertion failures caused by the root user having a home directory other than `/root` on some platforms (including macos and nixos), this tests the fully-expanded paths from bash and zsh against the root user's home directory by extracting it from /etc/passwd. This revisits #634.

[fail before nixos](https://github.com/oilshell/oil/files/4304497/nixos_before_fail.txt)
[fail before macos](https://travis-ci.org/abathur/oil/jobs/659933665#L3349)

[pass after nixos](https://github.com/oilshell/oil/files/4304496/nixos_after_pass.txt)
[pass after macos](https://travis-ci.org/abathur/oil/jobs/659973493#L3346)

And of course:
[pass before linux](https://travis-ci.org/abathur/oil/jobs/659933664#L3274)
[pass after linux](https://travis-ci.org/abathur/oil/jobs/659973492#L3274)

### DIFF
--- a/spec/brace-expansion.test.sh
+++ b/spec/brace-expansion.test.sh
@@ -170,8 +170,18 @@ echo {foo~,~}/bar
 # NOTE: osh matches mksh.  Is that OK?
 # ~/foo and ~bar
 HOME=/home/bob
-echo ~{/src,root}
-## stdout: /home/bob/src /root
+expected_root_home=$(awk -F: -v v="root" '{if ($1==v) print $6}' /etc/passwd)
+test_expansion(){
+  if [[ "$1" == "~/src" && "$2" == "~root" ]]; then
+    echo "$1 $2"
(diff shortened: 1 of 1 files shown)


andychu:
I feel like this is mangling the readability of the test -- let's leave it aside for awhile.  I would like to see a report like this for Nix and then decide what to do:

http://www.oilshell.org/spec-results/2020-03-07__09-38-00__lisa.wwz/_tmp/spec/osh.html

abathur:
Here's one for nixos.
[nixos-build-html.tar.gz](https://github.com/oilshell/oil/files/4307980/nixos-build-html.tar.gz)

andychu:
OK thanks, so there are 82 errors compared to 66, which isn't too bad.

I'm interested in  `diffable` change and then seeing what fails on   OS X without Nix first (which will probably be a lot more).  Tackling that harder, user-facing problem will probably tell us a lot about what we need to do inside Nix.

Issue #604, opened 2020-02-17
consider an IR for ${}


andychu:
These are confusing to disambiguate from the LST:

- `${!ref-default}` -- prefix_op and unary suffix op
- `${!prefix@}` -- prefix_op and nullary suffix_op
- `${!array_keys[@]}` -- prefix op and bracket op

We could compile it to something like:

```
braced_var_sub =   Ref(...) | Prefix(...) | Keys(...)
```

andychu:
NOTE: `sh_lhs_expr` is sort of an IR because it's used for both:

- `(( x = 42 ))`
- `x=42`

andychu:
Note: This is done for arithmetic, e.g. for #640 and also for #651 

Not sure about `${}`, that may not happen for awhile.

PR #657, opened 2020-03-15
[word_eval] Fix the ${!prefix@} failure for the second time

akinomyoga:
See the fourth item in https://github.com/oilshell/oil/issues/653#issuecomment-599087679.

> See the following test case. Even though it succeeds for the first time, it fails for the second time.
> 
> **test3.sh**
> ```bash
> #!/usr/bin/env bash
> 
> abc_hello=1
> function hello1 { echo "${!abc_@}"; }
> hello1
> hello1
> ```
> 
> ```console
> $ osh test3.sh
> abc_hello
>   function hello1 { echo "${!abc_@}"; }
>                              ^~~~
> test3.sh:4: fatal: Bad indirect expansion: ''
> ```

### DIFF
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -913,6 +913,7 @@ class AbstractWordEvaluator(StringWordEvaluator):
           e_die("Array %r can't be referred to as a scalar (without @ or *)",
                 var_name, part=part)
 
+    suffix_processed = False
     if part.prefix_op:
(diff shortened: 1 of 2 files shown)


andychu:
Wow this is embarrassing -- the perils of test-driven development with narrow tests!  :-/

I remember I filed #604 at this time because I found the code confusing.

Note to self: don't mutate the AST!  Thank you.

PR #658, opened 2020-03-15
[word_eval] Fix shifted slicing of ${@:offset:length}.

akinomyoga:
This is a quick fix for the third item in https://github.com/oilshell/oil/issues/653#issuecomment-599087679. I tried to make the change small, so it might not be so elegant.

Note: `dash` and `mksh` fail in `test/spec.sh word-split` for newly added tests because they don't support `${@:offset:length}`.

### DIFF
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -201,6 +201,7 @@ def _PerformSlice(val,  # type: value_t
                   length,  # type: int
                   has_length,  # type: bool
                   part,  # type: braced_var_sub
+                  arg0_val, # type: value_t
                   ):
   # type: (...) -> value_t
   UP_val = val
@@ -244,8 +245,17 @@ def _PerformSlice(val,  # type: value_t
         e_die("The length index of a array slice can't be negative: %d",
               length, part=part)
 
-      # NOTE: unset elements don't count towards the length.
       strs = []  # type: List[str]
+
+      # NOTE: "begin" for positional arguments ($@ and $*) counts $0.
+      if arg0_val is not None:
+        if begin == 0:
+          if not has_length or length > 0:
+            strs.append(arg0_val.s)
+        elif begin > 0:
+          begin -= 1
+
+      # NOTE: unset elements don't count towards the length.
       for s in val.strs[begin:]:
         if s is not None:
           strs.append(s)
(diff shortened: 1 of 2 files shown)


andychu:
Oh man I didn't know this -- have you seen it documented anywhere?

This is confusing to me:

```
$ set -- 1 2 3; echo ${@:2}
2 3
$ a=(1 2 3); echo ${a[@]:2}
3
```

But apparently zsh agrees with bash!  So I guess we should do it... (if multiple shells agree on a behavior I take that as an argument for it.)  zsh is usually the one that disagrees the most.

I see what you're doing, but let me think if there's a simpler way...

andychu:
Wow this is also confusing:

```
$ set -- 1 2 3; echo ${@};
1 2 3
$ set -- 1 2 3; echo ${@:0};
-bash 1 2 3
```

You can get more items by adding index 0 !!!

akinomyoga:
> Oh man I didn't know this -- have you seen it documented anywhere?

I think I remember that it is explained by a simple one sentence in the manual. But today I found in the manual a detailed description with examples. Maybe the description is added recently.

> From *[Shell Parameter Expansion (Bash Reference Manual)](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)*
>
> If parameter is ‘@’, the result is length positional parameters beginning at offset. A negative offset is taken relative to one greater than the greatest positional parameter, so an offset of -1 evaluates to the last positional parameter. It is an expansion error if length evaluates to a number less than zero.
>
> The following examples illustrate substring expansion using positional parameters:
>
> ```console
> $ set -- 1 2 3 4 5 6 7 8 9 0 a b c d e f g h
> $ echo ${@:7}
> 7 8 9 0 a b c d e f g h
> $ echo ${@:7:0}
>
> $ echo ${@:7:2}
> 7 8
> $ echo ${@:7:-2}
> bash: -2: substring expression < 0
> $ echo ${@: -7:2}
> b c
> $ echo ${@:0}
> ./bash 1 2 3 4 5 6 7 8 9 0 a b c d e f g h
> $ echo ${@:0:2}
> ./bash 1
> $ echo ${@: -7:0}
> ```

andychu:
Thanks a lot!  I merged and moved the test cases to `var-op-slice` (renamed from `var-op-other`).

PR #667, opened 2020-03-18
[test/spec] Add failing spec tests for fd manipulations

akinomyoga:


andychu:
Thanks for noticing that case and fixing it!

Issue #10, opened 2017-03-06
Allow easy installation


hcorion:
I attempted to try and create an Arch Linux PKGBUILD for this project and realized that setup.py install only installs python libc, and not oil. Could this be fixed?

andychu:
An Arch package would be great, but it's a little early for that.  As mentioned in issue #7 , cd isn't even working yet!

Right now OSH is passing on 248 tests and failing on 166:

http://www.oilshell.org/git-branch/master/11eaa05c/andy-home/spec/RESULTS.html

Once we get that closer to 95% or so, I think it will make sense to release an Oil 0.1, and then an Arch package will make sense.  I hope that will be in a couple months.

I will ping this bug then :)  Thanks for trying this.

andychu:
It's now packaged as a C program rather than a Python program:

http://www.oilshell.org/blog/2017/07/23.html

Please try it and let me know what goes wrong!  :)

PR #668, opened 2020-03-18
Implement printf %()T %.s %*s %.*s

### LINKED ISSUE #650: Implement printf %*s and %.*s
This basically uses the first argument as a replacement for the asterisk, i.e. `printf '%*s' 10 foo` should be equivalent to `printf %10s foo` and `printf '%.*s' 2 bar` should be equivalent to `printf '%.2s' bar`.

akinomyoga:
Resolve #650

- 44d5af8e: Support `%()T`
- 8142a783: Support `%6.s` (fix parsing of precision)
- 99ead0e7: Support `%*.*s`
- c5d87362: Update test case for `#` flag

## Note on `%()T`

In the implementation of `%()T`, I needed to reflect the exported shell variable `TZ` in the real environment variable `TZ`. [This](https://github.com/oilshell/oil/blob/44d5af8e/osh/builtin_printf.py#L261-L268) is the corresponding section. I'm not sure whether I did this in the right way. (Maybe a new method like `self.mem.GetExportedVar(name)` should be added to `class Mem`). Could you check and edit if necessary?

## Note on `%6.s`

In the current master branch, the parsing of `%6.s` is broken to generate the following errors (`s` is treated as a precision), so I added a check for the precision and treat missing precision as `0` (as POSIX requirements).

```sh
osh$ printf '%6.s\n' a
  %6.s\n
      ^
(source.ArgvWord word_spid:2):1: Invalid printf format character
osh$ printf '%6.ss\n' a
  File "/home/murase/prog/ext-github/oilshell.oil/bin/oil.py", line 965, in <module>
    sys.exit(main(sys.argv))
  File "/home/murase/prog/ext-github/oilshell.oil/bin/oil.py", line 911, in main
    return AppBundleMain(argv)
  File "/home/murase/prog/ext-github/oilshell.oil/bin/oil.py", line 884, in AppBundleMain
    status = ShellMain('osh', argv0, main_argv, login_shell)
  File "/home/murase/prog/ext-github/oilshell.oil/bin/oil.py", line 686, in ShellMain
    prompt_plugin, errfmt)
  File "/home/murase/prog/ext-github/oilshell.oil/core/main_loop.py", line 95, in Interactive
    is_return, _ = ex.ExecuteAndCatch(node)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/cmd_exec.py", line 1862, in ExecuteAndCatch
    status = self._Execute(node, fork_external=fork_external)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/cmd_exec.py", line 1807, in _Execute
    status, check_errexit = self._Dispatch(node, fork_external)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/cmd_exec.py", line 1061, in _Dispatch
    status = self._RunSimpleCommand(cmd_val, fork_external)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/cmd_exec.py", line 802, in _RunSimpleCommand
    return self.RunSimpleCommand(cmd_val, fork_external)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/cmd_exec.py", line 890, in RunSimpleCommand
    return self._RunBuiltin(builtin_id, cmd_val, fork_external)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/cmd_exec.py", line 524, in _RunBuiltin
    status = self._RunBuiltinAndRaise(builtin_id, cmd_val, fork_external)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/cmd_exec.py", line 467, in _RunBuiltinAndRaise
    status = builtin_func.Run(cmd_val)
  File "/home/murase/prog/ext-github/oilshell.oil/osh/builtin_printf.py", line 218, in Run
    precision = int(part.precision.val)
ValueError: invalid literal for int() with base 10: 's'
```

## Note on `printf # flag `

The tests for flag `#` in the current master branch doesn't fully describe the behavior, so I added and modified tests.

### DIFF
--- a/osh/builtin_printf.py
+++ b/osh/builtin_printf.py
@@ -87,9 +87,13 @@ class _FormatStringParser(object):
       self._Next(lex_mode_e.PrintfPercent)
 
     if self.token_type == Id.Format_Dot:
+      dot_spid = self.cur_token.span_id
       self._Next(lex_mode_e.PrintfPercent)  # past dot
-      part.precision = self.cur_token
-      self._Next(lex_mode_e.PrintfPercent)
+      if self.token_type == Id.Format_Num or self.cur_token.val == '0':
+        part.precision = self.cur_token
+        self._Next(lex_mode_e.PrintfPercent)
+      else:
+        part.precision = Token(Id.Format_Num, dot_spid, '0')
 
     if self.token_type == Id.Format_Type:
       part.type = self.cur_token
--- a/osh/builtin_printf.py
+++ b/osh/builtin_printf.py
@@ -5,7 +5,7 @@ builtin_printf
 from __future__ import print_function
 
 from _devbuild.gen.id_kind_asdl import Id, Kind
-from _devbuild.gen.runtime_asdl import cmd_value__Argv
+from _devbuild.gen.runtime_asdl import cmd_value__Argv, value_e, value__Str
 from _devbuild.gen.syntax_asdl import (
     printf_part, printf_part_t,
     source
@@ -13,6 +13,8 @@ from _devbuild.gen.syntax_asdl import (
 from _devbuild.gen.types_asdl import lex_mode_e, lex_mode_t
 
 import sys
+import time
+import os
 
 from asdl import runtime
 from core import error
@@ -27,7 +29,7 @@ from mycpp import mylib
 from osh import string_ops
 from osh import word_compile
 
-from typing import Dict, List, TYPE_CHECKING
+from typing import Dict, List, TYPE_CHECKING, cast
 
 if TYPE_CHECKING:
   from frontend.lexer import Lexer
@@ -40,6 +42,7 @@ if mylib.PYTHON:
   PRINTF_SPEC = arg_def.Register('printf')  # TODO: Don't need this?
   PRINTF_SPEC.ShortFlag('-v', args.Str)
 
+shell_start_time = time.time()
 
 class _FormatStringParser(object):
   """
@@ -219,7 +222,7 @@ class Printf(object):
               s = s[:precision]  # truncate
           elif typ == 'q':
             s = string_ops.ShellQuoteOneLine(s)
-          elif typ in 'diouxX':
+          elif typ in 'diouxX' or typ.endswith('T'):
             try:
               d = int(s)
             except ValueError:
@@ -227,6 +230,8 @@ class Printf(object):
                 # TODO: utf-8 decode s[1:] to be more correct.  Probably
                 # depends on issue #366, a utf-8 library.
                 d = ord(s[1])
+              elif len(s) == 0 and typ.endswith('T'):
+                d = -1
               else:
                 # This works around the fact that in the arg recycling case, you have no spid.
                 if word_spid == runtime.NO_SPID:
@@ -252,6 +257,22 @@ class Printf(object):
                 s = '%x' % d
               elif typ == 'X':
                 s = '%X' % d
+            elif typ.endswith('T'):
+              # set timezone
+              tzcell = self.mem.GetCell('TZ')
+              if tzcell and tzcell.exported and tzcell.val.tag_() == value_e.Str:
+                tzval = cast(value__Str, tzcell.val)
+                os.environ['TZ'] = tzval.s
+              elif 'TZ' in os.environ:
+                del os.environ['TZ']
+              time.tzset()
+
+              if d == -1: # now
+                d = None
+              elif d == -2: # shell start time
+                d = shell_start_time
+              s = time.strftime(typ[1:-2], time.localtime(d));
+
             else:
               raise AssertionError()
 
--- a/osh/builtin_printf.py
+++ b/osh/builtin_printf.py
@@ -287,7 +287,10 @@ class Printf(object):
                 # TODO: utf-8 decode s[1:] to be more correct.  Probably
                 # depends on issue #366, a utf-8 library.
                 d = ord(s[1])
-              elif len(s) == 0 and part.type.id == Id.Format_Time:
+              elif part.type.id == Id.Format_Time and len(s) == 0 and word_spid == runtime.NO_SPID:
+                # Note: No argument means -1 for %(...)T as in Bash Reference
+                #   Manual 4.2 "If no argument is specified, conversion behaves
+                #   as if -1 had been given."
                 d = -1
               else:
                 # This works around the fact that in the arg recycling case, you have no spid.
@@ -315,7 +318,14 @@ class Printf(object):
               elif typ == 'X':
                 s = '%X' % d
             elif part.type.id == Id.Format_Time:
-              # set timezone
+              # %(...)T
+
+              # Initialize timezone:
+              #   `localtime' uses the current timezone information initialized
+              #   by `tzset'.  The function `tzset' refers to the environment
+              #   variable `TZ'.  When the exported variable `TZ' is present,
+              #   its value should be reflected in the real environment
+              #   variable `TZ' before call of `tzset'.
               tzcell = self.mem.GetCell('TZ')
               if tzcell and tzcell.exported and tzcell.val.tag_() == value_e.Str:
                 tzval = cast(value__Str, tzcell.val)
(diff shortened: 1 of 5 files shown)


andychu:
Thanks this is great!  I pulled and ran the tests and they passed.

And thanks for pointing out the bug.  I think the existing code might not be behaving as much like a "parser" as I want it do -- I'm looking at why.

-----

Please separate the `|` rules into separate lines like this, since the lexer can't be converted to C otherwise:

```
-   R('[1-9][0-9]*|\*', Id.Format_Num),
+   R('[1-9][0-9]*', Id.Format_Num),
+   R('\*', Id.Format_Num),
    C('.', Id.Format_Dot),
    # We support dsq.  The others we parse to display an error message.
-   R('[disqbcouxXeEfFgG]|\([^()]*\)T', Id.Format_Type),
+   R('[disqbcouxXeEfFgG]', Id.Format_Type),
+   R('\([^()]*\)T', Id.Format_Type),
```

(`build/dev.sh fastlex` which is part of `build/dev.sh all` fails with `|` because of the custom regex translator I use. )

---

In addition I try to maintain different token types rather than looking at token values, see property #2 here:

http://www.oilshell.org/blog/2019/02/07.html#three-nice-properties-of-a-parser

So it would be nice if

- `*` is `Id.Format_Star`
- `%()T` is `Id.Format_Time`

These are defined in `frontend/id_kind.py` under `'Format'`.  You can add `Star` and `Time` if you think the code fits into the code extra.

This is more of a style thing but I think it makes the code cleaner.  If it causes complications let me know.

-----

One nice way to view the AST is to turn `if 0` to `if 1` in `Run()`.  That's how I view the separation between parsing and runtime.

----

OK I'm looking at this a little more... I think it's good but I think the width parsing was already messed up as you noticed.

andychu:
Oh I see there are tow problems:

1. The grammar at line 48 is wrong?  Because the number after `.` is optional right?
2. The code isn't following the grammar 


```
    fmt           = Format_Percent Flag? Num? (Dot Num)? Type
```

https://github.com/oilshell/oil/blob/master/osh/builtin_printf.py#L48


```
- /tmp/lSedbk_builtin_printf.py
+ osh/builtin_printf.py

@@ -85,6 +85,8 @@

      if self.token_type == Id.Format_Dot:
        self._Next(lex_mode_e.PrintfPercent)  # past dot
+       if self.token_type != Id.Format_Num:
+         p_die('unexpected token after .', token=self.cur_token)
        part.precision = self.cur_token
        self._Next(lex_mode_e.PrintfPercent)

```

Now I get this:

```
$ bin/osh -c 'printf "%6.ss\n" x'
  %6.ss\n
     ^
(source.ArgvWord word_spid:2):1: unexpected token after .
```

-------

So I think it is important to update the grammar and then follow the code.  But if you think that will mess things up let me know and I can try it.

------

Long term I want to share the printf lexer and parser with a static variant:

```
echo ${x %.3f}   # printf formatting
```

This is analogous to `$'\n'` and `echo -e '\n'` being static and dynamic variants.  Oil is trying to be more static.  Those actually share the same lexing rules!  (with some changes since bash's duplicate code led to slightly different languages in these two cases)

andychu:
Hm yeah I think the grammar should be something like:

```
  control = Num | Star
  fancy = Flag? control? (Dot control?)? Type
  fmt     = Format_Percent (fancy | Time)
```

instead of

```
  fmt           = Format_Percent Flag? Num? (Dot Num)? Type
```

Does that look right?

No actually that's not right because you can combine width and precision with time:
 
```
andy@lisa:~/git/oilshell/oil$ printf '%10(%Y-%m)T\n'
   2020-03
```

Maybe add those test cases to make sure they work?

andychu:
So then I think the grammar is more like:

```
  control = Num | Star
  fmt = Percent control? (Dot control?)? (Type | Time)
```

?

akinomyoga:
Thank you for your review!

> Please separate the | rules into separate lines like this

> In addition I try to maintain different token types rather than looking at token values

b695317 I added token IDs `Star`, `Time`, and `Zero`. `Zero` was added because `0` can appear in both *flags* context and *precision* context.

> So I think it is important to update the grammar and then follow the code.

27e435c Updated.

> Also another consequence of this is that I try to never synthesize Token values.

28c169fd Changed so that `Dot` is assigned to `part.precision` when a value is not supplied.

> No actually that's not right because you can combine width and precision with time:

> Maybe add those test cases to make sure they work?

1c84df0 Added a test case and supported precision for `%()T`.

## Multiple flags?

By the way, I noticed that the current implementation only accepts a single flag. Since Oil currently supports only `-` and `0`, which are mutually exclusive, there is no use case for specifying multiple flags, but basically the parser should accept multiple flags such as `%#010.8x`.

osh/builtin_printf.py
```diff
@@ -252,14 +314,31 @@ def Run(self, cmd_val):
                 s = '%x' % d
               elif typ == 'X':
                 s = '%X' % d
+            elif part.type.id == Id.Format_Time:
+              # set timezone
```
andychu: Can you explain this part in a high level comment?  I don't know these APIs very well and eventually they will have to be translated to C.

(Actually if you prefer to write in C, then `native/libc.c` has related functions and is fairly easy to modify.)

Does `strftime` read `os.environ['TZ']`?  I'm not sure how it works exactly.

-----

OK I looked over the docs a bit, I guess `time.tzset()` reads `os.environ['TZ']`.  That's kind of confusing.  Please add a comment about that.

osh/builtin_printf.py
```diff
@@ -252,14 +314,31 @@ def Run(self, cmd_val):
...
+                os.environ['TZ'] = tzval.s
+              elif 'TZ' in os.environ:
+                del os.environ['TZ']
+              time.tzset()
+
+              if d == -1: # now
+                d = None
+              elif d == -2: # shell start time
+                d = shell_start_time
+              s = time.strftime(typ[1:-2], time.localtime(d));
```
andychu: no semicolon needed

Also where does `-2` come from?  I'm having trouble following that.  May need more comments.

akinomyoga: I removed the semicolon and added comments in dcf99bd. Also, I changed to explicitly set `d = time.time()` (not relying on the default behavior of `time.localtime(None)`).

osh/builtin_printf.py
akinomyoga: dcf99bd Thank you! I added code comments.

akinomyoga:
41f1b6a I also added support of multiple flags.

andychu:
Great thank you!  This is one of the biggest contributions ever to Oil.

bash `printf` is so  complicated!  This is why I want a simple functions in Oil like:

```
var s = strftime('%Y-%m-%d') 
```

:-) 


I made this change to accomodate how we strip down the CPython interpreter (will be obsolete after C++ translation, but it's closer to what C looks like)

https://github.com/oilshell/oil/commit/172e65400c9c87b5adea3ed524d0889b6cdd0a61

Also I adjusted this zsh output:

https://github.com/oilshell/oil/commit/eba6506d83ff733fef5510dc0fbc72988b1feff6

Can you try building shells like this?  At least 2 people have done it before and it takes 2-3 minutes.

https://github.com/oilshell/oil/blob/master/test/spec-bin.sh

That is the "golden version" of the shell we test.  (Some of them are old but they are stable/known.  The spec tests tickle lots of verison-specific behavior.)

----

Thanks again!

akinomyoga:
Thank you!

> 172e654

FYI: I understand the motivation for using POSIX API and I'm not sure if it causes a real problem or not. but [Python documentation](https://docs.python.org/3/library/os.html#os.environ) recommends to call `setenv/unsetenv` indirectly through `os.environ` instead of directly calling them to keep consistency between the environment variables that Python caches and the real environment variables.

> Can you try building shells like this?

Oh, I'm sorry. I actually noticed the results of test cases are sometimes different but didn't think about it seriously. Now I built the shells! Thank you!

andychu:
Minor updates:

I removed a condition on TZ, and added some tests, see rationale here:

https://github.com/oilshell/oil/commit/fddcf679ec2ffdbf381a650fb31eaf145a7f9e10

Basically there is no equivalent of `posix.environ` in C++ -- it's just getenv, putenv, and unsetenv.

----

Also, I thought that TZ was like LANG.  LANG is respected even if it's NOT exported.  But that's not true of TZ.  https://github.com/oilshell/oil/issues/527

I looked in `lib/sh/strftime.c` in bash and it does something ad hoc and even has some `#ifdefs`.

I think the current behavior is fine but if you see any issues let me know.

-----

Right before I mentioned `spec-bin.sh` I moved it to `_deps/spec-bin`:

https://github.com/oilshell/oil/commit/ff303b6abab7970f82bff7fbdaab74ad08dc7673

Then I realized you might not have pulled in the last 5 minutes.  So if you have `_tmp/spec-bin` just move it to `_deps/spec-bin` and it should work fine.

I did this for the Travis build which caches the `_deps` directory: http://travis-ci.oilshell.org/jobs/

PR #670, opened 2020-03-21
[test/spec] Failing spec tests for rvalue dynamic var name in arith

akinomyoga:
Add tests to demonstrate a use case of dynamical rvalue/lvalue variable names in arithmetic expressions.

### DIFF
--- a/spec/arith.test.sh
+++ b/spec/arith.test.sh
@@ -532,7 +532,7 @@ xbar[5]=42
 ## N-I dash status: 2
 ## N-I dash stdout-json: ""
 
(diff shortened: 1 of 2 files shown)


andychu:
Ah OK I just added the 4 failing ones to cover the lvalue.  But this is for both lvalue and rvalue?

Maybe prefix the test case names with ble.sh: so we remember where they came from, since this file is very long.  It looks like it's matching the usage pattern of ble.sh which is good.

akinomyoga:
204b3d2b Thank you! I updated the test names

Issue #647, opened 2020-03-09
Output of declare -p / declare / declare -f / trap -p / trap (ble.sh)


akinomyoga:
Oil does not print the definition of variables for  `declare -p`. The script ble.sh uses `declare -p` to save states of scalar variables and arrays (`saved=$(declare -p foo bar)`) which will be later used  to restore the states by `eval` (`eval -- "$saved"`). I believe this is the expected usage of Bash `declare -p` (which is the reason that it prints the reusable form for `eval`).

```console
$ bash -c 'a=1; declare -p a; declare -p b'
declare -- a="1"
bash: line 0: declare: b: not found
$ bin/osh -c 'a=1; declare -p a; declare -p b'
a
$
```

Similarly Bash prints the definition of all the variables and functions with `declare` without arguments, but Oil does nothing for this.

```console
$ bin/osh -c 'a=1; declare'
$
```

In Bash, the function definition can be obtained by `declare -f`. But Oil does not provide the function body but only the function name. The script ble.sh uses `declare -f` to save/restore the functions or dynamically patch the behavior of existing functions. (If you are interested, maybe you can have a look at meta functions `ble/function#push`, `ble/function#pop`, and `ble/function#advice` in src/util.sh of the ble.sh source for save, restore, and patching, respectively).

```console
$ bash -c 'function aaa { echo aaa; }; declare -f aaa'
aaa ()
{
    echo aaa
}
$ bin/osh -c 'function aaa { echo aaa; }; declare -F aaa'
aaa
$
```

Is there any other way in Oil to obtain the body of existing shell functions? I tried `trap` but it neither output the body.

```console
$ bash -c 'function aaa { echo aaa; }; type aaa'
aaa is a function
aaa ()
{
    echo aaa
}
$ bin/osh -c 'function aaa { echo aaa; }; type aaa'
aaa is a function
$
```

Bash builtin `trap` also produces reusable form which can be used for save and restore of trap states. However, Oil produces not reusable form of outputs.

```console
$ bin/osh -c 'trap "echo hello" INT; trap -p'
2 _TrapHandler
```

In Bash, `trap` without arguments prints the current set of the registered traps in the same way as `trap -p`. But Oil produces an error.

```console
$ bin/osh -c trap
  trap
  ^~~~
[ -c flag ]:1: 'trap' requires a code string
$
```

andychu:
[...]

PR #671, opened 2020-03-21
[builtin/declare] Support -p

akinomyoga:
Support `declare` and `declare -p` (cf #647, but `declare -f`, `trap`, `trap -p` are not yet supported)

- Note: Bash quotes the values with double quotations, but this patch uses single quotations provided by `string_opt.ShellQuote`.
- Note: Bash ignores the flag `-g` when `-p` flag is specified, but this patch supports it.
- Note: Bash ignores the flags `-nrxaA` when `-p` flag and variable names are specified, but this patch supports them.
- Please see the changes in `spec/assign-extended.test.sh` for the detailed behavior

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -1387,10 +1387,10 @@ class Mem(object):
 
     return value.Undef()
 
-  def GetCell(self, name):
+  def GetCell(self, name, lookup_mode = scope_e.Dynamic):
     # type: (str) -> cell
     """For the 'repr' builtin."""
-    cell, _ = self._ResolveNameOnly(name, scope_e.Dynamic)
+    cell, _ = self._ResolveNameOnly(name, lookup_mode)
     return cell
 
   def Unset(self, lval, lookup_mode):
@@ -1497,6 +1497,31 @@ class Mem(object):
           result[name] = str_val.s
     return result
 
+  def GetAllCells(self, lookup_mode = scope_e.Dynamic):
+    # type: () -> Dict[str, cell]
+    """Get all variables and their values, for 'set' builtin. """
+    result = {}  # type: Dict[str, str]
+
+    if lookup_mode == scope_e.Dynamic:
+      scopes = self.var_stack
+    elif lookup_mode == scope_e.LocalOnly:
+      scopes = self.var_stack[-1:]
+    elif lookup_mode == scope_e.GlobalOnly:
+      scopes = self.var_stack[0:1]
+    elif lookup_mode == scope_e.LocalOrGlobal:
+      scopes = self.var_stack[0:1]
+      if len(self.var_stack) > 1:
+        scopes.append(self.var_stack[-1])
+    else:
+      raise AssertionError()
+
+    for scope in scopes:
+      for name, cell in iteritems(scope):
+        result[name] = cell
+    return result
+
+  def IsGlobalScope(self):
+    return len(self.var_stack) == 1
 
 def SetLocalString(mem, name, s):
   # type: (Mem, str, str) -> None
--- a/core/state.py
+++ b/core/state.py
@@ -1388,7 +1388,7 @@ class Mem(object):
     return value.Undef()
 
   def GetCell(self, name, lookup_mode = scope_e.Dynamic):
-    # type: (str) -> cell
+    # type: (str, scope_t) -> cell
     """For the 'repr' builtin."""
     cell, _ = self._ResolveNameOnly(name, lookup_mode)
     return cell
@@ -1498,7 +1498,7 @@ class Mem(object):
     return result
 
   def GetAllCells(self, lookup_mode = scope_e.Dynamic):
-    # type: () -> Dict[str, cell]
+    # type: (scope_t) -> Dict[str, cell]
     """Get all variables and their values, for 'set' builtin. """
     result = {}  # type: Dict[str, str]
 
@@ -1521,6 +1521,7 @@ class Mem(object):
     return result
 
   def IsGlobalScope(self):
+    # type: () -> bool
     return len(self.var_stack) == 1
 
 def SetLocalString(mem, name, s):
--- a/core/state.py
+++ b/core/state.py
@@ -1387,7 +1387,7 @@ class Mem(object):
 
     return value.Undef()
 
-  def GetCell(self, name, lookup_mode = scope_e.Dynamic):
+  def GetCell(self, name, lookup_mode=scope_e.Dynamic):
     # type: (str, scope_t) -> cell
     """For the 'repr' builtin."""
     cell, _ = self._ResolveNameOnly(name, lookup_mode)
@@ -1497,7 +1497,7 @@ class Mem(object):
           result[name] = str_val.s
     return result
 
-  def GetAllCells(self, lookup_mode = scope_e.Dynamic):
+  def GetAllCells(self, lookup_mode=scope_e.Dynamic):
     # type: (scope_t) -> Dict[str, cell]
     """Get all variables and their values, for 'set' builtin. """
     result = {}  # type: Dict[str, str]
--- a/osh/builtin_assign.py
+++ b/osh/builtin_assign.py
@@ -98,11 +98,23 @@ def _PrintVariables(mem, cmd_val, arg, print_flags, readonly = False, exported =
       decl += "=" + string_ops.ShellQuote(str_val.s)
     elif val.tag == value_e.MaybeStrArray:
       array_val = cast(value__MaybeStrArray, val)
-      body = ''
-      for element in array_val.strs:
-        if body: body += ' '
-        body += string_ops.ShellQuote(element or '')
-      decl += "=(" + body + ")"
+      if None in array_val.strs:
+        # Note: Arrays with unset elements are printed in the form:
+        #   declare -p arr=(); arr[3]='' arr[4]='foo' ...
+        decl += "=()"
+        first = True
+        for i, element in enumerate(array_val.strs):
+          if element is not None:
+            if first:
+              decl += ";"
+              first = False
+            decl += " " + name + "[" + str(i) + "]=" + string_ops.ShellQuote(element)
+      else:
+        body = ''
+        for element in array_val.strs:
+          if body: body += ' '
+          body += string_ops.ShellQuote(element or '')
+        decl += "=(" + body + ")"
     elif val.tag == value_e.AssocArray:
       assoc_val = cast(value__AssocArray, val)
       body = ''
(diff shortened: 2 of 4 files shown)


akinomyoga:
Also, there is a problem in implementing `declare -p` for an array that have unset elements. In Bash, arrays have a literal of the form `([index]=value)`, so unset elements can be skipped in the literals. However, Oil does not support the form `([index]=value)` for indexed arrays, so I don't know how to represent the array with unset elements. Current implementation outputs an empty string `''` for a value `None` as follows:

```bash
$ bash -c 'arr=(); arr[3]=foo; declare -p arr'
declare -a arr=([3]="foo")
$ osh -c 'arr=(); arr[3]=foo; declare -p arr'
declare -a arr=('' '' '' 'foo')
```

So the question is:

## Q. Is there any literal to represent an array with unset elements in Oil?

If Oil does not support such a literal, maybe the output of `declare` could be changed to the following form. Do you have any other ideas?

```bash
declare -a arr; arr[3]='foo'
```

spec/assign-extended.test.sh
```diff
@@ -99,6 +99,366 @@ echo $?
...
+test_var4='test_var1'
+test_var5='555'
+[readonly]
+declare -r test_var2='222'
+[export]
+declare -x test_var3='333'
+[local]
+test_var5='555'
+## END
+## OK bash STDOUT:
```
andychu: Maybe instead of listing exact output of OSH and bash, you can `eval` the output of `declare -p`?

I think that's the property you want, right?

It makes the test a little more meaningful if both bash and Oil pass the exact same way.

-----

Not all tests have to be like that , but I think at least one should no?

osh/builtin_assign.py
```diff
@@ -25,10 +26,108 @@
   from frontend import arg_def
 
 
+def _PrintVariables(mem, cmd_val, arg, print_flags, readonly = False, exported = False):
+  flag_g = getattr(arg, 'g', None)
```
andychu: These can all just be `arg.g`, `arg.n`, etc. right?  And then you don't really need the temporary

core/state.py
```diff
@@ -1497,6 +1497,31 @@ def GetAllVars(self):
...
+        scopes.append(self.var_stack[-1])
+    else:
+      raise AssertionError()
+
+    for scope in scopes:
+      for name, cell in iteritems(scope):
+        result[name] = cell
+    return result
+
+  def IsGlobalScope(self):
```
andychu: this should have `# type: () -> bool`

TODO: I think we can set up the continuous build to alert of that, since type checking is running there now

osh/builtin_assign.py
```diff
@@ -25,10 +26,108 @@
...
+    print_all = True
+    cells = mem.GetAllCells(lookup_mode)
+    names = sorted(cells)
+  else:
+    print_all = False
+    names = []
+    cells = {}
+    for pair in cmd_val.pairs:
+      name = pair.lval.name
+      if pair.rval and pair.rval.tag == value_e.Str:
```
andychu: It's better to use `tag_()` though it's unfortunately uglier.  That's because it translates to C++ better -- C++ doesn't have "virtual" fields like Python, only methods.  I plan to get rid of `.tag` and make it `.tag()`, but for now it's `.tag_()`

osh/builtin_assign.py
```diff
@@ -25,10 +26,108 @@
...
+      if cell.nameref: flags += 'n'
+      if cell.readonly: flags += 'r'
+      if cell.exported: flags += 'x'
+      if val.tag == value_e.MaybeStrArray:
+        flags += 'a'
+      elif val.tag == value_e.AssocArray:
+        flags += 'A'
+      if flags == '-': flags += '-'
+
+      decl = 'declare ' + flags + ' ' + name
```
andychu: generally speaking I use the style of creating a `parts = []` array and appending to it, then `print(''.join(parts))`.

But that is not super important now, if it passes the tests it seems fine

core/state.py
```diff
@@ -1497,6 +1497,31 @@ def GetAllVars(self):
           result[name] = str_val.s
     return result
 
+  def GetAllCells(self, lookup_mode = scope_e.Dynamic):
```
andychu: Also I use the style of `foo='default'` (no space).


 I suppose at some point we should do something about issue #1 (autoformatting for Python).  If enough contributors want it, I will be motivated to :)

akinomyoga:
Thank you for your fast review and useful comments!

> I guess it can switch to that form only if None is in the array?

I implemented 3a5eb656.

spec/assign-extended.test.sh
akinomyoga: > Maybe instead of listing exact output of OSH and bash, you can eval the output of declare -p?

Thank you for the suggestion! That's a good idea! I added such a test d1c85b43, but it is still a very simple test.

Actually, the problem is "*how to check the results*" of `eval` & `declare`: The existence, the flags, and the values of each variable should be checked (*i.e.*, printed for the test). In the case of indexed/associative array variables, the existence and the value of each element, and *non-existence* of redundant elements should also be checked (*i.e.*, printed). If one wants to test everything in the exact same way in Bash and Oil, it seems that one needs to implement another `declare -p` in pure shell scripts (which may involve other bugs/divergences on the detailed behavior of each shell), so I gave up.

> I think that's the property you want, right?

Yes! Thank you.

osh/builtin_assign.py
akinomyoga: This function `_PrintVariables` is shared by `class Export`, `Readonly`, and `NewVar`. Each has its own definition for `arg`. `Export`, `Readonly`, and `NewVar` supports flags `-n`, `-aA`, and `-gnrxaA`, respectively. For example, some have flag `g` and the others don't have the flag `g`. So, one first needs to check whether each attribute is defined in `arg` or not, or otherwise, I get `AttributeError`.

I thought about one possibility to create a dictionary at the caller side something like `_PrintVariables(..., arg={'g': None, 'n': None, 'r': None, 'x': None, 'a': arg.a, 'A': arg.A}, ...)`. But it looks like redundant codes. Also, we need to keep consistency between this function call and the definitions of `arg` of each classes; every time one updates the set of flags in `arg`, one needs to update this function call accordingly, which is not so maintainable.

Another possibility may be to support the same set of flags for `export` and `readonly` as `declare`.

Do you have an idea of a more clever way to implement this?

core/state.py
akinomyoga: Thank you! Updated 9e94f900.

osh/builtin_assign.py
akinomyoga: OK! Updated 0bd3c4a3.

osh/builtin_assign.py
akinomyoga: Thank you! I changed to use that strategy in d3dd9c97.

core/state.py
akinomyoga: Oh, thank you! Hm, [PEP 8](https://www.python.org/dev/peps/pep-0008/#other-recommendations) (mentioned in #1) says that spaces are not recommended for keyword arguments and default arguments.

> *PEP-8 - Other Recommendations*
>
> Don't use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter.

 Updated e2c89a00.

osh/builtin_assign.py
andychu: Ah OK I see.   This is OK for now but could use a comment saying that.

(I think I have an idea of how it will translate to C++ ... the dynamic types of `arg.X` are an issue I've been thinking about.)

spec/assign-extended.test.sh
andychu: Yes that's true, it's tricky.  One possibility is to `sed` to different variable names, and then check for equality.  Something like:

```
eval -- $(declare -p | grep ^myarray | sed s/^myarray=/restored=/)
# now check myarray == restored

arrays_equal ...   # I wrote a function that takes two lengths and then @A and @B concatenated
```

But yes this is a lot of trouble ...

Actually in Oil we may be able to help with this.  We simply have

```
if (A == B) {  # are string, arrays, associative arrays equal?
}
```

but that's in the future

osh/builtin_assign.py
```diff
@@ -25,10 +26,122 @@
...
+      if cell.nameref: flags.append('n')
+      if cell.readonly: flags.append('r')
+      if cell.exported: flags.append('x')
+      if val.tag_() == value_e.MaybeStrArray:
+        flags.append('a')
+      elif val.tag_() == value_e.AssocArray:
+        flags.append('A')
+      if len(flags) == 0: flags.append('-')
+
+      decl.extend(["declare -", ''.join(flags), " ", name])
```
andychu: It looks like you don't need `''.join(flags)` or `''.join(body)` below, because we join it at the end anyway.  But this is a minor detail

osh/builtin_assign.py
akinomyoga: OK!

> This is OK for now but could use a comment saying that.

It has been already merged, so could you write a comment for this? Thank you!

osh/builtin_assign.py
akinomyoga: Just for curiosity, how can I write in this case? The reason why I maintained other lists `flags` and `body` is because I wanted to use `len(flags)` and `len(body)`. Ah, OK. Maybe I could write something like the following, but I just didn't want to split a simple line into many lines.

```py
decl.append("declare -")
decl.extend(flags)
decl.extend([" ", name])
```

Or this? But the following code seems even less efficient because it induces addtional list generations.
```py
decl.extend(["declare -"] + flags + [" ", name])
```

andychu: I usually check guard space separation in a loop with `i != 0` , and then you don't need to check len(body).  And I think you can check `len(assoc_val.d)` for the last check.

But those are minor details and I think it's fine.  There is one more string allocation for `''.join(flags)` and one more for `''.join(body)`, but like you say it's easy to introduce other allocations elsewhere.  After translating to C++ the previous style of `+=` would introduce a lot more allocations.

I actually did go back and optimize allocations in C++ generated from Python for the parser!  e.g. the last step of http://www.oilshell.org/blog/2020/01/parser-benchmarks.html

But I think this is pretty close to optimal, and it's not a hot path anyway, i.e. `declare -p` is not that common.

andychu:
BTW the place where I expect it to matter was `osh/word_eval.py`.  It initally built up a  lot of intermediate lists and used `+` to concatenate them, but now it uses the `.append()` or `.write()` style.  There are a lot of tiny objects in this algorithm, as opposed to the escape codes in strings that other shells use:

https://github.com/oilshell/oil/wiki/OSH-Word-Evaluation-Algorithm

But I expect we can make it pretty fast ...

osh/builtin_assign.py
akinomyoga: OK! I see, it is aimed to be translated to C++. Thank you!

PR #672, opened 2020-03-21
[core/process] Fix/Implement redirections <>, 5>&-, 6>&5-, {fd}>file, etc.

akinomyoga:
## 2. 7ea281a7 Update test

The arguments don't match with the function declaration? I guess the test is old. If this is not the right fix, please let me know.

## 3. 8dd5a020 Add tests for redirections

This demonstrates bugs related to the subsequent fixes (4., 5., and 6.) and also includes tests for `20> file` and a Bash bug.

## 4. 10136ceb Fix "Bad file descriptor" on 3>&3
This is a bug fix. The file descriptor was broken with `echo 3>&3`. The following test case (in `spec/redirect.test.sh`) had been failing due to this bug.

```bash
#### : 3>&3 (OSH regression)
: 3>&3
echo hello
## stdout: hello
## BUG mksh stdout-json: ""
## BUG mksh status: 1
```

## 5. 17e2f052 Fix fd-restoring order in FdState.Pop

This is another fix. The order of restoring original fds in `saved` and `need_close` were mixed. The ordering should be preserved. The following test case (in `spec/redirect.test.sh`) had failed due to this problem:

```bash
#### : 3>&- << EOF (OSH regression: fail to restore fds)
exec 3> "$TMP/fd.txt"
echo hello 3>&- << EOF
EOF
echo world >&3
exec 3>&-
cat "$TMP/fd.txt"
## STDOUT:
hello
world
## END
```

## 6. f5c21786  Fix fd-leak on `: 5> file`
See the following test case. The file descriptor `9` was still alive after the end of the command.

```bash
#### : 9> fdleak (OSH regression)
: 9> "$TMP/fd.txt"
( echo world >&9 )
cat "$TMP/fd.txt"
## stdout-json: ""
```

## 7. 0af7781f Fix fd-leak on `exec 5>&1`

The saved fds are not released on `FdState.MakePermanent()`. See the following example. In Oil, there were remaining fds as many as the invokations of `exec 5>&1`.

```
bash$ echo /proc/self/fd/* | wc -w
9
bash$ for i in {1..100}; do exec 5>&1; done
bash$ echo /proc/self/fd/* | wc -w
10

osh$ echo /proc/self/fd/* | wc -w
8
osh$ for i in {1..100}; do exec 5>&1; done
osh$ echo /proc/self/fd/* | wc -w
108
```

I didn't add a test to `spec/redirect.test.sh` because I don't know how to make it portable (`/proc/self/fd` is available in Linux but not portable). I tried to use `ulimit -n 100` for this, but it turned out that the limit cannot be changed in Oil (see below). Maybe I could have implemented the test case to generate fds as many as `$(ulimit -n)`, but I'm afraid that `$(ulimit -n)` might be extremely large in some system (but I don't know actually).

```
osh$ ulimit -n 100
osh$ ulimit -n
1024
```

## 8. da16c453 Support redirection {fd}>, {fd}<, etc.

Also, it includes the support of more-than-single-digit redirections such as `20> file`. Actually, I'm not familiar with ASDL, so maybe I'm doing weird things although it works as expected superficially. I would appreciate it if you could check this commit in detail.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -103,19 +103,17 @@ class _FdFrame(object):
   def __init__(self):
     # type: () -> None
     self.saved = []  # type: List[Tuple[int, int]]
-    self.need_close = []  # type: List[int]
     self.need_wait = []  # type: List[Tuple[Process, Waiter]]
 
   def Forget(self):
     # type: () -> None
     """For exec 1>&2."""
     del self.saved[:]  # like list.clear() in Python 3.3
-    del self.need_close[:]
     del self.need_wait[:]
 
   def __repr__(self):
     # type: () -> str
-    return '<_FdFrame %s %s>' % (self.saved, self.need_close)
+    return '<_FdFrame %s>' % self.saved
 
 
 class FdState(object):
@@ -271,7 +269,7 @@ class FdState(object):
 
   def _PushClose(self, fd):
     # type: (int) -> None
-    self.cur_frame.need_close.append(fd)
+    self.cur_frame.saved.append((-1, fd))
 
   def _PushWait(self, proc, waiter):
     # type: (Process, Waiter) -> None
@@ -464,23 +462,23 @@ class FdState(object):
     frame = self.stack.pop()
     #log('< Pop %s', frame)
     for saved, orig in reversed(frame.saved):
-      try:
-        posix.dup2(saved, orig)
-      except OSError as e:
-        log('dup2(%d, %d) error: %s', saved, orig, e)
-        #log('fd state:')
-        #posix.system('ls -l /proc/%s/fd' % posix.getpid())
-        raise
-      posix.close(saved)
-      #log('dup2 %s %s', saved, orig)
-
-    for fd in frame.need_close:
-      #log('Close %d', fd)
-      try:
-        posix.close(fd)
-      except OSError as e:
-        log('Error closing descriptor %d: %s', fd, e)
-        raise
+      if saved == -1:
+        #log('Close %d', orig)
+        try:
+          posix.close(orig)
+        except OSError as e:
+          log('Error closing descriptor %d: %s', orig, e)
+          raise
+      else:
+        try:
+          posix.dup2(saved, orig)
+        except OSError as e:
+          log('dup2(%d, %d) error: %s', saved, orig, e)
+          #log('fd state:')
+          #posix.system('ls -l /proc/%s/fd' % posix.getpid())
+          raise
+        posix.close(saved)
+        #log('dup2 %s %s', saved, orig)
 
     # Wait for here doc processes to finish.
     for proc, waiter in frame.need_wait:
--- a/core/process.py
+++ b/core/process.py
@@ -315,6 +315,9 @@ class FdState(object):
             ok = False
           posix.close(target_fd)  # We already made a copy of it.
 
+        if ok:
+          self._PushClose(r.fd)
+
         # Now handle the extra redirects for aliases &> and &>>.
         #
         # We can rewrite
@@ -333,10 +336,6 @@ class FdState(object):
           elif r.op_id == Id.Redir_AndDGreat:
             if not self._PushDup(r.fd, 2):
               ok = False
-
-        # I don't think we need to close(0) because it will be restored from its
-        # saved position (10), which closes it.
-        #self._PushClose(r.fd)
 
       elif case(redirect_e.FileDesc):  # e.g. echo hi 1>&2
         r = cast(redirect__FileDesc, UP_r)
--- a/core/process.py
+++ b/core/process.py
@@ -127,7 +127,7 @@ class FdState(object):
 
   For example, you can do 'myfunc > out.txt' without forking.
   """
-  def __init__(self, errfmt, job_state, mem = None):
+  def __init__(self, errfmt, job_state, mem=None):
     # type: (ErrorFormatter, JobState, Mem) -> None
     """
     Args:
--- a/core/process.py
+++ b/core/process.py
@@ -20,6 +20,7 @@ from _devbuild.gen.runtime_asdl import (
     job_status, job_status_t,
     redirect_e, redirect_t,
     redirect__Path, redirect__FileDesc, redirect__CloseFd, redirect__MoveFd, redirect__HereDoc,
+    value, value_e, lvalue, scope_e, value__Str,
 )
 from asdl import pretty
 from core import util
@@ -125,8 +126,8 @@ class FdState(object):
 
   For example, you can do 'myfunc > out.txt' without forking.
   """
-  def __init__(self, errfmt, job_state):
-    # type: (ErrorFormatter, JobState) -> None
+  def __init__(self, errfmt, job_state, mem = None):
+    # type: (ErrorFormatter, JobState, Mem) -> None
     """
     Args:
       errfmt: for errors
@@ -136,6 +137,7 @@ class FdState(object):
     self.job_state = job_state
     self.cur_frame = _FdFrame()  # for the top level
     self.stack = [self.cur_frame]
+    self.mem = mem
 
   # TODO: Use fcntl(F_DUPFD) and look at the return value!  I didn't understand
   # the difference.
@@ -181,6 +183,20 @@ class FdState(object):
       raise OSError(*e.args)  # Consistently raise OSError
     return f
 
+  def _WriteFdToMem(self, fd_name, fd):
+    if self.mem:
+      self.mem.SetVar(lvalue.Named(fd_name), value.Str(str(fd)), scope_e.Dynamic)
+  def _ReadFdFromMem(self, fd_name):
+    if self.mem is None: return None
+
+    val = self.mem.GetVar(fd_name)
+    if val.tag == value_e.Str:
+      try:
+        return int(cast(value__Str, val).s)
+      except ValueError:
+        return None
+    return None
+
   def _PushSave(self, fd):
     # type: (int) -> bool
     """Save fd.
@@ -213,8 +229,8 @@ class FdState(object):
 
     return need_restore
 
-  def _PushDup(self, fd1, fd2):
-    # type: (int, int) -> bool
+  def _PushDup(self, fd1, fd2, fd2name = None):
+    # type: (int, int, string) -> bool
     """Save fd2, and dup fd1 onto fd2.
 
     Mutates self.cur_frame.saved.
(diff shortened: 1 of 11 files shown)


frontend/lexer_def.py
```diff
@@ -303,6 +303,17 @@ def IsKeyword(name):
   R(r'[0-9]*<>', Id.Redir_LessGreat),
   R(r'[0-9]*>\|', Id.Redir_Clobber),
 
+  R(r'\{[_a-zA-Z][_a-zA-Z0-9]*\}<', Id.Redir_Less),
```
andychu: I think this is identical to  `VAR_NAME_RE` right?  If so let's use it, maybe with an intermediate like:

```
FD_VAR_NAME = '\{' + VAR_NAME_RE + '\}'
```

andychu:
Second thought: maybe we should just have `string fd` for both `1` and `{fd}` cases, in `syntax.asdl` both files ?

And then in the evaluator can just check `isdigit()`.  If you don't think that will simplify the code, it's not that important.

osh/cmd_exec.py
```diff
@@ -639,7 +639,10 @@ def _EvalRedirect(self, n):
 
         # note: needed for redirect like 'echo foo > x$LINENO'
         self.mem.SetCurrentSpanId(n.op.span_id)
-        fd = consts.RedirDefaultFd(n.op.id) if n.fd == runtime.NO_SPID else n.fd
+        fd = n.fd
+        fd_name = n.fd_name
+        if fd == runtime.NO_SPID and not fd_name:
```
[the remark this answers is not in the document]
akinomyoga: Thank you! I defined `process.NO_FD` 1caf56d9. If you think `NO_FD` should be defined in a different place, please let me know.

frontend/lexer_def.py
akinomyoga: Thank you! I defined `FD_VAR_NAME` cc491ffb.

akinomyoga:
> As for the schema, I think a sum type like `dest = Name(Token name) | Num(int fd)` might be cleaner, but I think that can wait for another commit.

Thank you for all your review! Actually initially I thought about a similar one `int | string` but was not sure if it is a good idea or not.

> Second thought: maybe we should just have string fd for both 1 and {fd} cases, in just syntax.asdl, or in both files?

Thank you for the suggestion! I created commits for this suggestion (74f9d288 for `syntax.asdl` and 0f2b96be for `runtime.asdl`). You can have a look on these commits. If you don't like them, I can drop these commits. Of course, additional suggestions are also welcome.

----
I updated the default fd for `<>` b81d34ee. Also, I applied other cosmetic changes suggested in #671 [58fb6c85 for type annotations, 668d2bdf for `.tag_()`, and 0a096e68 for spaces in default arguments].

osh/cmd_parse.py
```diff
@@ -482,12 +483,18 @@ def ParseRedirect(self):
...
+    fdspec = ''
+    if op_tok.val[0] == '{':
+      index = op_tok.val.find('}')
+      if index < 0:
+        p_die('Invalid token after redirect operator', word=self.cur_word)
+      fdspec = op_tok.val[:index+1]
     else:
-      fd = runtime.NO_SPID
+      index = 0
+      while index < len(op_tok.val) and op_tok.val[index].isdigit():
```
andychu: Important issue: I think we need a limit on user descriptors so they don't collide with descriptors the shell uses?  Mentioned here:

https://www.aosabook.org/en/bash.html

> The other complication is one bash brought on itself. Historical versions of the Bourne shell allowed the user to manipulate only file descriptors 0-9, reserving descriptors 10 and above for the shell's internal use. Bash relaxed this restriction, allowing a user to manipulate any descriptor up to the process's open file limit. This means that bash has to keep track of its own internal file descriptors, including those opened by external libraries and not directly by the shell, and be prepared to move them around on demand. This requires a lot of bookkeeping, some heuristics involving the close-on-exec flag, and yet another list of redirections to be maintained for the duration of a command and then either processed or discarded.

Are you using descriptors over 9?  I haven't seen shell scripts using more, and POSIX only guarantees up to 9 I think.

Maybe we can limit it to  99 instead of 9 ?  But if there's no usage I would keep it at 9 until we encounter it.

core/process.py
```diff
@@ -102,29 +104,31 @@ def RemoveUserTrap(self, sig_num):
...
 
 
 class FdState(object):
   """This is for the current process, as opposed to child processes.
 
   For example, you can do 'myfunc > out.txt' without forking.
   """
-  def __init__(self, errfmt, job_state):
-    # type: (ErrorFormatter, JobState) -> None
+  def __init__(self, errfmt, job_state, mem = None):
```
andychu: mem=None here

andychu:
One thing that is annoying is that the use of unrestricted descriptors means you can't link arbitrary libraries into the shell that do I/O!  Like simply `fopen()` by a library could break shell scripts.  I found that out the hard way ...   I believe GNU readline has to take a lot of care about this.

For example, I thought it would be cool to link in `libevent` / `libev`  / `libuv` for some kind of event-driven programming / coroutines / state machines in shell.  But I think you can't do that if they use low-numbered file descriptors themselves, and `open()` defaults to the lowest FD if I recall correctly.

andychu:
Thinking about this more: I would say if ble.sh does **not** use descriptors over 9, let's just keep it at a single digit, and add a TODO to revisit that code later.  The other changes are more important.

If it does use descriptors over 9, we can discuss what the right restriction is.

akinomyoga:
> I think we need a limit on user descriptors so they don't collide with descriptors the shell uses?
>
> > The other complication is one bash brought on itself. Historical versions of the Bourne shell allowed the user to manipulate only file descriptors 0-9, reserving descriptors 10 and above for the shell's internal use. Bash relaxed this restriction, allowing a user to manipulate any descriptor up to the process's open file limit. This means that bash has to keep track of its own internal file descriptors, including those opened by external libraries and not directly by the shell, and be prepared to move them around on demand. This requires a lot of bookkeeping, some heuristics involving the close-on-exec flag, and yet another list of redirections to be maintained for the duration of a command and then either processed or discarded.

You don't have to worry about the collisions because Oil has already implemented above mentioned "*a lot of bookkeeping, some heuristics involving the close-on-exec flag, and yet another list of redirections to be maintained for the duration of a command and then either processed or discarded*" completely, and it should work without problems now. The above explanation is written in a way that it is a highly non-trivial matter, but that's an exaggeration. If you implemented it carefully, there is no problem.

> Are you using descriptors over 9?

This is a good question. Actually `ble.sh` **does** use file descriptors starting from 30 by default for its internal usage **in Bash 4.0 and below**. From Bash 4.1, `ble.sh` uses the construct `{fd}>` to let Bash determine the new file descriptors. The reason why it doesn't use 0-9 is that `ble.sh` want to keep this range (0-9) of file descriptors for users. The reason why the default value is starting from 30 is that I want to have some margins from the file descriptors 10 which Bash itself uses for its internal usage. In addition, the starting file descriptor of `ble.sh` internal usage can be configured by the shell variable [`bleopt_openat_base`](https://github.com/akinomyoga/ble.sh/wiki/Manual-%C2%A78-Miscellaneous#837-bleopt-openat_base-integer) in case it conflicts with another shell framework.

I think shell libraries/frameworks inevitably use larger values of file descriptors (as far as they don't require the redirection of the form `{fd}>`) because they need to leave 0-9 for applications. If you restrict the file descriptor range, the problem becomes more severe because an application and shell frameworks have to work with only 7 (excluding 1 and 2) file descriptors. Oil already manages file descriptors properly, so I don't see a reason to restrict the range to 0-9, 0-99 or whatever.

> For example, I thought it would be cool to link in libevent / libev / libuv for some kind of event-driven programming / coroutines / state machines in shell. 

Ah, OK... It's non-trivial. But can we control the range of the file descriptors used in those libraries? If we can't control it, the problem wouldn't be changed even if we restrict the available range to 0-9. For safe programming, I think one should always use the form `{fd}>`, etc. (although it is only available from Bash 4.1), and should not use the form `INT>`, etc.

core/process.py
akinomyoga: Thank you! I'm sorry, this was an oversight. Fixed 704ba1ec.

andychu:
OK yeah I guess you're right, as long as the program always uses `fd_state.Open()` then everything works, and third party libs are going to have a problem regardless.

I had to learn things about colliding file descriptors the hard way as described here [1], even though I think I had already read that bit about bash.

I agree that `{fd}` is the preferred construct for descriptors other than 0, 1, and 2.

I think it would be nice to have a better syntax for this, e.g. basically expose `fd_state.Open()` in Oil.  I will file a bug about that.

[1] http://www.oilshell.org/blog/2017/07/02.html

andychu:
FWIW this test passed on my machine, but on Travis bash did NOT fail, so the test failed.

https://github.com/oilshell/oil/commit/b6274796721e1fe7620c54513fce22c8ed673ac5

I guess this is the same problem where file descriptors are hidden global state that is not specified when opening a program.  In Oil I would deprecate it for #673 and that's another reason to avoid direct FD manipulation in shell (again http://www.oilshell.org/blog/2017/08/12.html).

(Although obviously in this case we need it for the test)

andychu:
Only other problem: this JSON test case is now failing, and I'm able to reproduce it on my machine:

http://travis-ci.oilshell.org/jobs/2020-03-21__19-53-14.wwz/_tmp/spec/oil-json.html

```
Error closing descriptor 0: [Errno 9] Bad file descriptor
osh I/O error: Bad file descriptor
```

Looking into it ...

akinomyoga:
> Oh it has to be of this, not sure exactly why but I'll look at it later...

I guess `fclose(stdin)` [ which performs `fflush(stdin)` and `close(fileno(stdin))` ] is missing. So the data buffered in `stdin` will be output after the file descriptor is closed by the shell redirection.

andychu:
OK that was a quicker fix than I expected.  The issue was that the local file returned by `posix.fdopen()` is closed when it goes out of scope due to CPython reference counting, but the redirect should be responsible for closing it.

This kind of thing shouldn't be an issue when it's translated to C++, and is another good reason to translate it.

BTW I'm not sure if `core/process.py` should be translated or rewritten.  It uses a lot of I/O which is hard to translate.  For example I extracted `osh/bool_stat.py`  which also does a lot of I/O for `test -f`, etc.   That one is very easy to rewrite in C++ so there's a stub in `cpp/osh_bool_stat.h`.  We'll have see about core/process.

PR #686, opened 2020-03-31
[test/spec] Failing spec tests for recursive arithmetic evaluations

akinomyoga:
`spec/ble-idioms.test.sh` mentioned in https://github.com/oilshell/oil/issues/653#issuecomment-605771723.

### DIFF
--- a/spec/ble-idioms.test.sh
+++ b/spec/ble-idioms.test.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+#### recursive arith: one level
+a='b=123'
+echo $((a))
+## stdout: 123
+## N-I dash status: 2
+## N-I dash stdout-json: ""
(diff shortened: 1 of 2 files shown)


andychu:
Great thanks!  It's interesting to see which shells support it, and as usual you find a bug or two :)

Since dash and yash don't support any of it, you can just delete the assertions and delete them from the arg list `test/spec.sh`.  One reason for the different spec files is so each file can run on different shells.   dash doesn't support most features that bash, mksh, etc. do.

akinomyoga:
995e256 I removed `dash` and `yash` from `test/spec.sh ble-idioms`. I left the assertions of these shells because `ble-idioms` can possibly contain other tests unrelated to recursive arithmetic evaluations in future and `dash` and `yash` may be revisited then.

PR #703, opened 2020-04-08
[test/spec] Spec tests for nocasematch

Crestwave:
I wasn't entirely sure what to put, but I figured it would be better to be too thorough than not thorough enough.

### DIFF
--- a/test/spec.sh
+++ b/test/spec.sh
@@ -710,7 +710,7 @@ extglob-match() {
 }
 
 nocasematch-match() {
-  sh-spec spec/nocasematch-match.test.sh \
+  sh-spec spec/nocasematch-match.test.sh --osh-failures-allowed 3 \
     $BASH $OSH_LIST "$@"
 }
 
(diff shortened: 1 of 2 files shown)


test/spec.sh
```diff
@@ -709,6 +709,11 @@ extglob-match() {
     $BASH $MKSH $OSH_LIST "$@"
 }
 
+nocasematch-match() {
+  sh-spec spec/nocasematch-match.test.sh \
```
andychu, line 713: Please `add --osh-failures-allowed 6` here.  That just indicates "we know about these failures, the whole test suite shouldn't fail"

Crestwave: Just to clarify, it only affects `case` and `[[` and the rest are to verify that it does not affect it.  As such, OSH currently passes 3 of the tests, so should it be `--osh-feailures-allowed 3`?

andychu: Ah yes, it should be 3.  You can leave the other tests since the look informative.

andychu:
Great thanks!

Issue #653, opened 2020-03-10
Try to parse and run ble.sh


andychu:
Why?

- it's a very substantial program using bash as a programming language, but it's *not* a batch program
- it's probably one of the hardest scripts to run (probably won't get it to run, but worth trying)
- it could be faster.  maybe that will help optimize Oil.  
  - I want Oil  to be fast at char-by-char processing

https://github.com/akinomyoga/ble.sh

andychu:
#620 was the original issue, already addressed `return 2>&1` etc.

andychu:
Two issues blocking parsing:

- several cases of #640 
- the `\$file` one

```
~/git/wild/ble.sh$ for name in */*.sh; do echo $name; ~/git/oilshell/oil/bin/osh -n $name | wc -l; done

...

memo/D1259.readbind.sh
84
out/ble.sh
            ((_ble_hook_c_$name=0))
                               ^
out/ble.sh:1644: Left-hand side of this assignment is invalid
0
src/benchmark.sh
2342
src/canvas.sh
    ((${_prefix}x=_pos[0]))
                 ^
src/canvas.sh:1521: Left-hand side of this assignment is invalid
0
src/color.sh
    (($_var=_ret))
           ^
src/color.sh:670: Left-hand side of this assignment is invalid
0
src/decode.sh
22515
src/def.sh
133
src/edit.sh
45409
src/history.sh
        ((_ble_builtin_history_rskip_dict[\$file]+=$2))
                                          ^
src/history.sh:701: Unexpected token in arithmetic context
0
src/util.sh
            ((_ble_hook_c_$name=0))
                               ^
src/util.sh:917: Left-hand side of this assignment is invalid
0
```

andychu:
Note: can parse `lib/core-syntax.sh` (6651 lines, one of the biggest) by commenting out 5 instances of dynamic LHS in arith

andychu:
* `src/util.sh` needs a lot of work.  This pattern is used all over the place.  
  * there is at least one instance of it in   our WILD corpus too
  * might want to take another look at #640 , or just allow parsing as a half-measure, and do
    evaluation later

andychu:
@akinomygoa

On #640 I added a temporary hack `shopt -s parse_unimplemented` that parses dynamic arithmetic but doesn't implement it.  So at least this idiom won't prevent you from using `osh -n`.  Now I can do:

```
[...]

PR #723, opened 2020-04-22
[core/process] Fix redirections

akinomyoga:
- 77e3af8 Add failing test cases which are fixed by the subsequent fixes.
- abe89a5 Fix *"#653 33. BUG: Redirection of 2 is persistent after `: 2>/dev/null >&30`"*
- 4ed53e1 Fix *"#653 37. BUG: `{fd}>&-` does not close the fd"*
- 7f86a1a Fix a problem that `>&100` does not fail even if fd 100 does not exist.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -441,6 +441,7 @@ class FdState(object):
       try:
         self._ApplyRedirect(r, waiter)
       except (IOError, OSError) as e:
+        self.Pop()
         return False  # for bad descriptor, etc.
       finally:
         self.errfmt.PopLocation()
--- a/core/process.py
+++ b/core/process.py
@@ -302,10 +302,11 @@ class FdState(object):
 
     elif loc.tag_() == redir_loc_e.Fd:
       fd = cast(redir_loc__Fd, UP_loc).fd
-      self._PushSave(fd)
 
     else:
       raise AssertionError()
+
+    self._PushSave(fd)
 
     return True
 
--- a/core/process.py
+++ b/core/process.py
@@ -257,6 +257,13 @@ class FdState(object):
         # already returned 3, e.g. echo 3>out.txt
         return NO_FD
 
+      # Check the validity of fd1 before _PushSave(fd2)
+      try:
+        fcntl.fcntl(fd1, fcntl.F_GETFD)
+      except IOError as e:
+        self.errfmt.Print('%d: %s', fd1, posix.strerror(e.errno))
+        raise
+
       need_restore = self._PushSave(fd2)
 
       #log('==== dup2 %s %s\n' % (fd1, fd2))
--- a/spec/redirect.test.sh
+++ b/spec/redirect.test.sh
@@ -551,3 +551,38 @@ hi
 hi
 hi
 ## END
+
+#### : >/dev/null 2> / (OSH regression: fail to pop fd frame)
+# oil 0.8.pre4 fails to restore fds after redirection failure In the
+# following case, the fd frame remains after the redirection failure
+# "2> /" so that the effect of redirection ">/dev/null" remains after
+# the completion of the command.
+: >/dev/null 2> /
+echo hello
+## stdout: hello
+## OK dash stdout-json: ""
+## OK dash status: 2
+## OK mksh stdout-json: ""
+## OK mksh status: 1
+# dash/mksh terminates the execution of script on the redirection.
+
+#### echo foo >&100 (OSH regression: does not fail with invalid fd 100)
+# oil 0.8.pre4 does not fail with non-existent fd 100.
+fd=100
+echo foo >&$fd
+## stdout-json: ""
+## status: 1
+## OK dash status: 2
+
+#### exec {fd}>&- (OSH regression: fails to close fd)
+# mksh, dash do not implement {fd} redirections.
+case $SH in (mksh|dash) exit 1 ;; esac
+# oil 0.8.pre4 fails to close fd by {fd}&-.
+exec {fd}>file1
+echo foo >&$fd
+exec {fd}>&-
+echo bar >&$fd
+cat file1
+## stdout: foo
+## N-I mksh/dash stdout-json: ""
+## N-I mksh/dash status: 1
--- a/spec/redirect.test.sh
+++ b/spec/redirect.test.sh
@@ -553,7 +553,7 @@ hi
 ## END
 
 #### : >/dev/null 2> / (OSH regression: fail to pop fd frame)
-# oil 0.8.pre4 fails to restore fds after redirection failure In the
+# oil 0.8.pre4 fails to restore fds after redirection failure. In the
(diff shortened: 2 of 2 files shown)


andychu:
Thanks!  It's not that surprising to me that there are still bugs left, since even after restructuring I don't grok every part of this code... I remember I mainly wrote it with strace :-/  The FD manipulations are still confusing to me.

I'm strace-ing bash and zsh, and they don't seem to do this check?

bash:

```
fcntl(1, F_GETFD)                       = 0
fcntl(1, F_DUPFD, 101)                  = 101
fcntl(1, F_GETFD)                       = 0
fcntl(101, F_SETFD, FD_CLOEXEC)         = 0
dup2(100, 1)                            = -1 EBADF (Bad file descriptor)
brk(0x1ce0000)                          = 0x1ce0000
```

zsh:

```
fcntl(3, F_DUPFD, 10)                   = 11
fcntl(11, F_GETFL)                      = 0x8000 (flags O_RDONLY|O_LARGEFILE)
rt_sigprocmask(SIG_BLOCK, [CHLD], [WINCH], 8) = 0
dup(100)                                = -1 EBADF (Bad file descriptor)
```

mksh has 2 different errors:

```
$ mksh -c 'echo foo >& 24'
mksh: >&24 : file descriptor too large
andy@lisa:~/git/oilshell/oil2$ mksh -c 'echo foo >& 23'
mksh: >&23 : bad file descriptor
``` 


OK I guess that's why it already fails for descriptor 99, but not 100... ?

----

But I guess the test cases are good so I should merge this and figure out the code again later ...

andychu:
Here is why Oil does now...

So I think there could be a pure check that `fd1` is not `_SHELL_MIN_FD` or greater?    Since 99 fails but not 100?

I'm not entirely sure about it, but I think that is what mksh does.

```
fcntl(1, F_DUPFD, 100)                  = 100
close(1)                                = 0
fcntl(100, F_SETFD, FD_CLOEXEC)         = 0
dup2(100, 1)                            = 1
fstat(1, {st_mode=S_IFCHR|0620, st_rdev=makedev(136, 45), ...}) = 0
write(1, "foo\n", 4foo
)                    = 4
dup2(100, 1)                            = 1
close(100)                              = 0
```

akinomyoga:
Thank you. I think I should have explained what happens here. The problem of not-failing `echo hello N>&M` occurs when `M` matches with the smallest unused number  (`>= _SHELL_MIN_FD`). In the current Oil,

1. First M is saved to an unused fd which is N
2. Then N is copied to M

Even if `N` is invalid before the redirection, it becomes valid by the step 1. But this new `N` is just a temporary backup of `M` so is not the fd that the user wanted to copy.

I checked `strace` of bash, zsh and mksh, they actually work around this problem with different approaches. None has this problem that the current Oil has.

### Bash

```
# strace for "echo hello >&12" (valid fd)

fcntl(1, F_GETFD)                       = 0
fcntl(1, F_DUPFD, 13)                   = 13
fcntl(1, F_GETFD)                       = 0
fcntl(13, F_SETFD, FD_CLOEXEC)          = 0
dup2(12, 1)                             = 1
fcntl(12, F_GETFD)                      = 0
write(1, "hello\n", 6)                  = 6
dup2(13, 1)                             = 1
fcntl(13, F_GETFD)                      = 0x1 (flags FD_CLOEXEC)
close(13)                               = 0

# strace for "echo hello >&13" (invalid fd)

fcntl(1, F_GETFD)                       = 0
fcntl(1, F_DUPFD, 14)                   = 14
fcntl(1, F_GETFD)                       = 0
fcntl(14, F_SETFD, FD_CLOEXEC)          = 0
dup2(13, 1)                             = -1 EBADF (Bad file descriptor)
dup2(14, 1)                             = 1
fcntl(14, F_GETFD)                      = 0x1 (flags FD_CLOEXEC)
close(14)                               = 0
```

Bash behaves like this:

1. First M is saved to the smallest unused fd that is `>= max(N+1, 10)`.
2. Then N is copied to M

Because Bash explicitly specifies X := max(N+1,10) (which is ensured to be larger than N) as `fcntl(fd1,F_DUPFD,X)`, there will be no collision of saved fd and N. This is a clever way but I don't know what happens when N matches with the largest available fd value.

### Zsh

```
# strace for "echo hello >&12" (valid fd)

dup(12)                                 = 3
fcntl(3, F_DUPFD, 10)                   = 15
fcntl(1, F_DUPFD, 10)                   = 16
close(1)                                = 0
dup2(15, 1)                             = 1
close(15)                               = 0
write(1, "hello\n", 6)                  = 6
dup2(16, 1)                             = 1
close(16)                               = 0

# strace for "echo hello >&15" (invalid fd)

dup(15)                                 = -1 EBADF (Bad file descriptor)
```

Zsh behaves like this:

1. First N is copied to an unused fd (X).
2. Then M is saved to another fd.
3. Finally X is copied to M

When N is invalid, zsh can detect it on the first step, so zsh can produce the error message. But this approach involves additional fd moves.

### Mksh

```
# strace for "echo hello >&12" (valid fd)

fcntl(12, F_GETFL)                      = 0x8402 (flags O_RDWR|O_APPEND|O_LARGEFILE)
fcntl(1, F_DUPFD, 24)                   = 25
fcntl(25, F_SETFD, FD_CLOEXEC)          = 0
dup2(12, 1)                             = 1
write(1, "hello\n", 6)                  = 6
dup2(25, 1)                             = 1
close(25)                               = 0

# strace for "echo hello >&13" (invalid fd)

fcntl(13, F_GETFL)                      = -1 EBADF (Bad file descriptor)
```

Mksh uses the same approach with this PR. It first checks if N is valid or not. Only when N is valid it tries to save M and copy N to M.

1. Check validify of N
2. Save M to an unused fd
3. Copy N to M

In addition, mksh prohibits the access to fds greater or equil to `_SHELL_MIN_FD` (24).

----

> So I think there could be a pure check that fd1 is not _SHELL_MIN_FD or greater? 

If you restrict the value of `fd1`, we don't have a way to access fds allocated by the redirection of the form `{fd}>`. `mksh` could restrict the value of `fd` because it doesn't have `{fd}>`.

```bash
$ exec {fd}> a.txt # <-- $fd becomes _SHELL_MIN_FD or larger.
$ echo hello >&$fd # <-- we need to access $fd (>= _SHELL_MIN_FD)
$ exec {fd}>&-
```

akinomyoga:
I added a test case 48474aa.

andychu:
OK thank you for the detailed answers!  I merged it.

I feel like the check could be optimized away in some cases but since I don't fully understand it yet, it's better to move forward.  The tests will help with future refactoring!

akinomyoga:
Thank you!

> I feel like the check could be optimized away

Actually, it will work without the check when `if fd1 >= _SHELL_MIN_FD`. But, with the check, we can eliminate redundant save/restore of fd2 in the case that `fd1` is invalid.

PR #724, opened 2020-04-22
[arith] Fix reversed digits in $((base#num)).

akinomyoga:
This fixes the problem that `echo $((10#0123))` outputs `3210`.

```bash
$ osh -c 'echo $((10#0123))'
3210
```


andychu:
Thanks!

PR #725, opened 2020-04-22
[arith] Support empty arith.

akinomyoga:
This fixes the following problem.

- *"#653 29. BUG: Cannot parse `${arr[@]::}`"*

Actually this was a N-I feature rather than a bug. It is related to an empty string in arithmetic context. I'm not sure if Oil should support this by default or not, but this is a patch to support empty strings in arithmetic context.

In this implementation, I derived a class `ArithParser` from `TdopParser` to add a special rule for empty arithmetic expressions.

### DIFF
--- a/frontend/option_def.py
+++ b/frontend/option_def.py
@@ -231,6 +231,7 @@ def _Init(opt_def):
 
   opt_def.Add('eval_unsafe_arith')  # recursive parsing and evaluation (ble.sh)
   opt_def.Add('parse_dynamic_arith')  # dynamic LHS
+  opt_def.Add('parse_empty_arith')  # empty arithmetic expressions
   opt_def.Add('compat_array')  # ${array} is ${array[0]}
 
   # Two strict options that from bash's shopt
--- a/frontend/parse_lib.py
+++ b/frontend/parse_lib.py
@@ -285,7 +285,7 @@ class ParseContext(object):
     lx = self._MakeLexer(line_reader)
     w_parser = word_parse.WordParser(self, lx, line_reader)
     w_parser.Init(lex_mode_e.Arith)  # Special initialization
-    a_parser = tdop.TdopParser(arith_parse.Spec(), w_parser, self.parse_opts)
+    a_parser = arith_parse.ArithParser(w_parser, self.parse_opts)
     return a_parser
 
   def MakeParserForCommandSub(self, line_reader, lexer, eof_id):
--- a/frontend/syntax.asdl
+++ b/frontend/syntax.asdl
@@ -211,7 +211,8 @@ module syntax
   attributes (int* spids)
 
   arith_expr =
-    VarRef %Token  -- variable without $
+    Empty
+  | VarRef %Token  -- variable without $
   | Word %compound_word  -- a string that looks like an integer
 
   | UnaryAssign(id op_id, arith_expr child)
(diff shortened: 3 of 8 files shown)


andychu:
Hm does any program need `$(( ))` and `(( ))`?  If not I'd rather keep it stricter.  dash and yash don't support it.  Although it's true they're in the minority, I can't think of any reason you would want `$(( ))` instead of say `$(( 0 ))`.

I'm also confused about  `${arr[@]::}`.  Doesn't that just give you an empty array always?  What use is it?

I suppose it's equivalent to `${arr[@]:0:0}`, and that is easier to read.

andychu:
I think this falls under the common subset principle: https://github.com/oilshell/oil/wiki/Language-Design-Principles

If there is a trivial rewrite (that may improve your program), it may be OK to disagree with other shells.  Generally speaking I think existing shells accept too much code without failing.  (That's why there are many `shopt -s strict_*` options in Oil.)

akinomyoga:
I don't think people will directly write something like `$(())`, but I assumed cases such as

```bash
let() { eval "(($*))"; }
## @fn process [param]
process() { let "$1"; }
process
```

Or,

```bash
## @fn func [extra_count]
func() {
  local -a extra_words=(1 2 3 4)
  eval "joined=\"abc\${extra_words[*]::$1}\""
}
func
```

I encountered a similar (but not exactly the same) case while testing `ble.sh`. Actually the exact `${arr[@]::}` case will not appear in `ble.sh`, so there is no problem with `ble.sh`.  Nevertheless, it is still possible some Bash scripts fall into this pattern. They can be easily patched to make it work without *empty arithmetic expressions* if we find the problem, but these patterns cannot be statically detected so we need to perform enough tests for scripts.

Another thing is that it may be more consistent with the treatment of empty variables in arithmetic expressions. The empty variables will be evaluated as 0 in arithmetic expressions.

```bash
$ osh -c 'var=; echo $((var))'
0
$ osh -c 'var=; echo $(($var))'
0
$ osh -c 'shopt -s eval_unsafe_arith; var=; echo $(($var))'
0
```

But I can understand this is too permissive in the view point of the language design. That is the reason that I wrote the following sentence in the first post.

> I'm not sure if Oil should support this by default or not

Maybe we can support it under `shopt -s eval_unsafe_arith` or another option?

akinomyoga:
I added `shopt -s parse_empty_arith` as an experiment. I can drop or rename this option anytime.

andychu:
Hm I'd rather wait until we have evidence that people need it.  Would you turn on this option in ble.sh, and where? 

(BTW in general I'm interested in what the next thing is to make some tests pass in ble.sh.  I guess you hit the FUNCNAME bug while testing?  And I still want to add basic coverage support so we can see how many lines are hit when tests pass.)

-----

I think the rewrites to avoid empty arithmetic are always improvements.  Keep in mind that you wrote one of the biggest shell programs in the world, so your knowledge of bash far exceeds the average user.  Oil is supposed to be for the average user, not the most sophisticated user! :-)

I'm also trying to keep the code small, which is why I [count lines on every release](http://www.oilshell.org/blog/2020/04/release-0.8.pre4.html#appendix-b-metrics-for-the-08pre4-release).

And `super()` doesn't translate to C++, etc.  (If it were a a very valuable feature that wouldn't be  an issue; I would make it translate myself.)

We're trying to avoid more global  options too:

https://github.com/oilshell/oil/wiki/Language-Design-Principles

akinomyoga:
OK! Thank you!

PR #728, opened 2020-04-23
[word_eval] Implement 'shopt -s compat_array'.

akinomyoga:
Ref #726 

- Fix a problem that the scalar form cannot be used in arithmetic expressions (9fc20e7f Test c2365f5f Fix):

```bash
$ osh -c '$((BASH_LINENO))'
  $((BASH_LINENO))
     ^~~~~~~~~~~
[ -c flag ]:1: fatal: Expected a value convertible to integer, got value.MaybeStrArray
```

- Implement `shopt -s compat_array` to allow `$array` (Test bf9f3328, Fix 67047709)

- Disallow `${#BASH_SOURCE}`, `${BASH_SOURCE:1:1}`, `${BASH_SOURCE#X}`, `${BASH_SOURCE-X}`, etc. by default (Fix b1a16c25). In this implementation, everything with additional suffix/prefix are disallowed.

### DIFF
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -331,7 +331,7 @@ class ArithEvaluator(object):
     # BASH_LINENO, etc.
     if val.tag_() in (value_e.MaybeStrArray, value_e.AssocArray) and lval.tag_() == lvalue_e.Named:
       named_lval = cast(lvalue__Named, lval)
-      if word_eval.CheckCompatArray(named_lval.name):
+      if word_eval.CheckCompatArray(named_lval.name, self.exec_opts):
         val = word_eval.ResolveCompatArray(val)
 
     # This error message could be better, but we already have one
@@ -358,7 +358,7 @@ class ArithEvaluator(object):
     # BASH_LINENO, etc.
     if val.tag_() in (value_e.MaybeStrArray, value_e.AssocArray) and node.tag_() == arith_expr_e.VarRef:
       tok = cast(Token, node)
-      if word_eval.CheckCompatArray(tok.val):
+      if word_eval.CheckCompatArray(tok.val, self.exec_opts):
         val = word_eval.ResolveCompatArray(val)
 
     # TODO: Can we avoid the runtime cost of adding location info?
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -41,6 +41,7 @@ from mycpp import mylib
 from mycpp.mylib import tagswitch, switch
 from osh import bool_stat
 from osh import word_
+from osh import word_eval
 
 import libc  # for fnmatch
 
@@ -327,6 +328,12 @@ class ArithEvaluator(object):
     lval = self.EvalArithLhs(node, runtime.NO_SPID)
     val = OldValue(lval, self.mem, self.exec_opts)
 
+    # BASH_LINENO, etc.
+    if val.tag_() in (value_e.MaybeStrArray, value_e.AssocArray) and lval.tag_() == lvalue_e.Named:
+      named_lval = cast(lvalue__Named, lval)
+      if word_eval.CheckCompatArray(named_lval.name):
+        val = word_eval.ResolveCompatArray(val)
+
     # This error message could be better, but we already have one
     #if val.tag_() == value_e.MaybeStrArray:
     #  e_die("Can't use assignment like ++ or += on arrays")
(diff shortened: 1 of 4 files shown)


osh/sh_expr_eval.py
```diff
@@ -327,6 +328,12 @@ def _EvalLhsAndLookupArith(self, node):
     lval = self.EvalArithLhs(node, runtime.NO_SPID)
     val = OldValue(lval, self.mem, self.exec_opts)
 
+    # BASH_LINENO, etc.
+    if val.tag_() in (value_e.MaybeStrArray, value_e.AssocArray) and lval.tag_() == lvalue_e.Named:
+      named_lval = cast(lvalue__Named, lval)
+      if word_eval.CheckCompatArray(named_lval.name, self.exec_opts):
+        val = word_eval.ResolveCompatArray(val)
```
andychu, line 335: Would this part be made obsolete if we made BASH_LINENO etc. immutable?

I guess that would make Oil technically incompatible, but I've been thinking about that.  Also it makes sense to tighten up things like `PATH=(a b c)` and `HOME=(a b c)` (they shouldn't be arrays)

This doesn't have to be done in this commit; I'm just curious / brainstorming the semantics

osh/word_eval.py
```diff
@@ -61,6 +61,26 @@
 # For compatibility, ${BASH_SOURCE} and ${BASH_SOURCE[@]} are both valid.
 # ${FUNCNAME} and ${BASH_LINENO} are also the same type of of special variables.
 _STRING_AND_ARRAY = ['BASH_SOURCE', 'FUNCNAME', 'BASH_LINENO']
+def CheckCompatArray(var_name, opts, compat_introspect=True):
```
andychu, line 64: I would name `compat_introspect` something more descriptive like `is_plain_var` or `is_plain_var_sub`

( The difference between SimpleVarSub and BracedVarSub is one thing I want to get rid of with #604 )

andychu:
Ah never mind about my lvalue comment.  I think this works now?

```
$ a=(1 0 0); (( a++ )); echo ${a[@]}
2 0 0
```

If so then I think we should add things like that as test cases.  I don't think the tests are hitting some code paths.

osh/sh_expr_eval.py
akinomyoga: As you have already written in another comment, this change also affects the other mutable variables when `shopt -s compat_array` is set.

akinomyoga:
> I think this works now?
>
> ```bash
> $ a=(1 0 0); (( a++ )); echo ${a[@]}
> ```

I added a test case to find that it actually doesn't work. Currently, it mutates the array to a scalar. I'll fix it later.

osh/word_eval.py
akinomyoga: I renamed it to `is_plain_var_sub`.

spec/ble-features.test.sh
```diff
@@ -576,3 +576,37 @@ v=tempenv1 f4_unset global,tempenv1
 [global,tempenv1,tempenv2,tempenv3] v: (unset) (unset 3)
 [global,tempenv1,tempenv2,tempenv3] v: (unset) (unset 4)
 ## END
+
+#### [compat_array] ${arr} is ${arr[0]}
+case ${SH##*/} in (dash|ash) exit 1;; esac # dash/ash does not have arrays
+case ${SH##*/} in (osh) shopt -s compat_array;; esac
+case ${SH##*/} in (zsh) setopt KSH_ARRAYS;; esac
+arr=(foo bar baz)
+echo "$arr"
```
andychu, line 585: I think this should also test `${arr}` since that's a different code path.  (as mentioned I want to collapse those)

andychu:
Oh I also had that other comment about testing `${arr}` too, maybe you can sneak it in the next commit if you're touching this...

In any case I'm interested to see what works in ble.sh is, so if there is something I can try let me know.  Should I just switch to the `osh` branch and try to run it with `osh` from head?

We can put instructions here too:

https://github.com/oilshell/oil/wiki/Running-ble.sh-With-Oil

akinomyoga:
Oh, you have already merged it. Thank you!

> OK I think the continuous build might fail but let's see what happens ...

I have actually fixed it now. I'll submit it in another PR.

> In any case I'm interested to see what works in ble.sh is, so if there is something I can try let me know.

Yes, now it is ready so I'm going to prepare something that can be tried out.

> We can put instructions here too:

OK.

spec/ble-features.test.sh
akinomyoga: I added the test for `${arr}` in another PR #733.

akinomyoga:
> We can put instructions here too:

I added a description in the Wiki page. You can follow the instructions [Update ble.osh](https://github.com/oilshell/oil/wiki/Running-ble.sh-With-Oil#update-bleosh) and [Try `read -e`](https://github.com/oilshell/oil/wiki/Running-ble.sh-With-Oil#try-read--e-by-blesh)

PR #729, opened 2020-04-23
[builtin/read] Fix 'IFS= read -n' bug.

akinomyoga:
This fixes the problem that when `read -n` is used with tempenv, the resulting variable is created in the tempenv frame so that one cannot access the results:

```bash
$ bash -c 'v= read -n 1 <<< X; echo "[$REPLY]"'
[X]
$ osh -c 'v= read -n 1 <<< X; echo "[$REPLY]"'
[]
```


andychu:
Thanks!

PR #730, opened 2020-04-23
[builtin/printf] Support printf %d \'.

akinomyoga:
`ble.sh` needs this to read <kbd>NUL</kbd> (which is the terminal sequence of <Kbd>C-@</kbd>) for emulated `read`.


andychu:
Thanks!

PR #731, opened 2020-04-23
[builtin/assign] Do not remove existing arrays by 'declare -a array'.

akinomyoga:
When `declare -a array` is performed on existing arrays, the array becomes empty. In this PR, the arrays are initialized only when the variable contains another type of value. This also fixes the same problem for associative arrays.

```bash
# Normal variables are not removed

$ bash -c 'var=hello; declare -r var; declare -p var'
declare -r var="hello"
$ osh -c 'var=hello; declare -r var; declare -p var'
declare -r var=hello

# Arrays and associative arrays are removed in Oil

$ bash -c 'arr=(1 2 3); declare -a arr; declare -p arr'
declare -a arr=([0]="1" [1]="2" [2]="3")
$ osh -c 'arr=(1 2 3); declare -a arr; declare -p arr'
declare -a arr=()
```

### DIFF
--- a/osh/builtin_assign.py
+++ b/osh/builtin_assign.py
@@ -377,15 +377,15 @@ class NewVar(object):
       flags |= state.ClearNameref
 
     for pair in cmd_val.pairs:
-      if pair.rval is None:
+      rval = pair.rval
+      if rval is None and (arg.a or arg.A):
+        old_val = self.mem.GetVar(pair.var_name)
         if arg.a:
-          rval = value.MaybeStrArray([])  # type: value_t
+          if old_val.tag_() != value_e.MaybeStrArray:
+            rval = value.MaybeStrArray([])
         elif arg.A:
-          rval = value.AssocArray({})
-        else:
-          rval = None
-      else:
-        rval = pair.rval
+          if old_val.tag_() != value_e.AssocArray:
+            rval = value.AssocArray({})
 
       rval = _ReconcileTypes(rval, arg, pair.spid)
       self.mem.SetVar(lvalue.Named(pair.var_name), rval, lookup_mode, flags=flags)
--- a/spec/assign.test.sh
+++ b/spec/assign.test.sh
@@ -643,3 +643,30 @@ declare -- foo=bar
 ## N-I dash stderr-json: ""
 ## stdout-json: ""
 
+#### declare -a arr does not remove existing arrays (OSH regression)
+case $SH in (dash) exit 99;; esac # dash does not support arrays
+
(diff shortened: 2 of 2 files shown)


akinomyoga:
With the patches #728..#731, now **the line editor works** (yet many parts of ble.sh including syntax highlighting, command execution, command history, visual bells, vim mode, completions, etc. are not working).

andychu:
Hm I'm curious where does ble.sh need to say `declare -a arr` on a variable that already exists?  Is this because you're changing the type of existing variables?

I tend to use "declare/local" to initialize once, and then mutate it later with `a=1` (no keyword), I guess like C and JavaScript.

Same with `local -a arr; a=other; local -a arr`, although I guess this change fixes that too.

-----

I feel like this falls mostly under:

https://www.oilshell.org/release/0.8.pre4/doc/known-differences.html#values-are-tagged-with-types-not-cells

akinomyoga:
> Hm I'm curious where does ble.sh need to say declare -a arr on a variable that already exists?

Actually, a problem occurred with associative arrays in `ble.sh` but not with arrays. When one wants to add a value to associative arrays which may or may not exist already, one needs to initialize an associative array in the case it doesn't exist already.

```bash
function user-setting {
  local dict_name=${1}_dict
  declare -gA "$dict_name"
  eval "$dict_name[\$2]=\$3"
}
user-setting hello aaa bbb
```

Of course, maybe one can rewrite it in the form like `is-assoc-array "$dict_name" || declare -gA "$dict_name"`. But all the other shells supporting arrays do the equivalent thing with simply `declare -a arr` or `declare -A dict`.

>  Is this because you're changing the type of existing variables?

No, this PR does not affect when the type of existing variable changes. This PR only changes the behavior *when the type is unchanged*, *i.e.*, when `declare -a arr` is performed on existing *arrays* and when `declare -A dict` is performed on existing *associative arrays*. The current Oil removes the existing arrays with `declare -a arr` which is inconsistent with the behavior of `declare var` which does not remove existing scalar variables or array variables.

Note that in Bash, the conversion from scalars to arrays or associative arrays are allowed, but the opposite direction and the conversion between associative arrays and arrays are not allowed. In the allowed cases, the value is always preserved in some way. When a scalar is converted to an array (associative array), the original value becomes the first element (the value for the key `'0'`).

> Same with local -a arr; a=other; local -a arr, although I guess this change fixes that too.

The above case will not be affected by the PR because it changes the type.

akinomyoga:
Thank you!

Issue #623, opened 2020-03-04
`help` crashes if not given arguments


jyn514:
```
$ bin/osh
osh$ help
Traceback (most recent call last):
  File "/home/joshua/Documents/oil/bin/oil.py", line 962, in <module>
    sys.exit(main(sys.argv))
  File "/home/joshua/Documents/oil/bin/oil.py", line 908, in main
    return AppBundleMain(argv)
  File "/home/joshua/Documents/oil/bin/oil.py", line 881, in AppBundleMain
    status = ShellMain('osh', argv0, main_argv, login_shell)
  File "/home/joshua/Documents/oil/bin/oil.py", line 683, in ShellMain
    prompt_plugin, errfmt)
  File "/home/joshua/Documents/oil/core/main_loop.py", line 95, in Interactive
    is_return, _ = ex.ExecuteAndCatch(node)
  File "/home/joshua/Documents/oil/osh/cmd_exec.py", line 1851, in ExecuteAndCatch
    status = self._Execute(node, fork_external=fork_external)
  File "/home/joshua/Documents/oil/osh/cmd_exec.py", line 1796, in _Execute
    status, check_errexit = self._Dispatch(node, fork_external)
  File "/home/joshua/Documents/oil/osh/cmd_exec.py", line 1050, in _Dispatch
    status = self._RunSimpleCommand(cmd_val, fork_external)
  File "/home/joshua/Documents/oil/osh/cmd_exec.py", line 791, in _RunSimpleCommand
    return self.RunSimpleCommand(cmd_val, fork_external)
  File "/home/joshua/Documents/oil/osh/cmd_exec.py", line 879, in RunSimpleCommand
    return self._RunBuiltin(builtin_id, cmd_val, fork_external)
  File "/home/joshua/Documents/oil/osh/cmd_exec.py", line 513, in _RunBuiltin
    status = self._RunBuiltinAndRaise(builtin_id, cmd_val, fork_external)
  File "/home/joshua/Documents/oil/osh/cmd_exec.py", line 456, in _RunBuiltinAndRaise
    status = builtin_func.Run(cmd_val)
  File "/home/joshua/Documents/oil/osh/builtin_misc.py", line 585, in Run
    span_id=cmd_val.arg_spids[1])
IndexError: list index out of range
```

andychu:
Hm I'm not getting that, but I did change some stuff related to the help build process that may be causing problems.

[...]

PR #794, opened 2020-07-22
Give a more helpful error if all `help` is missing

jyn514:
This is failing because `os_path` doesn't implement the `exists` function:

```
  File "/home/joshua/src/oil/osh/builtin_misc.py", line 598, in Run
    if not self.loader.Exists('_devbuild/help'):
  File "/home/joshua/src/oil/core/pyutil.py", line 91, in Exists
    return os_path.exists(os_path.join(self.root_dir, rel_path))
AttributeError: 'module' object has no attribute 'exists'
```

I'm not sure how to implement it in pure python, I thought I'd reach out for help.

Addresses #623.

### DIFF
--- a/core/pyutil.py
+++ b/core/pyutil.py
@@ -86,6 +86,9 @@ class _FileResourceLoader(_ResourceLoader):
       contents = f.read()
     return contents
 
+  def Exists(self, rel_path):
+    # type: (str) -> bool
+    return os_path.exists(os_path.join(self.root_dir, rel_path))
 
 class _ZipResourceLoader(_ResourceLoader):
   """Open resources INSIDE argv[0] as a zip file."""
--- a/core/pyutil.py
+++ b/core/pyutil.py
@@ -9,7 +9,7 @@ import zipimport  # NOT the zipfile module.
 
 from mycpp import mylib
 from pgen2 import grammar
-from pylib import os_path
+from pylib import os_path, path_stat
 
 import posix_ as posix
 
@@ -88,7 +88,7 @@ class _FileResourceLoader(_ResourceLoader):
 
   def Exists(self, rel_path):
     # type: (str) -> bool
-    return os_path.exists(os_path.join(self.root_dir, rel_path))
+    return path_stat.exists(os_path.join(self.root_dir, rel_path))
 
 class _ZipResourceLoader(_ResourceLoader):
   """Open resources INSIDE argv[0] as a zip file."""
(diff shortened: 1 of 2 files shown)


andychu:
Oh sorry this PR got buried below the bugs in my e-mail ...

We do have `pylib/path_stat.py exists()`, although that wouldn't work when it's in a zip file.  Let me think about what a good strategy is here

andychu:
Oh one possibility is just do hide it behind 

```
if mylib.PYTHON:
  import path_stat
  if path_stat.exists('_devbuild/help'):
    ...
```

That won't get translated to C++.  I think it's OK to add this one-off, as opposed to change the loader interface for something that only happens in dev builds.

jyn514:
> We do have pylib/path_stat.py exists()

Ooh that worked, thanks!

>  I think it's OK to add this one-off, as opposed to change the loader interface for something that only happens in dev builds.

`loader.Exists()` seems like it would be useful in other contexts. I don't have one in mind off the top of my head but adding to the interface seems better than adding hacks.

andychu:
But there is no implementation of Exists() for _ZipResourceLoader.  _FileResourceLoader and _ZipResourceLoader are polymorphic and should both implement _ResourceLoader.  

I'm pretty sure this will cause some MyPy type errors.  It may say something to the effect of _ResourceLoader doesn't have a method `Exists()`.

The way I check is `types/oil-slice.sh travis` (and I won't be surprised if there is a roadbump in running that that we need to fix, feel free to file a bug)

----

I think one other possibility is 

```
loader.HelpExists()
```

_ZipResourceLoader should always return True.  And then _FileResourceLoader returns if _devbuild/help exists.

That is probably a cleaner way to do it that obeys the "rules" of inheritance ...

It should still be hidden behind `if mylib.PYTHON`

andychu:
Hm actually it appears type checking passed

https://travis-ci.org/github/oilshell/oil/jobs/711840034

but I'm not sure why, as `self.loader` is of type `_ResourceLoader` so  the base class has to have the method

And `builtin_misc.py` should be type checked

andychu:
(Back from vacation)

I cloned this PR and I saw the type error I expected.  (I'm not sure why Travis didn't flag it -- something to look into later.)

```
andy@lisa ~/git/oilshell/oil2 (pr/794)$ types/oil-slice.sh travis

osh/builtin_misc.py:598: error: "_ResourceLoader" has no attribute "Exists"
Found 1 error in 1 file (checked 72 source files)

FAIL
```

How about the `HelpExists()` change I suggested?  I think that is actually the most natural interface for what we're trying to do.

jyn514:
I am not planning to follow up on this.

PR #811, opened 2020-08-11
Improve echo, read, mapfile documentation

dcghub:
I just wanted to check your opinion on these changes. I've tried to make texts as short and concise as possible, and I have tried to mark as unimplemented those that I _think_ they are.

The spec tests have been helpful, as you told me.

### DIFF
--- a/doc/help.md
+++ b/doc/help.md
@@ -883,6 +883,26 @@ CMD &
 CMD:
   The command to be run.
 
+<h4 id="fg">fg</h4>
+
+The fg builtin returns a job running in the background to the foreground.
+
+fg JOB?
+
+JOB:
+  Job ID to be moved to the foreground. If none is specified, the latest job
+  is chosen.
+
+<!--<h4 id="bg">bg</h4>
+
+The bg builtin resumes suspend job, while keeping it in the background.
+
+bg JOB?
+
+JOB:
+  Job ID to be resumed in the background. If none is specified, the latest job
+  is chosen. -->
+
 ### External
 
 <h4 id="kill">kill</h4>
--- a/doc/help.md
+++ b/doc/help.md
@@ -753,6 +753,25 @@ FLAGS:
   -L  If present in the path, follow symbolic links (default behavior)
   -P  Don't follow symbolic links, print the link instead of the reference.
 
+<h4 id="pushd">pushd</h4>
+
+The pushd builtin adds a directory to the directory stack, and changes the
+working directory to it. This is typically used along with popd and dirs.
+
+<!--pushd FLAGS DIR-->
+pushd DIR
+<!--pushd +/-NUM-->
+
+DIR:
+  File system directory where you want to move and which you want to stack.
+
+<!--FLAGS:
+  -n  Don't change the working directory, just manipulate the stack 
+NUM:
+  Rotates the stack the number of places specified. Eg, given the stack
+  '/foo /bar /baz', where '/foo' is the top of the stack, pushd +1 will move
+  it to the bottom, '/bar /baz /foo'-->
+
 ### Completion
 
 <h4 id="complete">complete</h4>
--- a/doc/help.md
+++ b/doc/help.md
@@ -796,22 +796,25 @@ FLAGS:
 
 <h4 id="complete">complete</h4>
 
-Register completion policies for different commands.
+The complete builtin registers completion policies for different commands.
 
 <h4 id="compgen">compgen</h4>
 
-Generate completion candidates inside a user-defined completion function.
+The compgen builtin generates completion candidates inside a user-defined
+completion function.
 
 It can also be used in scripts, i.e. outside a completion function.
 
 <h4 id="compopt">compopt</h4>
 
-Change completion options inside a user-defined completion function.
+The compopt builtin changes completion options inside a user-defined completion
+function.
 
 <h4 id="compadjust">compadjust</h4>
 
-Adjust COMP_ARGV according to specified delimiters, and optionally set
-variables cur, prev, words (an array), and cword.  May also set 'split'.
+The compadjust builtin adjusts COMP_ARGV according to specified delimiters,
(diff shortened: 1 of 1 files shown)


doc/help.md
```diff
@@ -580,14 +580,70 @@ PS1=$x.
...
+  \UHHHHHHHH Unicode character specified as an hexadecimal value with 1 to
+             8 hex digits
+
+<h4 id="readarray">readarray</h4>
+
+Alias for mapfile
+
+<h4 id="mapfile">mapfile</h4>
+
+The mapfile builtin reads lines from stdin into an array
```
andychu, line 636: Hm this only has `-t` implemented

https://github.com/oilshell/oil/blob/master/frontend/flag_def.py#L118

I would omit the (unimplemented) ones altogether (or somehow comment them out).  Because they might not be implemented for a long time!

andychu:
I guess to avoid getting caught up in open-ended designs for the format, I would merge something that only describes documented stuff. 

You can generally check just by doing say `mapfile -d` in OSH.

doc/help.md
dcghub: I have added a usage line and made sure that the options existed. I also have added a few more commands!

dcghub:
I have been adding some commands and I have added addressed some of your suggestions, let me know what's your opinion.

Minor setback: I haven't been able to test if "help foo" works...because the weird UnicodeEncodeError is happening again...

andychu:
Hm can you copy and paste the terminal session here?

https://github.com/oilshell/oil/issues/809


And show "git status" and "git diff".  And `which python`, `python -V` , etc.

doc/help.md
```diff
@@ -580,49 +580,330 @@ PS1=$x.
 
 These builtins take input and output.  They're often used with redirects.
 
+<h4 id="read">read</h4>
+
+The read builtin reads lines from stdin and stores them into a variable. Input
+is split according to the $IFS delimiter.
+
+read FLAGS VAR+
```
andychu, line 588: I think this should be `FLAGS* VARS*`

Because both flags and variables are optional.  I started using the regex of syntax of:

```
a*  0 or more
a+  1 or more
a?  0 or 1 (optional)
```

doc/help.md
```diff
@@ -580,49 +580,330 @@ PS1=$x.
...
+  -u FD     read from file descriptor FD instead of the standard input
+
 <h4 id="echo">echo</h4>
 
 The echo builtin prints its arguments to stdout, separated by a space, and
-terminated by a newline.  TODO: Describe flags, etc.
+terminated by a newline.
 
-<h4 id="read">read</h4>
+echo FLAGS STR
```
andychu, line 613: ditto, should be `FLAGS* STR*`

doc/help.md
```diff
@@ -580,49 +580,330 @@ PS1=$x.
...
 
 ### Run Code
 
+<h4 id="source">source</h4>
+
+The source builtin executes a script, which is run in the context of the
+running shell and can modify local variables (as opposed to running in a
+subshell, where it can't)
+
+source SCRIPT ARGS+
```
andychu, line 678: ARGS*

doc/help.md
```diff
@@ -580,49 +580,330 @@ PS1=$x.
...
+
+The shopt builtin configures the shell.
+
 ### Working Dir
 
+<h4 id="cd">cd</h4>
+
+The cd builtin changes the working directory.
+
+cd FLAGS DIR
```
andychu, line 734: FLAGS* DIR?

DIR can also be `-` I believe

doc/help.md
```diff
@@ -580,49 +580,330 @@ PS1=$x.
...
+
+<h4 id="mapfile">mapfile</h4>
+
+The mapfile builtin reads lines from stdin into an array.
+
+mapfile FLAGS ARRAY
 
-Reads lines from stdin into a variable -- $REPLY by default.
+ARRAY:
+  Array where the contents will be stored.
```
andychu, line 654: The description of the array feels a bit redundant.  While there should be consistency, it shouldn't be like "filling out a form".

Also there is a default value `MAPFILE`

doc/help.md
dcghub: Oooh, I noticed you were using them and tried to imitate it, but I didn't realize they were regex. That's a nice idea, will change it.

dcghub:
OK, here is a new iteration, I think I have addressed all your points, let me know if I did some mistake

As for the build problem...I swear I'm not joking, it now works again. I haven't rebooted and/or updated or changed anything AFAIK. I have tried to locate the problem in different branches or past commits, found nothing.

andychu:
Great thanks for doing this.   This helps get things going.... I am going to test it out and make another pass over it myself.

andychu:
Also I think the unicode issue must be that your editor is inserting Unicode characters like spaces or quotes ?  What editor are you using?

Again I would try `file` right when  you get the error.  It should be possible to at least figure out which file it's coming from

andychu:
I made an editing pass over read and echo

https://github.com/oilshell/oil/commit/1994d280b2138e5eeaeeb62989b900df7acda66a

It now renders better as markdown:

http://www.oilshell.org/preview/doc/help.html#read

The dual markdown/text thing is a little weird... I'm not sure if it will hold up, but so far it seems to.  That is, we can render the same text as HTML in the browser and in the terminal.  The conventions are sort of evolving, but it seems that the 4 space block is good for aligned flag descriptions.  And for the usage lnie.

andychu:
Some more edits:

https://github.com/oilshell/oil/commit/1de4b9f7776f836ffecf48d81c6bd482c15a7db9

I like how this is turning out!  More PRs in the same direction are appreciated :-)

dcghub:
Nice, thanks for the merge! I will take a look at your edits to follow your suggestions in the future.

dcghub:
I have some clues about the building problem - I will post them in the bug issue

PR #816, opened 2020-08-23
Add test

dcghub:
test is different to any other builtins, so I wanted to submit in a separate pull

### DIFF
--- a/doc/help.md
+++ b/doc/help.md
@@ -889,6 +889,103 @@ JOB:
 
 ### External
 
+<h4 id="test">test</h4>
+
+String tests:
+
+    test STR
+    test OP STR
+    test STR OP STR
+
+    [ STR ]
+    [ OP STR ]
+    [ STR OP STR ]
+
+
+File tests:
+
+    test TST FILE
+    test FILE1 OP FILE2
+
+    [ TST FILE ]
+    [ FILE1 OP FILE2 ]
+
+
+Arithmetic tests:
+
+    INT1 OP INT2
+
+    [ INT1 OP INT2 ]
+
+
+Evaluates a conditional expression and returns 0 (true) or 1 (false) as result.
+The number of elements in the expression depends on the type of test.
+
+String tests:
+
+    -n STR              True if STR is not empty. This is the default operation
+                        if no test is specified.
+    -z STR              True if STR is empty.
+    STR1 = STR2         True if the strings are equal.
+    STR1 != STR2        True if the strings are not equal.
+    STR1 < STR2         True if STR1 sorts before STR2 lexicographically.
+    STR1 > STR2         True if STR1 sorts after STR2 lexicographically.
+                        Note: < and > should be escaped, eg: \< and \>
+
+File tests:
+
+    -a FILE  True if FILE exists (recommended to use -e instead).
+    -b FILE  True if FILE is a block special file.
+    -c FILE  True if FILE is a character special file.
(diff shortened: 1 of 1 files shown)


doc/help.md
```diff
@@ -889,6 +889,86 @@ JOB:
 
 ### External
 
+<h4 id="[">[</h4>
```
andychu, line 892: Nothing will point here, so I think we should combine it with `test`.   The list of topics is in the index doc:

https://www.oilshell.org/release/0.8.pre10/doc/help-index.html

doc/help.md
```diff
@@ -889,6 +889,86 @@ JOB:
...
+    -c FILE  True if FILE is a character special file.
+    -d FILE  True if FILE is a directory.
+    -e FILE  True if FILE exists.
+    -f FILE  True if FILE is a regular file.
+    -g FILE  True if FILE has the sgid bit set.
+    -G FILE  True if current user's group is also FILE's group.
+    -h FILE  True if FILE is a symbolic link.
+    -L FILE  True if FILE is a symbolic link.
+    -k FILE  True if FILE has the sticky bit set.
+<!--    -N FILE  True if FILE was modified since last read (mtime newer than atime).-->
```
andychu, line 926: The comments in the middle cause some formatting problems in HTML.  Try building the HTML and seeing how it looks (and let me know if you can't)

BTW I found some smart unicode quotes in some files.  That could be related to the Unicode errors

doc/help.md
```diff
@@ -889,6 +889,86 @@ JOB:
...
+    INTEGER1 -ge INTEGER2  True if INTEGER1 is greater than INTEGER2.
+    INTEGER1 -lt INTEGER2  True if INTEGER1 is less than INTEGER2.
+    INTEGER1 -gt INTEGER2  True if INTEGER1 is greater than INTEGER2.
+
+Other operators:
+
+    -o OPTION  True if the shell option OPTION is set.
+    -v VAR     True if the variable VAR has been set.
+<!--    -R VAR     True if the variable VAR has been set and is a nameref variable. -->
+    EXPR1 -a EXPR2  True if expressions EXPR1 and EXPR2 are both true.
```
andychu, line 965: I would be tempted to leave these out as discouraged, and just say it follows POSIX.  But you can leave it alone, I will likely make an editing pass

doc/help.md
```diff
@@ -889,6 +889,86 @@ JOB:
...
+Tips:
+
+Don't forget to leave a space between the brackets and the expression. "[" is
+a command, not syntax, so the shell will fail to recognize it without the
+spaces.
+
+
+<h4 id="test">test</h4>
+
+    test EXPR+
```
andychu, line 908: What's EXPR?  It can't be repeated one or more times (`+`).

I think it's more accurate to write as:

```
test STR
test OP STR
test STR OP STR

[ STR ]
[ OP STR ]
[ STR OP STR ]
```

doc/help.md
dcghub: You are right, I didn't notice it. I have moved it at the end of the block, it doesn't affect the HTML output there.

doc/help.md
dcghub: That seems sensible, I added that and also equivalents for other tests, which I guess they are necessary?

andychu:
Hm I think it is a little long now, but I will edit it down.  Thanks!

PR #821, opened 2020-08-31
Docs

dcghub:
Another round, this completes the builtin section. 
A couple of things worth noticing here.
 * The printf section contains the same escape sequences as echo. Not a big deal, but it's duplication.
 * getopts is more of a small howto than a help text, but it seems appropriate in this special case, otherwise it's not easy to decipher how to use it. I also hope I didn't made any mistake, there were a lot of subtleties.

### DIFF
--- a/doc/osh-help.md
+++ b/doc/osh-help.md
@@ -932,6 +932,54 @@ Format specifiers:
                 epoch. It can also be -1 (current time, also the default value
                 if there is no argument) or -2 (shell startup time).
 
+#### getopts
+
+    getopts SPEC VAR ARG*
+
+Parses the options passed to the script. The character corresponding to the
+detected option is stored in VAR. By default, it parses the options passed
+to the shell ($1, $2...) unless an alternative options string ARG is passed.
+
+Spec:
+
+    SPEC contains the characters to be parsed as options. Eg, the string "abc"
+    will parse the options -a, -b and -c. If the options needs an additional
+    argument, a colon must be added after the character. Eg, the string "ab:c"
+    will parse arguments for -a, -b arg, and -c.
+
+    Arguments must be separated from options with spaces. ":" and "?" can't be
+    used as options.
+
+How to use:
+
+    getopts returns success if an option is found, and fails if the end of
+    options or an error is found. Every time it is run, it stores the character
+    corresponding to the detected option in VAR, and also places the index of
+    the next option in the variable `$OPTIND` (which is initialized to 1 every
+    time a shell starts). If an option requires an argument, the argument is
+    placed in the variable `$OPTARG`. This design is expected to be used in a
+    loop with a case statement, like this:
+
+    while getopts "ab:c" options; do
+        case $options in
+            a)
+                echo "-a option passed"
+                ;;
+            b)
+                echo "-b option passed with $OPTARG argument"
+                ;;
+            c)
+                echo "-c option passed"
+                ;;
+            *)
+                echo "Option $OPTARG: Unknown or argument missing"
+                ;;
+        esac
+    done
+
+    When an invalid option is found or a required argument is not found, VAR
+    is set to '?' and `$OPTARG` is unset.
+
 <h4 id="kill">kill</h4>
 
 TODO
--- a/doc/osh-help.md
+++ b/doc/osh-help.md
@@ -1006,6 +1006,28 @@ View on the web:
 
     https://www.oilshell.org/release/$VERSION/doc/
 
+#### hash
+
+    hash FLAG* NAME*
+
+Avoids searching for the location of commands by keeping a hash cache
+containing the path of previously executed commands.
+
+Name:
+
+    Name of the command whose path should be remembered, so they need not to be
+    searched for subsequent invocations. `$PATH` is used for the search, and
+    previously remembered paths are discarded. If no NAME and FLAG is supplied,
+    it prints the current hash cache.
+
+Flag:
+
+    -r       Discard all remembered locations.
+<!--    -d       Discard the remembered location of each NAME.
+    -l       Display output in a format reusable as input.
+    -p PATH  Inhibit path search, PATH is used as location for NAME.
+    -t       Print the full path of one or more NAME.-->
+
 <h3>Word Lookup</h3>
 
 ### Interactive
--- a/doc/osh-help.md
+++ b/doc/osh-help.md
@@ -550,30 +550,9 @@ Prints ARGs to stdout, separated by a space, and terminated by a newline.
 
 Flags:
 
-    -e  enable interpretation of backslash escapes
+    -e  enable interpretation of backslash escapes (see `escapes`)
     -n  omit the trailing newline
 <!--  -E  -->
-
-Backslash sequences recognized by -e:
-
-    \\         backslash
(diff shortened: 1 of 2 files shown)


doc/osh-help.md
```diff
@@ -805,6 +806,135 @@ these are discouraged.
 
 <!--    -R VAR     True if the variable VAR has been set and is a nameref variable. -->
 
+<a h4="printf">printf</h4>
```
andychu, line 809: These can all use markdown now:

```
#### printf
````

I changed the toolchain to support that.

(Also there is a typo here -- should be h4 tag, not a tag!)

File: doc/osh-help.md [hunk 2]
```diff
@@ -805,6 +806,135 @@ these are discouraged.
...
+    -v VAR  Write output in variable VAR instead of standard output.
+
+FMT:
+    String that can contain three types of objects: 1) characters that will be
+    printed, 2) escape sequences that are interpreted and then printed,
+    3) format specifiers that print the strings passed to printf as ARGS, but
+    after giving it a determinate format (if not enough ARGS are passed, zero
+    or a nul string will be used)
+
+Escape sequences:
```
andychu, line 825: I think it would make sense to create a new help topic for the character escapes.  So here is no duplication between echo and printf.

I can do that (because one surprising difference/bug is that `echo`   in bash uses  `\0ddd` while printf uses `\ddd`  !)

If you look at `_C_STRING_COMMON`  in `frontend/lexer_def.py`, it has the parts they share, and the parts they don't share are in `PrintfOuter` and `ECHO_E_DEF`

doc/osh-help.md
```diff
@@ -831,10 +961,125 @@ View on the web:
...
+Executes the shell builtin CMD, with arguments ARG, ignoring the existence of
+shell functions with the same name. Typically used when a shell builtin is
+reimplemented as a shell function, but the shell function needs to execute the
+builtin.
+
 ### Interactive
 
+<h4 id="alias">alias</h4>
+
+    alias CMD=STR
```
andychu, line 1033: This is subjective, but I would say `NAME=CODE` or `NAME=CODE_STR`

To me `CMD` could be a command string

doc/osh-help.md
```diff
@@ -805,6 +806,135 @@ these are discouraged.
...
+        floating-point literal.
+    %A  Same as %a, but print it like %E.
+    %(FORMAT)T  Prints date and time, according to FORMAT as a format string
+                for strftime(3). The argument is the number of seconds since
+                epoch. It can also be -1 (current time, also the default value
+                if there is no argument) or -2 (shell startup time).
+
+<h4 id="getopts">getopts</h4>
+
+    getopts STR VAR ARGS?
```
andychu, line 879: I think `ARG*` makes more sense here

and maybe instead of `STR` it should be `SPEC` , it's like a flag specifier

doc/osh-help.md
```diff
@@ -805,6 +806,135 @@ these are discouraged.
 
 <!--    -R VAR     True if the variable VAR has been set and is a nameref variable. -->
 
+<a h4="printf">printf</h4>
+
+    printf FLAG? FMT ARGS*
```
andychu, line 811: Even though there's only one flag, I think `FLAG*` is more correct, since

`printf -v one -v two %s arg`

 is allowed, and that is true for all builtins.  The flag parsers prefer the last value.

doc/osh-help.md
```diff
@@ -805,6 +806,135 @@ these are discouraged.
...
+Str:
+
+    STR contains the characters to be parsed as options. Eg, the string "abc"
+    will parse the options -a, -b and -c. If the options needs an additional
+    argument, a colon must be added after the character. Eg, the string "ab:c"
+    will parse arguments for -a, -b arg, and -c.
+
+    Arguments must be separated from options with spaces. ":" and "?" can't be
+    used as options. getopts will report errors; you can prepend a colon to the
+    beginning of the string (eg. ":ab:c") to enable silent error reporting.
```
andychu, line 895: It doesn't look like Oil does this.  We should only  document things that Oil does, not what bash does.   Generally it's a subset

doc/osh-help.md
```diff
@@ -831,10 +961,125 @@ View on the web:
...
+        'builtin' or 'file'.
+<!--    -a  Print all executables that can run CMD, including files, aliases,
+        builtins and functions. If used with -p, only the executable file will
+        be printed.-->
+ 
 <h3>Word Lookup</h3>
 
+<h4 id="command">command</h4>
+
+    command FLAG? CMD ARG*
```
andychu, line 1007: ditto, FLAG*

doc/osh-help.md
```diff
@@ -831,10 +961,125 @@ View on the web:
...
+is not provided, the current alias for CMD will be shown. If run without
+parameters, it will show a list of defined alias. To remove an existing alias,
+use `unalias`.
+
+Tips:
+
+Aliases are often used to create new compounded commands that are used as a
+shortcut, or to add default options to existing commands. For example,
+alias ports='ss -tulanp' will set up an alias for the command "port" which,
+when run, will display port information. alias ls='ls --color' will make ls
```
andychu, line 1045: I would discourage use of aliases, because they cause weird parsing problems, and you can almost laways use a function:

```
ports() {
  ss -tulanp "$@"
}

ls() {
  ls --color "$@"
}

```

or in Oil:

```
proc ports {
  ss -tulanp @ARGV
}

proc ls {
  ls --color @ARGV
}
```

doc/osh-help.md
```diff
@@ -831,10 +961,125 @@ View on the web:
...
+always display colors.
+
+Use of aliases by the shell can be disabled by prepending the character '\' to
+the command (eg, \ls), or quoting the command with single or doble quotes (eg,
```
andychu, line 1049: typo double

doc/osh-help.md
```diff
@@ -831,10 +961,125 @@ View on the web:
...
+"ls" or 'ls').
+
+Aliases are not used by the shell when the shell is not interactive.
+
+<h4 id="unalias">unalias</h4>
+
+    unalias CMD
```
andychu, line 1056: ditto, I could use `NAME` or `ALIAS_NAME`

I read `CMD` as "command", and commands are more general than aliases.

If you think of commands as the "first word", then functions, builtins, external programs, etc. are all commands

doc/osh-help.md
```diff
@@ -831,10 +961,125 @@ View on the web:
...
+Removes the alias for CMD.
+
+<!--Flag:
+
+    -a  Removes all existing aliases.-->
+
+<h4 id="history">history</h4>
+
+    history NUM
+    history FLAG ARGS
```
andychu, line 1067: I don't think this second form is implemented, we don't have any ARGS

doc/osh-help.md
dcghub: Sorry for the delay! Here is a new version
 * I added a new topic for the character escapes. I am afraid I wasn't able to find more differences between the echo and printf escapes, the code was too obscure for me.
 * I tried to fix the getopts "errors mode" part, but I'm not sure I did it right, this command is pretty weird and I wasn't able to debug the osh command
 * All the other suggestions were incorporated.

PR #825, opened 2020-09-04
Add Linux-only guard around HAVE_EPOLL

gromgit:
This should do the right thing across all platforms, especially macOS.

### DIFF
--- a/Python-2.7.13/pyconfig.h
+++ b/Python-2.7.13/pyconfig.h
@@ -197,7 +197,9 @@
 #define HAVE_DYNAMIC_LOADING 1
 
 /* Define if you have the 'epoll' functions. */
+#ifdef __linux__
 #define HAVE_EPOLL 1
+#endif
 
 /* Define to 1 if you have the `erf' function. */
 #define HAVE_ERF 1


andychu:
Thanks for the patch!

Can you show the compile error you're getting on OS X?

The better fix might be to set it to 0, since a shell doesn't need epoll.  The pyconfig has been gradually patched to be more portable.

andychu:
Ah OK now I see the Homebrew patch.  I just deleted them:

https://github.com/oilshell/oil/commit/e0e94f13ac7b8874a962af81b9949f46dbea321a

It seems like 0.7.0 should have been broken too, but oh well ...

andychu:
Thanks for testing Oil and let me know if you have any problems!

gromgit:
> Ah OK now I see the Homebrew patch. I just deleted them:
> 
> [e0e94f1](https://github.com/oilshell/oil/commit/e0e94f13ac7b8874a962af81b9949f46dbea321a)

That works too, thanks much!

> It seems like 0.7.0 should have been broken too, but oh well ...

It wasn't present in 0.7.0. 0.8.0 introduced `Python-2.7.13/Modules/selectmodule.c`, which is the sole consumer of `HAVE_EPOLL`.

andychu:
Ah yes, that makes sense.  Trivia: `read -t 0` in bash calls `select()`  !  It's different than `read -t 0.1`.

There's a lot of hidden stuff like that in bash, like `strftime()` hidden in `printf` ...

PR #836, opened 2020-10-02
treewide: use /usr/bin/env bash

grahamc:
I had some trouble with not all of the tests not passing, and I can take a look in to these later this evening. It is tricky to know if I broke them or not, since none of the builds work without this patch :).


benchmarks/testdata/test-cmd-util.sh
```diff
@@ -655,7 +655,7 @@ run_pod_tests() {
   kube::test::get_object_assert rc "{{range.items}}{{$id_field}}:{{end}}" ''
   ## kubectl create --edit can update the label filed of multiple resources. tmp-editor.sh is a fake editor
   TEMP=$(mktemp /tmp/tmp-editor-XXXXXXXX.sh)
-  echo -e "#!/bin/bash\n$SED -i \"s/mock/modified/g\" \$1" > ${TEMP}
+  echo -e "#!/usr/bin/env bash\n$SED -i \"s/mock/modified/g\" \$1" > ${TEMP}
```
andychu, line 658: It looks like the `sed` caught some testdata, which isn't desirable.

Let me try a  more focused `sed` command ... The continuous build will catch the broken tests for us

grahamc: Ah. I did see this change and it wasn't clear to me what the right choice was. Thank you.

andychu:
OK I fixed this, and even hooked up a lint check to the continuous build.  Let me know if you have more problems!

https://github.com/oilshell/oil/commit/ea3b5934dadb04a6dffe15ea2524b846fd75a991

grahamc:
Cool, thank you!

PR #855, opened 2020-10-31
Copy editing :)

drkameleon:


README.md
```diff
@@ -89,12 +89,12 @@ It's great for prototyping.
...
 free to edit them.  If you make a major change, let us know on Zulip!
 
-There are also READMEs in some subdirs, like `opy/` and `mycpp/`.
+There are also READMEs in some subdirectories, like `opy/` and `mycpp/`.
 
 If you're confused, the best thing to do is to ask on Zulip and someone should
 produce a pointer and/or improve the docs.
 
-Docs for **end users** are linked from each [release
+Docs for **end-users** are linked from each [release
```
andychu, line 97: Hm I don't think this should be hyphenated

I think when it's a noun, there should be no hyphen.  Arguably when it's used as an adjective it can have a hyphen.

https://ell.stackexchange.com/questions/116816/end-user-vs-end-user-correct-usage-of-hyphen

doc/oil-help.md
```diff
@@ -32,7 +32,7 @@ This section describes how to use the Oil binary.
     Usage: oil  [OPTION]... SCRIPT [ARG]...
            oil [OPTION]... -c COMMAND [ARG]...
 
-`bin/oil` is the same as `bin/osh` with a the `oil:all` option group set.  So
+`bin/oil` is the same as `bin/osh` with a the `oil:all` option groupset.  So
```
andychu, line 35: The meaning I was going for was "option group enabled".

`oil:all` is an "option group", and the whole thing can be enabled or disabled (set or unset).

doc/oil-help.md
```diff
@@ -143,7 +143,7 @@ Initializes a name to the Oil expression on the right.
     var s = 'mystr'        # equivalent to declare s=mystr
     var pat = / digit+ /   # an eggex, with no shell equivalent
 
-It's either a global or scoped to the current function.
+It's either global or scoped to the current function.
```
andychu, line 146: I think "a global" was OK here... that is basically short for "a global variable".

doc/oil-help.md
```diff
@@ -284,7 +284,7 @@ The shell-like syntax accepts the same syntax that a command can:
 
     +  -  *  /   //   %   **
 
-`div` is for integer math, while `/` is for floating point math.
+`div` is for integer math, while `/` is for floating-point math.
```
andychu, line 287: Please remove this whole sentence, as we no longer have `div`.

(I also prefer the unhyphenated version but it's moot :) )

doc/oil-help.md
```diff
@@ -429,7 +429,7 @@ Re-enable errexit.
 <!-- note: explicit anchor necessary because of mangling -->
 <h4 id="strict:all">strict:all</h4>
 
-Option in this group disallow problematic or confusing shell constructs.  The
+Option in this group disallows problematic or confusing shell constructs.  The
```
andychu, line 432: It should be "Options in this group disallow"

doc/oil-overview.md
```diff
@@ -149,7 +149,7 @@ Mentioned in the last post.  This is big!
 - Like shell and Python, there is a global scope and function scope..
 - shell functions don't live in their own scope?  This is like Lisp-1 vs
   Lisp-2.
-- There won't be dynamic scope as in shell.
+- There won't be dynamic scope as in the shell.
```
andychu, line 152: This is a fine distinction, but "in shell" to me means "in the shell language".  Sort of like "this works in Python", "this works in shell".

"the shell" sounds like something you type into, not a language.

andychu (commented):
Thanks for looking over this!   Mostly good except for some of the hyphen changes.

And one thing I need to is fine a Markdown spell checker ...

PR #857, opened 2020-11-04
[tea] Support AST generation for loops

isidentical:
- Implement 'for' statements in the grammar
- Support CST->AST conversion for loops (while/for)
- Add 'break'/'continue'/'return' as tea keywords.

### DIFF
--- a/oil_lang/expr_to_ast.py
+++ b/oil_lang/expr_to_ast.py
@@ -914,9 +914,75 @@ class Transformer(object):
 
     return sig
 
+  def func_item(self, node):
+    """
+    func_item: (
+      ('var' | 'const') name_type_list '=' testlist  # oil_var_decl
+
+      # TODO: for, if/switch, with, break/continue/return, try/throw, etc.
+    | 'while' test suite
+    | 'for' name_type_list 'in' test suite
+    | flow_stmt
+    | 'set' place_list (augassign | '=') testlist  # oil_place_mutation
+      # x  f(x)  etc.
+      #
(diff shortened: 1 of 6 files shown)


isidentical:
I couldn't run the tests on my local system @andychu, I presume you have some sort of way to test PRs so it is appreciated if you could try and report (I initally thought there is a CI, but apparently there is not.)

oil_lang/expr_to_ast.py
```diff
@@ -914,9 +914,75 @@ def Proc(self, pnode):
...
+    else:
+      raise NotImplementedError(node)
+
+  def func_items(self, pnode):
+    # type: (PNode) -> List[ast.stmt]
+    """
+    func_items: func_item (semi_newline func_item)* [semi_newline]
+    """
+    raw_items = pnode.children
+    if raw_items[-1].typ == grammar_nt.semi_newline:
```
andychu, line 965: I'm not sure this is necessary, since `::2` should do the right thing?

but yeah this code is inherently sort of fiddly that way so I think I will just merge and see what happens

andychu (approved):
Thanks this looks great!

oil_lang/expr_to_ast.py
isidentical: > I'm not sure this is necessary, since ::2 should do the right thing?

Well, since the pattern  `x, y, x, y, x, ...` is ensured; taking first, third, fifth item seemed make sense. Of course we could possibly right it in tons of different ways, including more explicit ones. But for this case, it seemed simple and powerful enough.

PR #895, opened 2021-01-25
Removing obsolete & unused readme-native

saurabhnigam:


andychu:
This is used!  There's still a separation between oil and oil-native unfortunately.  oil-native is the fast C++ version that should become the primary version eventually.

https://www.oilshell.org/release/0.8.7/

That is described on the blog a little bit: http://www.oilshell.org/blog/tags.html?tag=mycpp#mycpp

PR #922, opened 2021-04-05
doc/qtsv.md - fix typo

ilyash:


andychu:
Thanks!  By the way I might rename QTSV to QTT -- quoted, typed table.

It's like TSV but it can represent any string (including ones with tabs), and it can express the difference between floats, ints, strings, and bools.

I haven't yet implemented it but I am doing something right now that needs it: importing dumps like `find . -ls` into a sqlite database.

PR #930, opened 2021-05-05
idiom: fix run failures

happysalada:
running the idiom with the not_mutated example will currently yield a failure
```
 echo $not_mutated
       ^~~~~~~~~~~~
./backup.sh:5: fatal: Undefined variable 'not_mutated'
```

It seems to me that the examples should run as is.
Let me know if you need anything else.

### DIFF
--- a/doc/idioms.md
+++ b/doc/idioms.md
@@ -318,6 +318,7 @@ No:
 
 Yes:
 
+    var not_mutated = 'bar'
     forkwait {
       setvar not_mutated = 'foo'
     }


andychu:
Wow thanks for fixing this!  I'm pretty sure this just changed a day or 2 ago, but glad you caught it either way

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Variable.20Scope.20Thoughts

Feel free to join Zulip and leave any feedback on Oil :)

andychu:
The basic change was that I got rid of `setlocal`.  Now Oil just has `const` and `var` for declaration, and `setvar` and `setglobal` for mutation.  I think that is a lot simpler!

happysalada:
Thank you for merging this so quickly. Let me try some more of it!
I'll let you know if I find anything else!

Issue #934, opened 2021-05-07
Clarify if and errexit paragraph


happysalada:
In the idioms https://www.oilshell.org/release/latest/doc/idioms.html#if-and-errexit
the rule around if and brackets is not completely clear.
I would add at the end of that paragraph, something like
```
note that you can't add brackets around a function.
brackets are for evaluating conditionals only.

    if (! myfunc) {
        ^
   Syntax error in expression (near Id.Expr_Bang)
```

Let me know if that makes sense. 
The original error I came from was this

```
./run.sh dev
    if (! pg_isready) {
        ^
./run.sh:4: Syntax error in expression (near Id.Expr_Bang)
```

if that works for you, I'll make the PR, just let me know.

andychu:
I think a one-sentence link to "A Tour of Oil" that reminds the user that parens are significant would be very helpful!  I want to keep this doc brief, but it's worth mentioning.

andychu:
It's also sorta mentioned here: https://www.oilshell.org/release/0.8.10/doc/syntax-feelings.html#parens-expression

But I agree that the significant parens are unlike other languages, so it's worth highlighting

andychu:
Reopening this since we still haven't published a "Tour of Oil" yet !

I want to also thank you for the great feedback and say I didn't forget about this!  We have `build/doc.sh tour` which now actually validates the code in the doc.  And the next step is to actually finish writing it!

I have been travelling, so that has been unfortunately delayed, but this is one of the more important things coming up!

happysalada:
When you have a moment, could you detail the bigger parts that are left to write?
If you give me just a 1-10 bullet point list, I can dig into each of those and take a stab at it.
Like usual, If whatever I write doesn't end up being the actual thing but just serves as inspiration, I'm happy.

andychu:
Hm I think it would be useful to add more to the `oil_lang/testdata/tour.oil` file?  And to make sure it runs with `bin/oil`.  Remember there was an outline in the PR.

I copied some of that into the draft of `doc/tour.md`, so it does help!

happysalada:
Just to make sure I understand correctly.
Here is the original outline


    command language
        simple commands
            builtins
            external commands
            Ruby-like blocks -- e.g. cd /tmp { pwd }. This still needs more implementation.
        redirects
        pipelines
        control flow
            if
            case
            for
            while
        procs
        variable declaration and mutation
        maybe: fork and wait
    word language
        variable sub - $var
[...]

PR #935, opened 2021-05-08
initial add of tour of oil

happysalada:
I started on a tour of oil path.
It's a .sh file to be executable. I thought that way it could be embedded into learnXinY
I'm open about the format.
This is an initial PR, not everything is included for sure.
Feel free to nitpick on anything, I don't mind changing wording or format.
Hopefully this gets the discussion going.

References that should probably be included
https://www.oilshell.org/release/latest/doc/eggex.html
https://www.oilshell.org/release/latest/doc/idioms.html
https://www.oilshell.org/release/latest/doc/oil-help.html

missing topics
- read
- write
- test

Documents that can provide inspiration
https://learnxinyminutes.com/docs/tcsh/
https://learnxinyminutes.com/docs/bash/

fixes https://github.com/oilshell/oil/issues/934

Idea
- I quite like the `putting it together` part at the end of https://learnxinyminutes.com/docs/zfs/
I'm thinking of including a simple script like a git hook maybe? At the end to illustrate the concepts.

### DIFF
--- a/oil_lang/testdata/tour.oil
+++ b/oil_lang/testdata/tour.oil
@@ -1,0 +1,227 @@
+# Command language
+
+## Simple commands
+
+### variable declaration and mutation
+
+# Declaring a variable does NOT looks like this:
+# Variable="Some string" => name=val isn't allowed when shopt 'parse_equals' is on.
+# Hint: add 'env' before it, or spaces around =
+
+# Nor like this
+# Variable = "Some string" # => same error
+
+# this is the preferred way 
+# no need for capitalization like in bash
+var variable = "Some string"
+
+# also valid but not idomatic
+set variable = "Another string"
+
+# const to declare a constant
+const immutable = "definitive"
+
+# this will not work
+# set immutable = "not so sure" =>  Can't modify constant 'immutable'
+
+# mutate a variable with setvar
+var new = "a brave new world"
+setvar new = "perhaps it's already here"
+echo $new
+
+# setglobal will create or mutate a global variable
+setglobal the_answer = 42
+echo $the_answer
+
+# there are more ways to set variable by passing them
+# to procs. This will be expanded in the procs section
+
+### Variable substitution
+
+var regular_string = "I am just a normal string"
+echo $regular_string # quotes not needed
+#### if you want to do string interpolation
+echo "$regular_string and $regular_string"
+#### you need to do it another way if you want to substitute before a _
+# this will not work "$regular_string_"
+# as it will look for the regular_string_ variable
+echo "${regular_string}_" # but this will
+
+### Parameter expansion works just like in bash
+var some_string = "I am some string"
+echo ${some_string/another/some} # => I am another string
+
+### Strings
+
+# this will not work
+# echo "\n" => Invalid char escape in double quoted string
+# this prints the characters
+echo '\n'
+# use $ to print a new line
+echo $'\n'
+
+# unicode works like this
+var unicode = c'mu = \u{3bc}'
+echo $unicode
+# but not like this
+# var unicode = 'mu = \u{3bc}' =>  Strings with backslashes should look like r'\n', c'\n' or $'\n'
+
+### builtins
+
+#### echo
+# takes at most 1 argument and prints to stdout
+echo "Hello world!" # => Hello world!
+
+#### write
+# write outputs to stdout
+# write has a -- separator to not confuse command line arguments and what you want to print
+var i_am_a_string = "I'm just a string"
+write -- $i_am_a_string
+# if you have a string with a newline in it
+var i_have_a_newline = $'it's true I really do \n'
+# write works well with it, unlike bash's printf
+write -- $i_have_a_newline
+# you can print the content without the newline with -n
+write -n -- $i_have_a_newline
+
+#### test
+# test has been improved to be a little less crytic
+
+# test if something is a directory
+test --dir /tmp || echo "Not a dir"
+test -d /tmp # still works but less readable
+
+# test if something is a file
+test --file /tmp/maybe_a_file || echo "Not a file"
+
+### Procs
+
+### think of procs as a better version of bash's functions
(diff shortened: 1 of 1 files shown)


andychu:
The content looks good so far!  But I think that it should be one of two things:

- in the learn X in Y format and go in THAT repo (might be early, but it should happen eventually)
- in the same format as everything in `doc/`, and in our repo

I understand the desire to make it executable, and that's kind of a hole in our system.  But I don't want to accumulate TODOs since the docs are already a big undertaking and far from done.

Although it does seem very similar to the stuff here:

https://github.com/oilshell/oil/tree/master/oil_lang/testdata

It's possible that we can put the same content there AND it will be automatically run in the continuous build.  However that testdata is not published anywhere, which would be something else to do.

andychu:
I guess one possible option is to put it in `testdata/` for now, and iterate on the content (and probably fix some language bugs uncovered!)

The `_match('year')` thing is just unimplemented, and that's something that someone else brought up.  So it would be nice to fix that.

I think that will get us started.  And then at some point we can convert it to the `doc/` format so it will appear on the web.  (This could be a manual conversion or honestly it might not be hard to write a tool to do it.)

happysalada:
I don't have a problem with writing it as a md format.
I didn't know there was a testdata directory, considering visibility is limited, it seems better to have it as documentation.
I'll rewrite it and add the missing parts.
Let me know if you feel there are other things that should go in there.

andychu:
OK, I guess `.md` is OK but there is the problem that the code snippets have to be tested.  I think that the tour should quickly go over everything that's idiomatic:

- command language
  - simple commands
    - builtins
    - external commands
  - pipelines
  - control flow
    - if
    - case
    - for
    - while
  - procs
  - variable declaration and mutation
- word language
  - command sub
  - expression sub
  - inline function calls like `@split(x)` and `$join()`
- expression language
  - this is mostly Python
  - eggex


It can have a lot of links to other docs.  It's possible that this doc could get really long so we should find ways to keep it short and sweet.

We can leave out:

- exotic stuff like process sub
- rarely used shell stuff like `until`
- legacy shell stuff: `[[`, `$(( ))`, etc.

happysalada:
I moved things into tour.oil
Thank you for providing the outline.
I tried to include it with some indentation with several ##. This is not standard. I was thinking otherwise it could be included in separate functions ?

The only change I did to your outline was to bump up the variables section.

It's not complete, I can continue once we agree on the format.

andychu:
Sorry for the delay.  I think the  "Learn X in Y Minutes" style is pretty readable.  I was just suggesting a list of topics to cover -- and I think Learn Bash in Y minutes covers all of those.

If you aren't sure about a topic then you can just leave it out and I'll fill it in.  I think it's useful to make something that's readable to you or your friend, using the Learn X in Y style.

----

I added a `hello.oil` and a function to run it:

https://github.com/oilshell/oil/commit/1bdad7c849456d33cf6ba8711a657983b84b1de2

So we can run this with

    oil_lang/run.sh all-passing

andychu:
I'm feeling the need for the "Tour of Oil", e.g. see this comment responding to someone who said Oil's control flow was cryptic (?)

https://lobste.rs/s/d9xics/bourne_shell_bash_aren_t_right_languages#c_ygbqib

Maybe as a first pass you can just add the things that confused you in your initial feedback on issue #934?

That is just have a section on if statements.  And then we can fill it in gradually.

I have also made a few changes recently -- actually based on your feedback.  In `doc/strings.md` there is a bunch of new content and a quick reference:

https://github.com/oilshell/oil/blob/master/doc/strings.md#quick-reference

I think that is good material for the tour of Oil.  But I should probably add it because not everything works yet! :-/

happysalada:
I've added everything I used so far.
Still missing are for, while, blocks, fork.
I think it should be a pretty good start though.

Here are some comments
- I couldn't make the `echo -n $my_string` fail. My understanding of write is that it's supposed to act as a better echo. but I couldn't find use cases for it. I put what I knew about it, but I feel the case for `write` is weak as it is. I would be interested in having more examples
- There are currently 3 ways to use strings r' c' $'. I feel it would be good to recommend one, or differentiate properly with use cases for each. As I understand it $' is for using with newline strings, c' is null byte terminated strings. So as I understand the default recommendation should be $' ?
- The implicit run marked here does not work https://www.oilshell.org/release/latest/doc/idioms.html#if-and-errexit

Let me know if anything.

andychu:
1.  The issue with echo is that `echo $x` does NOT print an arbitrary string, which is a useful operation.  If x happens to be `-n` or `-e`, then it will be treated as a flag, and not printed.  There's no way to fix this in echo; it doesn't respect `--`.  `printf '%s' "$x"` is the POSIX way to do it.  But Oil provides `write -- $x` for simplicity.
2. Yes!   I just  got rid of `c''`.  Brainstorming on the Zulip thread: https://oilshell.zulipchat.com/#narrow/stream/202008-oil-documentation/topic/String.20Brainstorming.20.28quoting.20issue.20.28strict_errexit.20error.2E.2E.2E
3. What doesn't work?  Can you provide an repro of what you expected to work and what didn't?

andychu:
Good news is that the shell script actually catches bugs here, i.e. it caught the fact that we no longer accept `c''`, and a few other things:

https://github.com/oilshell/oil/commit/aae6eec4df42c1f9a761c62cad9f4963ac0710e1

So we want to sketch it first here to make sure it works, and then convert it to a proper doc

happysalada:
Feel free to re-order as you like, no worries from my side!

in this document it says that run is implicit https://www.oilshell.org/release/latest/doc/idioms.html#if-and-errexit
However if you try to put a proc in an if condition without run, it will throw an error
```
./test.sh:6: fatal: Can't run a proc while errexit is disabled. Use 'run' or wrap it in a process with $0 myproc
```
If you want to test it, replace `if run err_today` with `if err_today` in tour.oil

PR #945, opened 2021-05-19
add r' and write explanations

happysalada:
I added the explanations of r' and the echo flag confusion.

I understand that you are considering moving out of this format. No worries on my side.
Let me know if you are ok to submit this to learnXinY and I will make a PR.

### DIFF
--- a/oil_lang/testdata/tour.oil
+++ b/oil_lang/testdata/tour.oil
@@ -59,6 +59,8 @@ echo ${some_string/another/some} # => I am another string
 echo '\n'
 # use $ to print a new line
 echo $'\n'
+# oil intentionally adds a more explicit syntax for 'raw' strings
+echo r'\n' #
 
 # unicode works like this
 var unicode = $'mu = \u{3bc}'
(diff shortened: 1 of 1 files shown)


andychu:
Hm I think the `-n` issue too detailed to go in a "learn x in y".  That is a corner case, which Oil does need to consider and address

But I think it's just going to be too confusing to beginners, who are the audience for "learn x in y"

It's more of a footnote in a book in my opinion.

andychu:
Also I think we eventually need a "learn X in Y", but it's slightly early for it yet, as the language is still changing.

For example I have a bunch of renaming proposals here:

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Renaming.20Proposals

Including changing `run` to `try`, which would affect the doc.  I think we have to polish our own docs and then move it out into the world.

I would also consider getting feedback from /r/oilshell, lobste.rs, /r/ProgrammingLanguages, etc. about a "learn x in y" for Oil.  I do that with all my docs and it helps to have others read them over.  I can give feedback too but I'm also pretty close to the project, which paradoxically can be bad for docs

happysalada:
I agree! closing this for now.

PR #1033, opened 2021-12-04
Remove instances of "the the" from documentation.

emdash:
As discussed, some simple fixes to remove this common typo, based on a simple grep search.


andychu (approved):
Thanks

Issue #1003, opened 2021-11-16
limit size of process and jobs dictionary 


andychu:
In `core/process.py JobState`, we use a dictionary to map PID -> `Process` object, etc. for `waitpid()`.

This dictionary is currently unlimited in size, which should probably be considered a memory leak

emdash:
In `process.py`, we find the following (starting at line 1343)

```
  def MaybeRemove(self, pid):                                                                                                                                  
    # type: (int) -> None                                                                                                                                      
    """Process and Pipeline can call this."""                                                                                                                  
    # Problem: This only happens after an explicit wait()?                                                                                                     
    # I think the main_loop in bash waits without blocking?                                                                                                    
    log('JobState MaybeRemove %d', pid)                                                                                                                        
                                                                                                                                                               
    # TODO: Enabling this causes a failure in spec/background.                                                                                                 
    if 0:                                                                                                                                                      
      try:                                                                                                                                                     
        del self.jobs[pid]                                                                                                                                     
      except KeyError:                                                                                                                                         
        # This should never happen?                                                                                                                            
        log("AssertionError: PID %d should have never been in the job list", pid)                                                                              
```

Wondering if you can provide any insight into why the spec tests were failing.

andychu:
Hm unfortunately I don't remember, but I think what the `fg` issue revealed is that there is whole entire missing mechanism where we try to update the shell state but we're not doing on notification from the kernel.  So those 2 things can diverge so it's not surprising there is a bug here.

If you turn this on, can you reproduce the failure in `test/spec.sh background` ?  It might be flaky; not sure.

emdash:
I will investigate

emdash:
Enabling this does not seem to break `test/spec.sh background` at this point in time. So I guess I could fold that into the fix for #1004 / etc.

andychu:
Jobs dictionary looks done, but we still have the process dictionary I think

andychu:
@melvinw I think this is done?  Do you concur?

Not sure if we had a test for it, or if we just inspected the code

Thinking about it for release notes

andychu:
This was done with AddChildProcess / RemoveChildProcess() etc.

Though it could still be nice to have a test that makes sure we don't leak anything over long interactive shells, either this or other data structures

Issue #360, opened 2019-06-22
Implement the rest of job control


andychu:
- pipelines should be handled correctly (I only handled the vim case)
  - call setpgid() on every process in a pipeline
    - pstree -g is useful here (shows process group leaders)
  - killpg(..SIGCONT) instead of kill(...) in `fg` / `bg`
- stuff about the controlling terminal that I don't understand
  - I assume this is how the `bg` builtin is different
- output of `jobs` builtin needs polish -- `jobs -l`, etc.

Forked from #176

jyn514:
More things for the list:
- set `$!` to the pid of the last suspended process or background: https://stackoverflow.com/questions/1908610/how-to-get-pid-of-background-process
- set `%n` to the `n`th running background process

About `bg`: this is used for jobs that were started in the background, it's effectively the same as `<process> & kill -SIGTSTP $!`

jyn514:
I think controlling terminals are somehow related to the fact that `python -c input() &` shouldn't be able to read stdin: http://curiousthing.org/sigttin-sigttou-deep-dive-linux

andychu:
- $! should already work. 
- I added job numbering, but didn't implement `%n` syntax, which is now a trivial parsing problem
- Right now OSH supports `mycommand &`.  It's correctly "async".  But I don't think it's correctly in the "background", which is more of a terminal issue.  So `bg` needs to do that too.

jyn514:
This is a little strange ... I could swear it was working this morning, but now when I suspend vim, it immediately restarts. Here's an strace:
```
$ strace -p 4991 -f
...
[pid  5080] kill(0, SIGTSTP)            = 0
[pid  4991] <... wait4 resumed> 0x7fffd8896434, WSTOPPED, NULL) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
[pid  5080] --- SIGTSTP {si_signo=SIGTSTP, si_code=SI_USER, si_pid=5080, si_uid=1000} ---
[pid  4991] --- SIGTSTP {si_signo=SIGTSTP, si_code=SI_USER, si_pid=5080, si_uid=1000} ---
[pid  5080] recvmsg(5,  <unfinished ...>
[pid  4991] wait4(-1,  <unfinished ...>
[pid  5080] <... recvmsg resumed> {msg_namelen=0}, 0) = -1 EAGAIN (Resource temporarily unavailable)
... lots of `recvmsg` calls ...
```
[...]

PR #1037, opened 2021-12-06
[job control] Fix fg bug, and remove from jobs dictionary

emdash:
It seems that the `last_stopped_pid` instance variable of a processes's `JobState` is not cleared when a child process receives `SIGINT`. This patch explicitly clears this flag. Not sure if there's a better way to resolve this, but it seems to work.

```
osh-0.9.5$ fg                                                                                                                                                  
Continue PID 3199                                                                                                                                              
  C-c C-c                                                                                                                                                      
osh-0.9.5$ fg                                                                                                                                                  
No job to put in the foreground                                                                                                                                
osh-0.9.5$ 
```

Testing done:
- manually tested according to the bug descriptions
- spec tests pass
- all unit tests pass
- added `test/interactive.py` which a few variations of this bug
  - uses the python `pexpect` library
  - passes on this branch, fails on master
  - just a demo, not hooked into the test framework

I feel like it's 90% there, now it's just down to nits.


core/process.py
```diff
@@ -1461,6 +1461,8 @@ def WaitForOne(self, eintr_retry):
       # Print newline after Ctrl-C.
       if term_sig == signal_.SIGINT:
         print('')
+        if pid == self.job_state.last_stopped_pid:
+            self.job_state.last_stopped_pid = -1
```
andychu, line 1465: Thanks for looking at this!  I think the general idea of resetting last_stopped_pid is right, but probably not on Ctrl-C in general?

Note that there is only one instance of `JobState` in the whole interpreter.   I should probably rename it because it is confusing with `Job`.  So it isn't a 1:1 relationship.

Let me try to think of a test case where this wouldn't work... I think if you do something besides Ctrl-C it won't work.

Like you could resume `cat`, and instead of `Ctrl-C`, you can do `Ctrl-\`, or kill it from outside the shell.  Does the bug still happen then?

emdash: okay, I'll check... i am not quite sure how `SIGINT` propagates through the code, but this branch seems to be checking for `SIGINT` specfically, `KeybardInterrupt` is explicitly ignored above...

emdash: Seems to work if I use kill from inside the shell...

```
osh-0.9.5$ cat                                                                                                                                                 
  C-c C-z                                                                                                                                                      
[PID 4969] Stopped
osh-0.9.5$ kill -INT 4969                                                                                                                                      
osh-0.9.5$ fg                                                                                                                                                  
Continue PID 4969                                                                                                                                                                                                                                                                                      
osh-0.9.5$      
```


```
osh-0.9.5$ cat                                                                                                                                                 
  C-c C-z                                                                                                                                                      
[PID 6032] Stopped                                                                                                                                             
osh-0.9.5$ fg                  # ran kill -INT 4969 from outside the shell                                                                                                                                
Continue PID 6032                                                                                                                                                                                                                                                                                          
osh-0.9.5$  
```

This case fails:
```
osh-0.9.5$ cat                                                                                                                                                 
  C-c C-z                                                                                                                                                      
[PID 6077] Stopped                                                                                                                                             
osh-0.9.5$ kill -9 6077                                                                                                                                        
osh-0.9.5$ fg                                                                                                                                                  
Continue PID 6077                                                                                                                                              
osh I/O error: No such process                                                                                                                                 
brandon@rizzo:~/src/oil$
```
So I guess it does need to be more general.

andychu:
How about if the `fg` command resets the value to -1  ?  That seems to make sense, because `fg` continues a process.  After it's continued then it's no longer stopped.

I think there might be other places to track that, but that is a start ?

The other place might be if the kernel can notify us when a process goes from the stopped state to running?  I forget exactly how that works.   One thing I do is look at other shells do it

andychu:
Hm another possibility is that we just catch the "No such process" error, and then reset the state to -1 when that happens?  

When I search for "errno no such process", it turns out that is ESRCH.  So let's look what other shells do with ESRCH

https://man7.org/linux/man-pages/man3/errno.3.html

andychu:
dash does this, so I searched for "No current job" in the source code and found it.  Trying to figure out what it does


```
$ cat
^Z[1] + Stopped                    cat
$ fg
cat
^C
$ fg
dash: 5: fg: No current job
$

```

andychu:
Hm in `jobs.c`, the `fgcmd()` function in dash does call `set_curjob(..., CUR_RUNNING)`

Not sure exactly what that means, but it's a clue.

FWIW we archive other shell source code here; it's not too hard to find on the web

http://www.oilshell.org/blob/spec-bin/

andychu:
OK I think I see.

- `fgcmd()` calls `restartjob()`
- `restartjob()` calls `waitforjob()`
- `waitforjob()` ends up calling the wait syscall

So basically I think `fg` has to call the Waiter somehow, so **after it sends SIGCONT**, it gets notification of this from **the kernel**, and the internal data structures are updated correctly

emdash:
> How about if the `fg` command resets the value to -1 ? That seems to make sense, because `fg` continues a process. After it's continued then it's no longer stopped.
> 
> I think there might be other places to track that, but that is a start ?
> 
> The other place might be if the kernel can notify us when a process goes from the stopped state to running? I forget exactly how that works. One thing I do is look at other shells do it

I'm not sure that works if you never actually resume the process (e.g. sending SIGKILL from outside the shell), but the small tweak I just pushed does at least prevent OSH from crashing when this happens.

emdash:
Digging in further, it seems jobs are never removed from`JobState`. It defines `MaybeRemove`, but the actual `del` statement is if-defed out, and near as I can tell it's not called from anywhere. Instead, it seems that `NoneAreRunning` iterates over the job list each time it's called. I feel like I'm missing something.

emdash:
> OK I think I see. In `jobs.c`
> 
>     * `fgcmd()` calls `restartjob()`
> 
>     * `restartjob()` calls `waitforjob()`
> 
>     * `waitforjob()` ends up calling the wait syscall
> 
> 
> So basically I think `fg` has to call the Waiter somehow, so **after it sends SIGCONT**, it gets notification of this from **the kernel**, and the internal data structures are updated correctly

Hm, thanks for that explanation, it makes a lot more sense. The `Waiter` object is polling the kernel for changes to child processes via `posix.waitpid`... I feel a little bit more confident I'm in the right place here.

andychu:
Yeah as you noticed there is a bunch of stuff missing in job control, which is #1003 and #360.

The main reason is that I don't use job control myself (I use tmux).  And just to be clear the blog is saying that the interactive shell is a low priority, which includes job control.

But the idea was to implement a very minimal job control, where `Ctrl-Z` -> `fg` works.  I think that's what most people use if anything.

I think the current `fg` behavior is obviously dumb and not good enough.  (It would be nice if someone showed up with a working and complete job control implementation, but I think it is one of the hairiest parts of shell, that uses a lot of features from the kernel that nothing else does.  There is a quote about this in the Stevens UNIX book.)

So I would lean toward a more minimal fix.  We need to fix #1003 but I think that's separate?  I could be wrong.

andychu:
I tagged #1003  essential because it's a memory leak, so that would definitely be nice to fix too ...

emdash:
> Yeah as you noticed there is a bunch of stuff missing in job control, which is #1003 and #360.
> 
> The main reason is that I don't use job control myself (I use tmux). And just to be clear the blog is saying that the interactive shell is a low priority, which includes job control.

I sometimes use screen, if I'm working over ssh, but generally I prefer not to have the extra hassle around escape sequences when I'm working locally. I haven't used tmux... I learned screen a long time ago and just stick with it. But job control is useful for more than just multiplexing your terminal. It's still part of the sysadmin toolbelt, even if systemd is taking over.

> 
> But the idea was to implement a very minimal job control, where `Ctrl-Z` -> `fg` works. I think that's what most people use if anything.
> 
> I think the current `fg` behavior is obviously dumb and not good enough. (It would be nice if someone showed up with a working and complete job control implementation, but I think it is one of the hairiest parts of shell, that uses a lot of features from the kernel that nothing else does. There is a quote about this in the Stevens UNIX book.)

Yeah, I think if this at least doesn't work right, it will slow adoption. Job control does seem tricky, but there's tons of prior art out there, as you point out.

> 
> So I would lean toward a more minimal fix. We need to fix #1003 but I think that's separate? I could be wrong.

1003 is separate, but properly fixing both bugs seems to require a more robust approach to monitor the state of subprocesses. I think bash does this after every command, and maybe also periodically, based on the behavior I've observed over the years.

emdash:
Okay, I think I fixed the main issues with this latest update.
```
osh-0.9.5$ bin/osh                                                                                                                                             
osh-0.9.5$ exit                                                                                                                                                
JobState NotifyDone 30361                                                                                                                                      
osh-0.9.5$ cat                                                                                                                                                 
  C-c C-z                                                                                                                                                      
[PID 30370] Stopped                                                                                                                                            
osh-0.9.5$ fg                                                                                                                                                  
Continue PID 30370                                                                                                                                             
  C-c C-c                                                                                                                                                      
JobState NotifyDone 30370                                                                                                                                      
osh-0.9.5$ fg                                                                                                                                                  
No job to put in the foreground

This case doesn't seem quite right, but at least doesn't crash.
```
osh-0.9.5$ cat                                                                                                                                                 
  C-c C-z                                                                                                                                                      
[PID 30408] Stopped                                                                                                                                            
osh-0.9.5$ kill 30408                                                                                                                                          
JobState NotifyDone 30414  # <-- here, the PID is different.         
osh-0.9.5$ fg                                                                                                                                                  
Continue PID 30408              # <-- Ideally would say "No job to put in the foreground", but this PID is still in the table                                                                                                                            
JobState NotifyDone 30408                                                                                                                                      
osh-0.9.5$
```

**Comment by emdash** [comment]
Ready for review. 

Checked in the test script alongside the changes. "cleaned up" the test script a bit so it is easier to manipulate from the shell, but it remains a "stand-alone" script, not integrated into any other test apparatus.

**Comment by emdash** [comment]
Opened: https://github.com/pexpect/pexpect/issues/705 to hopefully get some clarification from the `pexpect` folks. We'll see what they say.

File: test/interactive.py [hunk 2]
```diff
@@ -0,0 +1,140 @@
+#! /usr/bin/python2
+
```
andychu, line 2: Can you write some comments about how to use this at the top?  And a link to the pexpect dependency and how you installed it, etc.

Also please follow the style of the other Python scripts

- `#!/usr/bin/env python2` (important for some distros)
-  2 space indents

Not sure exiting with g_failures is a good idea; I think 0 or 1 is more conventional.  If you have 256 failures it will overflow.

core/process.py
```diff
@@ -1215,6 +1215,34 @@ def GetLastStopped(self):
     # This be GetCurrent()?  %+ in bash?  That's what 'fg' takes.
     return self.last_stopped_pid
 
+  def NotifyContinue(self, pid, waiter):
+    # type: (int) -> None
+    if pid == self.last_stopped_pid:
+        self.last_stopped_pid = -1
```
andychu, line 1221: Please use 2 space indents consistently

osh/builtin_process.py
```diff
@@ -258,12 +258,7 @@ def Run(self, cmd_val):
     # TODO: Print job ID rather than the PID
     log('Continue PID %d', pid)
     posix.kill(pid, signal.SIGCONT)
-
-    job = self.job_state.JobFromPid(pid)
-    job.state = job_state_e.Running  # needed for Wait() loop to work
-    status = job.Wait(self.waiter)
-    #log('status = %d', status)
-    return status
+    return self.job_state.NotifyContinue(pid, self.waiter)
```
andychu, line 261: looks like this return statement isn't necessary

core/process.py
```diff
@@ -1462,11 +1476,13 @@ def WaitForOne(self, eintr_retry):
       if term_sig == signal_.SIGINT:
         print('')
 
+      self.job_state.NotifyDone(pid)
       proc.WhenDone(pid, status)
 
     elif WIFEXITED(status):
       status = WEXITSTATUS(status)
       #log('exit status: %s', status)
+      self.job_state.NotifyDone(pid)
```
andychu, line 1485: I think maybe we should rename these

self.job_state.WhenDone()

self.job_state.WhenContinued()

self.job_state.WhenStopped()

What do you think?

andychu (commented):
Thanks, this looks promising ... I still have to clone this and test it out, but I left some preliminary comments

As mentioned, please add some comments so I and other contributors know how to run it

core/process.py
emdash: Sure, that's what `Process` et. al. use. `Notify*` suggests some kind of callback, but everything here is synchronous.

osh/builtin_process.py
emdash: I did this to preserve the existing behavior, but I'll take a closer look.

test/interactive.py
emdash: Would you prefer this as a docstring or just as comments?

emdash: went with the docstring for now.

osh/builtin_process.py
emdash: I looked through the other classes in `builtin_process.py`, and it seems like they all their `Run` methods return some kind of status code. Am I missing something?

emdash:
Responded to review feedback, update pushed.

core/process.py
```diff
@@ -1215,6 +1215,34 @@ def GetLastStopped(self):
     # This be GetCurrent()?  %+ in bash?  That's what 'fg' takes.
     return self.last_stopped_pid
 
+  def WhenContinued(self, pid, waiter):
+    # type: (int) -> None
```
andychu, line 1219: OK I think I was confused because this is marked as returning None.

It will fail type checking (and this reminds me that you need a way to push to CI)

core/process.py
```diff
@@ -1196,7 +1196,7 @@ def __init__(self):
   #
   # [job_id, flag, pgid, job_state, node]
 
-  def NotifyStopped(self, pid):
+  def WhenStopped(self, pid):
     # type: (int) -> None
```
andychu, line 1200: this one also needs its type comment updated

andychu (commented):
OK let me pull this and run it locally, I think the logic makes sense

core/process.py
emdash: Is CI the only way get type checking, or can I run this locally? If so, I would add that to my workflow (and also update the wiki / etc to mention how this is done).

core/process.py
emdash: will do.

core/process.py
emdash: As an aside, depending on the framework you're using, I could see making a type-level distinction between various integer types:
- status codes
- process ids
- group ids
- user ids 
- file descriptors
etc.

andychu:
Right now I do `types/oil-slice.sh travis` to type check.  You just need MyPy installed -- I think `pip3 install mypy` should work.

Let me know if it works or not!

core/process.py
```diff
@@ -1215,6 +1215,34 @@ def GetLastStopped(self):
...
+    # needed for Wait() loop to work
+    job.state = job_state_e.Running
+    return job.Wait(waiter)
+
+  def WhenDone(self, pid):
+    # type: (int) -> None
+    """Process and Pipeline can call this."""
+    # Problem: This only happens after an explicit wait().
+    # I think the main_loop in bash waits without blocking?
+    log('JobState NotifyDone %d', pid)
```
andychu, line 1232: Is this log message for testing only?  (I'm not sure if it was there before)

Do we need it to test?    If so maybe hide it behind an environment variable?  So it doesn't happen in the normal flow.  

Ditto for any other log messages.

andychu (commented):
I pulled this and ran it, seems to work great! 

I just had to run `pip install pexpect`.   (I will add this to the continuous build once this is merged)

I made one comment about extraneous log messages.

Thanks for setting up the pexpect testing!  I think we can make use of this in the future :)

andychu:
Oh yeah when I run it interactively I get all the `JobState NotifyDone` messages.  So those should be either removed or hidden for testing

core/process.py
emdash: I preserved any log messages I as I found them, but I have no problem removing it. I will need to update the tests, of course.

emdash:
Removed that log message. I will leave logging (and log output redirection) as a separate issue.

andychu:
Great thanks for doing this!  This area definitely needed some love

And I want to hook these tests up to the CI !

PR #1042, opened 2021-12-08
Use system mypy if it is the only option

### LINKED ISSUE #1041: Typechecking script should use system version mypy if it is the only instance
Noticed that the scripts assume mypy is installed via pip.

Simple fix: #1042 

Broader question: how should one treat search paths for dependencies in general?

emdash:
Created to fix #1041

### DIFF
--- a/types/common.sh
+++ b/types/common.sh
@@ -3,15 +3,26 @@
 mypy_() {
   ### Version of mypy that PIP installs
 
-  # Try 2 locations.  There's a weird difference between pip3 install location
-  # on Debian vs. Ubuntu
-  local first=/usr/local/bin/mypy
-  local second=~/.local/bin/mypy
+  # Try 3 locations:
+  # - system prefix
+  # - global pip install prefix
+  # - user pip install prefix
+  #
+  # Which of these search paths should take precedence is an open
+  # question.
+  #
+  # XXX: Is there a shell idiom for parsing a colon-separated path
+  # list?
+  local system=/usr/bin/mypy
+  local pip_global=/usr/local/bin/mypy
+  local pip_user=~/.local/bin/mypy
 
-  if test -x $first; then
-    $first "$@"
+  if test -x $pip_user; then
+    $pip_user "$@"
+  elif test -x $pip_global; then
+    $pip_global "$@"
   else
-    $second "$@"
+    $system "$@"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


andychu:
Hm github is showing me all the diffs from your other change ... haven't seen that before

andychu:
Though I kinda think you want to use the same version of MyPy as the continuous build?  Why not install it from pip3?

emdash:
> Though I kinda think you want to use the same version of MyPy as the continuous build? Why not install it from pip3?

Mainly, forward thinking:
- containerized builds which target a specific base image
- downstream maintainers who want to run unit tests against their official packages
- "meta distros": nix, buildroot, yocto: these have their own requirements.
- some future user, who does the equivalent `apt-get build-dep osh`

Other considerations:
- At some point, PyPI might diverge from the version of mypy you choose to adopt.
- At some point, PyPI might become defunct
 - recent drama with `npm`
 - `pip search` is still broken

Not really trying to ignite a debate though.

emdash:
> Hm github is showing me all the diffs from your other change ... haven't seen that before

This is because github is bad at guessing the correct base branch. I think I can fix this.

andychu:
OK well how about just making a MYPY or MYPY_BIN variable you can  override?

Or if it's not in the $PATH, then fall back to the pip location.  I think the problem is that pip2 doesn't put it in $PATH on my machine, so that's why it's like that.

Also we should have a better name than `types/oil-slice.sh travis` ... I can do that part.

emdash:
> OK well how about just making a MYPY or MYPY_BIN variable you can override?
> 
> Or if it's not in the $PATH, then fall back to the pip location. I think the problem is that pip2 doesn't put it in $PATH on my machine, so that's why it's like that.
> 
> Also we should have a better name than `types/oil-slice.sh travis` ... I can do that part.

I was thinking the usual colon-separated list of search paths would make sense, just not sure what the shell idiom is for that. My guess is that you want some general mechanism to use for all build dependencies like this, I'm just not sure what it is.

andychu:
Actually we do have build/dev-shell.sh, which modifies $PATH, and is sourced by several of our scripts

Yeah let me take a stab at this.  I think I can make it work for both cases, and make it more consistent

andychu:
Actually I think this is OK, except just make the third option plain `mypy`, not /usr/bin/mypy.

That will find the one installed by your package manager

emdash:
Fair.

emdash:
Updated, moving on.

Issue #594, opened 2020-02-01
Generate parse tables for pgen2-native and hook it up to oil-native


andychu:
This is to translate Oil expressions (including eggexes) to C++.

In some ways this is retreading what CPython already has.  This is a good meaty issue that will move the project forward a lot, if anyone is good with C / C++ and wants to work with parsing algorithms.

andychu:
@melvinw did this in #1540 !   Woohoo!

https://github.com/oilshell/oil/commit/1fd961a23e91488829f5ded2f9949b5b46390b5c

andychu:
Released - https://www.oilshell.org/blog/2023/05/release-0.15.0.html

PR #1048, opened 2021-12-11
Create pgen-native, a stand-alone fork of python's pgen

emdash:
This is part of #594 

Got it building with a stand-alone makefile in a separate subdir

Status:
- can parse python expressions into a parse tree
- still coupled to the official python grammar


andychu:
Cool!  Can you make a short README.md that tells us how to run it ?

Python-2.7.13/Include/Python.h
```diff
@@ -1,4 +1,4 @@
-#ifndef Py_PYTHON_H
+o#ifndef Py_PYTHON_H
```
andychu, line 1: Looks like a typo here?

pgen-native/README.md
```diff
@@ -0,0 +1,23 @@
+This is work in progress attempt to split out the pgen2 from Python 2
```
andychu, line 1: just be clear `pgen2`  is the Python version we have in `pgen2/`, while pgen is the original.  

pgen-native is fine

andychu (commented):
OK this looks cool, so it's ready to merge?   If so I will do so and try it out

Thanks!

Python-2.7.13/Include/Python.h
emdash: good 'ol ctrl-x x o

emdash:
I am wondering what direction to head in, now that there's something which appears to be working...

emdash:
There's few questions I have:

.gitignore
```diff
@@ -3,6 +3,7 @@
 tags
 .vagrant
 .mypy_cache/
+_cache/
```
emdash, line 6: This was showing up as untracked, so I went ahead and added it.

.gitignore
```diff
@@ -33,3 +34,11 @@ TODO*.txt
 uftrace.data
 release-*.txt
 asdl.txt
+
+# pgen-specific stuff, this is probably temporary
```
emdash, line 38: This was driving me crazy, so I added it...

pgen-native/Makefile
```diff
@@ -0,0 +1,25 @@
+CFLAGS := -Wall -g
+
+OBJS := parser.o \
+	tokenizer.o \
+	grammar1.o \
+	grammar.o \
+        bitset.o \
```
emdash, line 7: Whoops, looks like despite my efforts, emacs still *really* wants to put tabs in my files.

pgen-native/Makefile
```diff
@@ -0,0 +1,25 @@
...
+        node.o \
+	python.o \
+	acceler.o \
+        listnode.o \
+	python_grammar.o \
+	parsetok.o \
+	main.o
+
+.PHONY: parser
```
emdash, line 16: Dependency tracking is a royal pain to set up in make, so I just made the main target .PHONY to force a rebuild every time. It really doesn't take long on modern hardware...

pgen-native/Python.h
```diff
@@ -0,0 +1,56 @@
...
+#include <assert.h>
+#include <ctype.h>
+#include <limits.h>
+#include <stdbool.h>
+#include <stddef.h>
+// XXX: Remove instances of these later
+#define PyAPI_FUNC(x) extern x
+#define PyAPI_DATA(x) extern x
+
+#define PY_SIZE_MAX SSIZE_MAX
```
emdash, line 18: I am pretty sure we could remove these defines and replace them with their definitions as appropriate.

pgen-native/Python.h
```diff
@@ -0,0 +1,56 @@
...
+
+// Define this, because it's true.
+#define PGEN
```
emdash, line 21: I would also like to just remove any conditionals of the form `#ifndef PGEN ...`

pgen-native/Python.h
```diff
@@ -0,0 +1,56 @@
...
+
+// Types
+typedef size_t Py_ssize_t;
```
emdash, line 24: Woops, this should be `ssize_t`. Fix that.

Then I would like to replace all instances of `Py_ssize_t` with `ssize_t`.

pgen-native/Python.h
```diff
@@ -0,0 +1,56 @@
...
+
+extern int Py_VerboseFlag;
+extern int Py_DebugFlag;
```
emdash, line 27: I'll leave these for now, they seem kinda useful. Should the `Py_` prefix become `Pgen_`? Or something else?

pgen-native/Python.h
```diff
@@ -0,0 +1,56 @@
...
+/* Argument must be a char or an int in [-128, 127] or [0, 255]. */
+#define Py_CHARMASK(c) ((unsigned char)((c) & 0xff))
+
+// not sure about this one
+#if USE_PY_CTYPE
+#define Py_ISALPHA(c)  (_Py_ctype_table[Py_CHARMASK(c)] & PY_CTF_ALPHA)
+#else
+#define Py_ISALPHA(c) isalpha(c)
+#define Py_ISALNUM(c) isalnum(c)
+#endif
```
emdash, line 54: `USE_PY_CTYPE` s not defined anywhere, I made it up. I am not sure how differently `Py_ISALPHA` behaves from posix `isalpha`, etc. 

This seems like something we should pay close attention to.

pgen-native/acceler.c
```diff
@@ -0,0 +1,127 @@
...
+
+/* The parser as originally conceived had disappointing performance.
+   This module does some precomputation that speeds up the selection
+   of a DFA based upon a token, turning a search through an array
+   into a simple indexing operation.  The parser now cannot work
+   without the accelerators installed.  Note that the accelerators
+   are installed dynamically when the parser is initialized, they
+   are not part of the static data structure written on graminit.[ch]
+   by the parser generator. */
+
```
emdash, line 12: You would think, after 25 years, that they'd have just integrated this stuff permanently...

pgen-native/bitset.c
```diff
@@ -0,0 +1,68 @@
...
+    return 1;
+}
+
+#if 0 /* Now a macro */
+int
+testbit(bitset ss, int ibit)
+{
+    return (ss[BIT2BYTE(ibit)] & BIT2MASK(ibit)) != 0;
+}
+#endif
```
emdash, line 48: I would just clobber this, since it's obviously not used.

pgen-native/bitset.h
```diff
@@ -0,0 +1,32 @@
...
+int samebitset(bitset bs1, bitset bs2, int nbits);
+void mergebitset(bitset bs1, bitset bs2, int nbits);
+
+#define BITSPERBYTE	(8*sizeof(BYTE))
+#define NBYTES(nbits)	(((nbits) + BITSPERBYTE - 1) / BITSPERBYTE)
+
+#define BIT2BYTE(ibit)	((ibit) / BITSPERBYTE)
+#define BIT2SHIFT(ibit)	((ibit) % BITSPERBYTE)
+#define BIT2MASK(ibit)	(1 << BIT2SHIFT(ibit))
+#define BYTE2BIT(ibyte)	((ibyte) * BITSPERBYTE)
```
emdash, line 27: these could be static inline functions

emdash: hell this whole file could just be a `std::bitset` if you're targeting C++...

andychu:
- I think whoever integrates this with `oil-native` will figure out the API.  I don't think we can plan it out ahead of time
- I wouldn't take too much out now, as long as it's fast to build and test, it's OK.  We should get it working with oil-native and then clean up the code
- Yes we need to somehow plug in a different lexer.  This might be one of the harder parts, as the lexer creates garbage collected tokens
- Usually I just have a shell script that calls test.c, or I use greatest.h.   See `test/cpp-unit.sh` and `*/*_test.cc`

Thanks!

pgen-native/grammar.c
```diff
@@ -0,0 +1,257 @@
...
+                    lb->lb_str);
+        }
+        else
+            printf("Can't translate STRING label %s\n",
+                lb->lb_str);
+    }
+    else
+        printf("Can't translate label '%s'\n",
+               PyGrammar_LabelRepr(lb));
+}
```
emdash, line 257: Are these the only functions from this file that we actually need?

Two things come to mind:
- if we generate c at build time, we don't need any of the mutator functions
- if we keep these functions, we could load grammar data from a file (msgpack, json, ini)

pgen-native/grammar1.c
```diff
@@ -0,0 +1,57 @@
...
+
+#include "Python.h"
+#include "pgenheaders.h"
+#include "grammar.h"
+#include "token.h"
+
+/* Return the DFA for the given type */
+
+dfa *
+PyGrammar_FindDFA(grammar *g, register int type)
```
emdash, line 12: the `register` keyword here is suspicious... if you target c++, I think you want to remove it

pgen-native/node.c
```diff
@@ -0,0 +1,165 @@
+/* Parse tree node implementation */
```
emdash, line 1: I assume you have your own node type? Or do you want to keep using this?

pgen-native/parsetok.h
```diff
@@ -0,0 +1,66 @@
...
+
+#ifndef Py_PARSETOK_H
+#define Py_PARSETOK_H
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "node.h"
+
+typedef struct {
```
emdash, line 12: move this to its own file? If you're targeting C++, we should use exceptions anyway...

pgen-native/parsetok.h
```diff
@@ -0,0 +1,66 @@
...
+
+PyAPI_FUNC(node *) PyParser_ParseStringFlagsFilename(const char *,
+					      const char *,
+					      grammar *, int,
+                                              perrdetail *, int);
+PyAPI_FUNC(node *) PyParser_ParseStringFlagsFilenameEx(const char *,
+					      const char *,
+					      grammar *, int,
+                                              perrdetail *, int *);
+
```
emdash, line 59: Do we want all these variants? Do we want the same set of flags?

pgen-native/pgenheaders.h
```diff
@@ -0,0 +1,40 @@
...
+#define newbitset _Py_newbitset
+#define newgrammar _Py_newgrammar
+#define pgen _Py_pgen
+#define printgrammar _Py_printgrammar
+#define printnonterminals _Py_printnonterminals
+#define printtree _Py_printtree
+#define samebitset _Py_samebitset
+#define showtree _Py_showtree
+#define tok_dump _Py_tok_dump
+#define translatelabels _Py_translatelabels
```
emdash, line 35: This is really weird to me... I think I understand what's happening, and I can guess at why, but I don't like it.

pgen-native/python.c
```diff
@@ -0,0 +1,82 @@
+/**
```
emdash, line 1: everything in this file is a placeholder, for now. Everything subject to change or removal, since these are just stubs.

emdash (commented):
I'm reviewing my own PR here, just to call attention to some things. This will also help me remember for later.

pgen-native/grammar.c
emdash: Note, looking at this in the preview, the selection is only a portion of what I intended, so this is a little confusing.

emdash: What i'm asking here is if we need to keep any of the functions which *change* a grammar file, since we never intend to create or manipulate them at runtime.

I'm all for eliminating any code we don't think we'll be calling.

pgen-native/pgenheaders.h
emdash: I would like to either remove these defines, so that the symbols retain their names as written, or else do a find-replace so that the we see the real identifiers.

emdash:
Wondering where this is at... on the one hand, I feel like there's a lot to do, on the other hand, you seemed excited to merge it.

andychu:
OK I built this and ran it, and seems to work!

Sorry for the delay in looking at it

We have to eventually divorce this from:

1. Python's grammar
2. Python's lexer
3. the AST creation or transformer

I think the grammar might be the easiest place to start but you can use your judgement

It will require some code generation

PR #1053, opened 2021-12-13
Add -l and --login flags as no-op

### LINKED ISSUE #1052: osh doesn't support -l (--login) flag
This is a flag in bash that says to run it as if it is a login shell. It is used by https://github.com/purcell/exec-path-from-shell as an example to try and make sure any parts of the path that are expected to be set are set.

Based on how osh sources start up files, I think that the correct thing to do in this case would be either to do nothing or to set `$0` to `-osh` instead of just `osh` but do nothing else (I've definitely seen scripts that check if `$0` starts with `-` or not and do things based on that, but whether `osh` wants to support that kind of stuff or not is questionable).

pdebelak:
Bash supports these flags to indicate that it should run as a login shell. This adds these flags as a no-op to osh since osh doesn't currently behave differently as a login vs. non-login shell.

Please let me know if this is the wrong place/way to test or fix this.

Fixes #1052

### DIFF
--- a/frontend/flag_def.py
+++ b/frontend/flag_def.py
@@ -206,6 +206,8 @@ OSH_SPEC.LongFlag('--help')
 OSH_SPEC.LongFlag('--version')
 
 OSH_SPEC.ShortFlag('-i')  # interactive
+OSH_SPEC.ShortFlag('-l')  # login - currently no-op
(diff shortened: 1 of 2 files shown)


andychu:
Great thank you!  Looks like the right test file.

Issue #1036, opened 2021-12-06
Implement failglob


andychu:
Raphael mentioned that they're using this

andychu:
Released! http://www.oilshell.org/release/0.9.6/

PR #1058, opened 2021-12-19
Implement failglob

pdebelak:
This implements failglob for osh. I'm not sure about the error message setup here, it could be too verbose, but bash prints a message for this in addition to setting a non-zero status. Also could be missing some contexts, but this appears to do the right thing for command context, loop context, and array literals.

In reference to #1036

### DIFF
--- a/core/error.py
+++ b/core/error.py
@@ -153,11 +153,8 @@ if mylib.PYTHON:
     argument.
     """
 
-  class FailGlob(Exception):
+  class FailGlob(_ErrorWithLocation):
     """Raised when a glob matches nothing when failglob is set."""
-
-    def __init__(self, glob):
-      self.glob = glob
 
 
   class Strict(FatalRuntime):
--- a/core/error.py
+++ b/core/error.py
@@ -158,7 +158,6 @@ if mylib.PYTHON:
 
     def __init__(self, glob):
       self.glob = glob
-      self.msg = 'No match: %s' % glob
 
 
   class Strict(FatalRuntime):
--- a/core/error.py
+++ b/core/error.py
@@ -153,6 +153,13 @@ if mylib.PYTHON:
     argument.
     """
 
+  class FailGlob(Exception):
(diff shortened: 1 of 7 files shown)


andychu:
Amazing, thank you!  You even changed the documentation :)

How about making the error message 

    'Pattern %r matched no files (failglob)'  % arg

to remind the user that they have failglob on.

Or maybe:

    'failglob: Pattern %r matched no files'  % arg

Also, one thing I noticed is that despite passing the  spec tests, this behaves slightly differently than bash.  But I think that is OK!

The issue is a semicolon vs. newline, which I think is extremely obscure.

Compare

```
shopt -s failglob
echo *.zz; echo status=$?  # bash doesn't execute the second part !!!

echo *.zz
echo status=$?  # bash executes this

```

I just added this to: https://github.com/oilshell/oil/wiki/Shell-WTFs

I noticed this in some other cases -- occasionally the `;` is significant.  But OSH in general doesn't respect it, and I don't think we should start.

I think it will probably mess up the code to try to respect it, whereas now the logic is clean.

But bash is the only shell that implements failglob.  So it might be nice to add a test case and mark OSH as OK.  Just so we have a record of the difference.

------

I think the error message is more important though ... the `;` is really trivia.  I believe it's due to bash constructing a fairly naive parse tree and then putting the stack unwinding in a specific spot, not a really intentional design decision.

pdebelak:
I've updated the error message to include the pattern and use the `failglob` prefix so it reads like `failglob: Pattern '*.ZZ' matched no files` and added a failing spec test for the bizarre `;` behavior. This should be ready for re-review.

Also, fun note, I keep typing `failblog` instead of `failglob`.

andychu:
Great thank you! 

I followed up with a commit to implement it for extended globs, which are separate from regular globs since they're more static (happen before word splitting).

I hope to do a release with this great work by the end of the week, or maybe next week!

PR #1065, opened 2022-01-04
Fix minor typo in core/completion.py

eliminmax:
While trying to figure out why [aliases and functions were not getting
suggested by the completion system](https://github.com/oilshell/oil/issues/1064), I noticed a small typo in a comment, that prevented it from showing up when I searched for appearances of `compgen` in the repository.


andychu:
Thanks!

PR #1095, opened 2022-03-03
Disallow null bytes inside of re2c rules.

Sketch98:
While building with build/dev.sh all, re2c complained about null bytes. After checking [the re2c manual](https://re2c.org/manual/warnings/warnings.html#wsentinel-in-midrule), i added \0 to the excluded bytes.

> _devbuild/tmp/osh-lex.re2c.h:652:37: error: sentinel symbol 0 occurs in the middle of the rule (note: if a different sentinel symbol is used, specify it with 're2c:sentinel' configuration) [-Werror-sentinel-in-midrule]
> _devbuild/tmp/osh-lex.re2c.h:720:37: error: sentinel symbol 0 occurs in the middle of the rule (note: if a different sentinel symbol is used, specify it with 're2c:sentinel' configuration) [-Werror-sentinel-in-midrule]
> _devbuild/tmp/osh-lex.re2c.h:886:37: error: sentinel symbol 0 occurs in the middle of the rule (note: if a different sentinel symbol is used, specify it with 're2c:sentinel' configuration) [-Werror-sentinel-in-midrule]
> _devbuild/tmp/osh-lex.re2c.h:1099:40: error: sentinel symbol 0 occurs in the middle of the rule (note: if a different sentinel symbol is used, specify it with 're2c:sentinel' configuration) [-Werror-sentinel-in-midrule]

### DIFF
--- a/frontend/lexer_def.py
+++ b/frontend/lexer_def.py
@@ -6,7 +6,7 @@ It consists of a series of lexer modes, each with a regex -> Id mapping.
 After changing this file, run:
 
     build/dev.sh all
-    
+
 or at least:
 
     build/dev.sh fastlex
@@ -58,7 +58,7 @@ def R(pat, tok_type):
 
 # See unit tests in frontend/match_test.py.
 # We need the [^\0]* because the re2c translation assumes it's anchored like $.
-SHOULD_HIJACK_RE = r'#!.*sh[ \t\r\n][^\0]*'
+SHOULD_HIJACK_RE = r'#![^\0]*sh[ \t\r\n][^\0]*'
 
 
 _SIGNIFICANT_SPACE = R(r'[ \t]+', Id.WS_Space)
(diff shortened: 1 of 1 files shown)


andychu:
Ah OK I think this is a good change, thank you.  What version of re2c are you using?

I think this warning doesn't appear with the version we're using, which is 1.0.3

I have this "vendored" copy here

https://github.com/oilshell/oil/blob/master/soil/deps-tar.sh#L26

I think the warning doesn't appear in that version, but maybe the bug does ...

Sketch98:
I grabbed re2c (version 3.0) from artix's repository because build/codegen.sh download-re2c and install-re2c.
I saw there is deps-tar.sh in the soil directory which is for building releases. Should developers also be using that for build tools?

andychu:
Ah OK, I fixed instructions on the Contributing wiki page to point to `soil/deps-tar.sh`.

More importantly, this change actually does fix bugs!  The re2c error is good because it catches reading past the end of a C string!

I had prevented that in the common cases, but these cases were left out and caught by re2c!  Thank you

https://github.com/oilshell/oil/commit/e5b7011335846081577f06beaf25a86380956dc6

PR #1098, opened 2022-03-05
Add support for jobs -p

### LINKED ISSUE #1017: Builtin jobs command missing -p flag
In `bash`/`dash`/`ash`, running `jobs -p` yields the pids of subshells. This is pretty helpful for scripting around with subshells (ex. killing all subshells with `kill $(jobs -p)`).

But currently in osh it seems like `-p` for jobs is unsupported.

In `osh`:
```sh
jobs -p
  jobs -p
       ^~
[ interactive ]:1: 'jobs' doesn't accept flag -p
```

glyh:
Support for `jobs -p`, aiming to close #1017 .

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -778,8 +778,8 @@ class Job(object):
     # Initial state with & or Ctrl-Z is Running.
     self.state = job_state_e.Running
 
-  def DisplayJob(self, job_id, f):
-    # type: (int, mylib.Writer) -> None
+  def DisplayJob(self, job_id, f, opts):
+    # type: (int, mylib.Writer, Dict[str, bool]) -> None
     raise NotImplementedError()
 
   def State(self):
@@ -835,15 +835,18 @@ class Process(Job):
     #return '<Process %s%s>' % (self.thunk, s)
     return '<Process %s %s>' % (_JobStateStr(self.state), self.thunk)
 
-  def DisplayJob(self, job_id, f):
-    # type: (int, mylib.Writer) -> None
+  def DisplayJob(self, job_id, f, opts):
+    # type: (int, mylib.Writer, Dict[str, bool]) -> None
     if job_id == -1:
       job_id_str = '  '
     else:
       job_id_str = '%%%d' % job_id
-    f.write('%s %d %7s ' % (job_id_str, self.pid, _JobStateStr(self.state)))
-    f.write(self.thunk.UserString())
-    f.write('\n')
+    if opts["p"]:
+      f.write('%d\n' % self.pid)
+    else:
+      f.write('%s %d %7s ' % (job_id_str, self.pid, _JobStateStr(self.state)))
+      f.write(self.thunk.UserString())
+      f.write('\n')
 
   def AddStateChange(self, s):
     # type: (ChildStateChange) -> None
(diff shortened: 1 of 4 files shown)


andychu:
This is great!

It would be nice to add spec tests for this.  I gave an example here:

https://github.com/oilshell/oil/commit/7491ef8bc2627ac6dd8aacc4f9996fa60ec60e86

The simplest thing to do would just be to copy it for `jobs -p`, but it would be nice if we can somehow do a more thorough test, e.g. `egrep '[0-9]+'` or something ...

glyh:
I guess this will fail, because I did not distinguish whether the command is used in a pipe.

glyh:
BTW, both of the two ways to run commands in background is failing on my computer:
```bash
oil$ jobs -p
oil$ sleep 1231231
^Z
[PID 185568] Stopped
  sleep 1231231
  ^~~~~
[ interactive ]:2: errexit PID 185564: Command failed with status -1
oil$ sleep 0.1 &
  sleep 0.1 &
            ^
[ interactive ]:3: TODO: Parse Redirect
```

andychu:
- Hm I filed #1101 for the first issue -- thanks for the report!  I want to fix this
- As for the TODO: parse direct, it's possible we should take that out
  - But `fork { sleep 1 }` is the Oil way of doing it


Let me know if you have more issues

glyh:
OK, seems like the failing of case 2 & 8 for background spec has none to do with this PR, because they are already failing in the master branch. The new testcase is working.

andychu:
Did you add a new test for `jobs -p`?  I don't see it in github

glyh:
Added one trivial test.

andychu:
I pulled this and tested it out.  One difference I see is when pipelines are jobs - it prints every PID, while other shells print the leader.

Try this:

```
sleep 10 | cat &
jobs -p
```

It will print 2 PIDs instead of 1.

I think this can be tested with the spec tests.  I should have added a pipeline to the test case I added.

----

I also think the code should probably be refactored to be something like:

```
if arg.p:
  self.job_state.PrintPids()
else:
  self.job_state.PrintJobs()
```

but that can be done separately ...

There will be a little duplication but I think it's cleaner.

andychu:
Hm I guess `jobs` without `-p` already has this issue ... it prints 3 lines instead of 2.  But that is obviously incomplete while `jobs -p` could be complete.  It would be nice if you can fix `jobs -p` but it can also go in another commit.  This is good progress!

andychu:
Oh this is the difference between `jobs -l` and `jobs`.  We only have `jobs -l` now.

So basically this implementation of `jobs -p` is behaving more like `-l`, and it should be a short format

glyh:
I've find different behaviors in other shells:

fish:
```
$ sleep 10 | cat & 
$ jobs
Job     Group   CPU     State   Command
1       549443  0%      running sleep 10 | cat &
$ jobs -p 
Process
549443
549444
$ jobs -p | wc
      2       2      14
```
Which is wierd.

zsh: 
```
$ sleep 10 | cat & 
[1] 550819 550820
$ jobs
[1]  + running    sleep 10 | cat
$ jobs -p 
[1]  + 550819 running    sleep 10 | 
              running    cat
$ jobs -p | wc
      2       9      66
```

bash:
```
$ sleep 10 | cat & 
[1] 551104
$ jobs
[1]+  Running                 sleep 10 | cat &
$ jobs -p 
551103
$ jobs -p | wc
      1       1       7
```

glyh:
I found they basically define `-p` as different commands:
Fish:
       • -p  or  --pid prints the process ID for each process
         in all jobs.
Bash:
       -p        Display only the process IDs for the process
                 group leaders of the selected jobs.
Zsh:
       -p        Display only the process IDs for the process
                 group leaders of the selected jobs.

glyh:
I'm now worried about the standard portable shell across different UNIX-likes may just not be a thing...... MacOS use zsh, GNU/Linux use bash, Andriod use mksh, Alpine Linux use BusyBox's ash, OpenBSD use pdksh, FreeBSD use tcsh. And they are incompatible ...... 

But I really think the portability of shell script itself should be a thing(And arguably its best selling point for the unix community)

andychu:
Also you can search for "jobs posix" in Google and it brings up this page:

https://pubs.opengroup.org/onlinepubs/007904875/utilities/jobs.html

> Display only the process IDs for the process group leaders of the selected jobs.

So that is clear.  What the code is doing now is displaying the PIDs for all processes in a pipeline.

The process group leader for a pipeline is the first process.  In other words the PGID of the process group is the PID of the leader.

The code is already checking `i == 0` for the leader so this shouldn't be too hard: https://github.com/oilshell/oil/blob/master/core/process.py#L977

As mentioned I would just make another function for -p, with a slight amount of duplication.  If you don't have time, I can do it

glyh:
Fixed this specific bug, Dash seems to be bugged.

andychu:
Great thank you!  I merged and will probably clean up the tests a little.

I noticed dash only prints in the interactive case too.  I prefer to avoid that if possible.  For POSIX functionality we take a survey of shells with the spec tests, and  that usually tells us the desired behavior

PR #1141, opened 2022-05-24
add task/usage table

bar-g:


bar-g:
Please adjust the `inherit_errexit`, I don't know it's supposed to be used.

andychu:
Ah so our documentation doesn't handle Markdown extensions like tables -- e.g. try `build/doc.sh split-and-render doc/foo.md`. 

I think the table should go on the wiki and then in the "Getting Started" we can link to all the relevant docs:


https://github.com/oilshell/oil/wiki/OSH-versus-Oil

https://github.com/oilshell/oil/wiki/Gradually-Upgrading-Shell-to-Oil


Comments

- `inherit_errexit` is on in `oil:basic`.  Try `osh -c 'shopt -s oil:basic; shopt -p oil:basic'`.  It isn't really relevant here -- it should silently work.
- OSH_HIJACK_SHEBANG is more like a debugging feature.  End users shouldn't use it.  The real solution is to change your shebang.

bar-g:
> Ah so our documentation doesn't handle Markdown extensions like tables -- e.g. try `build/doc.sh split-and-render doc/foo.md`.

So that's probably why the docs are rather poorly formatted with "no: ... yes: ..." where a table comparisons could work much better.

I just made my edit through the github webinterface, though. So, no doc builds.

Unfortunately, it seems the wiki doesn't support the formatting either.
https://github.com/oilshell/oil/wiki/Oilshell-Usage-(The-table-missing-in-getting-started.md)

Issue #636, opened 2020-03-08
Oil expression evaluator shouldn't be "metacircular"


andychu:
We're no longer depending on the CPython runtime.

Instead of `PyObject` we're going to use `runtime_asdl::value_t`

- [ ] Tuples aren't useful without funcs, without tuple keys.  Defer
- [ ] defer: list comprehensions, shouldn't parse
- [x] get rid of lambdas, shouldn't parse.  Maybe do it in `expr_to_ast.py`
  - expression language parser is shared with Tea language, but the evaluator isn't!
  - our evaluator has implicit string -> number coercion rules.  The Tea one will have runtime errors to simulate compile time.
    - although maybe this is an execution flag?  `shopt`
  - maybe: C++ function overloading at runtime?
- [x] defer:  heterogeneous list?   `%(one two)` is alias for `['one', 'two']` ?
- [x] get rid of `Array[T]()`
  - maybe that is better as `%Int[...]`, but in either case we don't have these typed arrays

Fully Recursive changes:

- [ ] `Array(value_t* elems)`
- [ ] `Dict(dict[str, value_t] items)`

Then we can do the List change, and maybe List comprehensions?

Things to ADD:

- [x] Block expressions?  -- now `var b = ^(echo one; echo two)`

andychu:
- Hm yeah we should probably get rid of list comprehensions and generator expressions

andychu:
These are related to how we're reusing CPython:

- #1092
- #954 
- #840 

And another one, due to Python's dict being converted to `value.AssocArray`

```
$ bin/oil -c 'var d ={name: [1,2,3]}; echo ${d[@]}'
Traceback (most recent call last):
  File "/home/andy/git/oilshell/oil/bin/oil.py", line 360, in <module>
    sys.exit(main(sys.argv))
  File "/home/andy/git/oilshell/oil/bin/oil.py", line 303, in main
    return AppBundleMain(argv)
  File "/home/andy/git/oilshell/oil/bin/oil.py", line 280, in AppBundleMain
    loader, line_input)
  File "/home/andy/git/oilshell/oil/core/shell.py", line 683, in Main
    cmd_flags=cmd_eval.IsMainProgram)
  File "/home/andy/git/oilshell/oil/core/main_loop.py", line 332, in Batch
    is_return, is_fatal = cmd_ev.ExecuteAndCatch(node, cmd_flags=cmd_flags)
[...]

PR #1145, opened 2022-05-25
doc/deprecations: sort out old/intermingled parts into revised structure

bar-g:


andychu:
Hm another way to organize this doc is around "Oil Basic" here:

https://www.oilshell.org/release/latest/doc/oil-help-topics.html

## Two Deprecations That Are Both Pretty Unlikely

- `parse_paren` --- `if ( )` and while `( )` no longer means subshell
- `parse_at`  --- `@()` no longer means extended glob in pure OSH; use `,()` instead.  (But this is legacy -- it's not allowed in `simple_word_eval`  !)

Oh I guess `simple_word_eval` means that extended globs are only allowed in patterns, not command lines!   So maybe this is fairly big, but not sure.  This just happened in November.

## MInor Breakage 

- `parse_raw_string` - `r'foo'` means `foo`, not  `rfoo`

## Minor Quoting Issues

- `parse_brace` means `cd '{'`
- `parse_at` means `'@quoted'` if that's what you mean
- `=word` would have to be `'=word'`, which is an unconditional change, because the `=` operator is unconditionally on. 
- Could be worth mentioning: `'var' 'const'` 'proc'   if you need those as command names  (for legacy reasons)

------------

Overall I'm pretty impressed with how little breakage there is!

And maybe "shell language deprecations" is the wrong title for this doc.  I think it's more like "Incompatibilities When Upgrading to Oil".  (as opposed to writing a **new** Oil script.)

andychu:
I would say the only thing that's really "deprecated" is extended glob in command lines because I didn't implement it under `simple_word_eval`.    (Extended globs in `[[` and `case` are supported.)

So arguably almost nothing is deprecated, but there are a half a dozen or so minor incompatibilities.  The biggest one is `if ()` and `while ( )`, and even those are quite uncommon in real shell code.  There is almost no reason to ever write those.

bar-g:
Sure, sounds all very well. I don't know any details to add obviously, can merely tell you later what I didn't understand, just've put in what I picked up to be worth to sort out. Just tried to give more meaning to the headings and add the soft/hard differentiation, to show even more that there isn't much breakage to be afraid of.

andychu:
Yes well thanks for pointing this out.  The feedback is useful

I think just limiting this PR to the minor edits I suggested could be good, or we could just wait until a rewrite (no promises on when that will happen -- I think the comment above is a pretty good start.)

It is also highly related to https://github.com/oilshell/oil/wiki/Gradually-Upgrading-Shell-to-Oil

andychu:
Another thing that crossed my mind: We should emphasize the benefits of upgrading!

1. Simple word evaluation / fixing quoting hell
2. Reliable error handling -- all the stuff in the error handling doc

The other 2 of "4 features that justify a new Unix shell" are:

3. static parsing for better error messages and tools
4. QSN -- reliable processing of untrusted data

Those two are always on, whereas the first two are on only with `oil:basic/upgrade`.

andychu:
I put the "flip side" on the wiki: https://github.com/oilshell/oil/wiki/Why-Use-Oil%3F

It is hard to explain because there are many different audiences for shell.

Honestly I feel like  on the home page there should be a big question:

- Do you like shell compatibility?  Why upgrade?
  - After all the benefits --   list these minor upgrade incompatibilites
- Do you want a brand new shell?  Why use Oil?

I get a lot of feedback from both audiences!

bar-g:
Yes, but I don't think listing and repeating all the related stuff everywhere helps. At some point you will need to add structure and mainly add, move to, or modify things in that tree (table of contents). List all the benefits in one place, but definitely not in the deprecations document.  Otherwise it quickly becomes like navigating blog spam. I don't think what you'd need, and want above, is doable by small external edits from another person.

> "shell language deprecations" is the wrong title for this doc. I think it's more like "Incompatibilities When Upgrading to Oil". (as opposed to writing a new Oil script.)

Though, that's like the definition for deprecations. If there are only very few, the page should just be really short or the short list even just made part of say a broader usage page.

bar-g:
Oh, and I did't read your last comment before sending my last one.

bar-g:
Ok, now I looked into your linked wiki page.

Personally, I find those bullet points much better to read than its blog post publication it refers to.
The blog post seems to differ or incomplete and missing the structure and formatting.

So I think that you should very perfectly publish such as basic documentation that uses bullet point lists to give basic overviews. They can be read by users of any interest. Just make sure to keep to these doc texts more focused on things and not any persons's perspectives. Those you may later address separately, with a structured doc already out and maintained.

Here is what that page made me think could be a useful structure to gather and present all features, while still allowing them to be graspable by different audiences and according to different interests:



```
# Oilshell's feature comparison

Oilshell compared with its legacy shells

  osh (POSIX and bash compatible mode)
     starts fewer processes
     ...

  oil:upgrade (feature enabled mode)
    supports native, typed expressions and real data structures* (not broken as in bash): const, var, setvar
    ... : proc
    cleaned up shell I/O idioms*: read --line, read --all, write)
    ensures reliable errexit (problems solved, enabled by defalt)
    provides error handling*: try ... case _status {...}
  
      * actually already active in osh mode (may only break in other shells without support)
  
  oil:all (clean slate and final migration target)
    only disables...
    only deprecates...
    


Oilshell compared with python
  
  oil
    has more powerful and simple loop constructs: for i, key, value in (dict) {...}


Oilshell compared with javascript

  oil
  
```

bar-g:
The idea being to present a feature in the comparison that most related to similarity and feature improvement.

bar-g:
I revised the example for you. Note that I think the indentation is important, so even plain text would be better for this type of doc than loosing the structure in a html stream.

bar-g:
Note, I think headers like "five or six unlikely" do not create a helpful category structure, also major/minor breakage is rather in the eye of the beholder. Therefore the layout above as suggested.

andychu (commented):
Oh looks like I never published these pending comments from yesterday

andychu:
Also what is the purpose of this change?  "document structure" is not a good description

I would like to see the preview, but on a high level I think the structure is "OK" -- the real problem with this doc is that it is 2 years old and needs to be updated to reflect all the changes we've settled on since then

andychu:
I just read over this doc on my iPad with fresh eyes ... It seems OK,  the only issues I see are:

1. "Right now" title should be "Upgrading to Oil" 
2. Minor Breakages could be at the top level, maybe called "Extra Quotes May Be Required" -- it should also document `cd \{`
3. Some missing info mentioned, like `parse_raw_string`

andychu:
I agree that we need a concise overview of features.  That is a good start, and it is pretty similar to what I drafted here:

https://github.com/oilshell/oil/wiki/Why-Use-Oil%3F

I added your comment at the top.

The reason it's all in blog format is somewhat on purpose -- because the Oil language is NOT done yet!

The OSH language is largely done, but Oil is not.

In particular I'm running into issues around #636.

I think around the time we fix #636 then we can advertise it more.  Also I need someone to help on the C++ translation.

Actually are you in the EU?  If you can help find a C++ / Python / compiler  engineer, that would be one of the best ways to move things forward

https://github.com/oilshell/oil/wiki/Compiler-Engineer-Job

We have a 50K euro grant to pay somebody!

bar-g:
I adapted the changes according to what I understood from your comments. But left out putting in things new to me which I don't really understand, e.g. no idea what that `cd` issue you mentioned is about.

Maybe see if it fits.

bar-g:
Concerning a consice overview of features, there is the same problem that I don't know enough to write it out myself. Would it help you if I put my draft from above on a wiki page with a code block or into a plain-text file?

But skimming through the reference I asked myself, does it get generated from the code?

Because if yes, then you could maybe also generate some kind of "positional fields" to each feature, based on the definition/occurrence? of its code (osh/oil:upgrade/oil:all). Then, together with some annotations/comments to denote origins/comparision of a feature (python/javascript/...), at least part of such a sorted feature overview could maybe be generated from the short one-liner descriptions that appear in the help-doc together with their "positional fields".

bar-g:
Just moved the plain-text feature overview to https://github.com/oilshell/oil/wiki/Why-Use-Oil%3F , to have an editable version.

bar-g:
> I would say the only thing that's really "deprecated" is extended glob in command lines because I didn't implement it under simple_word_eval.

Then you have removed/disabled or broken support.

https://en.wikipedia.org/wiki/Deprecation
The subshell syntax with parens is deprecated.

See preview:
https://github.com/oilshell/oil/blob/918167e9316224683ad8782f752abc6fbd6045cc/doc/deprecations.md

andychu:
I changed the title and rewrote this doc:

https://www.oilshell.org/preview/doc/upgrade-breakage.html

Let me know if this covers what you were thinking!  Thanks

bar-g:
> Deprecated -> Extendend globs

I believe that's not correct. Aren't extendend globs replaced or substituted, thus the old syntax disabled/unsupported?

I think the subshells outside of conditionals are actually deprecated (discouraged).

Move both of those up to their better corresponding section?

> Gated by Options

This is a rather general thing, for the clarity of documentation, maybe try to avoid all slang in favor of precise languange, and it's appreciated to have a coherent line of thought.

Rather than "this means this and that", for example, something like "Now extended globs @() notation no longer works, but they are still available using ,() notation."

bar-g:
Also move the new stuff from the summery to the proper section.

andychu:
OK I think you want it to make a distinction between "deprecated" and "unsupported".

But I think extended globs are "deprecated", because they're supported in `bin/osh`!

It think this is mostly a debate of the meaning of words ... 

Likewise maybe "Gated by" is an American thing ... I don't think it's "slang", but it could be different in other countries.  I will think about some more straightforward words

Thanks for the feedback

andychu:
Updated it, thanks:

https://www.oilshell.org/preview/doc/upgrade-breakage.html

https://github.com/oilshell/oil/commit/5da011e388907ce52e394f0db03295cf75dc864e

bar-g:
> distinction between "deprecated" and "unsupported"

sure: https://en.wikipedia.org/wiki/Deprecation
Both seem wrong to me:

If you call things that work in osh (and other shells) "deprecated", even though they must work there for compatibility?
And if you call things not working with oil:upgrade and oil:all "deprecated", even thought they are disabled in oil?

Just let the doc list all things disabled and deprecated **after enabling** oil:upgrade and you're fine. Maybe only use "discouraged"?
(But don't introduce `$(( x + 1 )), (( i++ )), [[ $s =~ $pat ]], and ${s%%prefix}` deprecations/discouragements only in a (therefore wrong) "summary".)

> With oil:upgrade Options

Does not seem to correspond to the code. In oil:all group maybe? parse_equals->hay?

bar-g:
if ( ) and while ( ) take expressions, not subshells
if ( ) and while ( ) take expressions, not subshell commands

bar-g:
r'c:\Users\' is a raw string
r'c:\Users\' is a raw string, not string join

andychu:
OK thanks, the heading changes are good.

https://www.oilshell.org/preview/doc/upgrade-breakage.html

I changed it to "Unsupported", since "deprecated" can imply the support will go away.   And extended globs will always be supported in `bin/osh`.

bar-g:
> I changed it to "Unsupported"

allright, now a doc structure starts appearing


More headings:

Extended Globs in Word Evaluation
Extended Globs in Word Evaluation, obsolete mini language

Or maybe use parens for the reasons everywhere:
Extended Globs in Word Evaluation (obsolete mini language)

##More Quotes May Be Needed
##More strings to quote (avoid parsing/errors)

###new keywords
###'@foo' (parse_at)  or parse_at_all?
###'cd '{'' (parse_brace)
###command '=' and '=x' (parse_equals)
* as it's disallowed anyway, what about using `=` as pretty print command instead of `_`, or discard (readable)

Still think it would be better to add another second level header "Discouraged Syntax", and below that listing the things from the "summary", at least as long as you don't have a separate page. But then maybe just rename the page to upgrade-changes? Oil is really not requiring that much changes, thanks to ...you.

bar-g:
Maybe:

Discouraged Syntax (linting notices)

bar-g:
Oh, `=` actually does work as pretty printer, already. :-) Definitely mention that nice side effect and justification.

Another thing that was in the old version is the deprecation of (...) subshells outside of the "changed syntax" in conditionals in idiomatic oil.

That could fit nicely into the additional, second-level section "Discouraged Syntax (linting notices, clean code)". Then it's getting a really nice `upgrade-changes.html`.

andychu:
Thanks for the feedback, I think it's OK now ... I don't want more headings.  We want to avoid drowning in details that will rarely happen :)

bar-g:
It's not about more headings at all, though. It's merely about changing the headings to use the two levels in a meaningful way, so they directly describe their section's content and the table of content becomes useful.

bar-g:
The ToC is the essence for the reader. Gold to grasp things, quick and succinct and to go in to read only where needed.

"More Quotes May Be Needed", "With oil:upgrade Options", "Unconditionally" don't tell much do the reader, so the reader must read details or just leaves it beeing (not a good doc).

Issue #1038, opened 2021-12-06
Contributors should be able to publish CI results


andychu:
Right now it requires an SSH key.

I think we need a simple CGI upload program that accepts 3 files: JSON, TSV, and ZIP.  (honestly PHP is the best at this for multi-file uploads).

And then it will call the Python program to rewrite the index. 

This will also solve #1015 

We have to be careful about the atomic rewrite of the index.

andychu:
We should also rename it to `build.oilshell.org` when we do this.  Then we can be securely "multi-tenant" without multiple subdomains

Actually it is bad if anyone can publish HTML to our subdomain ... so we still need some kind of auth.  I think a manual password file is OK.

Or maybe we can authenticate with github accounts somehow?

andychu:
Need four passwords with access to publish into 4 directories:

1. sourcehut/
1. github/
1. circle-ci/
1. cirrus-ci/

And then rewrite the dashboard in that directory.

I guess that means contributors can get the password?  As long as they can only add, it's mostly OK

andychu:
Should we have a "docs" task?

- See #1151 

It would be nice if it could just run alone ... without the `cpp` and everything

andychu:
We can simply add contributors to the repo, but I'd like them to be able to do so with less permissions ... I guess there should be a separate auth secret  for travis-ci.oilshell.org ?

- They don't need to be able to SSH and write files.  The `/status-api/github` thing
- But they should be able to publish JSON, TSV, and .wwz, and then rewrite the jobs index.
  - So I think this could go in some sort of PHP script that does auth and invokes shell ...
  - or really this would be a great use case for Oil ... maybe we should be ambitious and use a build of Oil

andychu:
This is a little messed up now, we have to add people explicitly as contributors

It would be nice to do this for PRs from anyone

andychu:
https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Contributors.20should.20be.20able.20to.20push.20to.20CI

PR #1151, opened 2022-05-30
doc/getting-started: basic sort, add usage table

bar-g:
Fixes: https://github.com/oilshell/oil/issues/1142

I now did everything from the things you mentioned in the issue as far as I could through the web-interface. In particular the markdown-within-html-table part.

Please test and adjust with your own site building toolchain.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -5,6 +5,9 @@ default_highlighter: oil-sh
 Getting Started
 ===============
 
+The [releases page](https://www.oilshell.org/releases.html) links to downloads
+and documentation (incl. this page) for every release.
+
 There are many ways to use Oil!
 
 - You can use it *interactively*, or you can write "shell scripts" with it.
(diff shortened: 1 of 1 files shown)


andychu:
Hm in reading over the doc, I agree  there is something that could be improved here....

But yeah I would lean toward making it `doc/usage-table.md` as mentioned in the bug.

Or maybe I would change my mind if I could see the output.  As mentioned I think we need a way for you to upload a preview

I do that here https://www.oilshell.org/preview/doc/error-handling.html

Another way to do that is to have a continuous build task to do it

#1038

bar-g:
Concerning the re-arranged text. Have you tried the "view file" under "files changed"?
https://github.com/oilshell/oil/blob/be671fe85f92bd594011b79e65639cacb1d52ceb/doc/getting-started.md
I think the CommonMark rendering would still need to be tested separately, though.

bar-g:
The link I posted to the preview seems static and not showing the last revision. But the latest previews are is in the tree dot (...) menu or the changed files.

bar-g:
> how does that help?

Moving install details to where they belong will move the usage info nearer to the top.

What's the problem with this preview?
https://github.com/oilshell/oil/blob/08c1b92f85bfbd509bd06963e31e2085f4a30bc7/doc/getting-started.md

Doesn't it provide you an overview about he changes, that you can compare?
If it's about testing in your toolchain, can't you paste the markdown in html-table source in there to check its rendering?

bar-g:
> 1.  It has to be tested and previewed
> 2.  The prose was also awkward / lengthy

Have you tried to open the online-editor? I checked "maintainers can edit", it might look better as the diff view.
https://github.com/bar-g/oil/edit/patch-2/doc/getting-started.md?pr=%2Foilshell%2Foil%2Fpull%2F1151

Unfortunately I'm not able to help with 1., about 2. I have tried to improve on what you already commented, would need new ideas/pointers.

PR #1159, opened 2022-06-08
Enforce the order of proc params. Closes #1023

### LINKED ISSUE #1023: Enforce the order of proc params
Almost done, but we should probably enforce the order more exactly:

    untyped* rest? typed*

_Originally posted by @andychu in https://github.com/oilshell/oil/issues/984#issuecomment-979675436_

jcob-sikorski:


andychu:
Can you add some of the tests here and see if they do what we want?

https://github.com/oilshell/oil/issues/1023#issuecomment-1130898600

This change should reduce the amount of code that is valid, i.e. it's extra validation

andychu:
Yeah also I believe there should be at least one more error in the code, which I don't see?

For example say you match `untyped* rest? typed*` (which can be matched by zero params) -- and then you have MORE PARAMS.  Those additional params are errors.

Does the code handle that?

PR #1175, opened 2022-06-24
Jesse/bugfix/warnings in const generated code

jjbandit:
@andychu I'm unfamiliar with how the `Undefined` values are used outside of what I've done here, though I did grep for them and thought this transformation should be fine.  At least we won't get garbage from those functions anymore on invalid inputs.

Other than those I shouldn't have changed functionality anywhere .. just pacified the compiler.

LMK if I should tweak anything

### DIFF
--- a/cpp/libc.cc
+++ b/cpp/libc.cc
@@ -82,7 +82,7 @@ List<Str*>* regex_match(Str* pattern, Str* str) {
   int match;
   const char* s0 = str0.Get();
   regmatch_t* pmatch = (regmatch_t*)malloc(sizeof(regmatch_t) * outlen);
-  if (match = (regexec(&pat, s0, outlen, pmatch, 0) == 0)) {
+  if ( (match = regexec(&pat, s0, outlen, pmatch, 0) == 0) ) {
     int i;
     for (i = 0; i < outlen; i++) {
       int len = pmatch[i].rm_eo - pmatch[i].rm_so;
--- a/cpp/libc.cc
+++ b/cpp/libc.cc
@@ -79,10 +79,10 @@ List<Str*>* regex_match(Str* pattern, Str* str) {
 
   int outlen = pat.re_nsub + 1;  // number of captures
 
-  int match;
   const char* s0 = str0.Get();
   regmatch_t* pmatch = (regmatch_t*)malloc(sizeof(regmatch_t) * outlen);
-  if ( (match = (regexec(&pat, s0, outlen, pmatch, 0) == 0)) ) {
+  int match = regexec(&pat, s0, outlen, pmatch, 0) == 0;
+  if (match) {
     int i;
     for (i = 0; i < outlen; i++) {
       int len = pmatch[i].rm_eo - pmatch[i].rm_so;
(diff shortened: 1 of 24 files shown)


cpp/libc.cc
```diff
@@ -82,7 +82,7 @@ List<Str*>* regex_match(Str* pattern, Str* str) {
   int match;
   const char* s0 = str0.Get();
   regmatch_t* pmatch = (regmatch_t*)malloc(sizeof(regmatch_t) * outlen);
-  if (match = (regexec(&pat, s0, outlen, pmatch, 0) == 0)) {
+  if ( (match = (regexec(&pat, s0, outlen, pmatch, 0) == 0)) ) {
```
andychu, line 85: Can you make this a separate assignment and then `if (match == 0)` ?  See `native/libc.c`, I did that there to silence this warning

(that is the Python version of this function, which is unfortunately duplicated)

frontend/consts_gen.py
```diff
@@ -375,6 +376,7 @@ def out(fmt, *args):
         out('  case %s: return %s;' % (a, b))
       out("""\
   }
+  return redir_arg_type_e::Undefined;
```
andychu, line 379: Actually these can all be assert(0) -- I tried and that shuts up the warning about unreachable.   

So then we don't need new undefined values

opy/_regtest/src/_devbuild/gen/types_asdl.py
```diff
@@ -14,6 +14,7 @@ class bool_arg_type_e(py_meta.SimpleObj):
 class redir_arg_type_e(py_meta.SimpleObj):
   ASDL_TYPE = TYPE_LOOKUP.ByTypeName('redir_arg_type')
 
+redir_arg_type_e.Unknown = redir_arg_type_e(0, 'Unknown')
```
andychu, line 17: Oh please revert these two opy/_regtest files ... That is an old snapshot of the code that doesn't need to be kept up to date.  It's only for testing the bytecode compiler (which we won't need after translation is done)

andychu (commented):
OK thanks for cleaning this up

jjbandit:
Alright cleaned up the stuff requested.

I also added a couple macros that are slightly more descriptive than `assert(0); `NotImplemented()` and `InvalidCodePath()`.

I like being able to distinguish between `assert(0)`s that are intentionally there such that they catch bugs (`InvalidCodePath`), and those that are there temporarily while catching up on implementation (`NotImplemented`).

I also typically have a switch (maybe Oil already has this) that strips out unnecessary debug/internal-only code for external (read: production) builds.  I put a comment by the `NotImplemented` macro to potentially beef up its utility in the future.  The same can be done on `InvalidCodePath` to intentionally cause a crash if security is a concern.

Happy to discuss whether you think these are appropriate for Oil.

andychu:
OK this looks good, except please revert the `opy/_regtest` file ... it is causing `other-tests` to fail.  (It does a checksum of the bytecode)

I have some `find` aliases to omit all the `_*` dirs, which might help

andychu:
Oh just to clarify that `opy/_regtest` file is never run and the code is never used -- I just copied some of our own code as a test for the old bytecode compiler!

jjbandit:
> OK this looks good, except please revert the `opy/_regtest` file ... it is causing `other-tests` to fail. (It does a checksum of the bytecode)
> 
> I have some `find` aliases to omit all the `_*` dirs, which might help

Oops .. I thought I reverted them all but missed that one.  Weird that `__attribute__((unused))` would change the bytecode, but whatever.  All good now.

PR #1180, opened 2022-06-29
Change out some assert(0) for InvalidCodePath() and add comments

jjbandit:
Just some random comments.  Any thoughts on the following @andychu ?

### DIFF
--- a/build/native.sh
+++ b/build/native.sh
@@ -17,7 +17,18 @@ source build/common.sh  # CLANGXX
 compile-quickly() {
   ### For the fast possible development experience
 
-  ninja _bin/clang-dbg/osh_eval
+  if test -f "$CLANGXX"; then
+    ninja _bin/clang-dbg/osh_eval
+  else
+    echo ""
+    echo " Error: Unable to locate clang at ($CLANGXX)"
+    echo ""
+    echo "        To install clang at the specified path, run the following commands:"
+    echo ""
+    echo "        soil/deps-binary.sh download-clang"
+    echo "        soil/deps-binary.sh extract-clang"
+    echo ""
+  fi
 }
 
 compiler-trace-build() {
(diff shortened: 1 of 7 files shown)


mycpp/gc_heap.cc
```diff
@@ -173,7 +173,7 @@ void Heap::Collect() {
           case Tag::Scanned:
             break;
           default:
-            assert(0);
+            assert(0); // NOTE(Jesse): Pretty sure this is InvalidCodePath();
```
andychu, line 176: I think this whole TODO_BUG block is hit by a bug in a test, probably mycpp/examples.

It's not enabled right now but we will probably hit it as we test the collector.

build/native.sh
```diff
@@ -17,7 +17,18 @@ source build/common.sh  # CLANGXX
...
+    ninja _bin/clang-dbg/osh_eval
+  else
+    echo ""
+    echo " Error: Unable to locate clang at ($CLANGXX)"
+    echo ""
+    echo "        To install clang at the specified path, run the following commands:"
+    echo ""
+    echo "        soil/deps-binary.sh download-clang"
+    echo "        soil/deps-binary.sh extract-clang"
+    echo ""
```
andychu, line 30: I tend to use a here doc to avoid all the quotes, but this is OK

(here docs mess up the indentation, so Oil has multi-line strings to give you the best of both worlds)

mycpp/mylib.h
```diff
@@ -387,6 +372,9 @@ class StrIter {
 };
 
 // TODO: Rewrite without vector<>, so we don't depend on libstdc++.
+//
+// TODO(Jesse): I like the sound of getting rid of std:vector
```
andychu, line 376: Yes the old `mylib.h` isn't garbage collected, so it just uses `std::vector`.  The new one will be garbage-collected so we can't use it.  That's in `gc_heap.h` / `mylib2.h`

andychu (approved):
OK seems good

mycpp/mylib.h
jjbandit: Ahhh, got it.

PR #1181, opened 2022-06-29
[cpp] Implement uname, pipe and Dict<K,V>::len()

jjbandit:
As far as I can tell these don't fix any tests, but these were the easiest `assert(0)` calls I could find to fix.

### DIFF
--- a/cpp/posix.h
+++ b/cpp/posix.h
@@ -4,6 +4,7 @@
 #define POSIX_H
 
 #include <unistd.h>
+#include <sys/utsname.h> // uname
 
 #include "mycpp/mylib.h"
 
@@ -56,7 +57,15 @@ inline Str* strerror(int err_num) {
 }
 
 inline Str* uname() {
-  assert(0);
+  Str *result = 0; // TODO(Jesse): Is it safe to return a 0 pointer here?
+
+  utsname un = {};
+  if (::uname(&un) == 0)
+  {
+    result = mylib::NewStr(un.sysname);
+  }
+
+  return result;
 }
 
 // TODO: write proper signatures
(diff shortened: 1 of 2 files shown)


andychu:
Hm I prefer if it's test driven, remember this bug has a lot of links: https://github.com/oilshell/oil/issues/1167

e.g. if you search for `assert` on this page, you'll see a whole bunch of `libc::realpath` and some others

https://www.oilshell.org/release/0.11.0/test/spec.wwz/cpp/builtins.html


----

Although I would have expected `uname()` to uncover a bug.  I looked into it and it looks like that function is unused!  

- I did `grep uname */*.py`, then found in `pyos.py` we're using in `OsType()` for the `$OSTYPE` variable
- Then I look at `cpp/core_pyos.h`, the corresponding C++.  And then the function returns TODO!  That should have been assert(0) probably.
  - so I imagine we need to delete uname and implement pyos.OsType().  
- if I do `grep OSTYPE  spec/*.sh` it reveals the test that should pass because of that


Not sure about Dict -- maybe that's because we're not actually using the length of a dict anywhere in the program ?

andychu:
Also remember once you see the `assert()` on the HTML report, you can do `test/spec.sh builtins -r 10 -p > out` to get a copy of it to debug, which should reproduce the assert

jjbandit:
Yeah after I did these ones I realized that they're not used anywhere.  My new strategy is to find the `assert(0)` calls that are actually used and fix those ones, which presumably would result in more tests passing.

jjbandit:
Going to close this PR -- I don't expect to merge these changes.

PR #1182, opened 2022-06-29
[cpp] Survey a fair number of assert(0) function calls

jjbandit:
Searched out a fair number of `assert(0)` functions and marked them in one of four ways:

1. Left it as `assert(0)` as I was unable to determine its purpose, or if the function is called.
2. Changed it to `InvalidCodePath()` if it's there to intentionally cause a crash.
3. Changed to `NotImplemented(); // Uncalled` if the function does not get called in the codebase.
4. Changed to `NotImplemented():` if the function gets called.

### DIFF
--- a/cpp/posix.h
+++ b/cpp/posix.h
@@ -62,11 +62,11 @@ inline Str* uname() {
 // TODO: write proper signatures
 // stat returns stat_result
 inline void stat() {
-  assert(0);
+  NotImplemented(); // Uncalled
 }
 
 inline void lstat() {
-  assert(0);
+  NotImplemented(); // Uncalled
 }
 
 inline Tuple2<int, int> pipe() {
@@ -84,12 +84,12 @@ inline int close(int fd) {
 }
 
 inline int putenv(Str* name, Str* value) {
-  assert(0);
+  NotImplemented();
 }
 
 // TODO: errors
 inline int chdir(Str* path) {
-  assert(0);
+  NotImplemented(); // Uncalled
 }
 
 inline int fork() {
--- a/cpp/qsn_qsn.h
+++ b/cpp/qsn_qsn.h
@@ -50,11 +50,11 @@ inline Str* UEscape(int codepoint) {
 }
 #else
 inline Str* XEscape(Str* ch) {
-  assert(0);
+  NotImplemented(); // Unused
 }
 
 inline Str* UEscape(int codepoint) {
-  assert(0);
+  NotImplemented(); // Unused
 }
 #endif
 
--- a/mycpp/demo/gc_heap.cc
+++ b/mycpp/demo/gc_heap.cc
@@ -276,6 +276,8 @@ class AtomEntry {
 class AtomTable {
  public:
   int Intern(Slice_2 slice) {
+    // TODO(Jesse): Does this get called?
+    // Is this NotImplemented() or InvalidCodePath() ??
     assert(0);
   }
 
--- a/mycpp/gc_heap.h
+++ b/mycpp/gc_heap.h
@@ -352,12 +352,16 @@ class Local {
 
   // IMPLICIT conversion.  No 'explicit'.
   Local(T* raw_pointer) : raw_pointer_(raw_pointer) {
+    // TODO(Jesse): Does this get called?
+    // Is this NotImplemented() or InvalidCodePath() ??
     assert(0);
     // gHeap.PushRoot(this);
   }
 
   // Copy constructor, e.g. f(mylocal) where f(Local<T> param);
   Local(const Local& other) : raw_pointer_(other.raw_pointer_) {
+    // TODO(Jesse): Does this get called?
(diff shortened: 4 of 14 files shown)


cpp/posix.h
```diff
@@ -84,12 +84,12 @@ inline int close(int fd) {
...
 
 inline int putenv(Str* name, Str* value) {
-  assert(0);
+  NotImplemented();
 }
 
 // TODO: errors
 inline int chdir(Str* path) {
-  assert(0);
+  NotImplemented(); // Uncalled
```
andychu, line 92: Hm OK I think this was changed to pyos.Chdir.  So it's probably best to just delete these functions.

cpp/qsn_qsn.h
```diff
@@ -50,11 +50,11 @@ inline Str* UEscape(int codepoint) {
 }
 #else
 inline Str* XEscape(Str* ch) {
-  assert(0);
+  NotImplemented(); // Unused
 }
 
 inline Str* UEscape(int codepoint) {
-  assert(0);
+  NotImplemented(); // Unused
```
andychu, line 57: I just added this as part of the GC translation, so that's why it's hidden behind !USING_OLD_QSN.

There is also USING_OLD_MYLIB.  So we need to work out a good scheme for porting to the GC runtime.

It is technically unused now but will be a different implementation for the GC.

The reason is that the existing functions just leak memory!

mycpp/demo/gc_heap.cc
```diff
@@ -276,6 +276,8 @@ class AtomEntry {
 class AtomTable {
```
andychu, line 276: Hm I think this whole thing is just unused.  

Ditto for the Local<T> and Param<T> thing.  We could either delete those or move them to an experimental file if it's confusing.

mycpp/gc_heap.h
```diff
@@ -633,11 +637,11 @@ class Str : public gc_heap::Obj {
   bool isupper();
 
   Str* upper() {
-    assert(0);
+    NotImplemented(); // Uncalled
   }
 
   Str* lower() {
-    assert(0);
+    NotImplemented(); // Uncalled
```
andychu, line 644: I think these must be called, if you 

`grep upper */*.py`

then it shows some usages. This is like `'mystr'.upper()` in Python

mycpp/gc_heap.h
```diff
@@ -1175,7 +1179,7 @@ class Dict : public gc_heap::Obj {
   V index_(K key) {
     int pos = position_of_key(key);
     if (pos == -1) {
-      assert(0); // NOTE(Jesse): Should we really crash if asking for a key not in a dict?
+      InvalidCodePath(); // NOTE(Jesse): Should we really crash if asking for a key not in a dict?
```
andychu, line 1182: I'm not sure exactly what we should do here -- I guess we do the minimal thing to make the shell work.  If the shell never uses it, then we never have to fix this :-P

mycpp/my_runtime.cc
```diff
@@ -232,7 +232,7 @@ Str* Str::strip() {
 
 // Used for CommandSub in osh/cmd_exec.py
 Str* Str::rstrip(Str* chars) {
-  assert(0);
+  NotImplemented(); // Uncalled
```
andychu, line 235: It is used, except the comment is now out of date.  The usage is in core/executor.py.

e.g. `grep rstrip */*.py`

andychu (commented):
Hm I think some of them are OK, but we should leave others as assert(0) if we don't know yet what to do.

mycpp/gc_heap.h
jjbandit: Agreed.  I left this comment in as more of a reminder for if we ever hit it.

mycpp/gc_heap.h
jjbandit: If I change those function names to `upper_` and `lower_` I don't get compile or link errors.

The ones in `mylib/mycpp.h` line 337 and 341 break when I rename them.

mycpp/demo/gc_heap.cc
jjbandit: I'm unbothered .. I got link errors when I checked if these things are called and didn't feel like digging deeper.  Might be worth getting rid of them if we're not planning on using them for anything.

cpp/qsn_qsn.h
jjbandit: Sounds good.  I assume for a fair number of the `Unused` tags this is the case

cpp/posix.h
jjbandit: "these functions" being all the ones in `posix.h` that are uncalled, or just `chdir` ?

jjbandit:
> Hm I think some of them are OK, but we should leave others as assert(0) if we don't know yet what to do.

Yeah the ones I didn't know about I left as-is.

PR #1183, opened 2022-06-29
[cpp] Add script to keep track of which spec-cpp tests pass/fail per-…

jjbandit:
So, this may be somewhat questionable/unconventional, but hear me out.

I want an easy way of tracking when changes I make have an effect registered by the test suite.

I run the test suite pretty frequently, but it's hard to discern when tests change state.  It's trivial to see if the total number of tests passing changed, but not individual tests.

So I thought to myself; I run `git diff` roughly a billion times per day.  What if `git diff` told me about when tests change state from failing to passing, or vice versa?

Enter `baseline.spec-cpp.sh`.  It simply writes a list of all passing tests to `baseline.spec-cpp.results` such that if that file's committed to version control, changes to the test suite will show up during `git diff`.

Admittedly, it's kind of a weird idea.  Not sure if it's good or bad; I've never heard of anyone doing this before.  The only downside I can think of is committing this 3k line file to the repo .. which frankly isn't that much of a downside.

Thoughts?

The major issue I would like to avoid is accidentally making 3 tests pass, and 1 test fail (or something of the like).  By looking strictly at the number of tests passing I wouldn't catch that sort of issue.


test/baseline.spec-cpp.sh
```diff
@@ -0,0 +1,20 @@
+#! /bin/bash
+
+
+#TODO(Jesse): Maybe make another results file for the failures?
+
+output_filename=./baseline.spec-cpp.results
+
+[ -f $output_filname ] && rm $output_filename
```
andychu, line 8: You should add our header to this script, also called "unofficial bash strict mode":

```
set -o errexit
set -o pipefail
set -o nounset
```

Because there is a typo in the "filname" variable and I think that's why the file has twice the number of reuslts I'd expect!

(Oil has this on by default, that was the subject of this blog post: https://www.oilshell.org/blog/2022/05/release-0.10.0.html)

test/baseline.spec-cpp.sh
```diff
@@ -0,0 +1,20 @@
...
+
+for filename in _tmp/spec/cpp/*.tsv; do
+
+  passes="$(cat "$filename" | grep "osh_\.cc\spass" | grep -P -o "\d+")"
+  for test_number in $passes; do
+    echo -n $test_number >> $output_filename
+    echo -n " "          >> $output_filename
+    echo -n $filename    >> $output_filename
+    echo ""              >> $output_filename
```
andychu, line 17: how about just:

echo "test_number $filename" >> $output_filename

andychu:
OK I see the motivation, because we have this in Python but not C++.

In `test/spec.sh`, you will see `--osh-failures-allowed N` passed to every test function.  So then that prevents the files from regressing.

But we don't have that for C++.  I was thinking about adding a similar mechanism but I haven't.  Although practically speaking I think implementing the assert(0) will not cause regressions.  We will need this at some point

test/baseline.spec-cpp.sh
jjbandit: Haha, yeah, not exactly sure why I took the long way round ..

test/baseline.spec-cpp.sh
jjbandit: Fixed

jjbandit:
Should be ready for :eyes: again @andychu

andychu:
Ah so the dev-minimal failure is because the shebang should start with `#!/usr/bin/env bash`

The reason is that it does break on other people's machines with the hard-coded path!

jjbandit:
> Ah so the dev-minimal failure is because the shebang should start with `#!/usr/bin/env bash`
> 
> The reason is that it does break on other people's machines with the hard-coded path!

Ahh, I knew about that one.  Fixed.

PR #1185, opened 2022-06-30
Jesse/feature/implement str upper str lower

jjbandit:
The commit log is a little weird here .. I'm not sure why Github can't figure out a bunch of those commits are in master already.

Anywho ..  I implemented Str::upper and Str::lower and they were hit by a few tests.

Note the `builtin-bracket` tests weren't in the `results` file because the test suite by default runs the `cxx-dbg`, but I wasn't compiling that anymore, I was just compiling the clang debug binary.  TL;DR I was running the test suite against a binary I built a few days ago.

Note on the above .. maybe we should run the test suite against the clang dbg binary?


mycpp/mylib.h
```diff
@@ -335,11 +347,34 @@ class Str : public gc_heap::Obj {
...
+  }
+
+  Str* transform(transform_op op)
+  {
+    char* new_buf = static_cast<char*>(malloc(len_));
+    for (int char_index = 0; char_index <= len_; ++char_index)
+    {
+      switch (op)
+      {
+        case transform_op::upper:
```
andychu, line 364: Hm this feels too abstract.  Why branch on every character?

I would just duplicate the `for` loop -- that code will be shorter and faster.  (I use `i` as the index pretty consistently)

Also I started moving stuff into the .cc files in general, although there is no hard rule.  That will avoid the forward declaration of NewStr.

andychu:
Oh please use `new Str()` for this PR.  I think we should only have `NewStr(int n)` and then `CopyStr(const char* n)`.

Because a `const char*` has to be copied in the GC regime -- we can't prepend a GC header to a `const char*` without copying.  So we should write our bindings so it is efficient in the new GC regime.

-------------

Let me think about the Clang issue ... I don't think it is that much faster with a parallel build, so maybe we should de-emphasize for simplicity.   If you are getting much faster compiles with it, let me know.

I don't want people to have to download Clang to contribute ... it should be an extra optimization.

mycpp/mylib.h
jjbandit: As far as being too abstract I'm somewhat indifferent.  Happy to duplicate the for loop if that's what you'd prefer.

Moving it into the .cc files is a good idea.  I'll do that tomorrow morning.

As far as speed goes, I'm pretty confidant this will run at very, very close to the same speed as independent loops.  Branch predictors on constant values are pretty good.  I might write a quick test program tomorrow to confirm/deny this suspicion.

jjbandit:
> Oh please use `new Str()` for this PR. I think we should only have `NewStr(int n)` and then `CopyStr(const char* n)`.
> 
> Because a `const char*` has to be copied in the GC regime -- we can't prepend a GC header to a `const char*` without copying. So we should write our bindings so it is efficient in the new GC regime.
> 
> Let me think about the Clang issue ... I don't think it is that much faster with a parallel build, so maybe we should de-emphasize for simplicity. If you are getting much faster compiles with it, let me know.
> 
> I don't want people to have to download Clang to contribute ... it should be an extra optimization.

Sounds good, I'll make that change in the AM.

jjbandit:
Took another pass at this @andychu.

I merged in your recent allocation API changes and took a stab at using BlankStr().  It's a bit clunky because the `data_` buffer is read-only, so we have to cast off the const-ness with a temp stack pointer.

Maybe we could add a function that returns a non-const pointer to the data buffer?  Open to suggestions.  Not a super urgent problem IMO.

mycpp/mylib.cc
```diff
@@ -208,6 +208,45 @@ Str* Str::join(List<Str*>* items) {
...
+    Str* self = this;
+    Str* result = 0;
+
+    /* StackRoots stroots = {&result, &self}; */
+
+    result = mylib::BlankStr(self->len_);
+
+    // TODO(Jesse): Have to cast off const-ness .. this is intolerable in the
+    // long-run.  How should we deal?
+    char *buffer = (char*)result->data_;
```
andychu, line 241: I added a `data()` method that does that, it's used in `core/qsn_qsn.h`

mycpp/mylib.cc
```diff
@@ -208,6 +208,45 @@ Str* Str::join(List<Str*>* items) {
...
+
+    for (int char_index = 0; char_index <= self->len_; ++char_index)
```
andychu, line 243: this should be `< self->len_`, ditto above

mycpp/mylib.cc
```diff
@@ -208,6 +208,45 @@ Str* Str::join(List<Str*>* items) {
...
+    {
+      buffer[char_index] = toupper(self->data_[char_index]);
+    }
+    return result;
+  }
+
+  Str* Str::lower() {
+    Str* self = this;
```
andychu, line 232: `mylib.cc` are the leaky implementations, so they don't need any StackRoots or `self` stuff.  This is just plain C++ code that leaks

I started the first GC bindings which is `core/qsn_qsn.h` -- there we have to take care.

And I started adding unit tests in `cpp/gc_bindings_test`, those are an absolute MUST for the GC code because of the potential for GC bugs.

So 

- the leaky code is `mylib.{h,cc}`
- the GC code is `{gc_heap,mylib2,my_runtime}.{h,cc}`

I'm thinking of renaming it to `mylib.leaky.h` and mylib.h` so it's clear.  We can name things so the goal is to get rid of everything "leaky"

mycpp/mylib.cc
jjbandit: Ahh, right.  I was thinking I was being clever by copying nulls over (since the buffers are defined to be `len_+1`) but we don't actually have to do that. BlankStr() allocates an extra byte for us, which is 0'd out, correct?

I guess doing that extra iteration wouldn't hurt anything, but it's kinda sketchy if we ever change the defined size of the buffers, right?

I'm mostly just talking out loud here to confirm/deny my understanding of the memory layout.

mycpp/mylib.cc
jjbandit: Yeah, I was thinking I'd just copy the implementation to the GC'd version of `Str` when the time came, and it would be nice to just uncomment a line to have it work .. but it does add a fair bit of noise.  I'll take that stuff out.

I do like the idea of unifying the naming a bit more, though I'm touching all of those files a lot right now so can we hold off on doing that till my current PRs are closed?  Wanna avoid a conflict-induced headache ;)

jjbandit:
Ready for :eyes: again @andychu

PR #1187, opened 2022-07-01
Jesse/bugfix/implement some not implemented functions

jjbandit:
@andychu Took a first pass at implementing a good number of the functions the test suite is hitting asserts in.

Still going to need to work on the `new Str()` calls a bit.. I don't think what I've done in this PR is going to work for the GC'd version.  Any suggestions there?  I don't know the API very well yet.

+10 test passes!  Progress :dancers:


cpp/core_pyos.h
```diff
@@ -32,20 +37,61 @@ class ReadError {
...
+    result = new Str(pw->pw_name);
+  }
+
+  return result;
 }
 
 inline Str* OsType() {
-  // uname()[0].lower()
-  return new Str("TODO");
+  Str *result = kEmptyString; // TODO(Jesse): Should we set this to "unknown" or something?
```
andychu, line 51: I think we should raise an `errno` exception in all these cases.  But I'm not sure if there is a good example of that in the C++ code.  I would make it `assert(0)` or `assert(!"TODO")` for now

cpp/core_pyos.h
```diff
@@ -32,20 +37,61 @@ class ReadError {
...
+  auto result = Tuple3<double, double, double>((double)t, (double)ru.ru_utime.tv_sec, (double)ru.ru_stime.tv_sec);
+  return result;
 }
 
 inline void PrintTimes() {
-  assert(0);
+  tms t;
+  if (times(&t) == -1)
+  {
+    printf("Error during time()");
```
andychu, line 75: ditto I think we need `errno` 

Actually now I see in `fcntl_.cc` it does `throw new IOError(errno)`, which I think does the right thing.

That exception should be caught an an error printed by the translated Python code

cpp/posix.h
```diff
@@ -84,7 +81,10 @@ inline int close(int fd) {
 }
 
 inline int putenv(Str* name, Str* value) {
-  NotImplemented();
+  int env_string_size = name->len_ + value->len_ + 1; // NOTE(Jesse): +1 for the '=' between them
+  char* env_string = reinterpret_cast<char*>(malloc(env_string_size));
```
andychu, line 85: I think `static_cast` works for malloc, because it's from void*

cpp/time_.h
```diff
@@ -9,21 +9,54 @@
...
 
-// TODO: Should these be bigger integers?
-inline int localtime(int ts) {
-  NotImplemented();
+// NOTE(Jesse): time_t is specified to be an arithmetic type by C++. On most
+// systems it's a 64-bit integer.  64 bits is used because 32 will overflow in
+// 2038.  Someone on a comittee somewhere thought of that when moving to 64-bit
+// architectures to prevent breaking ABI again; on 32-bit systems it's usually
+// 32 bits.  Point being, using anything but the time_t typedef here could
+// (unlikely, but possible) produce weird behavior.
```
andychu, line 25: Hmm then it seems like there will be a bug when the translated code consumes this ...

cpp/time_.h
```diff
@@ -9,21 +9,54 @@
...
-inline Str* strftime(Str* s, int ts) {
-  NotImplemented();
+inline Str* strftime(Str* s, time_t ts) {
+  Str* result = kEmptyString;
+
+  assert(s->IsNulTerminated());
+
+  tm* loc_time = ::localtime(&ts);
+
+  const int buf_size = 1024;
```
andychu, line 39: Hm this isn't always enough?

andychu (commented):
Great thanks, this mostly looks good, a few comments.

It's OK not to use any of the GC stuff since that's brand new... (and there are some issues).  This will help us when we port to the GC version.



At some point we need to run `test/lint.sh clang-format` on the code too, it is going to change all the `{` styles and so forth


jjbandit:
> Great thanks, this mostly looks good, a few comments.
> 
> It's OK not to use any of the GC stuff since that's brand new... (and there are some issues). This will help us when we port to the GC version.
> 
> At some point we need to run `test/lint.sh clang-format` on the code too, it is going to change all the `{` styles and so forth
> 
> https://github.com/oilshell/oil/wiki/Oil-Dev-Cheat-Sheet

Oh yeah, I have kind of a wonky style.  I'll run that script in a pre-commit hook or something so we don't get tons of style-related spew.

jjbandit:
Ready for :eyes: again @andychu

cpp/time_.h
andychu: Hm I was going to say that returning `""` isn't the right behavior when the strftime format string  is big, because it is exposed to the user through the `printf` builtin -- see `osh/builtin_printf`. 

So if they pass a long string then it looks like it will silently go to `""`.

But then I tested out bash and that's exactly what it does.

https://github.com/oilshell/oil/commit/737d716d91813a0d04cd5120b3ed89960fde4f87

I still kind of prefer a user visible error though.  Maybe I will merge this and see if we can use `e_die()`

PR #1192, opened 2022-07-02
Fix off by one

### LINKED ISSUE #1189: off by one error in strftime
I added this test for the fixed buffer size issue

https://github.com/oilshell/oil/commit/a380930cd9ea3fc9e4071d2822e8cdbf59213f30#diff-945146fdaafadf2255bb117643a1c713453e68fdf660a8801430bcf579278a87

And it's failing in an unexpected way -- all the string lengths are off by one.

http://travis-ci.oilshell.org/github-jobs/2022-07-02__03-36-17.wwz/_tmp/spec/cpp/builtin-printf.html

@jjbandit Can you take a look?

----

Also even though bash silently truncates, I think we should raise an error at `128` bytes.  Oil's philosophy is not to ignore errors like bash.

I think that can be done with the `e_die()` function

jjbandit:
Fixes #1189

### DIFF
--- a/cpp/time_.h
+++ b/cpp/time_.h
@@ -40,19 +40,7 @@ inline Str* strftime(Str* s, time_t ts) {
   char buffer[buf_size] = {};
 
   if (int size_of_result = strftime(buffer, buf_size, s->data_, loc_time)) {
-    // Adding 1 to size_of_result is safe, as long as the strftime function
-    // actually behaves as specified.
-    //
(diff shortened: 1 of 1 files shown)


jjbandit:
I've made this mistake a couple times now .. I keep forgetting that the string allocators take care of nulling the buffer that you pass them.

andychu:
Cool thanks!

Yeah when we unify the leaky and GC APIs this should be less annoying ... it's a little in flux but it will settle down

PR #1193, opened 2022-07-02
Jesse/bugfix/strftime static buffer

jjbandit:
So I'm not sure how much we care about this kind of thing, but this patch would improve the handling of strftime.

I noticed that the python version of osh handles very large buffers (the test suite times out after 3 megabytes, but it probably handles much larger ones than that).  The C++ version is faster, so in the same time we can handle roughly 10 megabytes.

I guess my philosophical question here is, for the C++ version, are we willing to accept more complexity in exchange for compatibility with the Python version, or would we rather keep things simple and be on-par or slightly better than BASH?

### DIFF
--- a/cpp/time_.h
+++ b/cpp/time_.h
@@ -6,6 +6,8 @@
 #include <time.h>
 
 #include "mycpp/mylib_leaky.h"
+#include "cpp/core_error.h"
+#include "cpp/core_pyerror.h"
 
 namespace time_ {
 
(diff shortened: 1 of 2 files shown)


jjbandit:
Oh, also, in plain English:

This patch dynamically re-allocates the buffer used to store the results of the strftime function call until it hits a ceiling of 10 megabytes.  It throws an error if it runs out of memory or hits the 10mb ceiling.

andychu:
Ahh so I thought we might need this originally, so then I tested bash and saw that it uses the static 128 buffer.  I think we should keep it simple for now.  

Although I would still like it to fail and not silently truncate like bash.  That should be one of the "language design principles"

Added it here: https://github.com/oilshell/oil/wiki/Language-Design-Principles

I will change the spec to ensure that

jjbandit:
> Ahh so I thought we might need this originally, so then I tested bash and saw that it uses the static 128 buffer. I think we should keep it simple for now.
> 
> Although I would still like it to fail and not silently truncate like bash. That should be one of the "language design principles"

Sounds good -- I'll close this PR and open a new one to just throw the error.

PR #1195, opened 2022-07-02
[cpp] Fix off-by-one in putenv

jjbandit:
Pretty straight-forward fix.  Putenv was truncating the value


cpp/posix.h
```diff
@@ -81,11 +81,15 @@ inline int close(int fd) {
...
-  return ::putenv(env_string);
+  // NOTE(Jesse): Technically this over-allocates by 1 byte, but by doing that
+  // we don't rely on the underlying Str* data_ buffer being defined as (Str::len_ + 1)
+  int env_string_size = name->len_ + value->len_ + 2;
+
+  Str* env_string = mylib::BlankStr(env_string_size);
+  char* env_string_buffer = env_string->data();
+
+  snprintf(env_string_buffer, env_string_size, "%s=%s", name->data_, value->data_);
+  return ::putenv(env_string->data());
```
andychu, line 92: Hm it occurs to me that this signature was always wrong.  It should be `void* putenv` and then `throw new IOError(errno)` as usual.

cpp/posix.h
```diff
@@ -81,11 +81,15 @@ inline int close(int fd) {
...
-  int env_string_size =
-      name->len_ + value->len_ + 1;  // NOTE(Jesse): +1 for the '=' between them
-  char* env_string = static_cast<char*>(malloc(env_string_size));
-  snprintf(env_string, env_string_size, "%s=%s", name->data_, value->data_);
-  return ::putenv(env_string);
+  // NOTE(Jesse): Technically this over-allocates by 1 byte, but by doing that
+  // we don't rely on the underlying Str* data_ buffer being defined as (Str::len_ + 1)
+  int env_string_size = name->len_ + value->len_ + 2;
+
+  Str* env_string = mylib::BlankStr(env_string_size);
```
andychu, line 88: It's in flux right now, but when we switch to GC, I think we will actually use malloc /free here.

Because then you don't need to worry about the buffer moving out from under `putenv()`!

That can only happen on allocations.  (And BTW I want to rename BlankStr to AllocStr to make that very clear.  Also AllocList, AllocDict, etc.)

cpp/posix.h
```diff
@@ -81,11 +81,15 @@ inline int close(int fd) {
 }
 
 inline int putenv(Str* name, Str* value) {
-  int env_string_size =
-      name->len_ + value->len_ + 1;  // NOTE(Jesse): +1 for the '=' between them
-  char* env_string = static_cast<char*>(malloc(env_string_size));
-  snprintf(env_string, env_string_size, "%s=%s", name->data_, value->data_);
-  return ::putenv(env_string);
+  // NOTE(Jesse): Technically this over-allocates by 1 byte, but by doing that
+  // we don't rely on the underlying Str* data_ buffer being defined as (Str::len_ + 1)
```
andychu, line 85: Hm I don't see what the problem was?  What was the bug that is fixed?

If so it's useful to demonstrate the bug in `cpp/leaky_bindings_test.cc`

I think you should be able to rely on the NUL terminator.  The length of the string is N and the length of the allocated buffer is N+1.  This should be true in both leaky and GC cases.

andychu:
Oh sorry I guess you said it was truncating ... it's hard for me to see too :-/   

Maybe the test can do `putenv()` and `getenv()` ??   This will help when we move to GC -- so we don't introduce similar bugs

cpp/posix.h
andychu: I think the comment is confusing me.  If it was truncating before, then why is this over-allocating by one byte?

Again I would favor the test rather than a comment

andychu:
Sorry one thing I realized is I moved around a lot of stuff in the last day or two related to the test.

- cpp/leaky_binding_test.cc is still in the same place, which is OK
- This commit moved the driver to `cpp/test.sh` -- https://github.com/oilshell/oil/commit/91793b673cfdeea8d881afce744d6a4bf29b6019
- and this commit moved the binary to `_bin/cxx-asan/cpp/` - https://github.com/oilshell/oil/commit/8ebd3a837feb786c0f870843569daef3aedef236

So hopefully that should be resolvable.

One important thing is that `greatest.h` accepts the `-t` for a test name substring to filter by.  That makes it faster to iterate on the test.

I added that here: https://github.com/oilshell/oil/wiki/Oil-Dev-Cheat-Sheet

jjbandit:
@andychu Bump.  Can we merge this or were you waiting for me to do something?

andychu:
Hmm so how about deleting the "overallocating" comment?  (I sent two comments at once and you might have missed the first)

Isn't it exactly right now because there was a bug?  

Sometimes I write things like `key_len + 1 + val_len + 1` to emphasize the 1 for the `=` and the 1 for the NUL.

Also the throw new IOError issue.  The existing code was misleading but this is a good time to fix it

andychu:
> env_string_size = name->len_ + value->len_ + 2;

Actually I think it should just be `name->len_ + 1 + value->len_` because you're passing it to AllocStr, right?

The test should catch any off by ones, and I think there's no reason to over-allocate

jjbandit:
The reason it over-allocates is because `snprintf()` requires you pass a buffer large enough for it to append a null.  So even though we pass a buffer that's already null-terminated by definition, we can't tell `snprintf` that.  There are two solutions.

a) over-allocate by 1 byte such that `snprintf` can write its own null into the buffer.
b) allocate the correct length, then pass `env_string_size+1` to snprintf.

I significantly prefer option (a) because if we change the `Str` allocated buffer to _not_ be defined as len+1 in the future
this will overwrite by 1 byte, which (without a fancy allocator) would be a very difficult bug to find.

jjbandit:
> Sometimes I write things like `key_len + 1 + val_len + 1` to emphasize the 1 for the `=` and the 1 for the NUL.

I like that.

> Also the throw new IOError issue. The existing code was misleading but this is a good time to fix it
Meaning we should throw an IOError if ::putenv fails?

andychu:
Yeah I think you missed my inline comments up top.  I think these are worth fixing now:

- the return value of putenv() should be `void*` and it should throw an exception on error (original code was wrong)
- optional since this is still the leaky version: when we pass DOWN to snprintf, we don't need a buffer managed by the GC runtime (AllocStr).  We will probably just use malloc/free.

But I think that will make the code clearer, it could look like

```
int str_length = key->len_ + 1 -> value_len;   // 1 for '='
char buf = malloc(... str_length + 1)
```

To me that is idiomatic and avoids any need for comments

jjbandit:
>     * the return value of putenv() should be `void*` and 

I still don't understand what the `void*` would be pointing to.  What would we return from this function?

> it should throw an exception on error (original code was wrong)

10-4 .. I can make that change


> But I think that will make the code clearer, it could look like
> 
> ```
> int str_length = key->len_ + 1 + value->len_;   // 1 for '='
> char buf = malloc(... str_length + 1)  // malloc since the GC runtime doesn't care about this string; it's for putenv
> putenv(buf...)
> free(buf)
> ```
> 
> To me that is idiomatic and avoids any need for comments

Yeah, I agree.  And frankly that's probably correct in the GC'd version too.  No reason to have that string go through the managed memory.

andychu:
Oh sorry I meant `void` not `void*`

If you look at `native/posix_.pyi`, you can see putenv returns `None` in Python, which is the equivalent of void in C++.

Also you can check the usage with `grep putenv */*.py`, and the return value isn't used (because there is none)

jjbandit:
> Oh sorry I meant `void` not `void*`

Ah, that makes more sense.  Updated.

mycpp/mylib2.cc
```diff
@@ -2,9 +2,9 @@
 
 #include "mylib2.h"
 
-#include <errno.h>   // errno global
 #include <unistd.h>  // isatty
 
+#include "myerror.h"
```
andychu, line 7: Try `grep 'clang-format off' cpp/*.h` and follow the  style there.   "myerror.h" goes at the top so that errno is not defined yet

Also why is the only change in the file?

I think it may solve the problem to put `putenv()` in `posix.cc` instead of `posix.h`.

mycpp/myerror.h
```diff
@@ -27,4 +27,6 @@ class OSError : public _OSError {
   }
 };
 
+#include <errno.h>
```
andychu, line 30: shouldn't need this if we follow the style

cpp/posix.h
```diff
@@ -80,12 +80,14 @@ inline int close(int fd) {
...
-  int env_string_size =
-      name->len_ + value->len_ + 1;  // NOTE(Jesse): +1 for the '=' between them
-  char* env_string = static_cast<char*>(malloc(env_string_size));
-  snprintf(env_string, env_string_size, "%s=%s", name->data_, value->data_);
-  return ::putenv(env_string);
+inline void putenv(Str* name, Str* value) {
+  assert(name->IsNulTerminated());
+  assert(value->IsNulTerminated());
+  int overwrite = 1;
+  bool err = ::setenv(name->data_, value->data_, overwrite) == -1;
```
andychu, line 87: The `bool` is confusing because often means success is true == 1.

-- I think we should use this C-style error handling in most places.  (We're not that consistent about it, e.g. in pipe() too)


```
int result = c_style_call()
if (result < 0) {
  throw new IOError(errno)
}
```

andychu (commented):
Hm the `#include` changes don't seem right ... I think it may solve the problem to put the `putenv` implementation `posix.cc` instead of `posix.h`

The `errno` thing is a bit annoying but I think this is the best solution given the constraints of Python and C

andychu:
Also thanks for changing it to `setenv()`, that is a lot cleaner!  

Seems like putenv is  problematic, not sure why Python wraps it like that ...

https://stackoverflow.com/questions/5873029/questions-about-putenv-and-setenv

It doesn't wrap `setenv`, I guess you're supposed to use `os.environ['foo'] = bar` which is confusing

jjbandit:
> It doesn't wrap setenv, I guess you're supposed to use os.environ['foo'] = bar which is confusing

Yeah, it's weird.  I read some suggestions to change pythons `putenv` to use `setenv()` under the hood and it was all very confusing as to why they didn't want to do that. Something-something `putenv()` lets you keep track of the buffer so Python can efficiently .. whatever.

> Try grep 'clang-format off' cpp/*.h and follow the style there. "myerror.h" goes at the top so that errno is not defined yet
>The errno thing is a bit annoying but I think this is the best solution given the constraints of Python and C

I noticed that a common pattern in the Oil code is to check for an error from a CRT function, then throw an exception wrapping errno.  That said, why not just `#include <errno>` at the end of `mycpp/myerror.h` so we don't have to have comments all over the place saying "<errno> has to be included _after_ myerror.h".  I've seen that comment in at least two places, and I've added it myself in a third after including `<errno>` before `myerror.h` and not understanding why things suddenly broke.

PR #1196, opened 2022-07-02
[cpp] Throw error if strftime writes too much data for target buffer

jjbandit:


andychu:
Hm weird this commit got lost ... Github apparently merged it into `soil-staging`, but it's not on `master`.

I'm going to reapply it.


https://github.com/oilshell/oil/commits/3a627231c82c1a39c5ae04b82ddaf1e1f7012cd9/cpp/time_.h?browsing_rename_history=true&new_path=cpp/time__leaky.h&original_branch=master

PR #1199, opened 2022-07-03
Jesse/bugfix/segfault handler

jjbandit:
This PR fixes critical bounds checking failures in Str::slice.  It was possible to pass invalid indexes to Str::slice and access invalid memory, sometimes resulting in segfaults.

I installed a segfault handler which prints an identifier to STDOUT so we can grep for "OSH_CPP_SEGFAULT" in the test output to easily find test crashes.  Currently there's one test crashing, which is also failing, and always has been.

This PR also exposes a bug in the `var-ref` tests that was relying on Str::slice crashing.  `var-ref 12` was passing, and is now failing.  I observed the affected test relying on the crash in the debugger.  Currently the test returns garbage for an unknown reason.


mycpp/mylib_leaky.h
```diff
@@ -157,15 +157,42 @@ class Str : public gc_heap::Obj {
...
       end = len_ + end;
     }
+
+    begin = std::min(begin, len_);
+    end = std::min(end, len_);
+
+    begin = std::max(begin, 0);
+    end = std::max(end, 0);
+
+    assert(begin >= 0);
```
andychu, line 180: I would write all of these

assert(0 <= begin && begin < len_);

Also it should be `< len_` not `<= len_`   (e.g. 0-based index )

And I think `min(begin, len_ - 1)`

andychu (commented):
This is great thanks!  Definitely a bad bug :-/

I didn't look at the whole thing yet, but I'm about to get on a train ... I will finish looking at it tonight.

mycpp/mylib_leaky.h
jjbandit: FWIW I actually strongly prefer 2 asserts with discrete conditions to one with a combined condition.  Makes it easier to see what the failure case was.

Re- the len_ conditionals; I'm actually quite sure the code is correct as-is.  There may be a different way of writing it, but I added more tests for the boundary conditions and it's working as it's written.  I did catch another corner case when writing those extra tests though. The one I caught was (begin > end).

andychu: OK sorry I can see that end can be `len`.  BTW here is how Python does it, which is shorter, but also has some asymmetry that is a little hard to read:

https://github.com/oilshell/oil/blob/master/Python-2.7.13/Objects/stringobject.c#L1700

mycpp/mylib_leaky_test.cc
```diff
@@ -519,6 +519,136 @@ TEST test_sizeof() {
...
+  }
+
+  {
+    Str* s1 = s0->slice(7, 5);
+    ASSERT(str_equals(s1, new Str("")));
+    printf("%s\n", s1->data());
+  }
+
+  {
+    Str* s1 = s0->slice(7, 6);
```
andychu, line 623: Some of these test cases seem very similar... I think  cases like  (-100 100 ) and (0 100)  actually add a bit of coverage?   there are many branches in this function

(I just got some code coverage tools working, I hope to put them in the CI)

andychu (commented):
Thanks, glad to have this check

TBH I find it a bit confusing but Python's code is also confusing, and seemingly spread out over 2 places:

https://github.com/oilshell/oil/blob/master/Python-2.7.13/Objects/stringobject.c#L1700

https://github.com/oilshell/oil/blob/master/Python-2.7.13/Objects/stringobject.c#L1138

One thing Python does is return the original instance when the slice is equal to it, to avoid allocation

BTW the same bug occurs in List::slice in the same file, i.e. `mylist[1:-1]` and `mystr[1:-1]` have the same behavior

cpp/core_pyos.cc
```diff
@@ -55,7 +55,41 @@ Str* ReadLine() {
...
+
+    segfault_handler_initialized = 1;
+  }
+}
+
 Dict<Str*, Str*>* Environ() {
+
+  // NOTE(Jesse): I put this in here because it just happens to be the first
+  // hand-written C++ I could find that the osh interpreter calls into.
+  complain_loudly_on_segfault();
```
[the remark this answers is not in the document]
andychu: Also the list of files is in `cpp/NINJA_subgraph.py` if that helps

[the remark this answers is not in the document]
jjbandit: > Also the list of files is in `cpp/NINJA_subgraph.py` if that helps

That does help! Thanks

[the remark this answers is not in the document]
jjbandit: > OK I think this belongs in a separate file like `cpp/dumb_alloc.cc`, and then we can call it from the `cpp-skeleton` in `build/translate.sh` as well.

Yeah good call.  I'll move it.

jjbandit:
> BTW the same bug occurs in List::slice in the same file, i.e. `mylist[1:-1]` and `mystr[1:-1]` have the same behavior

Yeah I've spent most of the last 20 hours or so debugging problems with `mycpp`.  The next thing I'm going to do is to comb over it carefully to try and ferret out obvious bugs.

andychu:
OK great, thanks for fixing this!

Issue #1188, opened 2022-07-01
Redirecting output in a non-interactive shell doesn't always work in oil-native


jjbandit:
I believe this to be the root cause of a good number of the failures in the test suite.  It's difficult to say how many, but I've seen this (or a similar) bug in several tests now.

`./_bin/clang-dbg/osh_eval -c "type mapfile > tmp" ` -- doesn't work.  We get stdout in our shell instead of in the file `tmp`

`./_bin/clang-dbg/osh_eval -c "type fdasfds > tmp"  2>&1` -- works.
`./_bin/clang-dbg/osh_eval -c "echo 'hi' > tmp" ` -- works.
`./_bin/clang-dbg/osh_eval -c "(type mapfile) > tmp" ` -- works.

I originally thought maybe it was builtins only, but this works.

`./_bin/clang-dbg/osh_eval -c "[ fdas -gt foofdsa] > tmp 2>&1"` -- works

Actually, on closer inspection, I think it may be only `stdout` of builtins is broken.  The above command redirects stderr.

`./_bin/clang-dbg/osh_eval -c "alias thing=ding && alias > tmp 2>&1"` -- doesn't work .. and redirects `stdout`.

Any thoughts on this one @andychu

andychu:
- First thought was that this is try/finally #1163 
- And second thought is YES, it definitely is because I see `sys.stdout.flush()` in the finally: block
  - so basically libc might print something into a buffer, and then the redirect is undone, and later the buffer is flushed.  So that will cause the symptom

I was wondering if we should not use libc buffered I/O like most shells?  Or if we should just do the flush() for now.

Let me think about it

andychu:
i.e. I remember putting those `sys.stdout.flush()` into fix these kind of bugs, but maybe it's better if we don't need it at all ...

I think most of our output comes from `mylib.Stdout().write()` ... which is buffered CFileWriter using libc `fwrite()`.

I think shells tend to use the `write()` syscall directly

Although actually there is already an `fflush()` fix in that function?  Hmmm

andychu:
So somehow  this may be an incomplete fix : https://github.com/oilshell/oil/blob/master/mycpp/mylib_leaky.cc#L420

jjbandit:
Closed by #1200

PR #1200, opened 2022-07-03
Turn off buffered IO

jjbandit:
During my marathon debug session yesterday I noticed turning off buffered IO fixes a large number of test cases.

I vote we just bite the perf bullet and turn buffered IO off, make the shell correct, then make it fast.

The 'make it fast' part probably involves rewriting all the IO code to use our own buffering anyways.


jjbandit:
Closes #1188

andychu:
Ah interesting I didn't know about this function ... I think it makes sense to just get past these failures, and yeah I think we will want to do it a slightly different way later.

The issue is that most shells use POSIX I/O like read() and write(), and not C/C++ I/O like fread/fwrite, which are buffered.  The shell predates libc :)  So our bindings are technically a little wrong, but I want/need Oil to have a little more polymorphism, so that's why they are like that.

Please add a comment about why those functions are there.  Does it have to be in Environ ??  Not at startup?

And please file a bug about changing it to read()/write() later.

Then I will merge.  (note I'll be out for most of the day but back online tonight!)

Thanks this is great!

jjbandit:
> The issue is that most shells use POSIX I/O like read() and write(), and not C/C++ I/O like fread/fwrite, which are buffered. The shell predates libc :) So our bindings are technically a little wrong, but I want/need Oil to have a little more polymorphism, so that's why they are like that.
>

Yeah I assume once we're doing performance if IO becomes a problem we'd want to do our own buffering and use `write` directly.

> Please add a comment about why those functions are there. Does it have to be in Environ ?? Not at startup?
>
Yeah actually I'd prefer to do them at startup .. this is just the first hand-written C++ I could find to put them in.  Is there already an initialize function somewhere?

> And please file a bug about changing it to read()/write() later.
> 

#1201

> Then I will merge. (note I'll be out for most of the day but back online tonight!)
> 
> Thanks this is great!

:ok_hand:

andychu:
Ah yes, please try putting it in the `cpp-skeleton` in `build/translate.sh`.

jjbandit:
> Ah yes, please try putting it in the `cpp-skeleton` in `build/translate.sh`.

Updated

PR #1202, opened 2022-07-03
Jesse/bugfix/rewrite str replace

jjbandit:
Note this relies on #1199 being merged before it will actually work, as it requires `Str::slice` to work properly.

### DIFF
--- a/mycpp/mylib_leaky.cc
+++ b/mycpp/mylib_leaky.cc
@@ -31,61 +31,118 @@ void println_stderr(Str* s) {
 //   - CPython has like 10 special cases for this!  It detects if len(old) ==
 //     len(new) as well.
 // - we're not handling internal NULs (because of strstr())
-Str* Str::replace(Str* old, Str* new_str) {
-  // log("replacing %s with %s", old_data, new_str->data_);
 
-  const char* old_data = old->data_;
-  const char* last_possible = data_ + len_ - old->len_;
+inline bool
+strings_match(Str *s1, Str *s2)
+{
+  bool result = true;
 
-  const char* p_this = data_;  // advances through 'this'
-
-  // First pass to calculate the new length
-  int replace_count = 0;
-  while (p_this < last_possible) {
-    // cstring-TODO: Don't use strstr()
-    const char* next = strstr(p_this, old_data);
-    if (next == nullptr) {
-      break;
+  if (s1 && s2 && s1->len_ == s2->len_)
+  {
+    for (int i = 0; i < s1->len_; ++i)
+    {
+      result = (result && (s1->data_[i] == s2->data_[i]));
     }
-    replace_count++;
-    p_this = next + old->len_;  // skip past
+  }
+  else
+  {
+    result = false;
   }
 
-  // log("done %d", replace_count);
+  return result;
+}
 
-  if (replace_count == 0) {
-    return this;  // Reuse the string if there were no replacements
+struct cursor
+{
+  char *start;
+  int at;
+  int end; // NOTE(Jesse): start[end] is undefined.  The cursor is empty when (at == end)
+};
+
+cursor make_cursor(Str *s)
+{
+  cursor result = {};
+  result.start = (char *)s->data_;
+  result.end = s->len_;
+  result.at = 0;
+  return result;
+};
+
+inline void
+copy_into_cursor(cursor &cur, char c)
+{
+  assert(cur.at < cur.end);
+  cur.start[cur.at++] = c;
+}
+
+bool
+copy_into_cursor(cursor &c, Str *s)
+{
+  bool result = (c.at + s->len_) <= c.end;
+
+  if (result)
+  {
+    for (int i = 0; i < s->len_; ++i)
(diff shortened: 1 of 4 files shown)


jjbandit:
Also worth noting I spent a fair bit of time debugging the previous implementation of `Str::replace`.  It wasn't very fun.. the previous implementation (while more efficient due to not allocating a lot due to `Str::slice`) was highly obtuse and very difficult to modify without breaking previously-working cases.

I was able to write a new implementation and get all tests to pass _almost_ without using the debugger.

andychu:
Hm what are some examples it doesn't handle correctly?  I'm not super surprised since this function is fiddly, but  I don't see a description of where it's wrong.  Though I think the general approach is sound

Looking at `mylib_leaky_test.cc`, I'm slightly surprised I didn't add more coverage for this :-/

andychu:
Also when I grep the code, it appears that the string to replace is always of length 1, which makes the problem easier... not sure why I didn't take advantage of that :-/


```
$ metrics/source-code.sh osh-files |xargs fgrep 'replace('
frontend/args.py:    name = name.replace('-', '_')
frontend/flag_spec.py:    attr_name = name.replace('-', '_')
osh/prompt.py:        ret.append(r.replace('$', '\\$'))
osh/string_ops.py:  s = s.replace('\r', '<INVALID CR>').replace('\n', '<INVALID NEWLINE>')
osh/word_eval.py:  return s.replace('\\', '\\\\')
osh/word_eval.py:          p = prompt.replace('\x01', '').replace('\x02', '')
```

andychu:
Yeah I am missing the context here -- how did you run into this?  Does it fix a spec test,  a crash, an invalid memory access, etc. ?

jjbandit:
> Hm what are some examples it doesn't handle correctly? I'm not super surprised since this function is fiddly, but I don't see a description of where it's wrong. Though I think the general approach is sound

Ah, sorry.  If you take a look at the first commit there are four commented tests; those are the cases that were failing

https://github.com/oilshell/oil/pull/1202/commits/8bb950f47bc9600f5bb33150482f8a6c58eecc14#diff-3ef9a8c92693967993d892382d9a80cf14aeba8268f624d0ba9ef95e6ae4bff0R672

jjbandit:
> Yeah I am missing the context here -- how did you run into this? Does it fix a spec test, a crash, an invalid memory access, etc. ?

After fixing Str::slice I was just suspicious and put asserts in it that failed.  Coincidentally, fixing it does fix four tests.  You can see which ones by scrolling to the bottom of https://github.com/oilshell/oil/pull/1202/files

andychu:
OK thanks, I'm going to run these tests against the existing version and see what went wrong.

I think the existing approach to allocate less is sound, even if there's a bug.  Though there should have been more tests because this seems like exactly the kind of code that needs a bunch of tests.  What's not tested is broken :-/

I think I was rushing through it and made end-to-end tests pass, and never wrote unit tests.  I was trying to see if this approach would even perform well at all!  (it did but now we need to go back and do it right)

----

BTW I think you can start using `using mylib::CopyStr;   ASSERT(str_equals(CopyStr ... ` in the tests instead of `new Str`.  That will make it easier to port to the GC version.

jjbandit:
> OK thanks, I'm going to run these tests against the existing version and see what went wrong.
> 
> I think the existing approach to allocate less is sound, even if there's a bug. Though there should have been more tests because this seems like exactly the kind of code that needs a bunch of tests. What's not tested is broken :-/

Yeah I'm thinking to avoid the billion allocations we do something like `Str::slice_inplace` which returns pointers into an allocated string.  The version of `Str::slice` that allocates/copies would basically be a thunk that calls `slice_inplace` and copies the result to a new buffer.  The obvious problem with that approach is that if the GC hits while we've got pointers into a GC'd buffer those then become pointers into the nether-lands of who knows what.  One solution we could have for this is do reference counting where we tell the GC to crash the program if we have live references into GC'd buffers.  Used along with `GC_EVERY_ALLOC` that would allow us to have those cheap buffer slices as well as reliably ferret out places we accidentally allocate when we've got temporary references to GC objects.  Thoughts?

It's worth noting that I'm talking about using library functions to write code that was previously a big pile of pointers.  The 'big pile of pointers' approach has the same problem as using library functions, in the sense that if the underlying data gets moved, they then point into invalid memory.  The 'big pile of pointers' implicitly never allocates because it's just using pointer arithmetic, and thus skirts the problem.  Using library functions would explicitly never allocate, and enforce that invariant by crashing in a specific way if it is violated.  Not sure if I'm articulating very well here, but this makes a lot of sense in my head.  Happy to discuss further.  I think this is important to get right.

> I think I was rushing through it and made end-to-end tests pass, and never wrote unit tests. I was trying to see if this approach would even perform well at all! (it did but now we need to go back and do it right)

Agreed!

andychu:
As mentioned in e-mail, I reused the tests here

https://github.com/oilshell/oil/commit/23cc7be97ad0ae43dfd266ad80881683979df97e

Still a couple follow-ups

PR #1203, opened 2022-07-04
Ensure that the spec test tsv output directory exists

watk:
Previously, running `test/spec-cpp.sh run-with-osh-eval smoke` would
fail with:

    IOError: [Errno 2] No such file or directory: '_tmp/spec/cpp/smoke.tsv'

Now it succeeds.

### DIFF
--- a/test/spec-cpp.sh
+++ b/test/spec-cpp.sh
@@ -61,6 +61,7 @@ run-with-osh-eval() {
   shift
 
   local base_dir=_tmp/spec/$SPEC_JOB
+  mkdir -p "$base_dir"
 
   # Run it with 3 versions of OSH.  And output TSV so we can compare the data.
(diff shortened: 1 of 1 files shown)


watk:
Just a tiny fix, @andychu

andychu:
Ah thanks, this is an oversight.  Although I think we should fix it in shell, e.g. `mkdir -p $base_dir` here:

https://github.com/oilshell/oil/blob/master/test/spec-cpp.sh

(the `test/spec-common.sh` function does that for Python, so the current code is inconsistent)

I think it is fairly standard for tools, e.g. `tar` won't create a file in a nonexistent directory.  Also technically there is a race condition here because we run spec tests in parallel (TOCTTOU)

watk:
yep, sorry, I should have taken a look around to see the analogue in spec-common.sh

PTAL

PR #1205, opened 2022-07-05
[build] Regenerate build.ninja on changes to the build graph

watk:
Hey @andychu, I got nerd sniped by this when I changed a compiler flag in NINJA_subgraph.py!

After doing this, I'm not so convinced it's a good idea, or worth the effort. Since the build is described by python scripts, I don't know if I've missed a file that contributes to it (e.g., NINJA-config.sh sources build/app-deps.sh, which could affect the output build.ninja even though I assume it doesn't). Maybe some discipline + smart tracking of python imports could do it? But, yeah, I'd probably just shelve this rather than do anything more complicated


andychu:
Ahh sorry the Ninja stuff is in flux now.  But I think it works pretty well if you remember to `./NINJA_config.py` and remember that generates `build.ninja`.

I tried to do something like this a week ago but backed out of it ...   I can't find the commit but there  wasn a "self edge" we had.

The problem I found is that if `build.ninja` doesn't exist the FIRST time, then Ninja didn't like that dependency.

How does this work?  It seems like nothing depends on `build.ninja`?  

-----

I'm going to be changing Ninja a bunch, and I think it will look like this:

```
build/dev.sh all  # generates Python source code
./NINJA-config.sh   # depends on the Python build in previous step to crawl deps at runtime
ninja _bin/cxx-dbg/osh_eval  # or any target
```

I think we should hold off until that has settled down, and then revisit this.

-----

I tagged some issues "opening up" for more contribution:

https://github.com/oilshell/oil/issues?q=is%3Aissue+is%3Aopen+label%3Aopening-up

This one had a bunch of good first issues: https://github.com/oilshell/oil/issues/1167

Although Jesse just fixed a lot of the low hanging fruit.  The `assert(0)` I see left is this one:

http://travis-ci.oilshell.org/github-jobs/2022-07-05__17-03-06.wwz/_tmp/spec/cpp/builtin-io.html#details-52-osh_.cc

That is probably relatively easy to write but a little hard to test.

------------

There is also this: https://github.com/oilshell/oil/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

although most are Python issues, not sure if you were more interested in helping with C++ (which we do need a lot of help on)

andychu:
I also want help with moving bindings, but I want to do this renaming first:

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Big.20Renaming.20Coming

Rename all the old code to `*.leaky.h`, and then only have "good new GC-d code" in `*.h`

watk:
> Ahh sorry the Ninja stuff is in flux now. But I think it works pretty well if you remember to `./NINJA_config.py` and remember that generates `build.ninja`. (I added NINJA-config.sh literally yesterday, so it doesn't do anything yet, sorry for the confusion.)

Yep sounds good, now that I've implemented this, I'll definitely remember how/when to regen build.ninja.

> I tried to do something like regenerate `build.ninja` automatically a week ago but backed out of it ... I can't find the commit but there was a "self edge" we had.
> 
> The problem I found is that if `build.ninja` doesn't exist the FIRST time, then Ninja didn't like that dependency.

Oh, well with these changes you have to manually run NINJA_config once to create the initial build.ninja. But from then on, it's automatic.

> How does this work? It seems like nothing depends on `build.ninja`?

Ah, this is a special ninja behaviour. From the build section of the manual: `If the top-level Ninja file is specified as an output of any build statement and it is out of date, Ninja will rebuild and reload it before building the targets requested by the user.`

> I think we should hold off until that has settled down, and then revisit this.

Yep, I'll close this for now

andychu:
Ah thanks for the reference to the manual

That could come in handy, but let's see when things have settled down

andychu:
Thinking about this more, the issue you pointed out is right

Even after my upcoming changes, if there are more Python deps, then people will have to run `NINJA-config.sh` anyway.  It is going to generate a bunch of implicit deps like `mycpp/NINJA` and `_build/NINJA`

I might have a blog post about this, but our "Python metaprogramming C++" does introduce some slightly unique build system issues

The rule will be: if you change Python imports, you have to run `./NINJA-config.sh`

So it's not that much of a stretch to say if you change `*/NINJA_subgraph.py`, you also run `./NINJA-config.sh`

This needs to be documented though.  Thanks for the reminder!

watk:
I don't 100% follow, but I believe you :) I was only inspired to add this because `gn` has it. And it's just a small QoL thing when working on Chrome, for example, to not be debugging something and realise you forgot to re-run `gn`. But I shouldn't have nerd sniped myself hah

PR #1208, opened 2022-07-06
[fix] Fix nullptr deref in List::reserve()

watk:
Caught by ubsan:

/home/c/dev/oil/mycpp/gc_heap.h:972:47: runtime error: member access within
null pointer of type 'Slab<int>'

Previously, List::reserve() was copying from the old slab to the new one
whenever the *newly updated* capacity was greater than zero. If the
capacity was previously zero, slab_ might be nullptr.

Now it copies only when the list is non-empty.

### DIFF
--- a/mycpp/gc_heap.h
+++ b/mycpp/gc_heap.h
@@ -957,23 +957,21 @@ class List : public gc_heap::Obj {
     auto self = this;
     StackRoots _roots({&self});
 
-    if (self->capacity_ < n) {
-      // Example: The user asks for space for 7 integers.  Account for the
-      // header, and say we need 9 to determine the obj length.  9 is
-      // rounded up to 16, for a 64-byte obj.  Then we actually have space
-      // for 14 items.
-      self->capacity_ = RoundUp(n + kCapacityAdjust) - kCapacityAdjust;
-      auto new_slab = NewSlab<T>(self->capacity_);
+    // Don't do anything if there's already enough space.
+    if (self->capacity_ >= n) return;
 
-      // slab_ may not be initialized constructor because many lists are
-      // empty.
-      if (self->capacity_ != 0) {
-        // log("Copying %d bytes", len_ * sizeof(T));
-        memcpy(new_slab->items_, self->slab_->items_, self->len_ * sizeof(T));
-      }
-      self->slab_ = new_slab;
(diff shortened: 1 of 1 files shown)


mycpp/gc_heap.h
```diff
@@ -957,23 +957,21 @@ class List : public gc_heap::Obj {
...
-
-      // slab_ may not be initialized constructor because many lists are
-      // empty.
-      if (self->capacity_ != 0) {
-        // log("Copying %d bytes", len_ * sizeof(T));
-        memcpy(new_slab->items_, self->slab_->items_, self->len_ * sizeof(T));
-      }
-      self->slab_ = new_slab;
+    // Don't do anything if there's already enough space.
+    if (self->capacity_ >= n) return;
```
watk, line 961: Note: as a micro-optimisation, this could be moved up to the top of this function. The question is: do you you want to follow a simple/conservative rule like: "member functions always start with `auto self = this; ..`". Or do you want it to be case-by-case?

andychu:
Oh cool!  Which test running with ubsan exposed this?  The only failure I see is `my_runtime_test`, and then I don't see this?

```
$ _bin/cxx-ubsan/mycpp-unit/my_runtime_test 
food
food
foobar
foobar
.value = [bar]
value = 42-42.
..mycpp/my_runtime_test.cc:132:17: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
mycpp/my_runtime_test.cc:135:7: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
i = -2147483648
.f1212d
f123123
f123x123
foobXXr

```

watk:
Really? The list_methods_test hits it. This prints the error for me: `_bin/cxx-ubsan/mycpp-unit/my_runtime_test -t list_methods_test`

mycpp/gc_heap.h
andychu: True, I'm not sure ... I think this is nice for now, to keep the code looking regular.  But the style may evolve as we make it correct!

andychu (approved):
Great thank you!  

I think the ubsan difference has to do with GCC vs. Clang -- I think the GCC ubsan is less mature  and so you probably have a newer version with more features (?)  

I will look into it and probably add a Clang variant

PR #1211, opened 2022-07-06
Init BraceGroup() with empty array

jjbandit:
Long story short, we've got to initialize these with an empty array or
the code that executes for loops tries to dereference 0 and crashes.

### DIFF
--- a/osh/cmd_parse.py
+++ b/osh/cmd_parse.py
@@ -1097,7 +1097,7 @@ class CommandParser(object):
     self._Eat(Id.Lit_RBrace)
     right_spid = word_.LeftMostSpanForWord(self.cur_word)
 
-    node = BraceGroup(doc_token, c_list.children, None)  # no redirects yet
+    node = BraceGroup(doc_token, c_list.children, [])  # no redirects yet
     node.spids.append(left_spid)
     node.spids.append(right_spid)
     return node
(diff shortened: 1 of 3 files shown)


andychu:
I'm trying to figure out why there was a mistranslation here.

When I do

```
test/spec.sh for-expr -r 2 -p > _tmp/f
bin/osh --ast-format text -n _tmp/f
```

and then `print(n1.redirects)` here:

https://github.com/oilshell/oil/blob/master/osh/cmd_parse.py#L1301

```

https://github.com/oilshell/oil/blob/master/osh/cmd_parse.py#L1301

Then I get an empty list, not None.

So this mistranslation could be masking other bugs.

I have run into this kind of thing before and want to get to the root cause, because it's probably causing other bugs that will hit later. 

Thanks for finding this!

**Comment by andychu** [comment]
So I did:

```
ninja _bin/cxx-uftrace/osh_eval 
uftrace record _bin/cxx-uftrace/osh_eval -n  _tmp/f  # the for-expr test case
uftrace replay

```

Then Ctrl-F for `ParseRedirectList` which is good, but then I can't find `ParseForExprLoop` which I'm expecting (?)

I want to narrow this down a bit ... why is the parser not producing the same data structure in Python and C++?

Did you use GDB to debug?    uftrace can be useful although sometimes you need to single step

PR #1212, opened 2022-07-06
Always throw from execve, as per python

jjbandit:


cpp/posix.h
```diff
@@ -133,10 +136,10 @@ inline void execve(Str* argv0, List<Str*>* argv, Dict<Str*, Str*>* environ) {
   _argv[n] = nullptr;
 
   ::execve(_argv0.Get(), _argv, nullptr);
+
+  throw error {};
```
andychu, line 140: Ahh yeah missing return value check :-/  We probably need a manual audit, I will file a bug about this


I think this should be

```
int result = ::execve(...)
if (result < 0) {
  throw new IOError(errno)
}
```

The reason is that we will catch errno and display an error message.  Which are not necessarily tested!  But I think it will be repro-able here.

-----


```
>>> posix_.execve('/zz', ['/zz'], {})
OSError: [Errno 2] No such file or directory
>>> 
```

andychu: Hm what would mycpp do?  enforce that it doesn't return?  I think it's hard to know that.  If you have suggestions definitely feel free to file bugs

What does  `[[noreturn]]` do?  Suppress control flow warnings?

jjbandit: Oh, actually, without a parser that also understands C++ my suggestion is impossible.  I suppose the best `mycpp` could do is insert "InvalidCodePath()` statements immediately following functions marked `#NO RETURN` .. which actually would have caught this bug and made it significantly easier to understand what was going wrong.  Might be a high-effort low-reward task though, considering there are only a few places that's used.

jjbandit: >What does [[noreturn]] do? Suppress control flow warnings?

For posterity, `[[noreturn]]` tells the compiler to error if the function does anything but throw an error.

PR #1214, opened 2022-07-06
Jesse/bugfix/list contains

jjbandit:


jjbandit:
I have no idea why the CI is failing.  I looked at the commands that are failing and they work on my machine.
 `ninja mycpp-all` also works on my machine.  @andychu any ideas?

andychu:
Cool thanks!  I will look at this tonight, and let's see about the CI

andychu (commented):
Thanks this looks good.

The build breakage is because of the `id_kind_asdl.h` #include -- it's a generated file that doesn't exist yet when the tests are run.  

What I do is click on the first red error and then search for "error", and that comes up

http://travis-ci.oilshell.org/github-jobs/2022-07-06__21-04-28.wwz/_tmp/soil/logs/cpp-unit.txt

-----

I suspect the simplest option is to move the `Kind::` specialization over to `cpp/preamble.h` below the `#Include`?  And get rid of the #include here

(It sort of violates the layering to have an "app type" like `Kind` inside the `mycpp` runtime library.)


Please reformat the code with `test/lint.sh format-oil` as well

PR #1226, opened 2022-07-15
Pass the environment to execve()

watk:
Now posix::execve() converts and forwards its environ argument to execve.

Reduces the spec-cpp delta cpp number by 26 (70 to 44) if I ran the tests correctly.

### DIFF
--- a/cpp/leaky_posix.h
+++ b/cpp/leaky_posix.h
@@ -121,18 +121,35 @@ inline mylib::LineReader* fdopen(int fd, Str* c_mode) {
 inline void execve(Str* argv0, List<Str*>* argv, Dict<Str*, Str*>* environ) {
   mylib::Str0 _argv0(argv0);
 
-  int n = len(argv);
+  int n_args = len(argv);
   // never deallocated
-  char** _argv = static_cast<char**>(malloc(n + 1));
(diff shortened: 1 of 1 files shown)


cpp/leaky_posix.h
```diff
@@ -121,18 +121,35 @@ inline mylib::LineReader* fdopen(int fd, Str* c_mode) {
 inline void execve(Str* argv0, List<Str*>* argv, Dict<Str*, Str*>* environ) {
   mylib::Str0 _argv0(argv0);
 
-  int n = len(argv);
+  int n_args = len(argv);
   // never deallocated
-  char** _argv = static_cast<char**>(malloc(n + 1));
+  char** _argv = static_cast<char**>(malloc((n_args + 1) * sizeof(char*)));
```
watk, line 126: I thought I'd fix this bug while I was here

andychu (approved):
Amazing, thank you !!

Glad you were also looking at the top entry in "delta cpp".  I was looking at it too and wondering why there was `None` in many places.

I think what happened is that I was making the test numbers go up in a rush, and added the `nullptr`.  I should have definitely added a TODO comment there!

And thanks for fixing the overflow bug :-/  I wonder if we should extract this into a function and let ASAN test it ... We can consider that when moving to the non-leaky GC bindings.

PR #1230, opened 2022-07-15
Throw OSError* from posix::execve() to match python

watk:
Previously this was throwing IOError (by value) for which there is no matching handler.

Now it throws OSError* to match its python equivalent (and exception handler in the translated code that calls it).

With this change there's no delta cpp for command_ spec tests!

### DIFF
--- a/cpp/leaky_posix.h
+++ b/cpp/leaky_posix.h
@@ -3,6 +3,10 @@
 #ifndef POSIX_H
 #define POSIX_H
 
+// clang-format off
+#include "mycpp/myerror.h"  // for OSError; must come first
+// clang-format on
+
 #include <errno.h>
 #include <stdlib.h>  // putenv
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


andychu:
Oh awesome!   Yeah I noticed this set of failures

-----

There is a fundamental issue where I want to treat OSError and IOError the same -- there is really no difference and it was removed in Python 3

I was thinking that we translate:

```
try
  foo()
except (OSError, IOError) as e:
  pass
```

to

```
try {
  foo()
} except  _OSerror {   // base class in myerror.h
  ;
}
```

to eliminate this distinction, but clearly it isn't done!


The motivation is that C++ does not have syntax for catching 2 types of exception at once.  You would have to duplicate the blocks.

But I guess we do not need to do that EVERYWHERE.  We can do that in ONLY the places where we have `(OSError, IOError)`.

I was thinking we want C++ code to throw just ONE of IOError and OSError, but maybe having both is OK

watk:
Oh, ok! Makes sense

watk:
Thanks for the friendly reviews :)

andychu:
Oh also I wonder if you know about the `-p` flag which prints the spec test case to stdout?  This helps me when debugging


```
test/spec-cpp.sh run-with-osh-eval smoke -r 0 -p > _tmp/0.sh
_bin/cxx-dbg/osh_eval _tmp/0.sh
```


I think I told Jesse about it, and just added it here:

watk:
Ah thanks, I had hacked up the runner script to do this ! Somehow missed the flag

PR #1234, opened 2022-07-18
Fix some bad math in Str::split causing invalid writes.

jjbandit:
I wrote a bunch of code to aid in debugging.  Not sure if we should keep it around, but I'm guessing it'll be useful again at some point so I've left it in for now.

### DIFF
--- a/mycpp/gc_heap.h
+++ b/mycpp/gc_heap.h
@@ -17,6 +17,17 @@
 
 #include "common.h"
 
+#define VALIDATE_ALLOCATION(allocation, num_bytes, test_value)        \
+  for (int i = 0; i < (num_bytes); ++i) {                             \
+    if (((unsigned char*)(allocation))[i] != (test_value)) {          \
+      unsigned long offset = (unsigned long)(allocation) -            \
+                             (unsigned long)gHeap.from_space_.begin_; \
+      printf("failed at address 0x%lx, offset (%lu) index (%u)\n",    \
+             (unsigned long)(allocation), offset, i);                 \
+      assert(0);                                                      \
+    }                                                                 \
+  }
+
 // Design Notes:
 
 // It's a semi-space collector using the Cheney algorithm.  (Later we may add a
@@ -207,7 +218,19 @@ class Heap {
   }
 
   void* Allocate(int num_bytes) {
+    // NOTE(Jesse): I'm not sure what we should do if we ask for 0 bytes, but
+    // this is the second bug I'm fixing due to it, so I'm putting this check
+    // and some assertions in.  Practically, we should probably crash if we
+    // try to write to a buffer that's of 0 size anyways.
+    //
+    if (num_bytes <= 0) {
+      return 0;
+    }
+    assert(num_bytes > 0);
+
     int n = aligned(num_bytes);
+    assert(n > 0);
+
     // log("n = %d, p = %p", n, p);
 
 #if GC_EVERY_ALLOC
@@ -215,7 +238,8 @@ class Heap {
 #endif
 
     if (free_ + n <= limit_) {  // Common case: we have space for it.
-      return Bump(n);
+      void* Result = Bump(n);
+      return Result;
     }
 
 #if GC_DEBUG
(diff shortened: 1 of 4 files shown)


andychu (commented):
Ah good catch, it looks like the last part was omitted from the total length, and prev_pos wasn't initialized correctly.

Are there unit tests for the split() function?

I remember this function being extremely fiddly, and I think that's why I mentioned it along with glob() returning `List<Str*>*`.

I think we might want some function like `AllocListOfStr(num_strings, total_length)` or something?  Not sure.


I think we should always fail if the requested allocation is 0 or less

andychu:
And yup this function looks like it is very lacking in test coverage , which is bad: 

http://travis-ci.oilshell.org/github-jobs/2022-07-15__15-58-32.wwz/_test/clang-coverage/unified/html/coverage/home/uke/oil/mycpp/my_runtime.cc.html#L425

andychu:
Oh actually no it has test coverage, but it's not run precisely because of the bug we were talking about !!!

So that's why the coverage reports it as not run.

mycpp/mylib2.h
```diff
@@ -155,7 +155,13 @@ class BufWriter : public Writer {
   // For cStringIO API
   Str* getvalue() {
     if (data_) {
+      // NOTE(Jesse): @copy_str_possible_gc_crash_no_stack_root
```
andychu, line 158: Yup I believe this is the bug here, which I was looking at last week

https://github.com/oilshell/oil/issues/1229

andychu:
One idea: can we have a `BoundedBuffer` abstraction?

- that you can  copy into repeatedly, but is as efficient as the existing code
- that omits checks in opt mode but does bounds checks in debug mode?

I think I was confused by the name of `Cursor`.  To me it's really a bounds-checked buffer

If that would be useful for both `replace()` and `split()`, and maybe `glob()`, I think it makes sense

jjbandit:
> One idea: can we have a `BoundedBuffer` abstraction?
> 
>     * that you can  copy into repeatedly, but is as efficient as the existing code
> 
>     * that omits checks in opt mode but does bounds checks in debug mode?
> 
> 
> I think I was confused by the name of `Cursor`. To me it's really a bounds-checked buffer
> 
> If that would be useful for both `replace()` and `split()`, and maybe `glob()`, I think it makes sense

Yeah it could be named anything really.  I call the struct `cursor` because that's what databases call it, but I agree that name is a bit confounding.

I also like to make a version of `assert()` that crashes in debug mode, but compiles out in release mode.. something like:

```
#if DEBUG_
#define debug_assert(condition) do {  if ((condition) == false) {  CauseRuntimeCrash;  }  } while (false)
#else
#define debug_assert(...)
#endif
```

That would be a useful thing for doing bounds checking on the writes.  I tend to throw asserts around like candy because I can think of them as free.

jjbandit:
Before I go to the trouble of rewriting this with a `BoundedBuffer` I'm going to futz around with the leaky version to see if I can get that to support both leaky and GC'd modes (using the `Param` thing)

andychu:
It doesn't have to be in this commit ... (and really shouldn't be)

I think we should land this fix as long as it has unit test coverage, and then we can added BoundedBuffer separately (maybe do it along with glob or something else, to make sure the API is right)

Ditto with  strip/rstrip/lstrip PR, although I would be interested in which test cases failed there

jjbandit:
Alright, yeah I'm good to merge this as-is.

mycpp/gc_heap.h
```diff
@@ -207,15 +218,28 @@ class Heap {
   }
 
   void* Allocate(int num_bytes) {
+    // NOTE(Jesse): I'm not sure what we should do if we ask for 0 bytes, but
+    // this is the second bug I'm fixing due to it, so I'm putting this check
+    // and some assertions in.  Practically, we should probably crash if we
```
tekknolagi, line 223: Round up to some minimal allocation size perhaps? We had a minimum allocation size of 2 words. One for the header, one for... something in the object body.

andychu:
FWIW we landed most of this change, but I changed the title to reflect that we could still add the assertion (and I think it's probably a good idea)

jjbandit:
Closing.  We're doing this now.

PR #1241, opened 2022-07-20
Fix invalid writes in Str::split

jjbandit:
https://github.com/oilshell/oil/pull/1234

Original PR here ^

Opening a new one because it's easier than taking all the spurious changes out of that one.

### DIFF
--- a/mycpp/my_runtime.cc
+++ b/mycpp/my_runtime.cc
@@ -440,9 +440,10 @@ List<Str*>* Str::split(Str* sep) {
   // Find breaks first so we can allocate the right number of strings ALL AT
   // ONCE. We want to avoid invalidating self->data_.
   int num_bytes = 0;
-  int prev_pos = 0;
+  int prev_pos = -1;
   std::vector<int> breaks;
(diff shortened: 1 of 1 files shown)


jjbandit:
@andychu Ready to merge

jjbandit:
@andychu 

Looks like the build is failing on this ..?
```
status	elapsed	task	script	action	result_html
status=128 max_status=128
128	0.017857	mycpp-git	mycpp/deps.sh	git-clone	-
```

Not exactly sure if that's a problem or just flaky network?

andychu (commented):
Great thank you! 

This will unblock the coverage for my_runtime_test

And then I'm wondering if `if (length)` has branch coverage for true and false?  But now we can see, I *think*

andychu:
Oh the funny thing is that the CI is failure is caused by the Docker containers being out of sync with what's in the repo.

i.e. Docker containers built at HEAD will work, but this branch is slightly behind HEAD.

So yeah this one slight flaw of Docker repo model, discussed in `#containers` on Zulip

Issue #1233, opened 2022-07-18
osh_eval bug: Forking causes osh to jump backwards in the script it's executing


watk:
I ran into this while debugging a bigger example for a spec test. Easy to repro though:
```bash
> cat >/tmp/f.sh <<'EOF'
echo start
$(:)
echo end
EOF

> _bin/cxx-dbg/osh_eval /tmp/f.sh
start
end
end
```

I spent some time debugging this and figured out that the child process `lseek()s` backward on the file stream for the script that's running. And since the fd is shared with the parent, the parent shell's state gets messed up. I took a look in the debugger to figure out why on earth the seek was happening and it turns out that it's glibc's atexit handler for file streams:
```
libc.so.6!__lseek64(int fd, off64_t offset, int whence)
libc.so.6!_IO_new_file_sync(FILE * fp)
libc.so.6!_IO_default_setbuf(FILE * fp, char * p, ssize_t len))
libc.so.6!_IO_new_file_setbuf(FILE * fp, char * p, ssize_t len)
libc.so.6!_IO_unbuffer_all()
libc.so.6!_IO_cleanup()
libc.so.6!__run_exit_handlers(int status, struct exit_function_list ** listp, _Bool run_list_atexit, _Bool run_dtors)
libc.so.6!__GI_exit(int status) 
posix::_exit(int status) (/home/c/dev/oil/cpp/leaky_posix.h:103)
process::SubProgramThunk::Run(process::SubProgramThunk * const this) (/home/c/dev/oil/_build/cpp/osh_eval.cc:9757)
...
```

Funnily enough `process.py` calls `_exit()` (which doesn't run atexit handlers): https://github.com/oilshell/oil/blob/7995c46e6556b77e3e8b37f61d6623c4f0e5a578/core/process.py#L716
but that code is translated into `exit()`:
https://github.com/oilshell/oil/blob/dec6fa0b911e9a6d03f5359e33a7de857812de54/cpp/posix.h#L127-L129

So is it fine to just change this latter one to `_exit()`?

OOI I took a look at what bash does: a) it doesn't use `FILE*`'s so wouldn't have any `atexit` handlers, and b) it closes the script fd in the child:
https://github.com/bminor/bash/blob/15409324f1974d41c183904ad575da7188058c1c/jobs.c#L2230

andychu:
Oh wow ... I didn't know glibc had an exit handler for file streams!    Eek

[...]

PR #1242, opened 2022-07-21
[cpp] posix::_exit() now calls _exit(), not exit() (#1233)

watk:
Calls in python to posix._exit() were being translated into calls to
exit(), which has different semantics. Now it calls _exit() as expected.

Testing:
```bash
> cat >/tmp/f.sh <<'EOF'
echo start
$(:)
echo end
EOF

# before this change
> _bin/cxx-dbg/osh_eval /tmp/f.sh
start
end
end

# with this change
> _bin/cxx-dbg/osh_eval /tmp/f.sh
start
end
```

@andychu Sorry if this is mentioned somewhere and I missed it. Will the CI automatically flag any regressions caused by a change like this? I'm not sure if I should be running something locally to verify or not

### DIFF
--- a/cpp/leaky_stdlib.h
+++ b/cpp/leaky_stdlib.h
@@ -86,7 +86,7 @@ inline int fork() {
 }
 
 inline void _exit(int status) {
-  exit(status);
+  ::_exit(status);
 }
 
 inline void write(int fd, Str* s) {
--- a/spec/bugs.test.sh
+++ b/spec/bugs.test.sh
@@ -153,3 +153,17 @@ echo status=$?
 status=1
 status=1
 ## END
+
+#### subshell while running a script (regression)
+# Ensures that spawning a subshell doesn't cause a seek on the file input stream
+# representing the current script (issue #1233).
+cat >tmp.sh <<'EOF'
+echo start
+(:)
+echo end
+EOF
+$SH tmp.sh
+## STDOUT:
+start
+end
+## END


andychu:
Thanks for doing this!  

I'm not sure why the CI isn't running for your PRs and they are for Jesse's

I added you as a collaborator to the project and maybe that will fix it

Let me look at his PRs first and then circle back... I do want to see what the spec tests results are!

andychu:
Ah OK I pulled your PR and then pushed to a branch, and now it's running the CI.

https://github.com/oilshell/oil/actions/runs/2708946912

Hopefully next time it will run automatically with your PRs, if you're a contributor

Let's see what happens!

watk:
Oh sweet thanks! 

I should add a spec test that fails before this change

andychu:
Everything passes with this change which is great:

I believe it's a good change just based on your explanation (and the crazy StackOverflow thread!!)

---

But yeah nothing changed here?

http://travis-ci.oilshell.org/github-jobs/2022-07-21__02-48-42.wwz/_tmp/spec/cpp/osh-summary.html

But maybe it would when we get rid of the

```
  setvbuf(stdout, 0, _IONBF, 0);
  setvbuf(stderr, 0, _IONBF, 0);
```

in the `main()` of  `osh_eval` ?

https://github.com/oilshell/oil/blob/master/build/translate.sh#L60

----

If you can think of a failing case that would be nice (and could catch future errors), but it's not essential to merge this

andychu:
Oh duh you could just copy this test case in to `spec/bugs.test.sh` 

```
echo start
$(:)
echo end
```

which you run as usual with

```
test/spec.sh bugs
test/spec-cpp.sh run-with-osh-eval bugs
```

(sorry doing too many things at once)

watk:
> Oh duh you could just copy this test case in to spec/bugs.test.sh

Oh perfect, thanks! Done in a new commit, and verified it fails before and passes after.

watk:
> I'm not sure why the CI isn't running for your PRs and they are for Jesse's

Could it be because I'm making changes in a fork but Jesse is pushing to the oil repo?

andychu:
Thank you!

I think I might need to change our YAML config: https://stackoverflow.com/questions/64802322/running-github-actions-on-the-code-in-a-pull-request-from-a-fork

PR #1249, opened 2022-07-22
Translate the wait builtin

watk:
I started translating `wait` and found that it was pretty much already translatable! 

I moved `class Wait` but I had to make one minor modification to it (GH is not smart enough to show the diff as a move + a  change).  I'll comment on the diff to point it out

Makes osh_eval native pass 14/16 of the background spec tests. Up from 0/16.

### DIFF
--- a/osh/builtin_process2.py
+++ b/osh/builtin_process2.py
@@ -4,18 +4,22 @@ builtin_process2.py
 """
 from __future__ import print_function
 
-from _devbuild.gen.runtime_asdl import cmd_value, cmd_value__Argv
-
-from core import vm
+from _devbuild.gen import arg_types
+from _devbuild.gen.runtime_asdl import (cmd_value, cmd_value__Argv,
+                                        wait_status__Cancelled,
+                                        wait_status__Pipeline,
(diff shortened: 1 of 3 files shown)


osh/builtin_process2.py
```diff
@@ -57,6 +61,133 @@ def Run(self, cmd_val):
...
+      # Does it look like a PID?
+      try:
+        pid = int(job_id)
+      except ValueError:
+        raise error.Usage('expected PID or jobspec, got %r' % job_id,
+                          span_id=span_id)
+
+      job = self.job_state.JobFromPid(pid)
+      if job is None:
+        self.errfmt.Print_("%d isn't a child of this shell" % pid,
```
watk, line 167: This `%d` used to be `%s`. `pid` is an int so there was a compiler error in the generated code

watk:
@andychu I don't think the CI ran on my branch. I think it ran on master?

andychu:
Awesome thank you!  

Not sure what's up with Github Actions, but we'll figure it out eventually ...

PR #1261, opened 2022-08-04
Jesse/refactor/move mylib old str code into leaky types

jjbandit:


mycpp/leaky_types.cc
```diff
@@ -2,16 +2,144 @@
...
+}
+
+Str* str_repeat(Str* s, int times) {
+  // Python allows -1 too, and Oil used that
+  if (times <= 0) {
+    return kEmptyString;
+  }
+  int len_ = len(s);
+  int new_len = len_ * times;
+  char* data = static_cast<char*>(malloc(new_len + 1));
```
andychu, line 36: I think this should now use `AllocStr(new_len)` and then copy into that buffer, but since we're doing mechanical transformations now, this is fine

Although I don't see what the point of CopyBufferIntoNewStr() is, because StrFromC() should suffice?  A buffer that is malloc'd is "from C", i.e. it is not garbage collected yet

mycpp/leaky_types.cc
```diff
@@ -2,16 +2,144 @@
...
+    if (!isspace(*p)) {
+      return false;
+    }
+    p++;
+  }
+  return true;
+}
+
+// for os_path.join()
+// NOTE(Jesse): Perfect candidate for bounded_buffer
```
andychu, line 88: Yes agree, we're seeing that pattern a lot

I think it makes sense to have a debug mode bounds checker, and then have it of in opt mode

mycpp/leaky_types.cc
```diff
@@ -32,6 +160,128 @@ bool OmitChar(uint8_t ch, int what) {
...
+  if (num_fill < 0) {
+    return this;
+  } else {
+    char* buf = static_cast<char*>(malloc(width));
+    char c = fillchar->data_[0];
+    memcpy(buf, data_, len_);
+    for (int i = len_; i < width; ++i) {
+      buf[i] = c;
+    }
+    return CopyBufferIntoNewStr(buf, width);
```
andychu, line 205: Ditto I think all of these will end up as `AllocStr(width)` and then mutating that buffer

I'm a little scared by the malloc(width) because it used to be malloc(width + 1), but AllocStr will solve that problem because it takes string length and not buffer length

andychu (approved):
OK thanks this looks good.  

Please run the auto-formatter -- `test/lint.sh format-oil`


----

I noticed some ASDL test failures in the CI

http://travis-ci.oilshell.org/github-jobs/2022-08-04__15-43-57.wwz/

You can fix those with this patch

```
$ git diff
diff --git a/asdl/test.sh b/asdl/test.sh
index ba9aa5b0..8a28f45c 100755
--- a/asdl/test.sh
+++ b/asdl/test.sh
@@ -52,6 +52,7 @@ gen-cpp-test() {
     asdl/runtime.cc \
     mycpp/mylib_old.cc \
     mycpp/gc_heap.cc \
+    mycpp/leaky_types.cc \
     _build/cpp/hnode_asdl.cc \
     $tmp_dir/typed_arith_asdl.cc \
     $tmp_dir/typed_demo_asdl.cc
@@ -91,6 +92,7 @@ gc-test() {
     mycpp/gc_heap.cc \
     mycpp/gc_builtins.cc \
     mycpp/gc_mylib.cc \
+    mycpp/leaky_types.cc \
     asdl/runtime.cc \
     _build/cpp/hnode_asdl.cc \
     $tmp_dir/demo_lib_asdl.cc \
```

I usually use `patch -p1 < mydiff`    to apply those

jjbandit:
Ready for merge @andychu .  I can't merge it for some reason.

andychu:
Great thanks!

The PRs have to be targetted to `soil-staging` to be merged, the default of `master` won't work.  So what I do is change the base and then merge in Github

I generally want to look at the PRs first, but for these mechanical changes maybe we can do something else ... i.e. we decided that the only thing that matters is tests, because we will make another pass over this

PR #1263, opened 2022-08-05
Jesse/refactor/continue porting old str layout to gcd layout

jjbandit:
I did two things here:

1) Move some code common to both `mylib_old` and `gc_heap` into headers that can be included by both of them
2) Changed all reads of  `Str::len_` to function calls `len(Str*)`

### DIFF
--- a/mycpp/error_types.h
+++ b/mycpp/error_types.h
@@ -1,0 +1,36 @@
+#ifndef ERROR_TYPES_H
+#define ERROR_TYPES_H
+
+class Str;
+
+class IndexError {};
+class ValueError {};
+class KeyError {};
+
+class EOFError {};
+
+class NotImplementedError {
+ public:
+  NotImplementedError() {
+  }
+  explicit NotImplementedError(int i) {  // e.g. in expr_to_ast
+  }
+  explicit NotImplementedError(const char* s) {
+  }
+  explicit NotImplementedError(Str* s) {
+  }
+};
+
+class AssertionError {
+ public:
+  AssertionError() {
+  }
+  explicit AssertionError(int i) {  // e.g. in expr_to_ast
+  }
+  explicit AssertionError(const char* s) {
+  }
+  explicit AssertionError(Str* s) {
+  }
+};
+
+#endif
--- a/mycpp/error_types.h
+++ b/mycpp/error_types.h
@@ -1,7 +1,11 @@
 #ifndef ERROR_TYPES_H
 #define ERROR_TYPES_H
 
+#ifdef LEAKY_BINDINGS
 class Str;
+#else
+using gc_heap::Str;
+#endif
 
 class IndexError {};
 class ValueError {};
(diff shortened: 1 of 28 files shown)


mycpp/error_types.h
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
andychu, line 1: I think we should stick with the #include guards, it looks like it's not standard

https://stackoverflow.com/questions/23696115/is-pragma-once-part-of-the-c11-standard

In practice I'm not sure if it will ever show up, though maybe someone wants to build Oil on a weird embedded system with a weird C++ compiler (if that even exists -- I know there are weird C compilers, not sure about C++)

mycpp/mylib_old.h
```diff
@@ -58,55 +61,17 @@ void print(Str* s);
...
 
-#ifdef MYLIB_LEAKY
+int len(Str *s);
+
 class Str : public gc_heap::Obj {
  public:
   Str(const char* data, int len)
       : gc_heap::Obj(Tag::FixedSize, gc_heap::kZeroMask, 0),
-        len_(len),
+        len__(len),
```
andychu, line 74: Hm this looked weird but I guess it's to expose all instances as compile errors, OK

andychu (commented):
OK cool I skimmed over this ... let's see what the CI says

andychu:
Hm the CI is running twice which is weird, but I pulled this PR and got this error

I think it has to do with LEAKY_TEST_MODE vs LEAKY_BINDINGS

```
$ cpp/test.sh gc-binding-test '' leaky
/tmp/ccOz2RyQ.o: In function `str_equals(Str*, Str*)':
/home/andy/git/oilshell/oil2/mycpp/mylib_old.h:570: undefined reference to `len(Str*)'
/home/andy/git/oilshell/oil2/mycpp/mylib_old.h:570: undefined reference to `len(Str*)'
/home/andy/git/oilshell/oil2/mycpp/mylib_old.h:571: undefined reference to `len(Str*)'
/tmp/ccOz2RyQ.o: In function `qsn::XEscape(Str*)':
/home/andy/git/oilshell/oil2/cpp/qsn.h:45: undefined reference to `len(Str*)'
collect2: error: ld returned 1 exit status

```

mycpp/mylib_old.h
jjbandit: Correct.  I didn't bother changing it back because (hopefully) we'll be deleting this Soon™

mycpp/error_types.h
jjbandit: Yeah I can change them.

jjbandit: I've actually had these include guards bite me before, which is why I prefer `pragma once`, but it's pretty unlikely

andychu:
Cool, it passed the CI now!  woohoo

PR #1264, opened 2022-08-05
Jesse/refactor/move more mylib code to leaky types

jjbandit:
Pretty mechanical change.  I've been `#if 0`ing out the duplicate code in `gc_builtins` instead of deleting it, but pretty much everything in that file can be deleted now, which is good.


andychu:
Oops, we raced ... I reviewed and was about to merge and then the `new Str` change came in ...

Let's see if the CI passes

andychu:
Hm actually something caused a failure here:

http://travis-ci.oilshell.org/github-jobs/2022-08-05__19-02-03.wwz/

http://travis-ci.oilshell.org/github-jobs/2022-08-05__19-02-03.wwz/_tmp/soil/logs/parse-errors.txt

The command to repro is now in the HTML.  Although sometimes it's easier to back out and redo it, rather than debug

jjbandit:
@andychu Any idea why the CI is complaining?  The commands it's complaining about work on my machine

jjbandit:
Ready to rumble @andychu !!

andychu:
OK cool this looks like big progress !!

I'm not sure I agree with some of the naming but we can talk about it later

PR #1265, opened 2022-08-06
Jesse/refactor/move str out of gc heap namespace

jjbandit:
I guess the only thing I really have to say about this is, it's going to get worse before it gets better.  Tests still pass, and I've got a plan for getting closer towards removing the `gc_heap` namespace.  Once that namespace is gone we should be able to collapse a lot of the little files I'm creating back down to bigger ones, but this intermediate step is the only method I've been able to make any progress with.

### DIFF
--- a/devtools/oil_gdb.py
+++ b/devtools/oil_gdb.py
@@ -152,10 +152,6 @@ class TypeLookup(object):
       #print('target name %r' % target.name)
       #print('target tag %r' % target.tag)
 
-      if target.name == 'gc_heap::Str':
-          return GcStrPrinter(val)
-
-      # TODO: remove Str when we remove mylib_old.cc
       if target.name == 'Str':
           return StrPrinter(val)
 
--- a/devtools/oil_gdb.py
+++ b/devtools/oil_gdb.py
@@ -153,7 +153,7 @@ class TypeLookup(object):
       #print('target tag %r' % target.tag)
 
       if target.name == 'Str':
-          return StrPrinter(val)
+          return GcStrPrinter(val)
 
       if target.name in self.sum_type_lookup:
           return AsdlPrinter(val, self.sum_type_lookup[target.name])
--- a/asdl/tool.py
+++ b/asdl/tool.py
@@ -89,7 +89,6 @@ using gc_heap::Obj;
 using gc_heap::Dict;
 using gc_heap::List;
 using gc_heap::NewList;
-using gc_heap::Str;
 using gc_heap::StrFromC;
 #endif
 """)
--- a/cpp/aligned.h
+++ b/cpp/aligned.h
@@ -1,0 +1,19 @@
+#ifndef ALIGNED_H
+#define ALIGNED_H
+
+// I'm not sure why this matters but we get crashes when aligning to 8 bytes.
+// That is annoying.
+// Example: we get a crash in cpp/frontend_flag_spec.cc
+// auto out = new flag_spec::_FlagSpecAndMore();
+//
+// https://stackoverflow.com/questions/52531695/int128-alignment-segment-fault-with-gcc-o-sse-optimize
+constexpr int kMask = alignof(max_align_t) - 1;  // e.g. 15 or 7
+
+// Align returned pointers to the worst case of 8 bytes (64-bit pointers)
+inline size_t aligned(size_t n) {
+  // https://stackoverflow.com/questions/2022179/c-quick-calculation-of-next-multiple-of-4
+  // return (n + 7) & ~7;
+  return (n + kMask) & ~kMask;
+}
+
+#endif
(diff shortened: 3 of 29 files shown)


devtools/oil_gdb.py
```diff
@@ -152,10 +152,6 @@ def __call__(self, val):
       #print('target name %r' % target.name)
       #print('target tag %r' % target.tag)
 
-      if target.name == 'gc_heap::Str':
-          return GcStrPrinter(val)
-
-      # TODO: remove Str when we remove mylib_old.cc
```
andychu, line 158: I think this should be changed to GcStrPrinter, or just revert this file ... we can take care of it when we hit it

andychu (commented):
Hm so you think you can get rid of gc_heap first, and that will make migrating to gc_heap::List and gc_heap::Dict easier?

I'm not sure I see it but if you have a plan that is good :)

I think the thing I was worried about is stuff that's implemented in leaky List / Dict, but UNimplemented in gc_heap::{List,Dict}

i.e. it feels like we could be in this state for awhile

devtools/oil_gdb.py
jjbandit: Updated

jjbandit:
> Hm so you think you can get rid of gc_heap first, and that will make migrating to gc_heap::List and gc_heap::Dict easier?
> 
> I'm not sure I see it but if you have a plan that is good :)
> 
> I think the thing I was worried about is stuff that's implemented in leaky List / Dict, but UNimplemented in gc_heap::{List,Dict}
>


> i.e. it feels like we could be in this state for awhile

It could be, but I'm pretty confidant I can get rid of the gc_heap namespace entirely in another day or two.  At least to the point where we can start collapsing the little includes back down to something closer to what it started as.

jjbandit:
> Hm so you think you can get rid of gc_heap first, and that will make migrating to gc_heap::List and gc_heap::Dict easier?

Yeah, that's correct.  Right now it's tough to add new stuff because there are a few different cases you have to satisfy, which are confusing at best if you're not intimately familiar with how the namespaces work.  It's particularly bad when stuff (like the containers) _change_ namespaces across LEAKY modes.

> I'm not sure I see it but if you have a plan that is good :)
> 
> I think the thing I was worried about is stuff that's implemented in leaky List / Dict, but UNimplemented in gc_heap::{List,Dict}
>

Yeah, we'll get to that once I'm done getting rid of this namespace.  That should actually be pretty fun to tackle.. I like implementing library code.  In terms of actually getting the code to compile .. it should be pretty easy to stub out any differences in APIs.  Though I'm not even sure that'll be necessary.

andychu:
Hm it seems like the new commit broke some stuff with cpp/aligned.h

http://travis-ci.oilshell.org/github-jobs/2022-08-06__19-47-17.wwz/

I'm OK with having small headers for awhile, to organize the refactoring, but yeah we should talk about the final structure at some point

I'd be OK with gc_str, gc_list, etc. but the idea was for there to be gc_types, gc_heap, gc_builtins, etc.

jjbandit:
> Hm it seems like the new commit broke some stuff with cpp/aligned.h
> 
> http://travis-ci.oilshell.org/github-jobs/2022-08-06__19-47-17.wwz/
>

Yeah I forgot to commit a file.  Rebuilding now.

> I'm OK with having small headers for awhile, to organize the refactoring, but yeah we should talk about the final structure at some point
> 
> I'd be OK with gc_str, gc_list, etc. but the idea was for there to be gc_types, gc_heap, gc_builtins, etc.

Yeah I remember we discussed/agreed on a strategy, which I tried really hard to stick to, but unfortunately it just didn't provide the level of granularity necessary to do this change.  We can go back to that eventually but for the moment I need to create a bunch of files.

The issue is that I'm de-duplicating code, which is interdependent on _other_ stuff within context, so it's really hard (maybe actually impossible) to just relocate it to a common place and have all the namespacing stuff work out.  Anyhow, thanks for coming to my TED talk.

jjbandit:
> > Hm it seems like the new commit broke some stuff with cpp/aligned.h
> > http://travis-ci.oilshell.org/github-jobs/2022-08-06__19-47-17.wwz/
> 
> Yeah I forgot to commit a file. Rebuilding now.
> 
> > I'm OK with having small headers for awhile, to organize the refactoring, but yeah we should talk about the final structure at some point
> > I'd be OK with gc_str, gc_list, etc. but the idea was for there to be gc_types, gc_heap, gc_builtins, etc.
> 
> Yeah I remember we discussed/agreed on a strategy, which I tried really hard to stick to, but unfortunately it just didn't provide the level of granularity necessary to do this change. We can go back to that eventually but for the moment I need to create a bunch of files.
> 
> The issue is that I'm de-duplicating code, which is interdependent on _other_ stuff within context, so it's really hard (maybe actually impossible, idk) to just relocate it to a common place and have all the namespacing stuff work out. Anyhow, thanks for coming to my TED talk.

andychu:
OK it passed the tests, so I merged it ... looks weird now but it's better to make progress

I should also figure out why Github Actions is running the duplicate checks ...

PR #1268, opened 2022-08-07
Jesse/refactor/move all leaky and nonleaky code into well defined places

jjbandit:
The big change here is having gotten enough common code out of the "mylib_old" and "gc_heap" files that they can actually now be renamed to "leaky_containers" and "gc_containers".  They're now self-contained implementations of their leaky/gc'd containers, and it's an error to try to compile gc'd containers against leaky mode.

That unlocked removing the 'gc_heap' namespace, which became trivial.


asdl/runtime.cc
```diff
@@ -4,19 +4,18 @@
 
 #ifdef LEAKY_BINDINGS
 #include "mycpp/mylib_old.h"
-using gc_heap::Alloc;
-using gc_heap::kZeroMask;
+
```
andychu, line 7: Oh do you know this file is separately built?

It's `build/translate.sh asdl-runtime`

the shell scripts might need to change, I can help with that

asdl/gc_test.cc
```diff
@@ -44,7 +43,7 @@ TEST pretty_print_test() {
 GREATEST_MAIN_DEFS();
 
 int main(int argc, char** argv) {
-  gc_heap::gHeap.Init(1 << 20);
+  gHeap.Init(1 << 20);
```
andychu, line 46: Hm I kind of like `gHeap` and `Obj` in a `gc_heap` namespace, but it's not that big a deal

On the other hand, Str,List,Dict are used by  generated code  (and duplicated)

cpp/NINJA-steps.sh
```diff
@@ -225,12 +225,16 @@ compile_and_link() {
...
+
   setglobal_compile_flags "$variant" "$more_cxx_flags" ""  # no dotd
 
   setglobal_link_flags $variant
 
   setglobal_cxx $compiler
 
+  echo ""
+  echo "Building: $cxx -o $out $flags $@ $link_flags"
+  echo ""
```
andychu, line 237: this will probably make the build a lot noiser

mycpp/gc_builtins.h
```diff
@@ -109,10 +107,16 @@ inline bool _cmp(Str* a, Str* b) {
   return str_cmp(a, b) < 0;
 }
 
+// TODO(Jesse): What does the following comment mean?  How does including
+// <algorithm> bloat types?  gc_heap.h already includes a ton of C++ headers so
+// I don't know how not including algorithm is a win.  Maybe this comment is
+// old..?
```
andychu, line 113: oh I think I just meant it would bloat the ASDL translation units, not the types

In general I want to avoid too much STL code in headers, but it's not our top priority and this comment can probably be deleted

We're re-organizing the code anyway which makes it somewhat obsolete

andychu (commented):
OK I agree with getting rid of MYLIB_LEAKY, that was definitely a smell 

If everything still works, it seems like a good intermediate step

cpp/NINJA-steps.sh
jjbandit: Oh yeah, I've been using it as a crutch for debugging builds.  I can comment it out before we merge this.

asdl/gc_test.cc
jjbandit: Yeah I was thinking it might make sense to namespace a couple _actual_ heap-related things like the allocator, etc.

asdl/runtime.cc
jjbandit: Looks like it still builds

PR #1270, opened 2022-08-08
Jesse/refactor/remove placement new for str

scallyw4g:
Not sure if we want to merge this @andychu, but this is one way of statically denying the use of the `new` operator.

Worth noting that we can't disallow the `Obj` constructor at the present time because that code still gets called in the oldstl paths .. ie during `new List` and friends.

WDYT?

### DIFF
--- a/mycpp/gc_builtins.cc
+++ b/mycpp/gc_builtins.cc
@@ -386,10 +386,10 @@ List<Str*>* Str::split(Str* sep) {
     int prev_pos = breaks[i - 1];
     int part_len = breaks[i] - prev_pos - 1;
     if (part_len > 0) {
+
       // like AllocStr(), but IN PLACE
-      int obj_len = kStrHeaderSize + part_len + 1;  // NUL terminator
-      Str* part = new (place) Str();                // placement new
-      part->SetObjLen(obj_len);                     // So the GC can copy it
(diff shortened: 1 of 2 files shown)


andychu:
Hmm, not sure ... I think we could just manually move `new` out of `gc_*` into `leaky_*`.   Actually  `grep -w new mycpp/gc_*` does a decent job !  There's only a handful

jjbandit:
The issue I'm getting at here is that 'placement new' is allowed, but regular 'new' is not.  Using regular 'new' results in .. best case a memory leak, worst case the GC will try to relocate/chase pointers in the struct and we get crashes or _very_ hard bugs.  The very hard bugs I'm thinking of are things like:

```

void main()
{
  BufferWriter *bufWriter = new BufWriter();

... enough code to forget bufWriter was `new` allocated ...

  functionThatTakesBufWriter(bufWriter);
}


void functionThatTakesBufWriter(BufWriter *bufWriter)
{
  StackRoots _roots({&bufWriter}); // << Very tricky bug.
}
```

Statically disallowing all use of `new` prevents us ever having the very-hard-bugs problem.  The code looks pretty similar, we just have "initializer functions" instead of calling constructors via `new`.

Noodle on it.. but I'd like to find a way of statically knowing that bug (and its friends) will never happen.

PR #1282, opened 2022-08-18
Remove OLDSTL list and consolidate code

jjbandit:
This removes the oldstl List implementation in favor of using the gc List


andychu:
OK thanks, this looks good progress ... I think it will be easier to talk on VC about some things

- the structure is a little weird, e.g. #including implementations. 
- the List::append() issue.

jjbandit:
Sure thing, I'm available today anytime.

Clif notes on:

1) #including implentations

I wanted to keep this PR to strictly the least number of possible mechanical changes to remove oldstl List.  My plan was to, once I'm done with Dict, make another pass over everything to merge all the small files I've produced into a larger, more sensible structure.

2) List::append

This is one occurrence of a more pervasive issue than I actually documented; there are a number of `Str*` specific specializations for a lot of templates, and more generic ones that are "intended" for primitive types (`int`, `float`, etc), but can (and do) get selected for heap allocated pointer types.  This is what lead to that pointer equality problem I fixed with Tuples.

We should definitely discuss fixing this;  it's fertile ground for very annoying bugs as there's currently no compile time or runtime checking for those bugs.

mycpp/comparator_impls.h
```diff
@@ -0,0 +1,68 @@
+#ifndef COMPARATOR_IMPLS_H
+#define COMPARATOR_IMPLS_H
+
```
andychu, line 3: I think this whole file can go in leaky_builtins.cc

I ran into the inline issue as well, I think that solve it

mycpp/gc_builtins.cc
```diff
@@ -335,114 +335,7 @@ Str* Str::replace(Str* old, Str* new_str) {
...
-      InvalidCodePath();
-    }
-    memcpy(p_result, items->index_(i)->data_, n);  // copy the list item
-    p_result += n;
-  }
-
-  assert(p_result[result_len] == '\0');  // GC should zero it
-  return result;
-}
+#include "mycpp/list_str_impl.h"
```
andychu, line 338: everything in list_str_impl.h can be in `leaky_builtins.cc`

comment on Zulip: "leaky" really means "common between OLDSTL and GC"

mycpp/gc_heap.h
```diff
@@ -188,6 +188,9 @@ class Heap {
     int n = aligned(num_bytes);
     // log("n = %d, p = %p", n, p);
 
+    // TODO(Jesse): Change to `assert(n >= sizeof(LayoutForwarded))`
+    assert(n >= 16);
```
andychu, line 192: this the size of Obj header + the required forwarded pointer, which is the minimum size a GC object could be


comment or use constant

sizof(Obj) + size(void*)

andychu (commented):
adjacent #includes probably need

```
// clang-format off
#include a.h
#include b.h
// clang-format on
```

then reformat with `test/lint.sh format-oil`

andychu:
Update description

- OLDSTL
- bug with i+=1
- assert on minimum allocation
- anything else

mycpp/list_str_impl.h
```diff
@@ -0,0 +1,110 @@
...
+
+  result = NewList<Str*>({});
+
+  int n = len(self);
+  int pos = 0;
+  int end = n;
+
+  while (true) {
+    // NOTE(Jesse): Perfect use case for cursor
+    int new_pos = find_next_occurance_of(self->data_, pos, end, sep_char);
```
andychu, line 88: occurrence

or  just

FindNext()

jjbandit:
Alrighty -- addressed the PR feedback.. should be good to merge so I can get started on Dict

andychu:
OK thanks, but please send out a PR to reformat the code and add the `// clang-format` stuff as mentioned

PR #1286, opened 2022-08-29
Fix bad increment in an iterator block

jjbandit:


jjbandit:
I was working off of `master`, which for some reason still has this bug.   Closing

PR #1288, opened 2022-08-30
Jesse/refactor/remove oldstl gc list

jjbandit:


jjbandit:
The vast majority of this PR is just shuffling code around and deleting dead or duplicate code.

I had to rewrite a portion of `execve` found in `leaky_stdlib.cc`.  It iterated over a dict by accessing the original vector.  Changing to DictIter() was sufficient.

A mechanical but semantic change I made was to rename `OLDSTL_BINDINGS` -> `LEAKY_ALLOCATOR`, because that's what it now represents.

I also started work on grooming the include graph because, frankly, it was a disaster.  It was impossible to tell what files include what, and when, because everything included everything else in a seemingly random order.  A while ago I realized that every file was included in every build, just some were (basically) entirely `#if`'d out.  I collected all the `#ifdef`s scattered about the codebase into a single file `runtime.h`, which includes everything a program (test or otherwise) might need to write against the API we're narrowing in on.

I also propose we get rid of the `LEAKY_ALLOCATOR` define and put it on a runtime switch so we can cut down on the build variants.  It's now on the order of a 1-line change to do that.

I didn't rename the `OLDSTL_RUNTIME` stuff in ninja_subgraph.py because I couldn't actually figure out how to do it and make it work.  Related, I didn't rename `oldstl_containers.cc`.  That file has a single `mylib` function left in it that needs to be relocated to it's appropriate home, but again, that's for a future PR.

Worth noting that (before this PR) `_tmp/spec/cpp/builtin-io.html` started getting garbage written into it (a single 0 byte), which is why it now shows up as a binary file in `test/baseline.spec-cpp.assertion_fails`.  This smells like another bad memory access or overwrite.  Maybe you could get to the bottom of that one @andychu .

mycpp/gc_mylib.cc
```diff
@@ -1,16 +1,31 @@
...
 
 #include <errno.h>
 #include <unistd.h>  // isatty
 
-#include "gc_builtins.h"
-
 mylib::BufWriter gBuf;
 
+
+// NOTE(Jesse): This was literally the only thing left in gc_builtins.cc so I
```
andychu, line 11: I think we should still have some sort of `builtins.{cc,h} `   file

i.e. the structure of containers/builtins/mylib still makes sense

andychu (commented):
OK this looks like good progress, the coverage and spec test numbers look good.

The builtin-io thing could either be a concurrency problem or a local change ... The CI runs `test/spec-cpp` serially, so it's possible concurrency introduces a weird bug.  Maybe we should consider running it into a container

OK I will merge this and then look around at what we need to do

PR #1296, opened 2022-09-01
Jesse/feature/use mmap always

jjbandit:
Pretty straight-forward change which simplified the Heap struct quite a bit.

I was able to remove `GC_NO_HACK` and `GC_PROTECT` by virtue of using `mmap()`.

First, we never need to clear pages because the OS does that when they're faulted in.

Second, we never need to call memprotect() because we jettison the entire mapping every time the collector hits.  We maintain the invariant that the program crashes if anyone tries to access the old zone; munmap makes sure of that.

### DIFF
--- a/cpp/NINJA-steps.sh
+++ b/cpp/NINJA-steps.sh
@@ -90,11 +90,11 @@ setglobal_compile_flags() {
 
     (gcstats)
       # unit tests use for gHeap.Report()
-      flags="$flags -g -D GC_PROTECT -D GC_STATS"
+      flags="$flags -g -D GC_STATS"
       ;;
 
     (gcevery)
-      flags="$flags -g -D GC_PROTECT -D GC_STATS -D GC_EVERY_ALLOC"
+      flags="$flags -g -D GC_STATS -D GC_EVERY_ALLOC"
       ;;
 
     (oldstl)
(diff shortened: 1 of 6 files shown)


jjbandit:
I have no idea why the CI is failing on this @andychu .. I'm getting failures in `mycpp/TEST.sh test-translator` for reasons that are very opaque to me

andychu:
I think this is  a real failure, from the log:

```
[41/192] example-task containers C++ _bin/cxx-opt/mycpp/examples/containers.mycpp _test/tasks/benchmark/containers.opt.task.txt _test/tasks/benchmark/containers.opt.log.txt
FAILED: _test/tasks/benchmark/containers.opt.task.txt _test/tasks/benchmark/containers.opt.log.txt 
```

then go to:

http://travis-ci.oilshell.org/github-jobs/2022-09-01__19-47-09.wwz/_test/tasks/benchmark/containers.opt.log.txt

Let me know if you can't reproduce locally

remember there is the `BENCHMARK=1` environment variable on each of the binaries

andychu:
The general idea looks like it makes sense, I'm not sure why the crash ... I would try running the all the benchmarks and all the tests and see if there is a pattern?

The benchmarks are different than the tests because they tend to allocate more and cause resizing

So that could be the cause of the bug ... it's possible it's exposing a latent bug

andychu:
This one might conflict a little ...

https://github.com/oilshell/oil/pull/1297

It exposes the 200 ms pause problem in debug mode, but it actually didn't cause a problem with tests, since most of them run in opt mode on CI.  (but not locally!   so you may see the delay)

andychu:
OK I think this boils down to needing a better test runner, which I queried about  yesterday:

https://lobste.rs/s/belqch/how_would_you_run_python_c_tests_parallel#c_eq5u6g

The error there but hidden in other errors, let's talk about it on VC

PR #1297, opened 2022-09-02
[mycpp runtime] Remove oldstl, -D LEAKY_ALLOCATOR, -D NO_GC_HACK

andychu:
- Deleted mycpp/leaky_heap.h
- Ran into issue with global NewList<> in _gen/frontend/consts.cc.
  Generate a GLOBAL_LIST() instead.
- Add an assertion to find Alloc<> before gHeap.Init()
- Get rid of NO_GC_HACK because the ASAN variant breaks with it!
  - Now we have a pause in cxx-dbg mode to get rid of (with mmap()).


andychu:
I started getting rid of OLDSTL, and then it turns out I had to also get rid of LEAKY_ALLOCATOR and NO_GC_HACK

It wasn't too bad

But now there is the 200 ms pause in `cxx-dbg` mode that `mmap()` should fix!

Let's talk about this tomorrow, along with the code reorg  ... the logical next step is the `mmap()`

andychu:
Spec test numbers unchanged: http://travis-ci.oilshell.org/github-jobs/2022-09-02__06-39-06.wwz/_tmp/spec/cpp/osh-summary.html

PR #1299, opened 2022-09-02
Redo mmap tricks

jjbandit:
Same change as before.  Doesn't crash anymore

### DIFF
--- a/mycpp/gc_heap.h
+++ b/mycpp/gc_heap.h
@@ -94,7 +94,6 @@
 // #defines for degbugging:
 //
 // GC_EVERY_ALLOC: Collect() on every Allocate().  Exposes many bugs!
-// GC_PROTECT: Use mprotect()
 // GC_VERBOSE: Log when we collect
 // GC_STATS: Collect more stats.  TODO: Rename this?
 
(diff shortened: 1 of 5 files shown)


mycpp/gc_heap.h
```diff
@@ -220,17 +203,10 @@ class Heap {
...
-    // succeed, copy the heap to it, then allocate the object.
+    // It's still too small.  Grow the heap.
     int multiple = 2;
-    while (from_space_.size_ + n > to_space_.size_ * multiple) {
-      multiple *= 2;
-    }
-    // log("=== FORCED by multiple of %d", multiple);
-    to_space_.Free();
-    to_space_.Init(to_space_.size_ * multiple);
+    Collect( (from_space_.size_+n) * multiple );
```
andychu, line 208: I think the `while` needs to be there ... 

Consider if you have a 256 KiB heap and then the next allocation is 1 MiB.  Then a single doubling won't be enough

We should probably have a test for that case, but it won't be an issue for mark and sweep, so meh

jjbandit: I added `n`, then doubled it, so we'd get 2.5MiB in the case you mentioned

andychu: OK I see... I'm not sure this an improvement, because the intention was that the heaps are always powers of 2.  They just keep doubling.

mmap() is going to allocate whole pages anyway, which are powers of 2.

So this is more complex because you could ask for 999 bytes for a string, which will get rounded up by the allocator, and then there's another implicit round up in mmap ...

But it does work and we're probably not using this code in the future, so OK

PR #1305, opened 2022-09-06
Jesse/feature/mark sweep

jjbandit:


mycpp/smartptr.h
```diff
@@ -2,8 +2,6 @@
 
 #include <assert.h>
 
-#include "mycpp/gc_heap.h"
```
andychu, line 5: Please revert this, since when `clang-format` sorts the #includes it causes a build error

The smartptr.h file uses gHeap, which is in gc_heap.h, so it should be included

I tried reverting the style change to the test only, and it caused a build error, so both files should be reverted

I think smartptr* has nothing to do with this change anyway -- it's unused in the codebase.


clang-format is basically enforcing this style:

https://github.com/include-what-you-use/include-what-you-use

andychu:
OK I reverted the smartptr changes, but there is this failure in building the tarball:

http://travis-ci.oilshell.org/github-jobs/2022-09-06__00-30-06.wwz/_tmp/soil/logs/test-tar.txt

Because the build system has to know which files are in the GC runtime, which can be changed here:

https://github.com/oilshell/oil/blob/master/mycpp/NINJA_subgraph.py#L180

i.e. the cheney_heap.cc file is not in the tarball now.  I think it's better just to add both files to the tarball and build.  We shouldn't be #including .cc files.

You might have to use two different class names, e.g. add MarkSweepHeap with an `#ifdef`

andychu:
Ugh now the build is failing worse ... I can see now that those weren't just style changes, but I think we need to agree on the use of the preprocessor

Now that we are done with the GC/OLDSTL unification, we should stick with `clang-format` and the  `.cc` and `.h`  structure it enforces

One of the main points of that style is that reordering headers shouldn't introduce build problems, which is something that happened here!

andychu:
FWIW the style of having #include "smartptr.h" be the first special #include    in smartptr.cc or smartptr_test.cc comes from IWYU

https://www.fluentcpp.com/2021/01/01/include-what-you-use/

https://news.ycombinator.com/item?id=26875353

i.e. that matches my experience with using it to reduce compilation times

jjbandit:
Maybe we should hop on a call to discuss?  This turned into a long response.  Sounds like you may be sharing some of my frustration with the build.

> One of the main points of that style is that reordering headers shouldn't introduce build problems, which is something that happened here!

I agree with that in principal, but there's a long way to go in practice for Oil.  The types have many serial dependencies, so include order matters.  It's a big job to go through the codebase to figure out what depends on what.  I've started doing that with `runtime.h`, but again, there's a long way to go.

> there is this failure in building the tarball [...] Because the build system has to know which files are in the GC runtime, which can be changed here:
> https://github.com/oilshell/oil/blob/master/mycpp/NINJA_subgraph.py#L180
> i.e. the cheney_heap.cc file is not in the tarball now. I think it's better just to add both files to the tarball and build. We shouldn't be #including .cc files.

First off, running the tarball script fails on my machine.  It does not specify what other scripts(s) I should run that set it up for success.  I tried running the `make-tar` task, which also fails with an unhelpful error.  I also looked through the output on Travis for hints but came up empty-handed.

I am unsure how to proceed here; I do not understand this part of the build system.  After reading the script you linked to (and the script it generates) I have no further ideas of what I might do to get the build to cooperate.  To me, it _looks_ like it should work.

The reasons I directly `#include`-ed the `.cc` files are:

1) The `MARK_SWEEP` switch was already there, so I didn't have to touch the build.
2) My rationale was that if I made a surgical change and left all the gc_heap code intact, I should be able to flip between the mark/sweep and cheney collectors by just changing the `MARK_SWEEP` define.  Also, doing it this way, I would never have to touch the build.
3) I didn't know what to add to the build to do it 'properly', so I chose the simple way that 'should work'.  Somehow, this extremely simple change _still_ managed to break the build.

> You might have to use two different class names, e.g. add MarkSweepHeap with an #ifdef

I strongly disagree.  Using the same class name will allow us to flip back-and-forth between mark/sweep and cheney, if we choose to do so.  Furthermore, I just spent a month untangling a pile of `#ifdef`s and I don't exactly love the thought of adding more.

> I think we need to agree on the use of the preprocessor

Agreed.  I think the main question I need answered to get us on the same page is:

"How do I put new code on a switch such that I can easily start porting old implementations over to a new format?"

jjbandit:
Maybe you're onto something with that pair-programming idea.

You can answer my questions about the build, and see how many problems I actually run into with it.

andychu:
OK yeah let's talk about it on a call ... I think we are pretty much in the state where we can use the "traditional" #include style, maybe you have worked on projects with different styles (e.g. unity builds)

andychu:
FWIW the structure I imagine is:

```
#ifdef MARK_SWEEP
  extern MarkSweepHeap gHeap;
#else
  extern CheneyHeap gHeap;
#endif
```

in the header, and then


```
#ifdef MARK_SWEEP
  MarkSweepHeap gHeap;
#else
  CheneyHeap gHeap;
#endif
```

in the C++ file

And then in `mycpp/NINJA_subgraph`, just add `cheney_heap.cc` and `marksweep_cc` to the lists

And then add the `_test.cc` files to the dictionary

But we can talk about it, because I'm curious why this isn't the "default"

Maybe it's a difference between C background vs. C++ background

jjbandit:
@andychu Looks like another case of flaky CI here?

andychu:
Oh yeah that is just github being down ... can't really do much about that

http://travis-ci.oilshell.org/github-jobs/2022-09-06__22-00-38.wwz/_tmp/soil/logs/ble-clone.txt

The logs are linked from the Github Actions `publish-html` output ... I removed a ton of spew the other day, let me remove a little more so it's clearer.

Then there will be more of a direct link from Github Actions to https://travis-ci.oilshell.org/

andychu:
OK I removed spew in a couple places, so next time it should be easy to click through from the PR -> GH Actions -> Our logs

Should I look at this now, or should we discuss it separately? 

I can probably meet tomorrow but there is a  chance I need to take care of some personal stuff

jjbandit:
Yeah it's ready @andychu

mycpp/gc_obj.h
```diff
@@ -0,0 +1,104 @@
+// Obj::heap_tag_ values.  They're odd numbers to distinguish them from vtable
```
andychu, line 1: Please add `#include` guards on all header files

Looks like at least gc_heap.h and gc_obj.h don't have them

There is actually a `cpp/run.sh` script which has a boilerplate generator

mycpp/gc_heap.h
```diff
@@ -1,285 +1,37 @@
...
-#endif
-
-  Space from_space_;  // space we allocate from
-  Space to_space_;    // space that the collector copies to
-
-  char* free_;   // next place to allocate, from_space_ <= free_ < limit_
-  char* limit_;  // end of space we're allocating from
+#define Terabytes(bytes) (Gigabytes(bytes) * 1024)
+#define Gigabytes(bytes) (Megabytes(bytes) * 1024)
+#define Megabytes(bytes) (Kilobytes(bytes) * 1024)
```
andychu, line 9: This  is gonna sound real pedantic, but we should distinguish between powers of 2 (KiB, MiB, GiB) and powers of 10 (KB, MB, GB)

So I would change this to `MiB` and `KiB`   (and delete G and T if we're not using them)

Kilobytes are powers of 10, just like Kilograms are (yes this changed some time in the last 20 years)

Some people don't care about the difference, but when you're measuring storage and networking speeds, it leads to real bugs and real outages, so I always keep a strict distinction

(Also it is annoying when you're buying a hard drive, etc.)

mycpp/gc_heap.h
```diff
@@ -1,285 +1,37 @@
...
-  // has better locality.
-  //
-  // femtolisp uses a global pointer to dynamically-allocated growable array,
-  // with initial N_STACK = 262144!  Kind of arbitrary.
+const int kMaxRoots = Kilobytes(4);
 
-  int roots_top_;
-  Obj** roots_[kMaxRoots];  // These are pointers to Obj* pointers
+#include "cheney_heap.h"
+#include "marksweep_heap.h"
```
andychu, line 15: Please put all the #includes at the top of the file (or delete if they don't need to be here? )  If it's just for the `extern`, there is no #include needed

I expanded this section:

https://github.com/oilshell/oil/wiki/Oil-Ninja-Build#include-style

mycpp/marksweep_gc_test.cc
```diff
@@ -0,0 +1,92 @@
...
+#include "mycpp/runtime.h"
+#include "vendor/greatest.h"
+
+TEST string_collection_test() {
+  Str *test_str = StrFromC("foo");
+  {
+    // NOTE(Jesse): This causes a crash when this gets compiled against the
+    // cheney collector w/ GC_EVERY_ALLOC.  I did verify it doesn't crash with
+    // the marksweep allocator but didn't want to figure out how to tell the
+    // build system to not compile these tests against the cheney collector
```
andychu, line 11: OK I think we need to figure out something around `dbg` mode, and `gcevery` variant, which uses the Cheney collector

One thing we could do is change the default??   Like `marksweep` is the default?

Maybe we're ready to do that

mycpp/marksweep_heap.h
```diff
@@ -0,0 +1,44 @@
...
+  Obj** roots_[kMaxRoots];  // These are pointers to Obj* pointers
+
+  u64 current_heap_bytes;
+  u64 max_heap_bytes;
+
+  // TODO(Jesse): This should really be in an 'internal' build
+  bool is_initialized_ =
+      true;  // mark/sweep heap doesn't need to be initialized
+
+  std::unordered_set<void*> AllAllocations;
```
andychu, line 40: Hm do you think this is better than a linked list? (which is traditional)


I think it probably does simplify some things, e.g. having extra linked list node types

mycpp/marksweep_heap.h
```diff
@@ -0,0 +1,44 @@
+#ifndef MARKSWEEP_H
+#define MARKSWEEP_H
+
+#include <new>
+#include <unordered_set>
+
+typedef uint64_t u64;
```
andychu, line 7: Please be consistent with the codebase and just use uint64_t

mycpp/marksweep_heap.h
```diff
@@ -0,0 +1,44 @@
...
+
+  u64 current_heap_bytes;
+  u64 max_heap_bytes;
+
+  // TODO(Jesse): This should really be in an 'internal' build
+  bool is_initialized_ =
+      true;  // mark/sweep heap doesn't need to be initialized
+
+  std::unordered_set<void*> AllAllocations;
+  std::unordered_set<void*> MarkedAllocations;
```
andychu, line 41: Please following the existing style of using member names like:

```
marked_allocations_
max_heap_bytes_
```

i.e. lower_case_ with trailing underscore

mycpp/marksweep_heap.cc
```diff
@@ -0,0 +1,114 @@
+#include "mycpp/runtime.h"
+
+void MarkSweepHeap::Init(int max_heap_bytes) {
```
andychu, line 3: It seems like there are a few interface mis-matches, like this one, and the `Collect(num_bytes)`, etc.

Maybe put a TODO there

mycpp/marksweep_heap.cc
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void* MarkSweepHeap::Allocate(int byte_count) {
+  void* Result = calloc(byte_count, 1);
+  assert(Result);
+
+  this->AllAllocations.insert(Result);
+
+  this->current_heap_bytes += byte_count;
+  if (this->current_heap_bytes > this->max_heap_bytes) {
```
andychu, line 14: What is the rationale behind the collection policy?

I think this might be OK to start with, but using a number of bytes doesn't seem obvious

Also I wouldn't call it "max_heap_bytes" as there is no maximum !  It's more like `collection_threshold_` or something

mycpp/marksweep_heap.cc
```diff
@@ -0,0 +1,114 @@
...
+    Obj* Root = this->roots_[root_index][0];
+
+    MarkAllReferences(Root);
+  }
+
+  for (auto it = AllAllocations.begin(); it != AllAllocations.end(); ++it) {
+    void* alloc = *it;
+
+    auto marked_alloc = MarkedAllocations.find(alloc);
+    bool alloc_is_dead = marked_alloc == MarkedAllocations.end();
```
andychu, line 95: seems simpler to just get rid of this var and put the `==` in the `if`

mycpp/marksweep_heap.cc
```diff
@@ -0,0 +1,114 @@
...
+
+    if (alloc_is_dead) {
+      free(alloc);
+    }
+  }
+
+  AllAllocations.clear();
+
+  for (auto it = MarkedAllocations.begin(); it != MarkedAllocations.end();
```
andychu, line 104: OK I think this could be a reason the linked list would be faster ... then we don't have to re-add all the live allocations to this set

andychu (commented):
Thanks, this looks like great progress!  Glad you were able to get something working so quickly

There are a lot of comments about style -- I will be more picky about this in the future, since the code is approaching "what it will look like" (no more OLDSTL experiment!), and because we're going to have more people in the codebase.

So it helps everyone read the code if it has a consistent style (e.g. member variable names, lower case, etc.)

Maybe I should set up `cpplint.py`, which basically enforces this: https://google.github.io/styleguide/cppguide.html

-----

The more substantive comments are about:

1. linked list vs. unordered_set for the list of all allocations
2. collection policy

We can talk about that on video if it's easier

I also want to look at exactly what tests are passing now, and think about how to do benchmarking

andychu:
I think a good thing to do pair programming on would actually be benchmarks, since it will involve a lot of shell + Python scripts

This is a great start, and as discussed we need some measurements to figure out where to go next

mycpp/marksweep_heap.cc
```diff
@@ -0,0 +1,114 @@
...
+    this->max_heap_bytes = this->current_heap_bytes * 3 / 2;
+  }
+
+  return Result;
+}
+
+void MarkSweepHeap::MarkAllReferences(Obj* obj) {
+  auto header = ObjHeader(obj);
+
+  this->MarkedAllocations.insert((void*)obj);
```
andychu, line 35: Please use C++ style casts, e.g. static_cast or reinterpret_cast 

I think that is consistent everywhere in the codebase

andychu:
Also not strictly related to this change, but another issue I've been thinking about is array vs. linked list for the root set

Most GCs I've seen seem to use a linked list for the root set, even though to me that seems slower

It avoids the bug you fixed where you exceed the maximum number of roots though!

The problem though is that we modify that list A LOT, every time `StackRoots` is called, which is twice in every function!

Another thing we could use benchmarks for: https://github.com/oilshell/oil/issues/1309

andychu:
Oh yeah another issue is what we talked about on this thread:

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Next.20on.20GC

The GC unit tests don't run under `sweepasan`

I just created that build variant, and didn't get the tests to pass

So you said you just got one test to pass while developing this?

Maye add that to `mycpp/TEST.sh test-runtime` so it doesn't regress

mycpp/gc_obj.h
jjbandit: Fixed

mycpp/gc_heap.h
jjbandit: Changed.  I left `GiB` in because I'll certainly use it for testing in the future.

mycpp/gc_heap.h
jjbandit: Fixed.

mycpp/marksweep_heap.h
jjbandit: The reason a linked-list is traditional is because people traditionally write their own allocators.  If you do that, the easiest method to keep track of free blocks is to use a linked list.

In _our_ case, `all_allocations_` could be either a hashtable or linked-list with minor tradeoffs in terms of memory and speed.

In _all_ cases, `marked_allocations_` _must_ be a hashtable (or some other structure with fast lookup) because we look up into it for every element in `all_allocations_`.

I made them both hashtables because we could see a (probably nominal) speedup if, at the end of collection, we clear `all_allocations_`, then swap the memory they point to.  Currently that swap is done in a loop at the end of the function.  I'm guessing that loop is going to be so fast (most of the time) it won't even show up on profiles, but decided I might as well leave the door open in case I'm wrong.

mycpp/marksweep_heap.h
jjbandit: Fixed

mycpp/marksweep_heap.h
jjbandit: Fixed

mycpp/marksweep_heap.cc
jjbandit: I'm not sure what you mean by 'interface mismatches' here.  Is it that the variable names are different?

The `Collect()` interface is semantically different, but there's no way to make them identical.  I put an assertion and a big block comment to explain it.

mycpp/marksweep_heap.cc
jjbandit: > What is the rationale behind the collection policy?

I typed something completely arbitrary.  We need to profile before we make decisions here.

> I think this might be OK to start with, but using a number of bytes doesn't seem obvious

I'm not sure how using a number of bytes is non-obvious.  You set a target heap size when you initialize the collector, then it tries to stay under that.  What would be more intuitive?

> Also I wouldn't call it "max_heap_bytes" as there is no maximum ! It's more like `collection_threshold_` or something

Good point, changed.

mycpp/marksweep_heap.cc
jjbandit: Put note on this previously.  This loop can go away

mycpp/marksweep_heap.cc
jjbandit: Fixed

mycpp/marksweep_gc_test.cc
jjbandit: > Also not strictly related to this change, but another issue I've been thinking about is array vs. linked list for the root set
> 
> Most GCs I've seen seem to use a linked list for the root set, even though to me that seems slower
> 
> It avoids the bug you fixed where you exceed the maximum number of roots though! (or actually not fixed, since we still have a max number, it's just higher)
> 
> The problem though is that we modify the root set A LOT, every time `StackRoots` is called, which is twice in every function!
> 
> Another thing we could use benchmarks for: #1309

Yeah I think what we want is a block allocator with a freelist that allocates blocks of, for instance, 64 cache boundary aligned pointers with a `Next` member to another block.  When we pop enough pointers to clear an entire block it goes on a freelist.  When we push 63 pointers and need another block we take one from the freelist, or `malloc` another one.

ie:
```
struct StackRootBlock
{
  void ** Pointers[63];
  StackRootsBlock *Next;
};

struct StackRootsBlockFreelist
{
  StackRootsBlock *FirstFreeBlock;
};

class StackRoots
{
  ... push/pop functions .. whatever ..

  StackRootsBlock *CurrentBlock;
  StackRootsFreelist BlockFreelist;
};
```

jjbandit:
> Also not strictly related to this change, but another issue I've been thinking about is array vs. linked list for the root set
> 
> Most GCs I've seen seem to use a linked list for the root set, even though to me that seems slower
> 
> It avoids the bug you fixed where you exceed the maximum number of roots though! (or actually not fixed, since we still have a max number, it's just higher)
> 
> The problem though is that we modify the root set A LOT, every time `StackRoots` is called, which is twice in every function!
> 
> Another thing we could use benchmarks for: #1309

Yeah I think what we want is a block allocator with a freelist that allocates blocks of, for instance, 64 cache boundary aligned pointers with a `Next` member to another block.  When we pop enough pointers to clear an entire block it goes on a freelist.  When we push 63 pointers and need another block we take one from the freelist, or `malloc` another one.

ie:
```
struct StackRootBlock
{
  void ** Pointers[63];
  StackRootsBlock *Next;
};

struct StackRootsFreelist
{
  StackRootsBlock *FirstFreeBlock;
};

class StackRoots
{
  ... push/pop functions .. whatever ..

  StackRootsBlock *CurrentBlock;
  StackRootsFreelist BlockFreelist;
};
```

jjbandit:
> Oh yeah another issue is what we talked about on this thread:
> 
> https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Next.20on.20GC
> 
> The GC unit tests don't run under `sweepasan`
> 
> I just created that build variant, and didn't get the tests to pass
> 
> So you said you just got one test to pass while developing this?
> 
> Maye add that to `mycpp/TEST.sh test-runtime` so it doesn't regress

Probably another good topic for VC

mycpp/marksweep_heap.cc
```diff
@@ -0,0 +1,127 @@
...
+  assert(Result);
+
+  this->all_allocations_.insert(Result);
+
+  this->current_heap_bytes_ += byte_count;
+  if (this->current_heap_bytes_ > this->collection_thresh_) {
+    Collect();
+  }
+
+  if (this->current_heap_bytes_ > this->collection_thresh_) {
```
andychu, line 18: Make a note that current_heap_bytes_ is mutated hence the duplicate check

Also can we update current_heap_bytes_ ?

PR #1316, opened 2022-09-16
Remove remaining `malloc` calls and globally enable MARK_SWEEP

scallyw4g:


scallyw4g:
Looks like everything's good to me.

Next step is incrementally turning on ASAN failures if we leak in tests

mycpp/marksweep_gc_test.cc
```diff
@@ -35,6 +35,35 @@ TEST string_collection_test() {
...
+    Str *test_str0 = nullptr;
+    Str *test_str1 = nullptr;
+    StackRoots _roots({&test_str0, &test_str1});
+
+    test_str0 = StrFromC("foo");
+    test_str1 = StrFromC("foo");
+
+    for (int i = 0; i < KiB(32); ++i)
+    {
+      if (i % 2)
```
andychu, line 48: What is the purpose of this test?  It could use some comments

mycpp/TEST.sh
```diff
@@ -240,11 +240,6 @@ compare-examples() {
 }
 
 test-sweep-asan-leaks() {
-  ninja _bin/cxx-sweepasan/mycpp/examples/fib_iter.mycpp
-  ASAN_OPTIONS='' _bin/cxx-sweepasan/mycpp/examples/fib_iter.mycpp
-
-  # CURRENTLY 9 of 22 fail
-  # This does NOT check if all pass
   ASAN_OPTIONS='' examples-variant '' sweepasan
```
andychu, line 243: How many of these pass with ASAN now?  All?  

I think this comment is saying that it doesn't exit 1 if there are failures, which we should fix

So we shouldn't delete the comment yet

andychu:
This mostly looks good, we can sync up about the rooting and how to enable ASAN leak detection gradually

mycpp/marksweep_gc_test.cc
scallyw4g: Oh it was just some random junk I typed in to try and find bugs.  I'll delete it, it's not testing anything specific

mycpp/TEST.sh
scallyw4g: There's still one failing test, but it's got nothing to do with the GC.  There's some pretty questionable code in the pretty-printers which leak and is not super easy to fix.

Issue #1317, opened 2022-09-21
asdl/format.py has heap-allocated BufWriter, which leak


andychu:
Because the internal `data_` pointer is malloc'd and we never call BufWriter::reset() which calls free()

andychu:
This is exposed by  `mycpp/examples/parse.py`

andychu:
This one is good because it has a test with a simple verification

andychu:
- the whole idea of having a `data_` pointer that's not GC managed on a GC managed object can be problematic
  - `data_` is just a heap-allocated `Str`
  - free() malloc(2 * old_size_) memcpy()

andychu:
Thiabaud got rid of some leaks by explicitly calling `reset()`, also needed to change `mycpp/mylib.pyi`

Is this right architecturally?

andychu:
@CoffeeTableEspresso I think I figured out a principled  solution in the shower (of course)

It's not valid under Cheney though

- `class BufWriter` should hold a new type `class Buf`, which is like `Str` but mutable
  - the `data_` or `buf_` pointer should have a **field mask** !  Forgot this part
- So `Buf` is a full fledged GC object
  - A sequence like this should be valid: malloc(), realloc(), realloc(), realloc(), free()
  - one issue: we call `calloc()` to ensure it's zero'd.  

----

Also to fix the quadratic problem it should be

```
realloc(10), realloc(20), realloc(40)  (doubling)

not

realloc(10), realloc(20), realloc(30)  realloc(40)
```

Basic knowledge that I've never really needed to use :)

andychu:
Or actually an easier option is to make `BufWriter` itself like `Str` -- it uses the "flexible array" pattern like `char buf_[1]` ?  And then placement new?

Although this could complicate generated code ... It couldn't use `Alloc<BufWriter>`, it would have to use a custom `NewBufWriter`

andychu:
Also we zero'd for Cheney and string NUL termination of C strings

But this is slightly harder with `realloc()`, so maybe we can get away with just doing NUL termination and not zeroing the whole buffer

https://stackoverflow.com/questions/2141277/how-to-zero-out-new-memory-after-realloc

andychu:
@CoffeeTableEspresso fixed this!

PR #1328, opened 2022-09-25
fix memory leak in ColorOutput

CoffeeTableEspresso:


andychu:
addresses this

- #1317 

But we both agree there are probably other leaks and maybe another solution

andychu:
Thanks for doing this!

I was thinking that the Buf type is GC'd, but maybe it doesn't need to be?  Let's talk about it

One thing we can do tonight is write some unit tests together

mycpp/leaky_mylib.h
```diff
@@ -125,12 +125,60 @@ class Writer : public Obj {
...
+    return data_ != nullptr;
+  }
+  void expand(Str* s) {
+    return expand(s->data_, ::len(s));
+  }
+  void expand(const char* s, int n) {
+    int orig_len = len_;
+    len_ += n;
+
+    data_ = static_cast<char*>(realloc(data_, len_ + 1));
```
andychu, line 146: Looking at this more closely, how does this `data_` buffer get freed in the garbage collected case?

Remember there were 2 cases:

1. the global case with `gBuf`, used by format strings, where you can call reset()
2. the garbage collected mylib.BufWriter instances used by `asdl/format.py`.  The garbage collector can't call `reset()`

I think it was `_bin/cxx-asan/mycpp/examples/parse.mycpp` that exposed this, or maybe it was in containers

Anyway we can talk about it tonight, thanks again!

PR #1335, opened 2022-09-28
[mycpp/runtime] Make roots arbitrarily sized with std::vector.

andychu:
Also make heap initialization consistent to KiB(1).


andychu:
Done

PR #1347, opened 2022-10-01
Fix Tuple{2,3,4} crashing when allocated on the heap

scallyw4g:


mycpp/tuple_types.h
```diff
@@ -1,11 +1,19 @@
...
 
+
+
 template <class A, class B>
 class Tuple2 {
  public:
-  Tuple2(A a, B b) : a_(a), b_(b) {
+  Tuple2(A a, B b) : a_(a), b_(b)
+  {
+    heap_tag_ = Tag::FixedSize;
```
andychu, line 11: I think it's better to use the `Obj()` constructor like Dict does

So then the body only has the `|=`

mycpp/tuple_types.h
```diff
@@ -34,6 +48,7 @@ class Tuple3 {
   }
 
  private:
+  OBJ_HEADER();
```
andychu, line 51: I think we should be more consistent about inheriting from Obj vs OBJ_HEADER(), which are roughly equivalent ways to do the same thing

The problem is multiple inheritance, which we use in ASDL

But here there is no inheritance so I think it's OK to inherit from Obj ?

cpp/leaky_stdlib.h
```diff
@@ -87,8 +87,9 @@ inline void _exit(int status) {
   ::_exit(status);
 }
 
-inline void write(int fd, Str* s) {
-  ::write(fd, s->data_, len(s));
+inline size_t write(int fd, Str* s) {
+  size_t size = ::write(fd, s->data_, len(s));
```
andychu, line 91: Is this to shut up a warning?

I think what we need to do is check `result < 0` and then throw an exception

That's how Python works so we should match the semantics

This will just push the ignored return value higher up

cpp/segfault_handler.h
```diff
@@ -3,7 +3,7 @@
 
 static int segfault_handler_initialized;
 
-#define WRITE_STRING_TO_STDERR(s) (write(2, (s), sizeof(s) - 1))
+#define WRITE_STRING_TO_STDERR(s) do {size_t ignored = write(2, (s), sizeof(s) - 1);} while (false)
```
andychu, line 6: Can this be static_cast<void*> instead?

https://stackoverflow.com/questions/689677/why-cast-unused-return-values-to-void

andychu (commented):
OK this is great, thanks

I think there might be a way to do this at compile time instead of runtime, but we can test with this first to see if it works

mycpp/tuple_types.h
scallyw4g: If you can figure out how to get the compiler to accept Tuple inheriting from Obj, be my guest.  I tried to get it to cooperate and it would not

mycpp/tuple_types.h
scallyw4g: Again: If you can figure out how to get the compiler to accept Tuple inheriting from Obj, be my guest. I tried to get it to cooperate and it would not

cpp/segfault_handler.h
scallyw4g: It's not `void*`, it's `void`, and I tried that, and it didn't work

scallyw4g:
> I think there might be a way to do this at compile time instead of runtime, but we can test with this first to see if it works

I didn't check the assembly, but I'm quite sure the compiler will compute the mask at compile time in `-02`

cpp/leaky_stdlib.h
scallyw4g: Alright, I'll revert this change.

PR #1348, opened 2022-10-01
[ASDL] Add field masks to generated C++ code

andychu:
- Add new hnode_test, but existing test doesn't pass yet.  TODO
- Fix broken prebuilt/ script

### DIFF
--- a/asdl/gc_test.cc
+++ b/asdl/gc_test.cc
@@ -6,7 +6,11 @@
 #include "prebuilt/asdl/runtime.mycpp.h"
 #include "vendor/greatest.h"
 
+using hnode_asdl::color_e;
 using hnode_asdl::hnode_t;
+using hnode_asdl::hnode__Array;
+using hnode_asdl::hnode__Record;
+using hnode_asdl::hnode__Leaf;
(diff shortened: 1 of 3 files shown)


andychu:
I'm not ready to commit this yet, I want to backtrack on a couple things

- make it easier to run unit tests
- move `asdl/*_test.cc` into the Ninja build.   The slowness is annoying me now!

We should really test the crap out of these, because it's better than testing it in the real program

andychu:
Also this doesn't fix any of the 6 `gcevery` tests, so maybe it doesn't work for some reason

There is a basic test that works, but the existing test still fails for unknown reasons

andychu:
Just as on my machine there's still this use-after-free, so I guess something is wrong

http://travis-ci.oilshell.org/github-jobs/2022-10-01__22-55-32.wwz/

andychu:
Oh duh, this test relies on the mycpp-translated ASDL runtime in `prebuilt/`, which has classes.

So it's possible  the ASDL masks are correct

But I still want to backtrack a bit and improve the tests

andychu:
landed but needs improvement

PR #1351, opened 2022-10-04
Refactor `BufWriter`

CoffeeTableEspresso:
This CL refactors `BufWriter` to prepare for changing it to use a garbage-collected buffer.


andychu:
Oh OK now I see there are 2 PRs open! 

OK I see there  are 3 failures instead of one -- http://travis-ci.oilshell.org/github-jobs/2022-10-04__02-21-59.wwz/_tmp/soil/logs/mycpp-translator.txt

That could be due to the `new` and `delete` , but we can talk about it !  Thanks

PR #1352, opened 2022-10-04
Add `Buf` class to `BufWriter`

CoffeeTableEspresso:
This CL encapsulates the data for `BufWriter` into a separate `Buf` class.

### DIFF
--- a/mycpp/leaky_mylib.cc
+++ b/mycpp/leaky_mylib.cc
@@ -13,7 +13,7 @@ mylib::FormatStringer gBuf;
 namespace mylib {
 
 Str* StrFromBuf(const Buf& buf) {
-    return ::StrFromC(buf.data_, buf.len_);
+  return ::StrFromC(buf.data_, buf.len_);
 }
 
 // NOTE: split_once() was in gc_mylib, and is likely not leaky
@@ -144,20 +144,19 @@ void Buf::Extend(Str* s) {
   int n = len(s);
 
   assert(cap_ >= len_);
-  if (cap_ - len_ < n + 1) {
-    cap_ = std::max(cap_ * 2, len_ + n + 1);
-    data_ = static_cast<char*>(realloc(data_, cap_));
-  }
+  cap_ = std::max(cap_ * 2, len_ + n);
+  data_ = static_cast<char*>(realloc(data_, cap_ + 1));
+
   memcpy(data_ + len_, s->data_, n);
   len_ += n;
   data_[len_] = '\0';
 }
 
 void Buf::Invalidate() {
-    free(data_);
-    len_ = -1;
-    cap_ = -1;
-    data_ = nullptr;
+  free(data_);
+  len_ = -1;
+  cap_ = -1;
+  data_ = nullptr;
 }
 
 //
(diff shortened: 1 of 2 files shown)


mycpp/leaky_mylib.h
```diff
@@ -125,12 +125,28 @@ class Writer : public Obj {
   virtual bool isatty() = 0;
 };
 
+class Buf {
+ public:
+  Buf(char* data, int len) : data_(data), len_(len) {}
+  explicit operator bool() { return len_ == 0; }
```
andychu, line 131: I prefer not to use any operator overloading (and I think that's in the Google style guide too)

This could be `IsEmpty()` ?

mycpp/leaky_mylib.h
```diff
@@ -125,12 +125,28 @@ class Writer : public Obj {
...
+Str* StrFromBuf(const Buf&);
+Buf AllocBuf(char*, int);
+
 class BufWriter : public Writer {
  public:
   BufWriter()
       : Writer(Tag::FixedSize, kZeroMask, sizeof(BufWriter)),
-        data_(nullptr),
-        len_(0) {
+        buf_(nullptr, 0) {
```
andychu, line 149: Seems like the constructor doesn't need any params, but we can also sort that out in the next PR

andychu (commented):
Thanks for doing this!   Looks like a good step toward having a garbage-collected Buf type

Jesse also pointed out that it could be `Slab<uint8_t>`, although I'm not sure if everything will line up.  Might be better to fix the leak first, and then refactor

mycpp/leaky_mylib.h
CoffeeTableEspresso: done

mycpp/leaky_mylib.h
CoffeeTableEspresso: done

mycpp/leaky_mylib.cc
```diff
@@ -132,18 +136,40 @@ bool CFileWriter::isatty() {
...
+//
+//
+
+void Buf::Extend(Str* s) {
+  int n = len(s);
+
+  assert(cap_ >= len_);
+  if (cap_ - len_ < n + 1) {
+    cap_ = std::max(cap_ * 2, len_ + n + 1);
+    data_ = static_cast<char*>(realloc(data_, cap_));
```
andychu, line 149: Sort of a nitpick, but I think it's easiest to reason about if `len_` and `cap_` both do NOT take into account NUL.

Simple `realloc(cap_ + 1)` in all cases

But then keep `+1` out of the `cap_ - len_` calculation, and the max calculation()

andychu (commented):
Thanks this looks great!

Let me know if running the formatter works:


deps/from-binary.sh download-clang
deps/from-binary.sh extract-clang

test/lint.sh format-cpp  # reformat the C++ code.  'git diff' to see the result.


mycpp/leaky_mylib.cc
CoffeeTableEspresso: done

CoffeeTableEspresso:
Formatting works for me

andychu:
Merged!  thanks a lot :)

PR #1359, opened 2022-10-09
[cpp] add misisng errno include

melvinw:
I had to make this change to get the oil-native build working with gcc 10.X


andychu:
Great thank you!

PR #1361, opened 2022-10-10
[cpp] some umask fixes

melvinw:
These should fix two failures found by `./test/spec-cpp.sh run-with-osh-eval builtins` (`get umask` and `set umask in octal`)

@andychu any reason we had `int umask(int)` instead of `mode_t umask(mode_t)`? The commit that tweaks that isn't essential, but I figured it was worth updating while I was in that region

### DIFF
--- a/mycpp/leaky_mylib.cc
+++ b/mycpp/leaky_mylib.cc
@@ -242,7 +242,9 @@ void FormatStringer::format_s(Str* s) {
 }
 
 void FormatStringer::format_o(int i) {
-  NotImplemented();
+  data_ = static_cast<char*>(realloc(data_, len_ + kIntBufSize));
+  int len = snprintf(data_ + len_, kIntBufSize, "%o", i);
+  len_ += len;
 }
 
 void FormatStringer::format_d(int i) {
--- a/osh/builtin_process2.py
+++ b/osh/builtin_process2.py
@@ -215,8 +215,11 @@ class Umask(vm._Builtin):
         # NOTE: This happens if we have '8' or '9' in the input too.
         stderr_line("osh warning: umask with symbolic input isn't implemented")
         return 1
-      else:
-        posix.umask(new_mask)
-        return 0
+      except TypeError:
+        # Should be unreachable
+        raise AssertionError()
+
+      posix.umask(new_mask)
+      return 0
 
     e_usage('umask: unexpected arguments')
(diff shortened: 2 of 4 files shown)


mycpp/leaky_mylib.cc
```diff
@@ -242,7 +242,9 @@ void FormatStringer::format_s(Str* s) {
 }
 
 void FormatStringer::format_o(int i) {
-  NotImplemented();
+  data_ = static_cast<char*>(realloc(data_, len_ + kIntBufSize));
```
melvinw, line 245: `kIntBufSize` might be wrong size?

andychu: I think this is fine, the comments around kIntBufSize seem to indicate it

(although they are giving me a headache now, probably should change it to 11 or 12 :) )

osh/builtin_process2.py
```diff
@@ -215,8 +215,11 @@ def Run(self, cmd_val):
...
         stderr_line("osh warning: umask with symbolic input isn't implemented")
         return 1
-      else:
-        posix.umask(new_mask)
-        return 0
+      except TypeError:
+        # Should be unreachable
+        raise AssertionError()
+
+      posix.umask(new_mask)
```
andychu, line 222: Ah OK I think this is because mycpp ignores "else" in try/except/else!

Because C++ doesn't have that construct

We could make that an error to find these bugs

andychu (commented):
Thanks, this looks good!

The issue with `mode_t` is that the code generated from Python  uses int, e.g. for `new_mask`

We're doing a very literal translation from Python to C++, and so there is some mismatch in types

I'm not sure this changes anything, but it probably doesn't make it worse either :-/

osh/builtin_process2.py
andychu: Great point, thanks for finding that!  I filed a bug about it

https://github.com/oilshell/oil/issues/1362

mycpp would be a good thing to talk about, I kind of want to rewrite it as Pea, with Python 3.10 pattern matching (which you mentioned! ):   https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/First.20step.20for.20Pea.3F

It's a bit messy, but I have been able to incrementally add all the features I want, like field masks last week.

(Jesse and I did a lot of working cleaning up and fixing the runtime in the last 3 months, but we haven't cleaned up the translator.  It's kind of burning a hole in my brain :) Seems fun but not sure I can justify it )

PR #1363, opened 2022-10-10
[cpp] remove some calls to malloc()

melvinw:
Passes the following relevant-sounding tests according to grep:
- `spec/builtin-printf`
- `spec/regex`
- `cpp/leaky-core-test`

### DIFF
--- a/cpp/leaky_core.cc
+++ b/cpp/leaky_core.cc
@@ -235,8 +235,8 @@ Str* ShowAppVersion(Str* app_name, _ResourceLoader* loader) {
 
 Str* BackslashEscape(Str* s, Str* meta_chars) {
   int upper_bound = len(s) * 2;
-  char* buf = static_cast<char*>(malloc(upper_bound));
-  char* p = buf;
+  Str* buf = AllocStr(upper_bound);
+  char* p = buf->data_;
 
   for (int i = 0; i < len(s); ++i) {
     char c = s->data_[i];
@@ -245,8 +245,8 @@ Str* BackslashEscape(Str* s, Str* meta_chars) {
     }
     *p++ = c;
   }
-  int len = p - buf;
-  return StrFromC(buf, len);
+  int len = p - buf->data_;
+  return StrFromC(buf->data_, len);
 }
 
 // Hack so e->errno will work below
--- a/cpp/leaky_core.cc
+++ b/cpp/leaky_core.cc
@@ -71,20 +71,13 @@ Dict<Str*, Str*>* Environ() {
     char* eq = strchr(pair, '=');
     assert(eq != nullptr);  // must look like KEY=value
 
-    int key_len = eq - pair;
-    char* buf = static_cast<char*>(malloc(key_len + 1));
-    memcpy(buf, pair, key_len);  // includes NUL terminator
-    buf[key_len] = '\0';
-
-    Str* key = StrFromC(buf, key_len);
-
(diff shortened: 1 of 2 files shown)


cpp/leaky_core.cc
```diff
@@ -252,8 +245,8 @@ Str* BackslashEscape(Str* s, Str* meta_chars) {
     }
     *p++ = c;
   }
-  int len = p - buf;
-  return StrFromC(buf, len);
+  int len = p - buf->data_;
+  return StrFromC(buf->data_, len);
```
andychu, line 249: There's an API called OverAllocatedStr() that encapsulates the pattern of:

- allocating an upper bound
  -  s = OverAllocatedStr(n)
- shrinking to fit after you know how many bytes were written
  -  s->SetObjLenFromStrLen(10)

I used it for read(), strftime() and a similar style for gethostname(), getcwd()   (grep should easily reveal them)

This will avoid copying the buffer -- e.g. StrFromC() copies and takes ownership

----

Honest question (looking for feedback): what do you think of this API?   Jesse found it to be error prone, while I think it is a natural bridge between C-style buffer interfaces and C++ GC objects, which avoids some copying

cpp/leaky_libc.cc
```diff
@@ -86,19 +86,17 @@ List<Str*>* regex_match(Str* pattern, Str* str) {
   int outlen = pat.re_nsub + 1;  // number of captures
 
   const char* s0 = str->data_;
-  regmatch_t* pmatch =
-      static_cast<regmatch_t*>(malloc(sizeof(regmatch_t) * outlen));
-  int match = regexec(&pat, s0, outlen, pmatch, 0) == 0;
+  Slab<regmatch_t>* pmatch = NewSlab<regmatch_t>(outlen);
```
andychu, line 89: Ah I think this one can be left alone because there's no leak?  Sorry I misphrased the goal -- get rid of leaks, not necessarily malloc

I think malloc and free() in the same function happens to be OK

Though interested in any feedback.

One way we could verify this is to make sure there is test coverage, and that ASAN catches no leaks when running the test (which we need to do for all tests)

andychu (commented):
Great thank you for doing this!

Now that I think about execve() -- I think it's OK to leak because there's no place to free, and the entire process gets replaced!

----

Also I added you as a contributor to the project, so I think if you send PRs from a copy of this repo (rather than a fork), it will run the CI on pull requests

You can also make a "draft pull request" if you just want to use the CI to run tests


Sorry I didn't know how to set it up the other way (open to suggestions).  I would prefer forks so everyone external can get CI results too, but that's how we've been working and it seems OK.

cpp/leaky_libc.cc
melvinw: Ah great. I was wondering if this on was necessary. I'll drop it

cpp/leaky_core.cc
melvinw: Ah great! I was a little bummed by the double alloc here. Glad there's a way to avoid.

I think the API is reasonable. Not really sure how else you'd do something like this without allocating twice, and I assume the wasted bytes from the shrink (i assume it's not actually giving back the slop) will be minimal and short lived. The name for the shrink op is a little awkward though. What if it were something like `s->Truncate(...)`, `s->Trim(...)`, or `s->Adjust(...)`?

melvinw:
gonna reopen this from a local branch

Issue #1354, opened 2022-10-06
new GC header (big/little endian)


andychu:
https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/New.20GC.20Header

andychu:
### Good Packing of 8 Bytes

- first "word"
  - 1 bit not vtable 
  - 7 bits ASDL type, which may include special cases for List<Obj> and Dict<Str, Obj> when we self host?
  - 24 bit object ID.  Implies 16 Mi objects.
- second "word"
  - 1 bit fixed or variable length
  - for variable length: 1 bit opaque or scanned
  - 30 bits field mask or length.  Implies 1 Gi string length or Slab length.

PR #1364, opened 2022-10-10
[cpp] remove some calls to malloc()

melvinw:
Trying #1363 again


andychu:
Merged, thank you!

----

FWIW the background on the weird SetObjLenFromStrLen() name is that we don't store both the Str length (for application code) and the Obj length (for the GC), as one can be derived from the other.  (This optimization is in the OCaml runtime, and I'm sure other places)

But we can rename it to `SetStrLen()` when we switch fully to mark and Sweep the new GC header -- #1354 

The reason for the difference is

- Cheney needs the length of ALL objects because it must copy all objects to the other semi-space
- Mark and Sweep needs the length of ONLY the objects to scan for pointers.
  - You don't need the length to call free()

PR #1365, opened 2022-10-10
[frontend] refactor signal code

melvinw:


cpp/leaky_osh_eval_stubs.h
```diff
@@ -38,12 +40,160 @@ namespace signal_def {
...
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGURG\0"), SIGURG));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGXCPU\0"), SIGXCPU));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGXFSZ\0"), SIGXFSZ));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGVTALRM\0"), SIGVTALRM));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGPROF\0"), SIGPROF));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGWINCH\0"), SIGWINCH));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGIO\0"), SIGIO));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGPOLL\0"), SIGPOLL));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGPWR\0"), SIGPWR));
+  sigs->append(Alloc<Tuple2<Str*, int>>(StrFromC("SIGSYS\0"), SIGSYS));
```
melvinw, line 79: glibc >=2.32 deprecated `sys_siglist`, otherwise we could probably just walk that instead of enumerating everything here...

andychu:
Cool, this looks like it will make some more things pass!

A few things:

- strcmp("HUP\0") technically isn't valid, because in Oil a string can have internal NUL.  (They are silently truncated when passed to syscalls). 
  -   So `str_equals()` which checks the length is better ... BUT
- I also think it might be better to generate this code, and produce global objects rather than GC-managed ones (?) -- not exactly sure as I haven't looked at this code in awhile.   IIRC the dash shell generates a C source file with the list of signals at build time, as well.
  - Other shells might not -- though I believe it does actually depend on Linux vs. BSD a bit ?   The existing Python code probably doesn't take that into account.
  - FWIW the purpose of GLOBAL_STR() and GLOBAL_LIST() is to generate layout-compatible global objects that aren't GC managed.  We don't yet have GLOBAL_TUPLE2 though

The code generation is probably a big yak shave, so if this unblocks other work, I'm happy to merge and move on to more critical problems

melvinw:
OK let's try another round on this. I ended up adding `frontend/signal_gen.py` to handle the C++ part instead of writing it by hand. What do you think?

I also noticed some of the other `_gen.py` files have modes for generating `mypy` code too. Should I add something similar here?

andychu:
Wow amazing!   Glad that all the build / code gen stuff seems to have been no problem :-)

Merged

PR #1366, opened 2022-10-10
[core] imperative-ize a few list comprehensions

melvinw:
Doesn't look like we know how to translate list comprehensions yet. Turning the ones in the accessors for `_pipeline_status` and `_process_sub_status` into vanilla loops brings us up to parity with `osh.py` for `spec/pipeline` and `spec/process-sub`

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -1926,11 +1926,13 @@ class Mem(object):
         return value.Str(self.this_dir[-1])  # top of stack
 
     if name in ('PIPESTATUS', '_pipeline_status'):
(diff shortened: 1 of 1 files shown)


andychu:
Oh great catch here!!  Thank you

Can you try the following:

```
tmp = [list comp]
return value.MaybeStrArray(tmp)
```

I believe mycpp can translate that.   It requires a temporary assignment, due to some limitations in its intelligence :)

We should fix that, I filed a bug here:

https://github.com/oilshell/oil/issues/1367

melvinw:
Nice! The temp var thing worked

PR #1373, opened 2022-10-13
[core] try to share more trap-related code

melvinw:
Thoughts? Taking a cue from this comment: https://github.com/oilshell/oil/blob/master/osh/builtin_trap.py#L34-L46

I marked a few comments in this patch with `XXX`. I think these things are somewhat awkward but may be fixable if my understanding of the code happens to be wrong (or if some refactoring happened at some point).


core/pyos.py
```diff
@@ -313,6 +313,16 @@ def SignalState_AfterForkingChild():
   signal.signal(signal.SIGTTIN, signal.SIG_DFL)
 
 
+def SignalState_PendingTraps():
+    # type: () -> List[Tuple[int, _TrapHandler]]
+    return []
+
+
+def SignalState_ClearPendingTraps():
```
andychu, line 321: A bit pedantic, but how about

- PendingTraps_Get()
- PendingTraps_Clear()

I think the `SignalState_` prefix was supposed for be for the "which signals the process responds to" state, and I think of pending traps as separate?

Maybe should rename that `SignalActions` or something

cpp/leaky_core.cc
```diff
@@ -18,6 +19,11 @@
 
 namespace pyos {
 
+const int kMaxSigs = 128;
+const int kMaxSigTraps = 128;
+builtin_trap::_TrapHandler* signal_traps[kMaxSigs][kMaxSigTraps];
```
andychu, line 24: Hm so why is it a 2 dimensional array?  

I haven't looked  at this code in awhile, but my memory is that the bash trap builtin has two kinds of "sig specs":

- a Unix signal like SIGINT / INT
- a "user space" hook like ERR or DEBUG

I guess right now those are mixed in the "self.traps" dictionary

I think the key weirdness is that we have state in that dictionary, and state in the global array

Let me think about this

mycpp/gc_list.h
```diff
@@ -500,9 +500,7 @@ List<Str*>* sorted(Dict<Str*, V>* d) {
 template <typename T>
 List<T>* list(List<T>* other) {
   auto result = NewList<T>();
-  for (int i = 0; i < len(other); ++i) {
-    result->set(i, other->index_(i));
-  }
+  result->extend(other);
```
andychu, line 503: Nice change :)

osh/cmd_eval.py
```diff
@@ -1433,6 +1433,21 @@ def _Dispatch(self, node, cmd_st):
...
+    # the trap code, but the type isn't accessible from the hand-written C++
+    pending = pyos.SignalState_PendingTraps()
+    if len(pending):
+      pending = list(pending)
+      pyos.SignalState_ClearPendingTraps()
+      for sig_num, handler in pending:
+          # TODO: Right now we have to invoke __call__() explicitly instead of
+          # using it as sugar. Perhaps we should teach mycpp to translate
+          # __call__() into operator() in C++
+          handler.__call__(sig_num, None)
```
andychu, line 1449: OK I see the key difficulty here ...

Though I'm a little surprised that the `__call__` translates to C++?  I didn't run it but it looks like the CI passes?

andychu (commented):
By and large this looks awesome!  Thanks for digging into this hairy area (and reading the comments!)

It looks like this passes a few more tests, which is great, so it will move us forward


Long term I'm thinking that signals expose some problems with the Python to C++ architecture

- is there any duplication of `self.traps` with the kernel state?  Not sure now that I think about it
- how to call back into the CommandEvaluator
  - we have this problem in a couple other places, and I think it's possible to expose a header?  If so I can help with that

Maybe we should discuss on our call tomorrow?

cpp/leaky_core.cc
melvinw: Ah yes, this is a little awkward. Basically the problem is, the current `_TrapHandler` type has no context for which signal (if not a hook) it was registered to. So, we keep a list of traps for each possible signal. This is pretty inefficient... If we added that missing context we could flatten this array and simplify a lot of the dependent code.

Luckily we don't have this problem for hooks, since the python code for that is portable and mycpp translates it cleanly

osh/cmd_eval.py
melvinw: Yeah, I was a little surprised by that too. It builds and passes `spec-cpp/builtins-trap` and a few manual tests, though

melvinw: Another option could be to expose more of `_TrapHandler` to the hand-written C++. Probably the same type of work, though. Just less of it since `_TrapHandler` is much simpler than `CommandEvaluator`

andychu:
Hm after the meeting I was thinking about this strategy:

- What's the minimal amount of code that makes the `spec/builtin-trap` tests pass?
  - I think it's probably not that much since it only tests EXIT and SIGUSR1 I think?
  - we could try to get that in
- And then what do we need to do to make `spec/stateful/*` pass?  That is a much bigger question
  -  that is blocked on the prompt issue now

But really I do think #1374 is fundamentally important; the existing code is not good for translating

core/shell.py
```diff
@@ -73,6 +74,30 @@
   from _devbuild.gen.runtime_asdl import Proc
 
 
+def _InitInteractiveShell(sig_state, display, my_pid):
```
andychu, line 77: I realize it's weird that SignalState has no members

But I think it's worth it because it's a "handle" for mutation of process state.  It's actually a wrapper for global state **in the kernel** !

So we should really name it SignalDisposition or something -- that's what the man page calls this global process state

Any piece of code that registers signal handler must have the  SignalDisposition instance

It's basically a style from functional programming or OOP (dependency inversion)

----

And then GlobalTrapState() is mostly separate, or it can hold the SignalDisposition as a member

osh/builtin_trap.py
```diff
@@ -95,12 +64,101 @@ def _GetSignalNumber(sig_spec):
...
+      del self.trap_nodes[sig_num]
+    # TODO: SIGINT is similar: set a flag, then optionally call user handler
+
+  def TakeRunList(self):
+      # type: () -> List[command_t]
+      new_run_list = []  # type: List[command_t]
+      pyos.ReserveHandlerCapacity(new_run_list)
+      ret = self.run_list
+      self.run_list = new_run_list
+      return ret
```
melvinw, line 137: This gets transpiled to the following, which _should_ help with one of the races you pointed out @andychu. Assuming pointer assignment on the underlying arch is atomic  
```
List<syntax_asdl::command_t*>* TrapState::TakeRunList() {
  List<syntax_asdl::command_t*>* new_run_list = nullptr;
  List<syntax_asdl::command_t*>* ret = nullptr;
  StackRoots _roots({&new_run_list, &ret});

  new_run_list = Alloc<List<syntax_asdl::command_t*>>();
  pyos::ReserveHandlerCapacity(new_run_list);
  ret = this->run_list;
  this->run_list = new_run_list;
  return ret;
}
```

cpp/leaky_core.h
```diff
@@ -51,25 +48,21 @@ class TermState {
...
-  void RemoveUserTrap(int sig_num) {
-    NotImplemented();
-  }
-  int last_sig_num = 0;
-
-  DISALLOW_COPY_AND_ASSIGN(SignalState)
+  virtual void Run(int) = 0;
 };
 
+void Sigaction(int sig_num, SignalHandler* handler);
```
melvinw, line 61: I _think_ I like this pattern. It will require **a lot** of discipline to make sure we keep the python handlers signal-safe, though...

melvinw:
Alright let's try this one again. I've cleaned it up a bit and it should pass all the tests now. It will also bring C++ and python to parity on `spec/builtin-trap`.

One thing that worries me a bit is the C++ version of `pyos.SignalHandler`. Since the handlers that implement it are all written in translatable python, the abstract class in `leaky_core.h` has to be an `Obj` (for `field_mask_`, etc...). This required a potentially hacky default constructor that will probably break the garbage collector once it's enabled...

core/process.py
```diff
@@ -888,7 +889,32 @@ def Start(self, why):
       raise RuntimeError('Fatal error in posix.fork()')
 
     elif pid == 0:  # child
-      pyos.SignalState_AfterForkingChild()
+      # XXX: If performance is a concern here we may not want to use
+      # `signal_def.GetNumber()` in the calls to `pyos.Sigaction()` below, as
+      # the code generated by `frontend/signal_gen` isn't super efficient
+      # because of the string comparisons.
```
melvinw, line 895: This is another important consideration. The performance hit on `fork()` may not be worth the ~10 lines of avoided code duplication...

mycpp/gc_list.h
```diff
@@ -154,7 +154,9 @@ template <typename T>
 void list_append(List<T>* self, T item) {
   StackRoots _roots({&self});
 
-  self->reserve(self->len_ + 1);
+  if (self->capacity_ - self->len_ < 1) {
```
andychu, line 157: Why add these checks?  reserve() already has an if statement in it

I believe it matches how vector::reserve(n) works -- it's a no-op if it's already big enough

core/process.py
andychu: I think you can just use SIGPIPE directly?

In Python you can do

from signal import SIGPIPE

and then in C++, the `leaky_preamble.h` hack should `#include <signal.h>`, which makes SIGPIPE visible

It's a little weird but I think an OK pattern

core/pyos.py
```diff
@@ -264,121 +263,26 @@ def InputAvailable(fd):
...
-  # https://docs.python.org/2/library/signal.html
-  # See Python/pythonrun.c.
-  signal.signal(signal.SIGPIPE, signal.SIG_DFL)
-
-  # Respond to Ctrl-\ (core dump)
-  signal.signal(signal.SIGQUIT, signal.SIG_DFL)
+    # XXX: We could have done something like `lambda sig, unused: handler.Run(sig)`
+    # in Sigaction(), but that doesn't type check because lambda returns Any and
+    # signal.signal() expects a Callable that returns None (and for some reason
+    # we can't provide type hints for lambdas)
```
andychu, line 275: Hm one idea is to just disable type checks inside the body, which I think is OK  since pyos.py doesn't need to translate.  It is basically an interface shim.

Anything that must translate needs strict types of course

cpp/leaky_core.cc
```diff
@@ -181,10 +184,31 @@ bool InputAvailable(int fd) {
   NotImplemented();
 }
 
-void SignalState_AfterForkingChild() {
-  signal(SIGQUIT, SIG_DFL);
-  signal(SIGPIPE, SIG_DFL);
-  signal(SIGTSTP, SIG_DFL);
+static void signal_handler(int sig_num) {
+  assert(gSignalHandlers != nullptr);
+  SignalHandler* handler = gSignalHandlers->get(sig_num);
```
andychu, line 189: There is likely a race here, because the main thread is writing and signal handlers are reading from this Dict

I think the "bottom-up" strategy of doing SharedTrapState and TSAN will help, and this code may end up looking different


This is simlar to the `std::map` demo in `cpp/data_race_test.cc` -- whether it's a race depends on the internal implementation of the `Dict`, and it almost certainly is (we could test it with TSAN!)

Also our dict implementation is likely to change (it doesn't have hashing yet), and it's definitely not intended to be used unsynchronized from different threads of control

cpp/leaky_core.cc
```diff
@@ -181,10 +184,31 @@ bool InputAvailable(int fd) {
...
+  assert(handler != nullptr);
+  handler->Run(sig_num);
+}
+
+void Sigaction(int sig_num, SignalHandler* handler) {
+  if (gSignalHandlers == nullptr) {
+    gSignalHandlers = Alloc<Dict<int, SignalHandler*>>();
```
melvinw, line 196: This could probably just happen at the top where this is declared...

cpp/leaky_core.cc
```diff
@@ -181,10 +184,31 @@ bool InputAvailable(int fd) {
...
 
-void SignalState_AfterForkingChild() {
-  signal(SIGQUIT, SIG_DFL);
-  signal(SIGPIPE, SIG_DFL);
-  signal(SIGTSTP, SIG_DFL);
+static void signal_handler(int sig_num) {
+  assert(gSignalHandlers != nullptr);
+  SignalHandler* handler = gSignalHandlers->get(sig_num);
+  assert(handler != nullptr);
+  handler->Run(sig_num);
```
andychu, line 191: This part isn't safe, as you alluded to in the comment

The handler can contain a `command_t` that will be run by the interpreter right?

It shouldn't be run directly from the signal handler, for several reasons, like it's not safe to malloc() from a signal handler

It should set a flag, and then the main loop will run it

(This is how CPython, bash, dash, etc. all work)

cpp/leaky_core.h
andychu: see comment above -- the Python handlers definitely won't be signal safe, as running them will allocate memory, collect garbage, etc.

osh/builtin_trap.py
```diff
@@ -95,12 +64,101 @@ def _GetSignalNumber(sig_spec):
 # Then hit Ctrl-C.
 
 
+class TrapState(pyos.SignalHandler):
```
andychu, line 67: I don't think the TrapState should inherit from pyos.SignalHandler

To clarify the code I might name it `SharedTrapState`  -- it should be ONLY the stuff that's shared between main thread and signal handler

It's the stuff we want to test for races.  I posted some messages on `#blog-ideas` about "functions and data" and ADTs that might help illuminate how I'm thinking about it

(And also evidence from CPython, bash, dash on `#oil-dev` )

andychu (commented):
Thanks for working on this!  (and diving into the probably the hairiest part of the code, as I mentioned)

As you hinted at on Zulip, I think this needs to be split up.  The C++ gc_list.h parts are one thing that can probably go in quickly.

-----

I think there are some "tactical refactorings" we can do first, e.g. when I looked at it yesterday, the `trap_nodes` thing was burning a hole in my head (the existing code doesn't have a great interface)

i.e. I think of it as "make the change easy (refactor), and then make the easy change (make it translatable to C++)

Personally I think the "bottom up" strategy with TSAN will help us think about the right interfaces

----

I believe this code has at least 1 race (maybe 2), and I think we want to convince ourselves that it doesn't before merging, and there are some in between steps we can do

core/process.py
melvinw: Ah great. That will simplify this

mycpp/gc_list.h
melvinw: Ah these were a bit preemptive. Should have read `reserve()` first. Will roll this back

cpp/leaky_core.cc
melvinw: > This part isn't safe, as you alluded to in the comment
> 
> The handler can contain a `command_t` that will be run by the interpreter right?

Ah no this isn't my intention here. Right now the only meaningful thing that implements `SignalHandler` is `TrapState` whose `Run()` simply appens the `command_t` for the trap to the run list for the main loop to run later

core/pyos.py
melvinw: I'm not _too_ torn up about not using `lambda` here, but it would help make it more obvious that this and the `SignalHandler` in `leaky_core.h` are the same thing. So, disabling the type check here is probably a net gain

cpp/leaky_core.cc
melvinw: Oh wow this is a pretty glaring hole in this...

melvinw: I _think_ I've convinced myself we don't need this Dict here. Pushed a fixup commit that we can talk about when we chat next. I think the challenge will be making the usage from python more obviously correct

andychu:
Thinking about it some more, the reason I was thinking about a separate Run() / Thunk() abstraction is also so we can call rl_resize_terminal()


https://github.com/oilshell/oil/blob/master/pyext/line_input.c#L667

Well I guess we need to find out / decide if that function is signal safe or not.  Should we call rl_resize_terminal() on the main loop, or in the signal handler?

What do you think?  We can talk about it tomorrow

andychu:
Also for tomorrow, thinking about:

- potential race between MarkObjects() on main thread and List::append in the signal handler
  -  I'm more or less convinced about TakeRunList, but it could use some comments, and I think we should test it
- why `if mylib.PYTHON` in a few places?
  - What does the PR claim to do?   I think we should split it up, but either way a detailed description would help.

andychu:
Brainstorming: what is the list of things we can potentially do in the signal handler or on the main thread?

1. execute a `command_t` , a user trap
2. set the last_sig_num -- I think this is for `wait()`.  This is probably fast and can go in the signal handler
3. call rl_resize_terminal()
4. set `width_is_dirty` on our own terminal display code

The first and second happen on any signal, potentially.

The third and fourth are for SIGWINCH, which can also have a `command_t` trap.

----

Also I'm thinking of potential renamings to be precise about "signal" vs. "trap" ...  Although it's confusing because a "shell trap handler" is run as a result of a signal

Related to that comment about SharedTrapState, which can hold SignalDisposition

Also maybe this is overboard, but I'm thinking about how to factor it so signals are optional -- e.g. Oil running as WASM, without POSIX.   WASI might not have signals ?  Though it probably doesn't work at all because it doesn't have fork() either

melvinw:
Another thought for tomorrow:

I think what I ultimately want to do here is establish a **very** simple interface between the kernel and the main loop. The current interface is close, but I think the `command_t` queue thing is still too much information. This PR in its current form is approaching that this abstract `SignalHandler` stuff is a step toward this, but I think what we really want to say is something along the lines of "please tell me if SIGX fires," **NOT** "please do run this function for me when SIGX fires". This is a slightly different way of thinking than what is encouraged by `signal(2)`, but I think it might work here.

What if we had something like the following (pardon the very literal names... they would change of course):
```
# Returns a list of all signals that have fired since the last invocation.
# The underlying state NEED NOT be an actual list.
# It could be, e.g., an atomic bitset or a large array of flags.
# The underlying state is cleared in the process.
# This is the ONLY place the kernel and the main loop should synchronize.
GetRecentSignals() -> List[int]

# Register a signal handler that flips the underlying state for the signal.
RegisterSignalnterest(int) -> None
```

Then:
- Main loop tells the kernel it is interested in a particular signal with`RegisterInterestInSingal(SIGX)`  (e.g. because it read `trap ... SIGX`) 
- We register a dead simple simple signal handler. Something like:
```
void handler(int sig_num) {
  gSignalFlags[sign_num] = 1; // atomic
}
```
- The main loop grabs the list of signals that have fired since the last tick with `fired = GetRecentSignals()`  and walks `fired`, handling signals in whatever way it wants (e.g.  evaluating `command_t`s from `TrapState.trap_nodes`).

I think this is closer in spirit to the very first version of this change but without having to tell the kernel or `pyos` about `command_t` or `List[command_t]`.

melvinw:
going to close this and send it out as a sequence of smaller, more targeted PRs

PR #1379, opened 2022-10-17
[mycpp/gc_list] fix copy constructor

melvinw:
Can't set a value at an index if it hasn't been reserved yet. Let's borrow extend() here to avoid duplication and bugs.

pulling this out of #1373


andychu:
Thank you for fixing this!

The unit test mentioned in VC would be great

melvinw:
Added a regression test

andychu:
Merged, thank you, sorry I missed this update  too :-/  I think the way Github sends notification e-mails didn't play well with GMail -- I will look more carefully in the future!

Maybe I should also clear out the old PRs so this becomes more obvious ... 

I'm not sure if setting "reviews" changes anything

PR #1380, opened 2022-10-18
[core] move trap-related state into SignalState

melvinw:
This is the first commit in a series that will refactor the signal and trap handling code to make it safely shareable between ptyhon and C++.

The first step in this process is to remove the illusion that the various users of the trap-related state each own it. We can make it more obvious that they're really borrowing global state by renaming the member(s) they all use to access it.

### DIFF
--- a/core/pyos.py
+++ b/core/pyos.py
@@ -22,6 +22,7 @@ from posix_ import WUNTRACED
 from typing import Optional, Tuple, List, Dict, cast, Any, TYPE_CHECKING
 
 if TYPE_CHECKING:
+  from _devbuild.gen.syntax_asdl import command_t
   from core.comp_ui import _IDisplay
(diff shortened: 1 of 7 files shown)


andychu:
Ohh somehow I didn't see this!  Sorry 

I saw the one about AfterForkingChild() and not this one ... very weird since I don't think I've missed a PR before :-/

Sorry for stomping on it, let me look now

andychu:
Hm I did get an e-mail notification about this -- I think what happened is that I thought this was the old PR, and didn't realize it was a new one.  Gah!  

Feel free to ping me if you don't get a response in 24 hours

andychu (commented):
This looks great, thank you!

Sorry again for missing it :-/

melvinw:
@andychu No sweat! I was taking a break from this yesterday anyway

PR #1381, opened 2022-10-18
[core] make post-fork signal setup translatable

melvinw:
This commit adds a wrapper around `signal.signal()` called `Sigaction()` and adds a C++ implementation for basic signal dispositions like SIG_DFL and SIG_IGN. We can use this to share some post-fork signal setup between python and C++.

A future commit will also use this to make some other signal-related code translatable.


andychu:
Thank you for splitting this up, looks great!

Did you mean to mark it "draft"?  Looks like I can't merge it that way.

Some people use the protocol "LGTM" to approve a merge, I think  you may be able to: LGTM :)

melvinw:
Oh whoops!  This should be ready now

PR #1384, opened 2022-10-20
[osh] move more trap mutations into SignalState

melvinw:
This will make some pending refactors a bit easier.


melvinw:
(meant to open this as a draft. ignore for now)

cpp/leaky_core.h
```diff
@@ -57,11 +57,24 @@ class TermState {
...
+  }
+  builtin_trap::_TrapHandler* GetHook(Str* hook_name) {
+    return hooks->get(hook_name);
+  }
+  void AddUserHook(Str* hook_name, builtin_trap::_TrapHandler* handler) {
+    hooks->set(hook_name, handler);
+  }
+  void RemoveUserHook(Str* hook_name) {
+    hooks->remove(hook_name);
+  }
```
melvinw, line 77: These will be replaced by translated code in the next PR. Just here to keep the build and tests happy in the meantime

melvinw:
ok this is ready now

core/pyos.py
```diff
@@ -354,12 +376,16 @@ def AddUserTrap(self, sig_num, handler):
...
     else:
       signal.signal(sig_num, handler)
+    self.traps[sig_num] = handler
     # TODO: SIGINT is similar: set a flag, then optionally call user _TrapHandler
 
   def RemoveUserTrap(self, sig_num):
     # type: (int) -> None
     """For user-defined handlers registered with the 'trap' builtin."""
     # Restore default
+    if sig_num in self.traps:
```
andychu, line 386: FWIW we have `mylib.dict_remove()`

https://github.com/oilshell/oil/blob/master/mycpp/mylib.py#L140

The benefit is that in C++ we won't involve any exceptions, which can be slow in the "main path" of the code

i.e. this is a shim, and in C++ there is another method that ensures key doesn't exist

core/pyos.py
```diff
@@ -345,6 +346,27 @@ def InitInteractiveShell(self, display, my_pid):
...
+    # type: () -> int
+    """Return the last signal that fired"""
+    return self.last_sig_num
+
+  def GetHook(self, hook_name):
+    # type: (str) -> _TrapHandler
+    """Return the handler associated with hook_name"""
+    return self.hooks.get(hook_name, None)
+
+  def AddUserHook(self, hook_name, handler):
```
andychu, line 359: I think technically hooks have nothing to do with signals and concurrent shared state -- they are purely "user space" / main loop


If we want to be really "pure" we could put them in a separate class, which is only used by `builtin_trap`

But that can always be done later, if at all

andychu (commented):
Looks good!

FWIW this issue is related to printing source code in `trap`:

https://github.com/oilshell/oil/issues/1082

core/pyos.py
melvinw: Ah nice I'll use that

core/pyos.py
melvinw: Oh I like this! Removes the need for that awkward temporary code-dup I commented on in `leaky_core`

melvinw:
All set! Have another look?

core/pyos.py
```diff
@@ -301,7 +302,8 @@ def __init__(self):
     self.sigwinch_handler = None  # type: SigwinchHandler
     self.last_sig_num = 0  # MUTABLE GLOBAL, for interrupted 'wait'
     # signal/hook name -> handler
-    self.traps = {}  # type: Dict[str, _TrapHandler]
+    self.hooks = {}  # type: Dict[str, _TrapHandler]
```
andychu, line 305: looks like this dict can be removed now

andychu (approved):

core/pyos.py
melvinw: ah good catch

andychu:
Merged, thanks!

PR #1385, opened 2022-10-21
[core,osh] make SignalState translatable

melvinw:
This is the second-to-last PR in the signal series. It's size is primarily because of the cut-and-paste of `SignalState` from `pyos` into `builtin_trap`. Otherwise it's mostly some house-keeping. After this the API between the main loop and pyos/kernel is just the six small functions below. The rest of the signal handling code is shared with the hook handling code in the main loop and is all translatable python. I think I like this API, but I'm open to comments, e.g. `SetSigwinchCode()` is a little awkward, but I think it's the least awkward option.
```
def Sigaction(sig_num, handler):
  # type: (int, Any) -> None
  """Set the disposition of a signal. handler is either SIG_DFL or SIG_IGN"""

def RegisterSignalInterest(sig_num):
  # type: (int) -> None
  """Have the kernel notify the main loop about the given signal"""

def GetPendingSignals():
  # type: () -> List[int]
  """Transfer ownership of the current queue of pending signals to the caller."""

def LastSignal():
  # type: () -> int
  """Returns the number of the last signal that fired"""

def SetSigwinchCode(code):
  # type: (int) -> None
  """
  Depending on whether or not SIGWINCH is trapped by a user, it is expected to
  report a different code to `wait`. SetSigwinchCode() lets us set which code is
  reported.
  """

def InitShell():
  # type: () -> None
  """Opaquely initialize state""" 
```

This brings `oil-native` and `opy` up to parity on `spec/builtin-trap`! A follow up PR will do some additional house-keeping to make some interactive shell and SIGWINCH stuff translatable too.

### DIFF
--- a/core/pyos.py
+++ b/core/pyos.py
@@ -266,26 +266,56 @@ def InputAvailable(fd):
 
 UNTRAPPED_SIGWINCH = -1
 
-class SigwinchHandler(object):
-  """Wrapper to call user handler."""
 
-  def __init__(self, display, sig_state):
-    # type: (_IDisplay, SignalState) -> None
-    self.display = display
-    self.sig_state = sig_state
-    self.user_handler = None  # type: _TrapHandler
+class _SignalHandler(object):
+  """
+  A singleton that implements a basic generic signal handler that enqueues
+  signals for asynchrnous processing as they are fired.
+  """
+  _instance = None
+
+  signal_queue = []  # type: List[int]
+  last_sig_num = 0  # type: int
+  sigwinch_num = UNTRAPPED_SIGWINCH
+
+  def __new__(cls):
+    # type: () -> None
+    if cls._instance is None:
+        cls._instance = super(_SignalHandler, cls).__new__(cls)
+
+    return cls._instance
 
   def __call__(self, sig_num, unused_frame):
     # type: (int, Any) -> None
-    """For Python's signal module."""
+    if sig_num == signal.SIGWINCH:
+      self.last_sig_num = self.sigwinch_num
+    else:
+      self.last_sig_num = sig_num
 
-    # SENTINEL for UNTRAPPED SIGWINCH.  If it's trapped, self.user_handler
-    # will overwrite it with signal.SIGWINCH.
-    self.sig_state.last_sig_num = UNTRAPPED_SIGWINCH
+    self.signal_queue.append(sig_num)
 
-    self.display.OnWindowChange()
-    if self.user_handler:
-      self.user_handler(sig_num, unused_frame)
+  def TakeSignalQueue(self):
+    # type: () -> List[int]
+    # A note on signal-safety here. The main loop might be calling this function
+    # at the same time a signal is firing and appending to
+    # `self.signal_queue`. We can forgoe using a lock here
+    # (which would be problematic for the signal handler) because mutual
+    # exclusivity should be maintained by the atomic nature of pointer
+    # assignment (i.e. word-sized writes) on most modern platforms.
+    # The replacement run list is allocated before the swap, so it can be
+    # interuppted at any point without consequence.
+    # This means the signal handler always has exclusive access to
+    # `self.signal_queue`. In the worst case the signal handler might write to
+    # `new_queue` and the corresponding trap handler won't get executed
+    # until the main loop calls this function again.
+    # NOTE: It's important to distinguish between signal-saftey an
+    # thread-saftey here. Signals run in the same process context as the main
+    # loop, while concurrent threads do not and would have to worry about
+    # cache-coherence and instruction reordering.
+    new_queue = []  #  type: List[int]
+    ret = self.signal_queue
+    self.signal_queue = new_queue
+    return ret
 
 
 def Sigaction(sig_num, handler):
@@ -299,10 +329,9 @@ class SignalState(object):
 
   def __init__(self):
     # type: () -> None
-    self.sigwinch_handler = None  # type: SigwinchHandler
-    self.last_sig_num = 0  # MUTABLE GLOBAL, for interrupted 'wait'
+    self.display = None  # type: _IDisplay
     # signal/hook name -> handler
-    self.traps = {}  # type: Dict[int, _TrapHandler]
+    self.traps = {}  # type: Dict[int, command_t]
     # appended to by signal handlers
     self.nodes_to_run = []  # type: List[command_t]
 
@@ -331,8 +360,9 @@ class SignalState(object):
 
     # This is ALWAYS on, which means that it can cause EINTR, and wait() and
     # read() have to handle it
-    self.sigwinch_handler = SigwinchHandler(display, self)
-    signal.signal(signal.SIGWINCH, self.sigwinch_handler)
+    self.display = display
+    signal.signal(signal.SIGWINCH, _SignalHandler())
+    _SignalHandler().sigwinch_num = UNTRAPPED_SIGWINCH
 
     # This doesn't make any tests pass, and we might punt on job control
     if 0:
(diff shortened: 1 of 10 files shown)


cpp/leaky_core.cc
```diff
@@ -16,6 +16,25 @@
...
 
+static List<int>* AllocSignalQueue() {
+  List<int>* ret = NewList<int>();
+  ret->reserve(kMaxSignalsInFlight);
+  return ret;
+}
+
+static List<int>* signal_queue = nullptr;
+static int last_sig_num = 0;
+static int sigwinch_num = UNTRAPPED_SIGWINCH;
```
melvinw, line 27: I _think_ the writes to these should be safe. If we want to be safe we can mark them as explicitly atomic or word-align them so they get treated similarly to pointers

cpp/leaky_core.cc
```diff
@@ -16,6 +16,25 @@
...
+
+static void signal_handler(int sig_num) {
+  assert(signal_queue != nullptr);
+  signal_queue->append(sig_num);
```
melvinw, line 31: Did we decide if `List::append()` was signal-safe in the reserved case? If not, I can switch to a static array here and just build the list in `GetPendingSignals()`

andychu: I think we should wrap all the global state in a class / struct

Then we can do the TSAN thing more logically.  (You could probably do it without, but for me it's easier to reason about)

I believe these are exactly the same in terms of signal safety

```
sigwinch_num = foo;

gSignalState.sigwinch_num = foo;
```

i.e. the global is at a known address, and so if the offset of the member, so the `.` doesn't actually matter for atomicity

core/pyos.py
```diff
@@ -266,26 +261,56 @@ def InputAvailable(fd):
...
+
+  def __new__(cls):
+    # type: () -> _SignalHandler
+    if cls._instance is None:
+        cls._instance = super(_SignalHandler, cls).__new__(cls)
 
-    # SENTINEL for UNTRAPPED SIGWINCH.  If it's trapped, self.user_handler
-    # will overwrite it with signal.SIGWINCH.
-    self.sig_state.last_sig_num = UNTRAPPED_SIGWINCH
+    return cls._instance
```
andychu, line 281: Can we just create a global like:

```
gSignalHandler = SignalHandler()

```

and then 

```
def LastSignal():
  return gSignalHander.LastSignal()
```

This is just a style thing but I think it makes the  code more consistent

andychu (commented):
Thanks, this looks good!  I like that it is making more code translatable

I made 2 suggestions around global state in C++ and Python -- the C++ one I think will help with TSAN, and the Python one is a style thing that will make it more consistent with C++

Basically I just want to use classes and methods, and then a single instance in both cases (no singleton)

If there are some other changes on top of this, we could do that refactoring later, but I think there are some benefits

melvinw:
pushed some patches for you comments @andychu

core/pyos.py
```diff
@@ -294,109 +315,43 @@ def Sigaction(sig_num, handler):
...
-    else:
-      signal.signal(sig_num, handler)
-    self.traps[sig_num] = handler
-    # TODO: SIGINT is similar: set a flag, then optionally call user _TrapHandler
+def GetPendingSignals():
+  # type: () -> List[int]
+  """Transfer ownership of the current queue of pending signals to the caller."""
+  global gSignalHandler
+  assert gSignalHandler is not None
+  return gSignalHandler.TakeSignalQueue()
```
andychu, line 331: Why not use just one name?  e.g. EITHER GetPendingSignals or TakeSignalQueue (I like the latter)

cpp/leaky_core.cc
```diff
@@ -179,12 +181,64 @@ bool InputAvailable(int fd) {
   NotImplemented();
 }
 
+SignalHandler::SignalHandler()
+    : signal_queue(), last_sig_num(0), sigwinch_num(UNTRAPPED_SIGWINCH) {
+}
+
+void SignalHandler::EnqueueSignal(int sig_num) {
+  assert(signal_queue != nullptr);
+  signal_queue->append(sig_num);
```
andychu, line 190: We should have an assert() that's it's less than the kMaxSignals, so we don't allocate in the handler


(We haven't really worked out if asserts should be on in release mode; currently they are.)

cpp/leaky_core.h
```diff
@@ -54,35 +56,29 @@ class TermState {
...
-  }
-  int last_sig_num = 0;
-  Dict<int, builtin_trap::_TrapHandler*>* traps;
-  List<syntax_asdl::command_t*>* nodes_to_run;
+  SignalHandler();
+  void EnqueueSignal(int sig_num);
+  List<int>* TakeSignalQueue();
 
-  DISALLOW_COPY_AND_ASSIGN(SignalState)
+  List<int>* signal_queue;
```
andychu, line 65: Please add a trailing underscore for members -- we're using the style consistently

(It basically comes from https://google.github.io/styleguide/cppguide.html , which is pretty  widely used )

cpp/leaky_core.cc
```diff
@@ -179,12 +181,64 @@ bool InputAvailable(int fd) {
   NotImplemented();
 }
 
+SignalHandler::SignalHandler()
+    : signal_queue(), last_sig_num(0), sigwinch_num(UNTRAPPED_SIGWINCH) {
+}
+
+void SignalHandler::EnqueueSignal(int sig_num) {
```
andychu, line 188: Might want to call this `UpdateVars()`  or something, because it does more than enqueue a signal

That's why I was thinking `SignalSafe::Update()` or `SignalShared::Update()` are more descriptive

i.e. this class isn't really a signal handler -- it's a bunch of state that gets mutated by a signal handler!

osh/builtin_trap.py
```diff
@@ -83,6 +55,96 @@ def RemoveUserHook(self, hook_name):
...
+    # TODO: SIGINT is similar: set a flag, then optionally call user _TrapHandler
+
+  def TakeRunList(self):
+      # type: () -> List[command_t]
+      """Transfer ownership of the current queue of pending trap handlers to the caller."""
+      sig_queue = pyos.GetPendingSignals()
+
+      run_list = []  # type: List[command_t]
+      for sig_num in sig_queue:
+        node = self.traps.get(sig_num, None)
```
andychu, line 134: FWIW it looks like self.traps is no longer read (or written) by the signal handler

So my instinct is to move it out of here, but it's not a big deal

i.e. I was thinking the ONLY members of the `SignalSafe` / `SignalShared` are the ones that can potentially be raced on

I guess one place to put it is to unify it with HookState (could be back to TrapState LOL)

andychu (commented):
Great thanks for making the changes! 

I made some style/naming comments which are suggestions, but I think the assert() should be done

cpp/leaky_core.h
melvinw: Ah right. Good catch

cpp/leaky_core.cc
melvinw: Good catch! This must have gotten lost in the shuffle

core/pyos.py
melvinw: I like the latter too. Not sure why I gave these separate names...

osh/builtin_trap.py
melvinw: Agh this one is awkward. If we move `SignalState.traps` over to `HookState` (or `TrapState` :smile:), `SignalState`'s member functions will basically just be passthrough to `pyos`... Then `TakeRunList()` now either needs to move too or `SignalState` needs to have a `HookState` as a member so it can find the `command_t`s for each signal. The semantics of `AddUserTrap()` would also get kind of weird. What if we combine `SignalState` and `HookState` into `TrapState` and relegate all the racey bits to `pyos` (where they already live)?

melvinw:
Pushed a patch for the style stuff and the missing assert. Not sure what the right approach for the `SignalState` vs `HookState` thing is, though... Perhaps we punt?

osh/builtin_trap.py
andychu: I was actually thinking that we can have one class with racey shared state (self.run_list) and then one class with the main thread state (self.hooks, self.traps)

But I think that will be clearer when we do the TSAN thing (which I suppose is not that urgent, since this looks pretty clean to me)

So this seems good for now

andychu (approved):

andychu:
Merged, thanks!

PR #1392, opened 2022-10-25
[cpp] make print_time() consistent with bash

melvinw:
The CPU load column doesn't exist there and the entries are separated by newlines by default. Also print to stderr.


andychu:
Thanks for fixing this!

PR #1393, opened 2022-10-26
Make Buf use GC

CoffeeTableEspresso:


frontend/consts_gen.py
```diff
@@ -162,7 +162,7 @@ def GenCharLookup(func_name, lookup, f, required=False):
 
   for char_code in sorted(lookup):
     f.write("  case '%s':\n" % CChar(char_code))
-    f.write('    return StrFromC("%s");\n' % CChar(lookup[char_code]))
+    f.write('    return StrFromC("%s", 1);\n' % CChar(lookup[char_code]))
```
andychu, line 165: I think if you merge master into your branch, then Github won't show these diffs?

That might be a good idea since the change has been open for awhile

mycpp/TEST.sh
```diff
@@ -90,8 +90,8 @@ examples-variant() {
     (asan|rvroot)
       # TODO: make examples/parse pass!
       # https://github.com/oilshell/oil/issues/1317
-      if test $num_failed -ne 1; then
-        echo "FAIL: Expected 1 failure in ASAN"
+      if test $num_failed -ne 0; then
+        echo "FAIL: Expected 0 failure in ASAN"
```
andychu, line 94: woohoo!

mycpp/leaky_mylib.cc
```diff
@@ -8,8 +8,18 @@ mylib::FormatStringer gBuf;
 
 namespace mylib {
 
-Str* StrFromBuf(const Buf& buf) {
-  return ::StrFromC(buf.data_, buf.len_);
+Str* StrFromBuf(const Buf* buf) {
+  return ::StrFromC(buf->data_, buf->len_);
```
andychu, line 12: This makes a copy, but we can look at removing that in a follow-up

mycpp/leaky_mylib.cc
```diff
@@ -137,51 +147,73 @@ bool CFileWriter::isatty() {
...
+    b->len_ = buf_->len_;
+    b->data_[b->len_] = '\0';
+    // free(buf_->data_);
+    buf_ = b;
+  }
+}
+
+void BufWriter::Extend(Str* s) {
+  ExpandBufCapacity(len(s));
+  buf_->Extend(s);
```
andychu, line 195: It seems like both of these lines do a memcpy() ?  Is that necessary?

What about `MarkSweepHeap::Reallocate()` which calls `realloc()`?

mycpp/leaky_mylib.h
```diff
@@ -166,10 +170,24 @@ class BufWriter : public Writer {
...
+
+  void ExpandBufCapacity(int n);
+  void Extend(Str* s);
+
+  bool BufIsEmpty() {
+    return buf_ == nullptr;
+  }
+
+  Buf* buf_;
+  bool buf_is_valid_ = true;
```
andychu, line 183: do we need this member?

mycpp/leaky_mylib.cc
```diff
@@ -8,8 +8,18 @@ mylib::FormatStringer gBuf;
...
-  return ::StrFromC(buf.data_, buf.len_);
+Str* StrFromBuf(const Buf* buf) {
+  return ::StrFromC(buf->data_, buf->len_);
+}
+
+Buf* NewBuf(int cap) {
+  void* place = gHeap.Allocate(sizeof(Buf) + cap + 1);
+
+  auto* b = new (place) Buf();
+  b->cap_ = cap;
```
andychu, line 19: style issue: It seems like we could make cap_ a constructor param rather than initializing to zero and then mutating it here

`len_` could also rely on the constructor

andychu (commented):
Thanks for doing this!  I took a quick look and may not have understood everything

I think the main question is if we should use realloc(), I don't see why not?

mycpp/leaky_mylib.cc
CoffeeTableEspresso: Ack

PR #1394, opened 2022-10-27
[frontend] make str lens explicit in lookup tables

melvinw:
We currently use the `StrFromC(char *)` form for `Str` creation in the generated character lookup tables for C++. This means '\0' gets interpreted as the empty string instead of a string of length one with a NULL byte in it.  Using the `StrFromC(char *, int)` form fixes this. For characters other than '\0' this should make no difference.


melvinw:
This should fix a few spec tests in `builtin-io` and `nul-bytes`

andychu:
Great catch, thank you !!

PR #1395, opened 2022-10-29
Create two basic spectests for argparse

Melkor333:
I finally found some time to do something about the spec tests for argparse.

Now I'm really unsure if it more or less resembles what we discussed, I think some of your points went over my head plus I read through the python argparse manual and took a lot of inspiration. The idea is that the hay definition should feel as much as possible like the python argparse definitions.

I intentionally only implemented 2 spec tests because I'm very unsure if they make sense! I'll add more spec tests for different features once these are approved 🙂 

I will do a final rebase (with proper commit naming) lateron

### DIFF
--- a/spec/oil-builtin-argparse.test.sh
+++ b/spec/oil-builtin-argparse.test.sh
@@ -4,14 +4,11 @@
 #
 # The following is as close as possible to the python argparse which seems to work well
 
-#### Argparse basic statement
+#### Argparse boolsche option and positional
 
 #I assume the hay definitions are given.
 ArgSpec myspec {
-  description = '''
-     Reference Implementation
-  '''
-  arg -v --verbose { type = Bool, help = "Verbose" }
+  arg -v --verbose { type = Bool }
   # TODO: can position be implicit (in given order?) or is the order not maintained in hay?
   arg firstPositional
 }
@@ -26,7 +23,7 @@ argparse (myspec, args, :opts)
 (List)   []
 ## END
 
-#### Argparse print automatic option "help"
+#### Argparse basic help message
 ArgSpec myspec {
   description = '''
      Reference Implementation
--- a/spec/oil-builtin-argparse.test.sh
+++ b/spec/oil-builtin-argparse.test.sh
@@ -5,43 +5,55 @@
 # The following is as close as possible to the python argparse which seems to work well
 
 #### Argparse boolsche option and positional
+hay define ArgSpec
+hay define ArgSpec/Arg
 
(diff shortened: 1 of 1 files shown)


andychu:
Thank you, I'm traveling now and will try to look at this tonight!

spec/oil-builtin-argparse.test.sh
```diff
@@ -2,34 +2,48 @@
...
 
-#### Argparse Prototype
-
-hay define argparse
-
-# Oops, we're running into this problem ...
+#I assume the hay definitions are given.
+ArgSpec myspec {
+  arg -v --verbose { type = Bool }
+  # TODO: can position be implicit (in given order?) or is the order not maintained in hay?
```
andychu, line 12: The order can be maintained (though we have to figure out how to do it)

Maybe add two positional args like `src` and `dest`?

spec/oil-builtin-argparse.test.sh
```diff
@@ -2,34 +2,48 @@
...
+#### Argparse basic help message
+ArgSpec myspec {
+  description = '''
+     Reference Implementation
+  '''
+  prog = "program-name"
+  arg -v --verbose { type = Bool, help = "Verbose" }
+  arg firstPositional
+}
+var args = ['-h' 'posone']
```
andychu, line 35: This might be missing a comma

Try

`test/spec.sh oil-builtin-argparse` to test it

It would be useful to see the errors

andychu (commented):
Thanks!  Somehow I forgot I added a skeleton back in May

But this looks like a good improvement!  I think I agree with all the changes

spec/oil-builtin-argparse.test.sh
Melkor333: I saw only now that `test/spec.sh` allows `-f`, otherwise it only shows "FAIL" 😄  Thanks for spotting the typo

Melkor333:
I re-added the `hay define..` lines so that we get a (more or less) reasonable error code when running the spec tests. The current reason for failure still points to a syntax error, but I think that's okay:
```
case	line	oil	
  0	 7	FAIL	Argparse boolsche option and positional
case: 0
[oil stdout] Expected '{\n  "verbose": true,\n  "src": "src/path",\n  "dst": "dst/path"\n}\n[\n\n]\n', got ''
[oil status] Expected 0, got 2
oil stdout:

oil stderr:
  argparse (myspec, args, :opts)
                          ^
[ stdin ]:10: Syntax error in expression (near Id.Arith_Colon)
```

PR #1397, opened 2022-11-05
[mycpp/runtime refactor] Prepare BufWriter for 2 optimizations

andychu:
Potential optimizations:

1. Use realloc() with MarkSweepHeap::Reallocate()

- Inline small private methods into BufWriter::write().  I think it's easier to see what happens to buf_ all in one place.
- Rename ExpandBufCapacity() -> EnsureCapacity() because it doesn't always "Expand".  It now takes a capacity integer arg, and returns a Buf*, which is similar to realloc().

2. "Zero copy" of mutable Buf -> immutable Str

- Move is_valid_ state out of Buf, into BufWriter.  Check it in both write() and getvalue().
- Remove bogus comment I added about ASAN error. This was because the test was using BufWriter incorrectly, and the assertions weren't tight enough.

Also:

- Add test coverage for the kEmptyString case.

This is a pure refactoring PR -- no behavior change.

### DIFF
--- a/mycpp/bump_leak_heap.cc
+++ b/mycpp/bump_leak_heap.cc
@@ -23,6 +23,10 @@ void* BumpLeakHeap::Allocate(int num_bytes) {
   return p;
 }
 
+void* BumpLeakHeap::Reallocate(void *p, int num_bytes) {
+  return Allocate(num_bytes);
+}
+
 void BumpLeakHeap::Report() {
   log("[BumpLeakHeap]");
   log("  num allocated = %10d", num_allocated_);
--- a/mycpp/bump_leak_heap.cc
+++ b/mycpp/bump_leak_heap.cc
@@ -24,7 +24,11 @@ void* BumpLeakHeap::Allocate(int num_bytes) {
 }
 
 void* BumpLeakHeap::Reallocate(void* p, int num_bytes) {
-  return Allocate(num_bytes);
+  // TODO:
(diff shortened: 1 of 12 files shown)


mycpp/bump_leak_heap.cc
```diff
@@ -23,6 +23,10 @@ void* BumpLeakHeap::Allocate(int num_bytes) {
   return p;
 }
 
+void* BumpLeakHeap::Reallocate(void* p, int num_bytes) {
+  return Allocate(num_bytes);
```
CoffeeTableEspresso, line 27: This discards the data from the previous pointer.

mycpp/leaky_mylib.cc
```diff
@@ -153,66 +153,60 @@ void Buf::Extend(Str* s) {
...
-    return;
-  }
-
+// TODO: realloc() to new capacity instead of creating NewBuf()
+Buf* BufWriter::EnsureCapacity(int capacity) {
   assert(buf_->cap_ >= buf_->len_);
 
-  if (buf_->cap_ < buf_->len_ + n) {
+  if (buf_->cap_ < capacity) {
     // buf_->cap_ = std::max(buf_->cap_ * 2, buf_->len_ + n);
```
CoffeeTableEspresso, line 165: Delete commented out lines?

CoffeeTableEspresso (approved):

andychu:
merged with CLI

PR #1400, opened 2022-11-06
add job control to oil-native

melvinw:
Just needed a few small fixes to make `builtin_process.py` translatable to get this working

### DIFF
--- a/cpp/leaky_stdlib.cc
+++ b/cpp/leaky_stdlib.cc
@@ -5,6 +5,7 @@
 
 #include <errno.h>
 #include <fcntl.h>      // open
+#include <signal.h>     // kill
 #include <sys/stat.h>   // umask
 #include <sys/types.h>  // umask
 #include <sys/wait.h>   // WUNTRACED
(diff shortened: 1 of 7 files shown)


cpp/leaky_stdlib.cc
```diff
@@ -111,6 +112,10 @@ void execve(Str* argv0, List<Str*>* argv, Dict<Str*, Str*>* environ) {
   InvalidCodePath();
 }
 
+void kill(int pid, int sig) {
+  assert(::kill(pid, sig) == 0);
```
andychu, line 116: To follow Python, this should check the return code and then

```
 throw Alloc<OSError>(errno);
```

Some of the other functions do this, and we should clean up ones that don't


(Annoying thing: Python 2 sometimes throws IOError and sometimes OSError, and it's not documented.   So I had to look at pyext/posixmodule.c  and read posix_error() to see which one it was! )

andychu (commented):
Awesome!   Thanks for doing this

cpp/leaky_stdlib.cc
melvinw: Ah nice. This is less aggressive than assert-failing here. Added

andychu:
Thanks!

PR #1405, opened 2022-11-10
implement Obj for a few types

melvinw:
The classes changed in this commit (or their sub-classes) are all allocated at some point or another in the code using `gc_heap::Alloc<T>()`. Since they don't have an object header, this leads to crashes or odd behavior when the garbage collector inspects them (e.g. they might have an invalid `heap_tag_`).


melvinw:
@andychu does this make sense? Noticed this while working on retval rooting. Kind of guessed at details like field masks and heap tags. Could use guidance on those

melvinw:
There are also a few cases where this change lead to use using multiple inheritance... The style guide strongly discourages this in most cases, but this might be fine?

cpp/leaky_core_error.cc
```diff
@@ -0,0 +1,72 @@
+// leaky_core_error.cc
```
andychu, line 1: FYI sorry about the weird structure of this code, I noted it in the "smells" thread

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/C.2B.2B.20Code.20Smells

This seems fine now but we will change the structure of errors a bit

cpp/leaky_core_error.h
```diff
@@ -16,61 +16,27 @@ using syntax_asdl::Token;
 using syntax_asdl::word_part_t;
 using syntax_asdl::word_t;
 
-class Usage : public std::exception {
+class Usage : public std::exception, Obj {
```
andychu, line 19: We might be able to avoid multiple inheritance by using the `OBJ_HEADER()` macro

I'm honestly not sure if multiple inheritance gives the desired layout -- it could be worth writing a test for this

cpp/leaky_frontend_match.cc
```diff
@@ -22,6 +22,13 @@ Tuple2<Id_t, int> OneToken(lex_mode_t lex_mode, Str* line, int start_pos) {
   return Tuple2<Id_t, int>(static_cast<Id_t>(id), end_pos);
 }
 
+SimpleLexer::SimpleLexer(MatchFunc match_func, Str* s)
+    : Obj(Tag::FixedSize, SimpleLexer::field_mask(), sizeof(SimpleLexer)),
```
andychu, line 26: Hm should we use this `::field_mask()` style everywhere?

It's kind of interesting that you can define it inline

cpp/leaky_frontend_match.h
```diff
@@ -21,14 +21,17 @@ Tuple2<Id_t, int> OneToken(lex_mode_t lex_mode, Str* line, int start_pos);
...
  public:
-  SimpleLexer(MatchFunc match_func, Str* s)
-      : match_func_(match_func), s_(s), pos_(0) {
-  }
+  SimpleLexer(MatchFunc match_func, Str* s);
   Tuple2<Id_t, Str*> Next();
   List<Tuple2<Id_t, Str*>*>* Tokens();
 
+  static constexpr uint16_t field_mask() {
+    return maskbit(offsetof(SimpleLexer, match_func_)) |
```
andychu, line 31: Hm I think the match function isn't a GC object, so this should be removed

Unit tests can verify this

andychu (commented):
Thanks for doing this!  It generally looks like the right idea

I think it is worth testing the correctness of the masks, because  debugging and finding  it "in production"  is painful

We haven't settled on style of doing that, but maybe we need to

This was the idea behind the "Big ASAN / Rooting Scrub" , i.e. I guess we need a "object header scrub" too

andychu:
OK I added `cpp/core_test.cc` and the static_assert showed that the multiple inheritance does break the layout

So the solution is just to remove `std::exception` -- I'm not sure why we had that in the first place!

---

I think I will add some skeletons for unit tests -- we need fine-grained tests to do the "Big ASAN / Rooting Scrub", which relies on correct GC headers too

Issue #1315, opened 2022-09-14
Build sandboxing could detect missing deps


andychu:
Examples:

- #1313 
- preprocessed metric missed ASDL header deps -- just fixed
- lexer_gen, etc.

caveats here:

https://github.com/oilshell/oil/wiki/Oil-Ninja-Build

andychu:
Ran into this problem here

https://github.com/oilshell/oil/commit/15693a6a86317ed3fa95b8bac5bbf7a374a25b51

PR #1407, opened 2022-11-11
[cpp] start to proliferate RootingScope

melvinw:
Using some of the more trivial bindings in `pyos` to test my understanding of things here.


melvinw:
If these look correct I'll tack on the `read()` bindings too

melvinw:
`cpp-small` seems to be failing in an unrelated way...

```
  (asdl_main c) -> _gen/frontend/types.asdl_c.h
  (frontend/consts_gen c) -> _gen/frontend/id_kind.asdl_c.h
  (lexer_gen) -> _build/tmp/frontend/match.re2c.txt
_build/tmp/frontend/match.re2c.txt -> (re2c) -> _gen/frontend/match.re2c.h
[1/2] clang coverage '' cpp/qsn_test.cc _build/obj/clang-coverage/cpp/qsn_test.o
[2/2] LINK clang coverage _bin/clang-coverage/cpp/qsn_test _build/obj/clang-coverage/cpp/qsn_test.o _build/obj/clang-coverage/mycpp/gc_mylib.o _build/obj/clang-coverage/mycpp/bump_leak_heap.o _build/obj/clang-coverage/mycpp/marksweep_heap.o _build/obj/clang-coverage/mycpp/leaky_containers.o _build/obj/clang-coverage/mycpp/leaky_builtins.o _build/obj/clang-coverage/mycpp/leaky_mylib.o
RUN _bin/clang-coverage/cpp/qsn_test > _test/clang-coverage/cpp/qsn_test.log
OK
[1/14] make-pystub _bin/shwrap/consts_gen frontend/consts_gen.py _devbuild/gen/hnode_asdl.py _devbuild/gen/types_asdl.py asdl/pybase.py asdl/runtime.py core/error.py core/pyerror.py frontend/builtin_def.py frontend/id_kind_def.py frontend/option_def.py mycpp/mylib.py pylib/collections_.py
[2/14] asdl_main cpp  frontend/syntax.asdl _gen/frontend/syntax.asdl _gen/frontend/syntax.asdl_debug.py
[3/14] consts_gen cpp _gen/frontend/id_kind.asdl
[4/14] clang coverage '' cpp/leaky_frontend_tdop.cc _build/obj/clang-coverage/cpp/leaky_frontend_tdop.o
FAILED: _build/obj/clang-coverage/cpp/leaky_frontend_tdop.o 
build/ninja-rules-cpp.sh compile_one clang coverage '' cpp/leaky_frontend_tdop.cc _build/obj/clang-coverage/cpp/leaky_frontend_tdop.o _build/obj/clang-coverage/cpp/leaky_frontend_tdop.o.d
In file included from cpp/leaky_frontend_tdop.cc:1:
cpp/leaky_osh.h:6:10: fatal error: '_gen/frontend/syntax.asdl.h' file not found
#include "_gen/frontend/syntax.asdl.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
[5/14] clang coverage '' cpp/leaky_pgen2.cc _build/obj/clang-coverage/cpp/leaky_pgen2.o
[6/14] clang coverage '' cpp/leaky_osh.cc _build/obj/clang-coverage/cpp/leaky_osh.o
```

andychu:
Oops sorry, yeah my build refactoring must have missed dependencies

which leads to non-deterministic failures when there is parallelism

We need some kind of build sandboxing to detect this 100%, it seems weird that most build systems don't have it (?)

andychu:
OK I fixed this: https://github.com/oilshell/oil/commit/15693a6a86317ed3fa95b8bac5bbf7a374a25b51

And #1315 tracks that build issue

cpp/leaky_core.cc
```diff
@@ -26,6 +26,7 @@ Tuple2<int, int> WaitPid() {
 }
 
 Tuple2<int, int> Read(int fd, int n, List<Str*>* chunks) {
+  RootingScope r_;
```
andychu, line 29: Let's use a leading underscore `_r` consistently

cpp/core_test.cc
```diff
@@ -23,6 +28,109 @@ TEST exceptions_test() {
...
+  ASSERT(os_type != nullptr);
+
+  utsname un = {};
+  ASSERT(uname(&un) == 0);
+  ASSERT(str_equals(StrFromC(un.sysname), os_type));
+
+  PASS();
+}
+
+TEST pyos_readbyte_test() {
```
andychu, line 68: These are from leaky_core_test.cc right?

I think we should remove those

Or is it hard to make them run in both modes?

melvinw: Yeah, there's still some work left to get this working with `StackRoots`. Simply adding it wherever I added `RootingScope` wasn't enough

andychu: OK I think we may have an unresolved problem here

I'm not sure if it's worth making them run in both modes, but I also don't want to lose test coverage for too long

We might just need to "tear the band-aid off" and go for it, but we should have all the ducks in a row

andychu:
Hmmm.. I think it's worth doing some tests to see if we can just abandon the old rooting scheme

Rather than having this elaborate dance of:

1. add two kinds of rooting simultaneously
2. make the tests work with both kinds
3. remove the old kind

That is probably a lot of wasted work, when the old rooting is only active  in tests/examples now

andychu:
I guess what we could do is make `-D RET_VAL_ROOTING` the default

I think it will just leak memory because everything is owned by one frame

But that's OK?

The tests will blow up in memory usage, and we are measuring that, but it's probably OK

https://www.oilshell.org/release/0.12.9/benchmarks.wwz/mycpp-examples/

melvinw:
All three options sound workable to me. I'm in no particular rush to get this merged, though. So, I'm happy to wait until we figure out why `StackRoots` is broken if that ends up being the right thing to do (though, after reading your comments it's less clear to me if that's the case...)

andychu:
OK I'm thinking about whether we can just turn on `-D RET_VAL_ROOTING` everywhere, and let it leak to start

And then we can just migrate tests to the new rooting scheme.  I think it's better not to have 2 active at once

I will do an experiment

melvinw:
Great! Just saw that land on master. I'll tidy up this branch now

cpp/core_test.cc
melvinw: Now that your change is in I moved everything from `leaky_core_test` into here and made the relevant bindings unleaky.

melvinw:
I'll do the signal state bindings in a separate PR

cpp/qsn.h
```diff
@@ -43,9 +45,11 @@ inline Str* UEscape(int codepoint) {
   // 3 for \u{
   // 6 for codepoint
   // 1 for }
+  RootingScope r_;
```
andychu, line 48: These two should also be `_r`

andychu (commented):
Thanks for updating the PR!  one small thing

cpp/qsn.h
melvinw: Ah nice catch! Will update

melvinw:
Should be ready to go now

andychu:
Great thanks, I'm going to do a little renaming and then we should be able to go wild on this

- RootingScope -> RootsFrame
  - Add NO_ROOTS_FRAME() to show that we decided against it
- Remove all` leaky_` prefixes

PR #1409, opened 2022-11-13
[cpp/qsn] Optimize rooting for core/qsn functions

andychu:
Remove the old rooting style.

Also update cpp/README.md with the new structure.

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -3,44 +3,53 @@
 
 Structure of this directory:
 
+Correspond to Python standard library modules:
+
+    stdlib{.h,.cc,_test.cc}
+      import fcntl
+      import time
+      import posix  # forked in pyext/posixmodule.c
+
 Correspond to repo directories:
 
-    core.{cc,h}
-    osh.{cc,h}
+    core{.h,.cc,_test.cc}
+
+    osh{.h,.cc,_test.cc}
       from osh import arith_parse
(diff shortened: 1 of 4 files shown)


andychu:
What do you think?  It seems clear "on inspection" for this case, although it's not super clear how to deal with longer and more complex functions

melvinw (approved):
Looks good! Nice catch on being able to pass the rooting down through to the heap libs. This makes me wonder then... When exactly does it make sense to be pushing new rooting frames? Only when there a multiple allocations?

Mostly just thinking out loud... I think the approach of putting `RootsFrame` everywhere then staring at individual functions and optimizing later (like this) will work out

andychu:
Yeah I think it would be nice to write down the rules in a doc somewhere,  like `doc/gc_rooting.md` or something

I think it's clear in this case, but not obvious in general.  The structure of the function affects it a lot

- multiple allocations
- conditionals

This is why I want a test-heavy approach with 100% coverage, which should be doable since we have 381 or so functions (actually less, because some are constructors which never require rooting)

So the hope is that if we make mistakes, the tests will catch them

And we also need to watch out for leaks, which is what I've been thinking about

----

I agree it's better to be conservative at first.  And that's what mycpp is going to do, because it can't be that smart at first


thanks for the comments!

Issue #1007, opened 2021-11-18
flaky process sub spec test


andychu:
Case #7 failed with SIGPIPE for some reason

```
7 process subs and pipelines together

[osh stdout] Expected '1\n2\n1\n2\n3\nstatus=4\nprocess_sub=2 3\npipeline=0 4\n__\n', got '1\n2\n1\n2\n3\nstatus=4\nprocess_sub=2 3\npipeline=141 4\n__\n'

stdout:

1
2
1
2
3
status=4
process_sub=2 3
pipeline=141 4
```

This is on the 24 core release machine

andychu:
http://travis-ci.oilshell.org/github-jobs/2021-12-24__20-11-32.wwz/_tmp/spec/survey/process-sub.html#details-7-osh

Ran into this again

andychu:
http://travis-ci.oilshell.org/srht-jobs/2022-03-21__15-51-20.wwz/_tmp/spec/survey/process-sub.html

```
1
2
1
2
3
status=4
process_sub=2 3
pipeline=141 4
```

same SIGPIPE

andychu:
saw it again on sourcehut: http://travis-ci.oilshell.org/srht-jobs/2022-07-09__17-10-03.wwz/_tmp/spec/survey/process-sub.html

andychu:
Hit this on the 0.12.7 release

andychu:
Hit it in CI

http://travis-ci.oilshell.org/github-jobs/2022-11-10__15-54-16.wwz/_tmp/spec/survey/process-sub.html

andychu:
https://www.oilshell.org/release/0.13.1/

andychu:
Hit this again

http://travis-ci.oilshell.org/github-jobs/2023-04-02__15-03-21.wwz/_tmp/spec/survey/process-sub.html

andychu:
Hit this during 0.15.0 release

PR #1410, opened 2022-11-13
[cpp] use RootsFrame in pyos signal bindings

melvinw:


cpp/core.h
```diff
@@ -55,7 +55,7 @@ class TermState {
   }
 };
 
-class SignalHandler {
+class SignalHandler : public Obj {
```
melvinw, line 58: Without this we lose track of the new signal queue and get a use-after-free

cpp/core.cc
```diff
@@ -211,48 +214,62 @@ void SignalHandler::Update(int sig_num) {
...
 void SetSigwinchCode(int code) {
-  gSignalHandler.sigwinch_num_ = code;
+  NO_ROOTS_FRAME(FUNC_NAME);  // no allocations here
+  assert(gSignalHandler != nullptr);
+  gSignalHandler->sigwinch_num_ = code;
 }
 
 void InitShell() {
-  gSignalHandler.signal_queue_ = AllocSignalQueue();
+  NO_ROOTS_FRAME(FUNC_NAME);  // Alloc() does it
```
andychu, line 270: Actually I think we could have a frame here, because otherwise Alloc<SignalHandler> would be "owned" by the caller of InitShell, and then it would not get freed for awhile

Also I think we may need `gHeap.AddGlobalRoot(gSignalHandler)` -- there are only a few globals in the program, but we have to prevent them from getting freed by the GC

cpp/core.cc
```diff
@@ -211,48 +214,62 @@ void SignalHandler::Update(int sig_num) {
...
   struct sigaction act = {};
   act.sa_handler = signal_handler;
   assert(sigaction(sig_num, &act, nullptr) == 0);
 }
 
 List<int>* TakeSignalQueue() {
-  return gSignalHandler.TakeSignalQueue();
+  NO_ROOTS_FRAME(FUNC_NAME);  // SignalHandler::TakeSignalQueue() does it
+  assert(gSignalHandler != nullptr);
+  auto* ret = gSignalHandler->TakeSignalQueue();
```
andychu, line 253: I would just use the original style of `return gSignalHandler...`

cpp/core.cc
```diff
@@ -211,48 +214,62 @@ void SignalHandler::Update(int sig_num) {
...
   List<int>* ret = signal_queue_;
   signal_queue_ = new_queue;
   return ret;
 }
 
 void Sigaction(int sig_num, sighandler_t handler) {
+  NO_ROOTS_FRAME(FUNC_NAME);  // no allocations here
   struct sigaction act = {};
   act.sa_handler = handler;
   assert(sigaction(sig_num, &act, nullptr) == 0);
```
andychu, line 235: While we're in here, I think we could change these to throw IOError or OSError -- it shouldn't matter which one, because we catch IOError_OSError

cpp/core.h
andychu: OK interesting, yeah I can see why having a global pointing to a GC managed object is probelmatic

But don't you think we need to also root the global?

andychu (commented):
Thanks for diving into this!

andychu:
Thinking about this a little more, I think this will happen to work ONLY if InitShell() is called by main(), so that the objects it allocates are "owned" by main (in other words, main() is responsible for de-rooting)

(Well it also depends on whether main() has a RootsFrame, and is osh_eval::main() has one, etc.)

Anyway I guess we should come up with a good test for that, and maybe it can be separate from this change

But it's worth thinking about and at least adding a TODO about

The thing I'm concerned with is gSignalHandler getting free() before the program ends

cpp/core.cc
melvinw: Ah whoops. This was leftover from some debugging I was doing

melvinw:
> Thinking about this a little more, I think this will happen to work ONLY if InitShell() is called by main(), so that the objects it allocates are "owned" by main (in other words, main() is responsible for de-rooting)

Precisely. I figured this was a safe assumption, but it's probably a good idea to make this more robust.

> The thing I'm concerned with is gSignalHandler getting free() before the program ends

Yeah, now that I think about it, all you'd probably need to lose this is to move the call to `InitShell()` into some helper (which mycpp would add a `RootsFrame` to) instead of calling it directly from `Main()`. There'd be no obvious connection to the breakage when looking at that diff at the python level. Pretty annoying debugging session just waiting to happen...

I can tack an attempt at `AddGlobalRoot()` to this PR and we can nail down the specifics.

melvinw:
Added special rooting for globals in 78d6d4c8f198329780b9b12269eae4e5ea256a67. What do you think?

mycpp/marksweep_heap.h
```diff
@@ -75,6 +76,13 @@ class RootSet {
     stack_[num_frames_ - 1].push_back(root);
   }
 
+  void AddGlobalRoot(Obj* root) {
```
andychu, line 79: This is the right idea, but we should have tests for the new API.  I think it can go in marksweep_heap_test.cc:

1. it could just make something that assigns a GC object to a global
2. And then it should crash under `gcevery mode`
3. and then add AddGlobalRoot() , which should fix the crash I believe

mycpp/marksweep_heap.h
```diff
@@ -97,6 +105,9 @@ class RootSet {
   // root_set_[2] is being pushed/popped/modified.
   std::vector<std::vector<Obj*>> stack_;
   int num_frames_ = 0;  // frames 0 to N-1 are valid
+
+  // Track globals separately from the stack to avoid mistakes.
+  std::vector<Obj*> globals_;
```
andychu, line 110: I think we should use `stack_[0]` instead, which will simplify the code.  Notice that we initialize `num_frames_(1)` for this frame.

This frame is currently

- the one that is handled by DoProcessExit(bool) -- it pops the last frame and Collect()
- That is, It's  for objects that are rooted until the end of main() -- i.e. returned by a function called by main

It can also be the "global" frame

I think the same tests will pass, but the code will be simpler

melvinw: Ah ok cool. I thought about using `stack_[0]` but wasn't super confident that we could count on it never being accidentally popped. Will update this and add a test

melvinw: In retrospect, this assert should have been sufficiently convincing... https://github.com/oilshell/oil/blob/master/mycpp/marksweep_heap.h#L74

andychu:
This failed because of a flaky test:

- #1007 

I will fix that

andychu:
Thanks for doing this!    Merged

Hm somehow when I looked at the PR a couple days ago it looked like the feedback hadn't  been addressed?  Not sure what happened

Somehow I saw an old diff

PR #1411, opened 2022-11-13
[cpp] add RootsFrame to libc bindings

melvinw:


cpp/libc.cc
```diff
@@ -12,17 +12,20 @@
...
 Str* gethostname() {
+  RootsFrame _r{FUNC_NAME};
   Str* result = OverAllocatedStr(HOST_NAME_MAX);
   int status = ::gethostname(result->data_, HOST_NAME_MAX);
   if (status != 0) {
     throw Alloc<OSError>(errno);
   }
   // Important: set the length of the string!
   result->SetObjLenFromC();
+  gHeap.RootOnReturn(result);
```
andychu, line 23: I think we can rely on OverAllocatedStr to root the object it returns


I think the key point is that there is no other object allocated by a function this gethostname() calls, so there's nothing else to "de-root" when this function returns

What do you think?

I think our testing should catch any problems here?  One possible thing to take advantage of is that a manual Collect() returns the number of live objects

andychu: I guess a possible consideration is having a leak in the case where OSError is thrown ?

But I think that works OK because the RootsFrame are exception safe

andychu (commented):
This is great, thanks!

cpp/libc.cc
melvinw: > I think we can rely on OverAllocatedStr to root the object it returns
>
>I think the key point is that there is no other object allocated by a function this gethostname() calls, so there's nothing else to "de-root" when this function returns
>
>What do you think?

Ah sure. I wrote this PR after the decision to be conservative by default about this, but It's pretty obvious that `RootsFrame` isn't necessary here. Let's drop it.

> I guess a possible consideration is having a leak in the case where OSError is thrown ?

Yeah I'm not sure what the best pattern is for these. I've noticed in some tests we do `gHeap.RootInCurrentFrame(...)` in the `catch` for this, but I wasn't sure if that was the standard. If we rooted them in the throwing frame we might lose them if whatever catches the exception calls `Alloc()` before reading the exception.

Should we have some special exception-only rooting mechanism kind of like the thing you proposed for globals on the other PR?

andychu: We haven't dealt that much with exceptions yet, but I don't think it's too bad because:

- I don't think there is a single instance of `catch` in hand-written code
- In generated code mycpp can ensure the invariant that the exception is rooted right when it's caught, so there's no possibility of allocating/GC'ing before that

Does that make sense?

melvinw: Ah great. I wasn't sure that I could count on safe catches. This could be another thing worth adding to the dev docs to avoid ambiguity.

PR #1414, opened 2022-11-18
[cpp] add retval rooting to frontend libs

melvinw:


melvinw:
Seems to be reasonable coverage of the heap code paths in `cpp/frontend_{flag_spec,match}.cc` already, though I'll try to add a new one for `SimpleLexer::Tokens()`.

I noticed `cpp-coverage` doesn't report anything for `frontend_flag_spec.cc` (perhaps because of the `CPP_UNIT_TESTS` thing) They happen to be measured elsewhere?

andychu:
Ah yes, there is a bug in `test/coverage.sh` -- it's not including the coverage from the `-D CPP_UNIT_TESTS`  directory !  I can fix that

cpp/frontend_match_test.cc
```diff
@@ -8,14 +8,15 @@ namespace Id = id_kind_asdl::Id;
 TEST match_test() {
   match::SimpleLexer* lex = match::BraceRangeLexer(StrFromC("{-1..22}"));
 
-  while (true) {
-    auto t = lex->Next();
-    int id = t.at0();
+  List<Tuple2<Id_t, Str*>*>* toks = lex->Tokens();
```
andychu, line 11: Did this test have to change?

It seems like we're not testing `Next()` anymore?

Can we have both tests for coverage?   i.e. the old one and the new one


I may have forgotten how this code works ...

melvinw: Ah `Tokens()` calls `Next()` here: https://github.com/oilshell/oil/blob/master/cpp/frontend_match.cc#L44-L46

melvinw: This change wasn't for any fundamental reason, though. Just a cheap way to bump up coverage since the old loop from the test and `Tokens()` are nearly identical

andychu:
Cool thanks!  

I just fixed the coverage report bug: http://travis-ci.oilshell.org/github-jobs/2022-11-18__07-19-17.wwz/_test/clang-coverage/unified/html/index.html

PR #1417, opened 2022-11-21
use MutableStr

CoffeeTableEspresso:


mycpp/gc_mylib.h
```diff
@@ -132,16 +132,42 @@ class Writer : public Obj {
   virtual bool isatty() = 0;
 };
 
-class Buf;  // forward declaration
+class MutableStr;
 
-Str* StrFromBuf(const Buf&);
+class Buf {
+ public:
+  // The initial capacity is big enough for a line
```
andychu, line 139: this comment should be a TODO

mycpp/gc_mylib.h
```diff
@@ -132,16 +132,42 @@ class Writer : public Obj {
...
+
+  char* data();
+  char* end();
+  int capacity();
+
+  MutableStr* str_;
+  int len_;
+
+  /*
+  // TODO: move this state into BufWriter
```
andychu, line 156: Why not have just BufWriter and MutableStr, instead of BufWriter, Buf, and MutableStr, as this comment suggests?

andychu (commented):
Thanks for working on this!  

My initial reaction is that it seems like there should be 2 classes instead of 3 ... what do you think?  The extra indirection makes things harder to follow IMO, i.e. abstractions/types that are only used in one place can be problematic (they don't abstract over things)

mycpp/gc_mylib.h
```diff
@@ -132,16 +132,13 @@ class Writer : public Obj {
...
-Str* StrFromBuf(const Buf&);
-Buf* NewBuf(int);
+class MutableStr;
 
 constexpr uint16_t maskof_BufWriter();
 
 class BufWriter : public Writer {
  public:
-  BufWriter() : Writer(Tag::FixedSize, maskof_BufWriter(), sizeof(BufWriter)) {
+  BufWriter() : Writer(Tag::FixedSize, maskof_BufWriter(), sizeof(BufWriter)), str_(0), len_(0) {
```
andychu, line 141: for pointer types, please spell the constant `nullptr` instead of `0`

mycpp/gc_mylib.cc
```diff
@@ -129,66 +140,42 @@ bool CFileWriter::isatty() {
...
+}
 
-  auto* b = new (place) Buf(cap);
-  return b;
+int BufWriter::capacity() {
+  return str_ ? len(str_) : 0;
 }
 
-void Buf::Extend(Str* s) {
+void BufWriter::Extend(Str* s) {
```
andychu, line 160: I would be inclined to inline Extend() into write(), since it's only used in one place and is very short, but that is just a suggestion

andychu (commented):
Looks great!  Thanks for doing this

FWIW I will probably follow up with the change to start it at >= 80 chars (because I realize how much GC pressure is an issue -- skipping the allocation of 10, 20, 40, etc. bytes should be a fairly big win in practice)

mycpp/gc_mylib.h
CoffeeTableEspresso: Done

mycpp/gc_mylib.cc
CoffeeTableEspresso: Done

mycpp/gc_mylib.h
CoffeeTableEspresso: Done

andychu:
Uh I think we hit a race condition with Github PRs!   I got an e-mail, and then merged, and don't see the `nullptr` change

PR #1421, opened 2022-12-05
History in XDG_DATA_HOME, configurable via Environment variable

eliminmax:
I created a modified version of 0.12.8, then ported my changes to 0.12.9, that changes the history file path to have the following behavior:

If `$HISTFILE_OIL`/`$HISTFILE_OSH` environment variable for the current language is set, use that.
Otherwise, if `$XDG_DATA_HOME` is set, save to `"$XDG_DATA_HOME"/oil/history_oil` or `"$XDG_DATA_HOME"/oil/history_osh`.
If neither option is set, fall back to `~/.local/share/oil/history_oil` or `~/.local/share/oil/history_osh`.

### DIFF
--- a/core/shell.py
+++ b/core/shell.py
@@ -611,7 +611,6 @@ def Main(lang, arg_r, environ, login_shell, loader, line_input):
         display = comp_ui.MinimalDisplay(comp_ui_state, prompt_state, debug_f)
       if 'HISTFILE_%s' % lang.upper() in environ:
         history_filename = environ.get('HISTFILE_%s' % lang.upper())
-        print(repr(history_filename))
       elif 'XDG_DATA_HOME' in environ:
         history_filename = os_path.join(
           environ.get('XDG_DATA_HOME'), 'oil/history_%s' % lang)
--- a/oil-version.txt
+++ b/oil-version.txt
@@ -1,5 +1,4 @@
-0.12.9-hist
-
+0.12.9
 # The first line of this file is the Oil version, and the rest is ignored.
 # It's used at build time for the release tarball, and at runtime for oil
 # --version / osh --version.


eliminmax:
Exactly - I prefer to keep any history files in one directory (specifically `~/.local/share/histfiles`. If you'd rather not include that, that's fine.

andychu:
OK sounds good, please remove the other env var and just use XDG_DATA_HOME

And revert the `oil-version.txt` change -- that is only for releases

Thank you

andychu:
Hm actually it sounds like XDG_DATA_HOME isn't enough if you want `~/.local/share/histfiles` ?

Setting that would mean a lot of other non-history data is stored there too

eliminmax:
Using this patch, set the `HISTFILE_OSH` and `HISTFILE_OIL` variables to point to `~/.local/share/histfiles/osh` and `~/.local/share/histfiles/oil`, respectively, and set similar variables or configuration options for other programs. That's why I wanted the other environment variables to be available as options.

eliminmax:
I noticed, once I installed my patched version on a bare-bones Linux From Scratch system I'd made, that if the directory of the history file is missing, regardless of what it is, it silently fails to save history, so I think my patch needs a bit more work anyway - will close for now, and reopen once I've finished my college finals and had a chance to fix that.

andychu:
Hm what shell do you normally use, and how do you point it at your `histfiles` dir?  

Hm looks like bash and zsh BOTH support HISTFILE

https://linuxhint.com/check-zsh-history/

But does that mean bash and zsh overlap?  I don't think they can actually read from the same file

---

I think there is something to do here, but I'm not sure what the right behavior is

andychu:
I think one thing we could do is have OSH support plain `$HISTFILE` and Oil support something else

andychu:
Hi, I'm thinking about this again.  I'm thinking of doing it like this:

- OSH respects `$HISTFILE` , which is a POSIX variable, and default is `~/.config/oils/osh_history`
- YSH respects `$HISTFILE_YSH` , default is `~/.config/oils/ysh_history`

We can do XDG_DATA_HOME later, but I wonder if that is enough for your use case ?

(Project was slightly renamed - https://www.oilshell.org/blog/2023/03/rename.html )

andychu:
Although I guess `~/.local/share/oils/{osh,ysh}_history` can be considered cleaner, since it's not config

bash and zsh put it in `~`

Fish follows that style, so I think we can do that

```
ls ~/.local/share/fish/
fish_history  generated_completions/
```

PR #1424, opened 2022-12-13
[core] add interface for readline

melvinw:
This commit replaces direct use of the line_input module for readline access with a class that mirrors the readline API. This should make it easier to use readline from the C++ code.

This gets rid of a good chunk of the remaining `Any`s in the `core/shell.py` paths.


melvinw:
@andychu What do you think? I did some manual tests with `bin/osh` and readline seems to work as expected (let's see how CI feels...). This still needs integration with readline in the C++ part, but wanted to make sure the approach was reasonable before spending time on that.

The placement in `pyutil` is pretty arbitrary. Open to moving this elsewhere...

andychu:
This looks excellent!  Thank you

I think we should move it out of `core/` because once we implement the C++ stuff, I think there it will be conditionally compiled because GNU readline might not be detected at `./configure` time

I think we could call it `core/pyreadline.py` and `cpp/core_pyreadline.{h,cc}` or something?

I guess "core" is kinda arbitrary, you could also put it in "frontend", since most of readline is about getting input from the user

So `frontend/pyreadline.py` and `cpp/frontend_pyreadline.{h,cc}`, which is quite verbose ... hm

Well we can think about it and do it in the next PR

andychu:
FWIW the `./configure` script for the Python app bundle build currently outputs `HAVE_READLINE` in shell, but not in `.h`

We should probably make one for C++, like `./configure2` for now ...     (That's ugly, another naming problem to solve)


```
l$ ./configure 



andy@andy-VirtualBox:~/git/oilshell/oil$ cat _build/detected-config.sh 
READLINE_DIR=
HAVE_READLINE=1
PREFIX=/usr/local
DATAROOTDIR=/usr/local/share


andy@andy-VirtualBox:~/git/oilshell/oil$ cat _build/detected-config.h
#define SIZEOF_INT 4
#define SIZEOF_LONG 8
#define SIZEOF_VOID_P 8
#define SIZEOF_SHORT 2
#define SIZEOF_FLOAT 4
#define SIZEOF_DOUBLE 8
#define SIZEOF_SIZE_T 8
```

PR #1425, opened 2022-12-14
rename readline wrapper

melvinw:
This is largely a cut-paste/rename job. However, the build changes in the second commit are worth discussing.

@andychu It looks like `_build/detected-config.h` is only used to build python for the ovm release. Is it too awkward to piggy-back on that for the c++ build?


configure
```diff
@@ -353,6 +353,27 @@ EOF
   cc_statement HAVE_STAT_TV_NSEC2 \
     'struct stat st; st.st_mtimespec.tv_nsec = 1; return 0;' \
     '#include <sys/stat.h>'
+
+  if test "$FLAG_without_readline" = 1; then
```
andychu, line 357: OK just reminding myself how this works ... right now `_build/detected-config.sh` has HAVE_READLINE=1, and is used by `build/ovm-compile.sh`

Eventually I think we get rid of that ...

cpp/frontend_pyreadline.cc
```diff
@@ -0,0 +1,82 @@
+#include "frontend_pyreadline.h"
+
+#include <assert.h>
+
+#include "_build/detected-config.h"
```
andychu, line 5: This header file also has a lot of "OVM" stuff to get rid of, but that's OK for now

cpp/frontend_pyreadline.cc
```diff
@@ -0,0 +1,82 @@
...
+
+namespace py_readline {
+
+Readline::Readline() : Obj(Tag::Opaque, kZeroMask, sizeof(Readline)) {
```
andychu, line 9: FWIW I started putting `kNoObjLen` for the length instead of `sizeof`

The Cheney allocator needed the length for all objects to copy, while Mark and Sweep doesn't, so it's unused

(the length would only be needed for variable length objects like Slab -- for list and dict)


At some point we should make them all consistent

frontend/py_readline.py
```diff
@@ -0,0 +1,93 @@
+"""
+py_reader.py: Code that's only needed in Python.  C++ will use other mechanisms.
```
andychu: should be `py_readline.py`

andychu (approved):
Looks great!  I edited a comment in one file

Other than that all the comments are FYI

andychu:
Great thank you!

PR #1427, opened 2022-12-17
[mycpp] add StrFormat()

melvinw:
This will make it easier to support dynamic format strigns when translating modules like core/comp_ui. It will also allow us to skip gnerating format string functions in mycpp, which should lead to smaller binaries (and less bespoke code to scan when debugging).

This commit just adds a library. A follow-up will update mycpp to generate code that uses it.

### DIFF
--- a/mycpp/gc_str.cc
+++ b/mycpp/gc_str.cc
@@ -1,8 +1,14 @@
 #include <ctype.h>  // isalpha(), isdigit()
+#include <stdarg.h>
+
+#include <regex>
 
 #include "mycpp/runtime.h"
 
 GLOBAL_STR(kEmptyString, "");
+
+static const std::regex gStrFmtRegex("([^%]*)(?:%([0-9]*)(.))?");
+static const int kMaxFmtWidth = 256;  // arbitrary...
 
 int Str::find(Str* needle, int pos) {
   int len_ = len(this);
@@ -461,4 +467,105 @@ List<Str*>* Str::split(Str* sep) {
   }
 
   return result;
+}
+
+static inline Str* _StrFormat(const char* fmt, va_list args) {
+  int n = strlen(fmt);
+  auto beg = std::cregex_iterator(fmt, fmt + n, gStrFmtRegex);
+  auto end = std::cregex_iterator();
+
+  char int_buf[kMaxFmtWidth];
+  std::string buf;
+  for (std::cregex_iterator it = beg; it != end; ++it) {
+    const std::cmatch& match = *it;
+
+    const std::csub_match& lit_m = match[1];
+    assert(lit_m.matched);
+    const std::string& lit_s = lit_m.str();
+    buf.append(lit_s.c_str());
+
+    int width = 0;
+    bool zero_pad = false;
+    const std::csub_match& width_m = match[2];
+    const std::string& width_s = width_m.str();
+    if (width_m.matched && !width_m.str().empty()) {
+      if (width_s[0] == '0') {
+        zero_pad = true;
+        assert(width_s.size() > 1);
+        assert(StringToInteger(width_s.c_str() + 1, width_s.size() - 1, 10,
+                               &width));
+      } else {
+        assert(StringToInteger(width_s.c_str(), width_s.size(), 10, &width));
+      }
+      assert(width >= 0 && width < kMaxFmtWidth);
+    }
+
+    char const* str_to_add = nullptr;
+    const std::csub_match& code_m = match[3];
+    const std::string& code_s = code_m.str();
+    if (!code_m.matched) {
+      assert(!width_m.matched);  // python errors on invalid format operators
+      goto done;
+    }
+    assert(code_s.size() == 1);
+    switch (code_s[0]) {
+    case '%': {
+      assert(code_s.size() == 1);
+      str_to_add = code_s.c_str();
+      break;
+    }
+    case 's': {
+      Str* s = va_arg(args, Str*);
+      str_to_add = s->data();
+      zero_pad = false;  // python ignores the 0 directive for strings
+      break;
+    }
+    case 'r': {
+      Str* s = va_arg(args, Str*);
+      s = repr(s);
+      str_to_add = s->data();
+      zero_pad = false;  // python ignores the 0 directive for strings
+      break;
+    }
+    case 'd':  // fallthrough
+    case 'o': {
+      int d = va_arg(args, int);
+      assert(snprintf(int_buf, kMaxFmtWidth, match.str().c_str(), d) > 0);
+      str_to_add = int_buf;
+      break;
+    }
+    default:
+      break;
+    }
+    assert(str_to_add != nullptr);
+
+  done:
+    if (str_to_add != nullptr) {
+      int n = strlen(str_to_add);
+      if (n < width) {
+        for (int i = 0; i < width - n; ++i) {
+          buf.push_back(zero_pad ? '0' : ' ');
+        }
+      }
+      buf.append(str_to_add);
+    }
+  }
+
+  return StrFromC(buf.c_str());
+}
+
+Str* StrFormat(const char* fmt, ...) {
+  va_list args;
+  va_start(args, fmt);
+  Str* ret = _StrFormat(fmt, args);
+  va_end(args);
+  return ret;
+}
+
+Str* StrFormat(Str* fmt, ...) {
+  va_list args;
+  va_start(args, fmt);
+  Str* ret = _StrFormat(fmt->data(), args);
+  va_end(args);
+  return ret;
 }
--- a/mycpp/gc_str.cc
+++ b/mycpp/gc_str.cc
@@ -469,9 +469,8 @@ List<Str*>* Str::split(Str* sep) {
   return result;
 }
 
-static inline Str* _StrFormat(const char* fmt, va_list args) {
-  int n = strlen(fmt);
-  auto beg = std::cregex_iterator(fmt, fmt + n, gStrFmtRegex);
+static inline Str* _StrFormat(const char* fmt, int fmt_len, va_list args) {
+  auto beg = std::cregex_iterator(fmt, fmt + fmt_len, gStrFmtRegex);
   auto end = std::cregex_iterator();
 
   char int_buf[kMaxFmtWidth];
@@ -501,6 +500,7 @@ static inline Str* _StrFormat(const char* fmt, va_list args) {
     }
 
     char const* str_to_add = nullptr;
+    int add_len = 0;
     const std::csub_match& code_m = match[3];
     const std::string& code_s = code_m.str();
     if (!code_m.matched) {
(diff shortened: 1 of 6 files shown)


mycpp/gc_str.cc
```diff
@@ -482,3 +485,97 @@ List<Str*>* Str::split(Str* sep) {
...
+    case 'r': {
+      str_to_add = va_arg(args, Str*);
+      str_to_add = repr(str_to_add);
+      zero_pad = false;  // python ignores the 0 directive for strings
+      break;
+    }
+    case 'd':  // fallthrough
+    case 'o': {
+      int d = va_arg(args, int);
+      assert(snprintf(int_buf, kMaxFmtWidth, match->str().c_str(), d) > 0);
```
melvinw, line 543: I was surprised to learn that `itoa()` et. al. aren't super portable, so we use `snprintf()` here. It feels a little awkward, though...

mycpp/gc_str.cc
```diff
@@ -482,3 +485,96 @@ List<Str*>* Str::split(Str* sep) {
 
   return result;
 }
+
+// TODO: We should reduce the number of allocations and copies done here.
+static inline Str* _StrFromFormat(const char* fmt, va_list args) {
```
melvinw, line 490: To keep things consistent with python's `%` operator, I'm pretty liberal with assertions in this function. I could be convinced we should just return `nullptr` and have the callers check for failures. It may require changing a lot of call sites, though...

mycpp/gc_str.cc
```diff
@@ -482,3 +488,96 @@ List<Str*>* Str::split(Str* sep) {
...
+  int n = strlen(fmt);
+  auto beg = std::cregex_iterator(fmt, fmt + n, gStrFmtRegex);
+  auto end = std::cregex_iterator();
+  mylib::BufWriter* w = Alloc<mylib::BufWriter>();
```
andychu, line 497: can be `auto*`

Although actually how about using `std::string`  for a growable string, and then create a string at the end?


BufWriter will create GC pressure when it grows

Also then you don't have to write StrFromC() fragments

I suppose you have one extra copy at the end (since BufWriter::getvalue() no longer copies) but I think std::string is still better, since we know the lifetime

mycpp/gc_str.cc
```diff
@@ -482,3 +488,96 @@ List<Str*>* Str::split(Str* sep) {
...
+  for (std::cregex_iterator match = beg; match != end; ++match) {
+    auto& lit = (*match)[1];
```
andychu, line 499: for readability I think NOT using auto would be better, since I don't know this API

IMO it's nice to have the type shown exactly once, not zero or 2 times

mycpp/gc_str.cc
```diff
@@ -482,3 +488,96 @@ List<Str*>* Str::split(Str* sep) {
...
+    w->write(StrFromC(lit.str().c_str(), lit.str().size()));
+
+    bool zero_pad = false;
+    int width = 0;
+    auto& width_m = (*match)[2];
+    if (width_m.matched && !width_m.str().empty()) {
+      if (width_m.str()[0] == '0') {
+        zero_pad = true;
+        assert(width_m.str().size() > 1);
+        width = atoi(&width_m.str()[1]);
```
andychu, line 510: atoi() is generally buggy, and we have some functions that wrap strtol() in gc_builtins

But I guess for this case it should be OK since the regex in validates it?  That is worth a command

mycpp/gc_str.cc
```diff
@@ -482,3 +488,96 @@ List<Str*>* Str::split(Str* sep) {
...
+        w->write(padding);
+      }
+      w->write(str_to_add);
+    }
+  }
+
+  return w->getvalue();
+}
+
+Str* StrFromFormat(const char* fmt, ...) {
```
andychu, line 569: I might be tempted to call these functions Sprintf() or StringFormat() or StrFormat()

mycpp/gc_str.cc
```diff
@@ -482,3 +488,96 @@ List<Str*>* Str::split(Str* sep) {
...
+// TODO: We should reduce the number of allocations and copies done here.
+static inline Str* _StrFromFormat(const char* fmt, va_list args) {
+  int n = strlen(fmt);
+  auto beg = std::cregex_iterator(fmt, fmt + n, gStrFmtRegex);
+  auto end = std::cregex_iterator();
+  mylib::BufWriter* w = Alloc<mylib::BufWriter>();
+  for (std::cregex_iterator match = beg; match != end; ++match) {
+    auto& lit = (*match)[1];
+    assert(lit.matched);
+    w->write(StrFromC(lit.str().c_str(), lit.str().size()));
```
andychu, line 501: how about making a variable from `lit.str()` and `width_m.str()` etc.?   I think that would read more nicely

andychu (commented):
I scanned over this, not sure I understood it all since I haven't used the C++ regex API

I think the std::string could make it easier to read perhaps, and should create less GC pressure

mycpp/gc_str.cc
melvinw: Ah great. We might as well use the helpers

mycpp/gc_str.cc
melvinw: Yeah, I'd really like to avoid unnecessary GC pressure here. I'll adapt to use `std::string`

mycpp/gc_str.cc
melvinw: I like `StrFormat()`. Will update

melvinw:
Thanks for the notes @andychu! Have another look?

mycpp/gc_str.cc
```diff
@@ -462,3 +468,105 @@ List<Str*>* Str::split(Str* sep) {
...
+      if (n < width) {
+        for (int i = 0; i < width - n; ++i) {
+          buf.push_back(zero_pad ? '0' : ' ');
+        }
+      }
+      buf.append(str_to_add);
+    }
+  }
+
+  return StrFromC(buf.c_str());
```
andychu, line 555: This should also pass the length, to avoid the 1-argument version of StrFromC() calling strlen() again on it

mycpp/gc_str.cc
```diff
@@ -462,3 +468,105 @@ List<Str*>* Str::split(Str* sep) {
 
   return result;
 }
+
+static inline Str* _StrFormat(const char* fmt, va_list args) {
+  int n = strlen(fmt);
```
andychu, line 473: this strlen() could also be moved to the 2 callers -- in one case it passes len() which is pre-computed, in the other it calls strlen()

mycpp/gc_str.cc
```diff
@@ -462,3 +468,105 @@ List<Str*>* Str::split(Str* sep) {
...
+    const std::csub_match& code_m = match[3];
+    const std::string& code_s = code_m.str();
+    if (!code_m.matched) {
+      assert(!width_m.matched);  // python errors on invalid format operators
+      goto done;
+    }
+    assert(code_s.size() == 1);
+    switch (code_s[0]) {
+    case '%': {
+      assert(code_s.size() == 1);
```
andychu, line 513: remove duplicate assert

mycpp/gc_str.cc
```diff
@@ -462,3 +468,105 @@ List<Str*>* Str::split(Str* sep) {
...
+      str_to_add = code_s.c_str();
+      break;
+    }
+    case 's': {
+      Str* s = va_arg(args, Str*);
+      str_to_add = s->data();
```
andychu, line 519: Hm technically this does not follow Python semantics, where NULs aren't treated as special, unlike in C

```
>>> s = '\0 foo'
>>> 'format %s' % s
'format \x00 foo'
```

mycpp/gc_str.cc
```diff
@@ -462,3 +468,105 @@ List<Str*>* Str::split(Str* sep) {
...
+    case 'd':  // fallthrough
+    case 'o': {
+      int d = va_arg(args, int);
+      assert(snprintf(int_buf, kMaxFmtWidth, match.str().c_str() + lit_s.size(),
+                      d) > 0);
+      str_to_add = int_buf;
+      break;
+    }
+    default:
+      break;
```
andychu, line 539: I would also add an assert(0) here, even though you have one below -- it reads more nicely


```
>>> '%z' % 'foo'
ValueError: unsupported format character 'z' (0x7a) at index 1
```

andychu (commented):
Thanks, looks much better this way!  I think a few tweaks will make it more consistent with C++/Python, and only use C-style NUL terminated strings at the "edges"

- str_to_add can be `std::string` too, which respects NUL bytes
- use pointer + length  in the helper, instead of assuming NUL-terminated, avoiding extra strlen()

andychu:
Oh also just a note about error checking, no change:

- I think it's good for mycpp to continue statically parsing the format strings it can, just like C++ compilers do.  It should fail if it finds a malformed static one
- For dynamic strings, I think it's OK if the runtime just does assert(0) for now
  - although I guess we haven't decided if asserts will always be on in the final build -- they are now ...

mycpp/gc_str.cc
melvinw: Ah! nice catch

mycpp/gc_str.cc
melvinw: ah great catch. i'll add a test for this

melvinw:
Pushed a few changes.

> - str_to_add can be std::string too, which respects NUL bytes

I added a local to track the length of the chunk to be appended instead (this also removes the need for that `strlen()` at the end of the loop). I think using `std::string` could lead to unnecessary heap allocations and copies without careful use of `std::move()`. Let me know what you think of the current code.

mycpp/gc_str.cc
```diff
@@ -462,3 +468,108 @@ List<Str*>* Str::split(Str* sep) {
...
+        assert(StringToInteger(width_s.c_str(), width_s.size(), 10, &width));
+      }
+      assert(width >= 0 && width < kMaxFmtWidth);
+    }
+
+    char const* str_to_add = nullptr;
+    int add_len = 0;
+    const std::csub_match& code_m = match[3];
+    const std::string& code_s = code_m.str();
+    if (!code_m.matched) {
```
andychu, line 506: This part confused me for awhile ...  

I guess one problem with the asserts is that it's hard to test the error cases, hm.  

I just checked C's printf is that it doesn't seem to have error cases!  It just gives you garbage :-P

Hm actually can we just get rid of the goto and make this `break` ?

Because nothing in the `done:` block will get executed?

mycpp/gc_str.cc
```diff
@@ -462,3 +468,108 @@ List<Str*>* Str::split(Str* sep) {
...
+      break;
+    }
+    default:
+      assert(0);
+      break;
+    }
+    assert(str_to_add != nullptr);
+
+  done:
+    if (str_to_add != nullptr) {
```
andychu, line 548: I think if we change `goto done` to break, we don't need this nullptr check, because every other case will append something

andychu (commented):
Looks great, sorry noticed one last thing ... 

It's possible I'm not reading the code correctly, but in the case it could use some comments around the `goto done` part

mycpp/gc_str.cc
melvinw: Oh you're right! This `goto` is completely unnecessary, since the literal part is appended directly to `buf`. Pushed a commit to remove it

andychu:
Great thank you!

PR #1429, opened 2022-12-18
[core] make comp_ui translatable

melvinw:
Everything but the readline and completion dependencies can be translated now. Let's fold them into `shell_native.py`.

This PR also includes some commits that add C++ versions of the builtins needed by `comp_ui`.

### DIFF
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -4,13 +4,14 @@ comp_ui.py
 from __future__ import print_function
 
 import atexit
-import sys
 
 from core import ansi
 from core import completion
 import libc
 
-from typing import Any, List, Optional, Dict, IO, TYPE_CHECKING
+from mycpp import mylib
+
+from typing import Any, List, Optional, Dict, TYPE_CHECKING
 if TYPE_CHECKING:
   from frontend.py_readline import Readline
   from core.util import _DebugFile
@@ -44,12 +45,13 @@ def _PromptLen(prompt_str):
     elif c == '\x02':
       escaped = False
     elif not escaped:
-      display_str += c
+      # mycpp: rewrite of +=
+      display_str = display_str + c
   last_line = display_str.split('\n')[-1]
   try:
     width = libc.wcswidth(last_line)
   # en_US.UTF-8 locale missing, just return the number of bytes
-  except (SystemError, UnicodeError):
+  except UnicodeError:
     return len(display_str)
   if width == -1:
     return len(display_str)
@@ -90,24 +92,24 @@ class _IDisplay(object):
   """Interface for completion displays."""
 
   def __init__(self, comp_state, prompt_state, num_lines_cap, f, debug_f):
-    # type: (State, PromptState, int, IO[bytes], _DebugFile) -> None
+    # type: (State, PromptState, int, mylib.Writer, _DebugFile) -> None
     self.comp_state = comp_state
     self.prompt_state = prompt_state
     self.num_lines_cap = num_lines_cap
     self.f = f
     self.debug_f = debug_f
 
-  def PrintCandidates(self, *args):
-    # type: (*Any) -> None
+  def PrintCandidates(self, unused_subst, matches, unused_match_len):
+    # type: (Optional[str], List[str], int) -> None
     try:
-      self._PrintCandidates(*args)
-    except Exception as e:
+      self._PrintCandidates(unused_subst, matches, unused_match_len)
+    except Exception:
       if 0:
         import traceback
         traceback.print_exc()
 
   def _PrintCandidates(self, unused_subst, matches, unused_match_len):
-    # type: (Optional[Any], List[str], Optional[Any]) -> None
+    # type: (Optional[str], List[str], int) -> None
     """Abstract method."""
     raise NotImplementedError()
 
@@ -126,14 +128,15 @@ class _IDisplay(object):
     # Doesn't apply to MinimalDisplay
     pass
 
-  def PrintRequired(self, msg, *args):
-    # type: (str, *Any) -> None
-    # This gets called with "nothing to display"
-    pass
+  if mylib.PYTHON:
+    def PrintRequired(self, msg, *args):
+      # type: (str, *Any) -> None
+      # This gets called with "nothing to display"
+      pass
 
-  def PrintOptional(self, msg, *args):
-    # type: (str, *Any) -> None
-    pass
+    def PrintOptional(self, msg, *args):
+      # type: (str, *Any) -> None
+      pass
 
   def OnWindowChange(self):
     # type: () -> None
@@ -148,10 +151,9 @@ class MinimalDisplay(_IDisplay):
   It could be useful if we ever have a browser build!  We can see completion
   without testing it.
   """
-  def __init__(self, comp_state, prompt_state, debug_f, num_lines_cap=10,
-               f=sys.stdout):
-    # type: (State, PromptState, _DebugFile, int, IO[bytes]) -> None
-    _IDisplay.__init__(self, comp_state, prompt_state, num_lines_cap, f,
+  def __init__(self, comp_state, prompt_state, debug_f):
+    # type: (State, PromptState, _DebugFile) -> None
+    _IDisplay.__init__(self, comp_state, prompt_state, 10, mylib.Stdout(),
                        debug_f)
 
     self.reader = None
@@ -164,7 +166,7 @@ class MinimalDisplay(_IDisplay):
     self.f.write(self.comp_state.line_until_tab)
 
   def _PrintCandidates(self, unused_subst, matches, unused_match_len):
-    # type: (Optional[Any], List[str], Optional[Any]) -> None
+    # type: (Optional[str], List[str], int) -> None
     #log('_PrintCandidates %s', matches)
     self.f.write('\n')  # need this
     display_pos = self.comp_state.display_pos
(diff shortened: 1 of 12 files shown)


mycpp/gc_builtins.cc
```diff
@@ -255,3 +257,29 @@ bool str_equals0(const char* c_string, Str* s) {
     return false;
   }
 }
+
+int hash(Str* s) {
+  std::hash<void*> h;
+  return h(static_cast<void*>(s));
```
melvinw, line 263: this could potentially just be
```suggestion
  return reinterpret_cast<int>(s);
```

core/comp_ui.py
```diff
@@ -44,12 +45,14 @@ def _PromptLen(prompt_str):
     elif c == '\x02':
       escaped = False
     elif not escaped:
-      display_str += c
+      display_str = display_str + c
```
andychu, line 48: hm maybe add a comment `# mycpp rewrite of +=`  here

I did that in a few other places where I didn't quite like how it turned out

core/comp_ui.py
```diff
@@ -149,8 +153,8 @@ class MinimalDisplay(_IDisplay):
   without testing it.
   """
   def __init__(self, comp_state, prompt_state, debug_f, num_lines_cap=10,
-               f=sys.stdout):
-    # type: (State, PromptState, _DebugFile, int, IO[bytes]) -> None
+               f=mylib.Stdout()):
```
andychu, line 156: Hm default args don't get translated to anything; this is probably silently wrong

Also constructors are translated a bit differently than others

In a few places I had to use this `Init_*` method pattern, like `parse_ctx.Init_Trail()`

it is sort of a hack around keyword args.   You put the default initialization in the constructor, and then change the callers to call `Init_Field()` when they want to override

---

I'm  definitely open to better options!

core/comp_ui.py
```diff
@@ -189,22 +193,23 @@ def _PrintCandidates(self, unused_subst, matches, unused_match_len):
...
 
 def _PrintPacked(matches, max_match_len, term_width, max_lines, f):
-  # type: (List[str], int, int, int, IO[bytes]) -> int
+  # type: (List[str], int, int, int, mylib.Writer) -> int
   # With of each candidate.  2 spaces between each.
   w = max_match_len + 2
 
   # Number of candidates per line.  Don't print in first or last column.
-  num_per_line = max(1, (term_width-2) // w)
+  num_per_line = max(1, int((term_width-2) / w))
```
andychu, line 212: Hm this might be equivalent, but I think it's also a good idea to translate `//` to `/` in mycpp

Let me see if I can add that

core/comp_ui.py
```diff
@@ -313,7 +320,7 @@ def __init__(self,
                prompt_state,  # type: PromptState
                debug_f,  # type: _DebugFile
                readline,  # type: Optional[Readline]
-               f=sys.stdout,  # type: IO[bytes]
+               f=mylib.Stdout(),  # type: mylib.Writer
```
andychu, line 323: ditto, default args aren't translated

core/comp_ui.py
```diff
@@ -368,13 +375,13 @@ def _ReturnToPrompt(self, num_lines):
...
   def _PrintCandidates(self, unused_subst, matches, unused_max_match_len):
-    # type: (Optional[Any], List[str], Optional[Any]) -> None
+    # type: (Optional[str], List[str], int) -> None
     term_width = self._GetTerminalWidth()
 
     # Variables set by the completion generator.  They should always exist,
     # because we can't get "matches" without calling that function.
     display_pos = self.comp_state.display_pos
-    self.debug_f.log('DISPLAY POS in _PrintCandidates = %d', display_pos)
+    self.debug_f.write('DISPLAY POS in _PrintCandidates = %d' % display_pos)
```
andychu, line 384: I think this needs a `\n` to match log()

I guess we should get rid of the method `DebugFile::log()` in favor of `%` ?   

I have been meaning to do that for some other instances

varargs in different functions are annoying to translate -- I think we could be consistent and use `%` everywhere

core/comp_ui.py
```diff
@@ -394,7 +401,7 @@ def _PrintCandidates(self, unused_subst, matches, unused_max_match_len):
     # This could be more accurate but I think it's good enough.
     comp_id = hash(''.join(matches))
     if comp_id in self.dupes:
-      self.dupes[comp_id] += 1
+      self.dupes[comp_id] = self.dupes[comp_id] + 1
```
andychu, line 404: please add `# mycpp rewrite of +=` too

cpp/libc.cc
```diff
@@ -163,4 +165,22 @@ Tuple2<int, int>* regex_first_group_match(Str* pattern, Str* str, int pos) {
   return tup;
 }
 
+int wcswidth(Str* str) {
+  int len = mbstowcs(NULL, str->data(), 0);
+  if (len == -1) {
+    throw Alloc<UnicodeError>(StrFromC("mbstowcs error: Invalid UTF-8 string"));
+  }
+  wchar_t unicode[len + 1];
```
andychu, line 173: hm it occurs to be that the Python code is using a dynamically sized array?

I think it was contributed

I would have used malloc() / free()

I guess the reason people don't like it because there's the possibility of stack overflow

(Although I guess many other shells use stack allocation AFAIK)

core/comp_ui.py
```diff
@@ -44,12 +45,14 @@ def _PromptLen(prompt_str):
...
       escaped = False
     elif not escaped:
-      display_str += c
+      display_str = display_str + c
   last_line = display_str.split('\n')[-1]
   try:
     width = libc.wcswidth(last_line)
   # en_US.UTF-8 locale missing, just return the number of bytes
-  except (SystemError, UnicodeError):
+  except SystemError:
```
andychu, line 53: Reading the code in `pyext/libc.c` it seems it never throws SystemError?  And our C++ doesn't

So I guess delete this, it was dead code ?

mycpp/gc_builtins.cc
andychu: To match Python, we should hash the CONTENTS of the string

The use of hash() in `comp_ui.py` is weird and I wonder if it could be improved (open to suggestions)

But I think the shortest solution is to hash the sting for now

andychu (commented):
Thanks for doing this!

Let me see if I can add a couple things to mycpp

andychu:
I just added support for `//` which works the same for positive numbers only, not negative (!)

https://github.com/oilshell/oil/commit/1e386ec9870d4e088ddcc20ef6f89628d8b45e85

cpp/libc.cc
melvinw: Huh yeah this is odd... I just copied this from `pyext/libc.c` and didn't even see this. We should definitely use `malloc()`/`free()` here

mycpp/gc_builtins.cc
melvinw: oh you're right... for some reason i thought `hash(str) == id(str)` in python. i'll update this

core/comp_ui.py
melvinw: oh wow, nice catch! will do adopt `Init_*()` here

core/comp_ui.py
melvinw: ah yeah. i have a plan to convert the other `log()`s to `write()`s as i translate other files. then i'll remove `DebugFile::log()` once all of the call sites have been translated

core/comp_ui.py
melvinw: Actually, it turns out nobody overrides these kwargs right now... so the latest commit just removes them and hard-codes the defaults. Figured we can add a way to set them in the future if/when we need it. What do you think?

mycpp/gc_builtins.h
```diff
@@ -16,6 +16,10 @@ class _ExceptionOpaque : public Obj {
 };
 
 // mycpp removes constructor arguments
+class Exception : public _ExceptionOpaque {};
+
+class SystemError : public _ExceptionOpaque {};
```
andychu, line 21: I think SystemError can be removed now

mycpp/gc_builtins.h
```diff
@@ -59,6 +63,22 @@ inline RuntimeError::RuntimeError(Str* message)
...
+ public:
+  explicit UnicodeError(Str* message);
+  Str* message;
+
+  static constexpr uint16_t field_mask() {
+    return maskbit(offsetof(UnicodeError, message));
+  }
+};
+
+inline UnicodeError::UnicodeError(Str* message)
```
andychu, line 77: minor: if the constructor is in the header, we should put them in the `class { }`

The reason there are some constructors outside `class { }` was because I thought we needed to do that for field masks!    We couldn't define the mask function without the compiler saying it hadn't seen the fields yet.

But I think you showed otherwise, so I want to go back and change the style everywhere!

core/comp_ui.py
```diff
@@ -44,12 +45,15 @@ def _PromptLen(prompt_str):
...
+      # mycpp: rewrite of +=
+      display_str = display_str + c
   last_line = display_str.split('\n')[-1]
   try:
     width = libc.wcswidth(last_line)
   # en_US.UTF-8 locale missing, just return the number of bytes
-  except (SystemError, UnicodeError):
+  except UnicodeError:
+    return len(display_str)
+  except OSError:
```
andychu, line 56: Please remove OSError here, since I think we should have a separate path for allocation failure

An OSError should only be for things Python raises OSError for

(It actually raises `MemoryError` for out of memory, but I'm not sure if we should do that as well)

cpp/libc.cc
```diff
@@ -163,4 +165,33 @@ Tuple2<int, int>* regex_first_group_match(Str* pattern, Str* str, int pos) {
...
 
+int wcswidth(Str* str) {
+  int len = mbstowcs(NULL, str->data(), 0);
+  if (len == -1) {
+    throw Alloc<UnicodeError>(StrFromC("mbstowcs error: Invalid UTF-8 string"));
+  }
+
+  auto* unicode = static_cast<wchar_t*>(malloc(len + 1));
+  if (unicode == nullptr) {
+    // NOTE: glibc malloc() sets errno to ENOMEM on failure, but customer
```
andychu, line 176: I think we should either just `assert(unicode != nullptr)` or throw a custom exception and for allocation failure and handle it at the top level, with a custom message

Right now we are using malloc() in a few places, and those aren't checked

I think this is conceptually distinct and much rarer than IOError/OSError, so I'm OK with deferring it for awhile

andychu (commented):
Looks great, just some minor tweaks!

core/comp_ui.py
andychu: Yup sounds good!  This code probably has some generality that isn't used

----

BTW we have a hack for ONE optional arg on FUNCTIONS/methods if you ever need it, but I don't think it works for constructors

It's actually a hard-coded list in `mycpp/cppgen_pass.py` :-P    (another thing I want to get rid of, but it works)

mycpp/gc_builtins.h
melvinw: ah great. i was wondering why the surrounding code was like this. will send a follow up to make them consistent with the new style

mycpp/gc_builtins.h
melvinw: ah yes

melvinw:
updated!

andychu:
Great thanks!

PR #1430, opened 2022-12-18
[mycpp] translate % to StrFormat()

melvinw:


melvinw:
I don't understand the logic in `cppgen_pass.py` super well yet, but this seems to work. @andychu is this missing anything?

melvinw:
Pushed a few changes. Have a look?

Turns out adding support for arbitrary expressions as format strings was pretty easy. Just had to call `accept()`

mycpp/gc_builtins.cc
```diff
@@ -4,13 +4,17 @@
...
+  for (int i = 0; i < len(s); ++i) {
+    fputc(s->data()[i], stdout);
+  }
   fputs("\n", stdout);
 }
 
 // Like print(..., file=sys.stderr), but Python code explicitly calls it.
 void println_stderr(Str* s) {
-  fputs(s->data(), stderr);
+  for (int i = 0; i < len(s); ++i) {
```
andychu, line 15: Ah I see, I guess this was necessary to make tests pass with NUL bytes!


When we do the C I/O -> POSIX I/O changes, I think this will end up using `man 2 write`

```
write(0, s->data(), len(s))
write(0, "\n", 1)
```

andychu (commented):
Looks great, thank you!

andychu:
I added `mycpp/examples/invalid_format_strings.py` in this commit (currently in soil-staging, will be on master)



https://github.com/oilshell/oil/commit/a4607aa7b138130020ad85c429f52663e2049750


Which can be run with `mycpp/TEST.sh test-invalid-examples`


A flaw with this testing method and mycpp is that it only allows ONE error per file

It would be better if you could assert that the file has 5 errors, but unfortunately no

PR #1437, opened 2022-12-30
[translation] share fanos between C++ and pyext

melvinw:
A future commit will need this change to make core/main_loop.py translatable.

### DIFF
--- a/cpp/fanos.cc
+++ b/cpp/fanos.cc
@@ -1,0 +1,55 @@
+// fanos.cc
+
+#include "cpp/fanos.h"
+
+#include <assert.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+#include "cpp/fanos_shared.h"
+
+namespace fanos {
+
+static constexpr int kNumFds = 3;
+static constexpr int kSizeOfFds = sizeof(int) * kNumFds;
+
+void send(int sock_fd, Str* blob) {
+  int fds[kNumFds] = {-1, -1, -1};
+  char* value_err = NULL;
+  int err_code = 0;
+  fanos_send(sock_fd, blob->data(), len(blob), fds, &err_code, &value_err);
+  if (err_code != 0) {
+    throw Alloc<IOError>(err_code);
+  }
+  if (value_err != NULL) {
+    throw Alloc<ValueError>(StrFromC(value_err));
+  }
+}
+
+Str* recv(int sock_fd, List<int>* fd_out) {
+  int fds[kNumFds] = {-1, -1, -1};
+  char* value_err = NULL;
+  int err_code = 0;
+  int data_len = -1;
+  char* data_buf = fanos_recv(sock_fd, fds, &data_len, &err_code, &value_err);
+  if (err_code != 0) {
+    throw Alloc<IOError>(err_code);
+  }
+  if (value_err != NULL) {
+    throw Alloc<ValueError>(StrFromC(value_err));
+  }
+
+  for (int i = 0; i < 3; i++) {
+    fd_out->append(fds[i]);
+  }
+  if (data_len == 0) {
+    return nullptr;
+  }
+
+  return StrFromC(data_buf, data_len);
+}
+
+}  // namespace fanos
--- a/cpp/fanos.cc
+++ b/cpp/fanos.cc
@@ -18,38 +18,38 @@ static constexpr int kSizeOfFds = sizeof(int) * kNumFds;
 
 void send(int sock_fd, Str* blob) {
   int fds[kNumFds] = {-1, -1, -1};
-  char* value_err = NULL;
-  int err_code = 0;
-  fanos_send(sock_fd, blob->data(), len(blob), fds, &err_code, &value_err);
-  if (err_code != 0) {
-    throw Alloc<IOError>(err_code);
+  FanosError err = {0};
+  fanos_send(sock_fd, blob->data(), len(blob), fds, &err);
+  if (err.err_code != 0) {
+    throw Alloc<IOError>(err.err_code);
   }
-  if (value_err != NULL) {
-    throw Alloc<ValueError>(StrFromC(value_err));
+  if (err.value_err != NULL) {
+    throw Alloc<ValueError>(StrFromC(err.value_err));
   }
 }
 
 Str* recv(int sock_fd, List<int>* fd_out) {
+  FanosError err = {0};
+  FanosResult res = {nullptr, -1};
   int fds[kNumFds] = {-1, -1, -1};
-  char* value_err = NULL;
-  int err_code = 0;
-  int data_len = -1;
-  char* data_buf = fanos_recv(sock_fd, fds, &data_len, &err_code, &value_err);
-  if (err_code != 0) {
-    throw Alloc<IOError>(err_code);
+  fanos_recv(sock_fd, fds, &res, &err);
+  if (err.err_code != 0) {
+    throw Alloc<IOError>(err.err_code);
   }
-  if (value_err != NULL) {
-    throw Alloc<ValueError>(StrFromC(value_err));
+  if (err.value_err != NULL) {
+    throw Alloc<ValueError>(StrFromC(err.value_err));
   }
 
   for (int i = 0; i < 3; i++) {
     fd_out->append(fds[i]);
   }
-  if (data_len == 0) {
+  if (res.len == 0) {
     return nullptr;
   }
 
-  return StrFromC(data_buf, data_len);
+  Str* ret = StrFromC(res.data, res.len);
+  free(res.data);
+  return ret;
 }
 
 }  // namespace fanos
--- a/cpp/fanos_shared.c
+++ b/cpp/fanos_shared.c
@@ -9,8 +9,8 @@
 #include <sys/socket.h>
 #include <unistd.h>
 
-static const int kNumFds = 3;
-static const int kSizeOfFds = sizeof(int) * kNumFds;
+#define NUM_FDS 3
+#define SIZEOF_FDS (sizeof(int) * NUM_FDS)
 
 // error string helper
 static void set_err(char** dst, const char* msg) {
@@ -53,7 +53,7 @@ void fanos_send(int sock_fd, char* blob, int blob_len, const int* fds,
   union {
     /* ancillary data buffer, wrapped in a union in order to ensure
        it is suitably aligned */
-    char buf[CMSG_SPACE(kSizeOfFds)];
+    char buf[CMSG_SPACE(SIZEOF_FDS)];
     struct cmsghdr align;
   } u;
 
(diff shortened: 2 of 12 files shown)


core/pyfanos.py
```diff
@@ -0,0 +1,11 @@
+import fanos
+
+from typing import List
+
+def send(sock_fd, blob):
+  # type: (int, str) -> None
+  fanos.send(sock_fd, blob)
+
+def recv(sock_fd, fd_out):
```
melvinw, line 9: we seem to use a pretty limited subset of the kwargs for `fanos.send()` and `fanos.recv()`. so, i only added the ones we need here. will require some work to fold in the others

andychu: OK I had to remind myself why these exist!   You're referring to the `fd0` `fd1` `fd2` args in `pyext/fanos.pyi` , which is only for `send()`.

It appears the issue is that Oil is the "headless" or "FANOS" SERVER.   So it receives field descriptors.

The CLIENT would send 3 file descriptors -- stdin, stdout, stderr.  (I just hard-coded 3 to make it easy.)

So I think it's probably fine to do this incrementally, let me read over the rest ...

andychu (commented):
Thanks for doing this! 

I'm thinking about the general issue of `pyext/` Python extensions vs. `cpp/` bindings.  We have 3 that are duplicated:

- libc
- fastlex
- posix_

And then two that are not yet ported:

- line_input wrapper for GNU readline
- fanos, the protocol for the headless shell

This change is consistent with what we've done already, which is basically to duplicate them

But the last two are the most problematic, since they have a lot of hairy logic.

Have you looked at `line_input` ?  what do you think of it?

---

I guess my main concern is that FANOS is the one that's most likely to change ... whereas libc and POSIX really won't change.

And also when I  read back on this code, I don't remember how it works, even though I wrote it :)

----

Also is there some way we can hook up tests?  Right now we have `pyext/fanos_test.py`

I'm wary of having code that is duplicated and also without unit tests

I'm not sure what the answer is -- it might be good to discuss on our next call

melvinw:
> Have you looked at line_input ? what do you think of it?

`line_input` didn't look too bad. The only somewhat hairy looking parts were some of the `libedit` special casing for macos and the callback stuff. All very solvable, though.

> Also is there some way we can hook up tests? Right now we have pyext/fanos_test.py
>
> I'm wary of having code that is duplicated and also without unit tests

Ah yes. I was hoping we might have some coverage of this from the spec/gold tests (Are there any for headless mode? I haven't checked). But having some unit tests here would be nice.

As I've been porting the stuff from pyext, the changes have all been pretty superficial. It looks like there are three main parts to each of these bindings:
1. The fiddly code to "unmarshal" arguments
2. The bits that build and throw exceptions
3. The actual logic of the binding

So far (3) has been very easy to tease apart from (1) and (2).

Personally, I'm not too worried about duplication here (as long as the tests are also duplicated). Like you said, it's not that much code, and I get the impression that all of these things are very stable when compared to the rest of the shell and runtime. However, if we really wanted to avoid duplication we could write really bare pure-c versions of the essential logic for each of the bindings and helpers. Then in `cpp` and `pyext` we could just have wrappers around them that deal with (1) and (2) above. This might require slight refactoring of (3) in some cases, but it could be worthwhile. What do you think?

andychu:
Yeah I looked into some other options like SWIG / pybind11, but I think they are way overkill

So I think for now we can have some duplication and it's OK

Hm I'm looking around for tests -- now I see `client/run.sh demo`, which I uses the protocol, but not in the context of Oil.

----

I think having a pure C version of fanos that is wrapped by Python and C++ is a good idea in general .  Though it's probably not worth it now, given that the whole feature is a bit experimental

So I guess it depends how close we are to unifying shell.py and shell_native.py?  It feels like there are still some hard problems to solve there

andychu:
I guess to have a more direct path forward, I would say if we can add tests similar to `pyext/fanos_test.py` then it's mergable

The duplication is OK since there aren't good options I see, but the new code should have tests.  Ideally there would be some way to share the tests

Also I somewhat prefer if it's in a file outside core.cc.    Right now we have

- pyext/libc.c ->  cpp/libc.cc
- pyext/posixmodule.c -> cpp/stdlib.cc (since `import posix` is part of the stdlib)
- pyext/fastlex.c  isn't necessary in C++
- pyext/line_input.py -> wrapped by frontend/py_readline.py.  I guess in this case we changed the interface

----

On a related note, is it possible to avoid the tiny pyfanos wrapper?

Could we do something like:

```
fanos.send(fd, payload)
fanos.send_with_filenos(fds, payload, 0, 1, 2)
```

So then we can use the same interface directly?



--

BTW happy new year!

andychu:
Thinking about it a little more, the idea of having a "normal" C or C++ library and wrapping it with both Python and garbage-collected C++   does have some appeal

In that case I think we could just keep the existing `fanos_test.py` and then the C++ version would be a trivial wrapper, so it wouldn't need testing.

The lib could `bool success` instead of exceptions, and maybe has an out param for error strings.  I think `std::string` might be fine to use if necessary

----

Not sure how this would turn out, it's an idea ... I'm open to other solutions that preserve the test coverage

melvinw:
Happy new year!

I'll give the bare c/c++ library w/ wrappers approach a shot in the next couple of days. I'll also push some changes to separate this from `cpp/core.cc`. I may need to bug you about some build system stuff for this, but let's see how far I get...

> On a related note, is it possible to avoid the tiny pyfanos wrapper?

Probably? I noticed for the posix bindings we use this `import posix_ as posix` pattern. There's a similar one for `time` too (`import time as time_`). I'll sniff around the build scripts to see how those work and apply here.

andychu:
OK definitely hit me up on Zulip if you run into issues or want to brainstorm ... I don't know exactly how it should look

- The `posix_` name is because I don't want `import posix` to pick up the standard library module.  Python has a bit of ambiguity around PYTHONPATH in that respect
- I think `import time as time_` is because in the C++ code, there might have been a `time` symbol that conflicted with a namespace?  Not sure exactly

Either way I tend to add trailing `_` to avoid name conflicts (even though admittedly that is also our style for member vars)

So I think fanos shouldn't have any name conflicts, I think it can just be `fanos::recv()` and so forth

andychu:
Ah I noticed the build errors due to some Python tarball stuff 

I can probably help fix that tomorrow

That stuff is a bit "incidental" and should hopefully go away in the long term

melvinw:
Ah great. Thanks! I was just about to dig into those logs

andychu:
OK yeah let me try to fix it up tomorrow, I think it should be easy in most cases

But yeah the python tarball is a bit of hacky rats nest

C++ much less so :)

andychu:
Oh actually I pulled this branch and did `build/py.sh all`

```
cpp/fanos_shared.c:13:31: error: initializer element is not constant
 static const int kSizeOfFds = sizeof(int) * kNumFds;
                               ^~~~~~
error: command 'x86_64-linux-gnu-gcc' failed with exit status 1

```

So I think maybe it is a C vs C++ thing -- Python extensions are built more like C ?

Let me see

andychu:
OK I fixed that, let's see if passes CI

Sorry I thought it was some weird shell script build bug haha, but it was C vs. C++

andychu:
```
../py-yajl/yajl.c:164:15: note: (near initialization for 'yajl_methods[0].ml_meth')
/usr/bin/ld: /tmp/ccgHmrBI.o: in function `func_send':
fanos.c:(.text+0x75): undefined reference to `fanos_send'
/usr/bin/ld: /tmp/ccgHmrBI.o: in function `func_recv':
fanos.c:(.text+0x142): undefined reference to `fanos_recv'
collect2: error: ld returned 1 exit status
```

OK  this is closer to what I was thinking of

andychu:
OK, I fixed one silly build bug (it didn't like the `.c` extension), and fixed up the "ovm tarball" build

I want to get rid of the latter once we move fully to C++!

Thanks, let's see if it passes now

cpp/fanos_shared.c
```diff
@@ -0,0 +1,212 @@
...
+#include <string.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+#define NUM_FDS 3
+#define SIZEOF_FDS (sizeof(int) * NUM_FDS)
+
+// error string helper
+static void set_err(char** dst, const char* msg) {
+  *dst = (char*)malloc(strlen(msg) + 1);
```
andychu, line 17: Hm will this cause a leak?

How about we just say that all the error strings must be GLOBAL CONSTANTS?  And I think they are.  There is no formatting

Then we can just assign them directly, and avoid malloc?

( It would be  nice to have ASAN test this, but I'm not sure how to do it with Python's distutils build.  Should  be pretty possible though )

andychu (commented):
Looks great now, thanks!

I kinda wish we had some kind of (char*, len) thing to reduce the number of  params

But I think we should live with it for now, and eventually wrap our garbage-collected C++ directly with Python somehow

I'm thinking about the JSON binding `py-yajl` too

andychu:
Also not sure if it helps, but a possible style is

```
struct FanosError {
  int err_num;
  char* value_error;
};
```

And then the functions could be called like

```
FanosError e = { 0 };  // zero-initialize
send(..., &e)
if (e.err_num != 0)  {
}
if (e.value_error) {
}
```

cpp/fanos_shared.c
melvinw: Ah good question... The user of `fanos_*()` owns `*dst`, so I figured something in python would handle freeing it once the exception was destroyed. The docs for `PyErr_SetString()` (https://docs.python.org/3/c-api/exceptions.html#c.PyErr_SetString) aren't super clear on how the `const char*` arg is treated. Though, it would definitely be odd if it were freeing a `const` that it didn't own... I like the globals idea, though. I'll check that it's safe to do and switch to it

melvinw:
Ah yes this would reduce some output param bloat. Will try it out

cpp/fanos.cc
```diff
@@ -0,0 +1,55 @@
...
+    throw Alloc<ValueError>(StrFromC(value_err));
+  }
+}
+
+Str* recv(int sock_fd, List<int>* fd_out) {
+  int fds[kNumFds] = {-1, -1, -1};
+  char* value_err = NULL;
+  int err_code = 0;
+  int data_len = -1;
+  char* data_buf = fanos_recv(sock_fd, fds, &data_len, &err_code, &value_err);
```
andychu, line 37: Oh I think data_buf might leak here?  Because it is malloc'd in fanos_recv.  

StrFromC() is always a copy -- the copy becomes owned by the GC.  But the original still exists


Since fanos_shared.c has two callers, I think it's a good idea to document the memory management responsibilities in the interface,  I guess in the header file

---

Also just like the struct for errors, it might be useful to have

```
struct FanosResult {
  char* data;
  int len;
}
```

and call the same way:

```
FanosResult result = { 0 };
FanosError e = { 0 };

if (fanos_recv(sock_fd, fds, &result, &e)) {  // success or failure
  Str* ret = StrFromC(result->data, result->len);
  free(result->data);  
  ...
} else {
  // Handle errors
}
```


I tend to use the "out params" style a lot in C, since it is pretty general and functions written in the style compose pretty easily

andychu: I suppose errno could also be the return value, so you check

`if (fanos_recv(...) != 0)`

I guess whatever makes the call stack more consistent / shorter

melvinw: Passing the error struct as an output param felt slightly cleaner to me

melvinw:
Thanks for the build fixes and nice catch on those leaks! Pushed a fix. Let me know what you think

cpp/fanos.cc
```diff
@@ -0,0 +1,55 @@
...
+  }
+  if (err.value_err != NULL) {
+    throw Alloc<ValueError>(StrFromC(err.value_err));
+  }
+
+  for (int i = 0; i < 3; i++) {
+    fd_out->append(fds[i]);
+  }
+  if (res.len == 0) {
+    return nullptr;
```
andychu, line 47: It feels like this should be kEmptyString, not nullptr, but not sure

I will pull it and check it out ... could be some missing test coverage in `fanos_test.py` around this case

andychu:
Yup we lacked test coverage on the empty string case !  I fixed the confusion between empty string and EOF (which wasn't documented either :-/ )

https://github.com/oilshell/oil/commit/472bf7d088e8a6a52e5dd48c1059ad29a5d51d13

Thanks for diving into this, that was a big help !   It feels good that we've retained the tests

PR #1438, opened 2022-12-30
[mycpp] add some missing member initializers

melvinw:
g++-10 complains otherwise.

### DIFF
--- a/mycpp/gc_dict.h
+++ b/mycpp/gc_dict.h
@@ -39,30 +39,20 @@ class Dict {
  public:
   Dict()
       : GC_CLASS_FIXED(header_, field_mask(), sizeof(Dict)),
-        len_(),
-        capacity_(),
-        entry_(),
-        keys_(),
-        values_() {
-    assert(len_ == 0);
-    assert(capacity_ == 0);
-    assert(entry_ == nullptr);
-    assert(keys_ == nullptr);
-    assert(values_ == nullptr);
+        len_(0),
+        capacity_(0),
+        entry_(nullptr),
+        keys_(nullptr),
+        values_(nullptr) {
   }
 
   Dict(std::initializer_list<K> keys, std::initializer_list<V> values)
       : GC_CLASS_FIXED(header_, field_mask(), sizeof(Dict)),
-        len_(),
-        capacity_(),
-        entry_(),
-        keys_(),
-        values_() {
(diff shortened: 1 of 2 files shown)


mycpp/gc_dict.h
```diff
@@ -37,7 +37,13 @@ List<T>* ListFromDictSlab(Slab<int>* index, Slab<T>* slab, int n) {
...
  public:
-  Dict() : GC_CLASS_FIXED(header_, field_mask(), sizeof(Dict)) {
+  Dict()
+      : GC_CLASS_FIXED(header_, field_mask(), sizeof(Dict)),
+        len_(),
+        capacity_(),
+        entry_(),
+        keys_(),
+        values_() {
     assert(len_ == 0);
```
andychu, line 47: OK I think it probably makes sense to switch to this style ... The implicit zero initialization was left over from the Cheney heap, where you allocate within a big block that is zero'd.

Right now `MarkSweepHeap` actually use calloc() which zeroes the buffer, but I think we could switch to `malloc()` and rely on C++ initialization.  The only cases I know where we need custom initialization are the flexible arrays `data_[1]`  in `Str` and `Slab`.

mycpp/gc_dict.h
```diff
@@ -46,7 +52,12 @@ class Dict {
...
-      : GC_CLASS_FIXED(header_, field_mask(), sizeof(Dict)) {
+      : GC_CLASS_FIXED(header_, field_mask(), sizeof(Dict)),
+        len_(),
+        capacity_(),
+        entry_(),
+        keys_(),
+        values_() {
     assert(len_ == 0);
     assert(capacity_ == 0);
     assert(entry_ == nullptr);
```
andychu, line 63: Please remove all the assert() and DCHECK() as well, since they're now redundant

Personally I would be explicit with `len_(0)` and `keys_(nullptr)`, since I  didn't know you could leave them out !  :-)

andychu, line 63: Please remove all the assert() and DCHECK() as well, since they're now redundant

Personally I would be explicit with `len_(0)` and `keys_(nullptr)`, since I  didn't know you could leave them out !  :-)

melvinw:
Alrighty. Updated to explicitly set the values and removed those asserts

PR #1444, opened 2023-01-07
[translation] unify shell.Main() and shell_native.Main()

melvinw:
This will make it easier to track the difference between the osh and osh-native


core/shell.py
```diff
@@ -122,7 +122,7 @@ def SourceStartupFile(fd_state, rc_path, lang, parse_ctx, cmd_ev, errfmt):
   rc_line_reader = reader.FileLineReader(f, arena)
   rc_c_parser = parse_ctx.MakeOshParser(rc_line_reader)
 
-  with alloc.ctx_Location(arena, source.SourcedFile(rc_path)):
+  with alloc.ctx_Location(arena, source.SourcedFile(rc_path, -1)):
```
andychu, line 125: The -1 could be runtime.NO_SPID

(the location is supposed to be where the `source` statement happens, but for the rcfile there's no `source`.  Might want to polish this later)

andychu (commented):
Wow this is awesome!  Thank you

I took a quick look

melvinw:
Nice! Folded in fixes for your comments while fixing CI failures. I think this is ready for a proper review now :)

mycpp/gc_mylib.h
```diff
@@ -105,6 +107,9 @@ class CFileLineReader : public LineReader {
   virtual int fileno() {
     return ::fileno(f_);
   }
+  void close() {
+    fclose(f_);
+  }
```
melvinw, line 112: should probably add `f_ = nullptr` here...

melvinw:
Interestingly this is has introduced a regression in [builtin-completion](http://travis-ci.oilshell.org/github-jobs/2023-01-07__23-49-34.wwz/_tmp/spec/cpp/builtin-completion.html)...

melvinw:
Ah nevermind! `osh_.cc` has been failing all along. It's just that `osh_.py` passes most of that suite now

core/shell.py
```diff
@@ -528,14 +547,16 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
...
+    assert line_reader is not None
     if flag.location_start_line != -1:
       line_reader.SetLineOffset(flag.location_start_line)
 
   arena.PushSource(src)
 
   # TODO: assert arena.NumSourcePaths() == 1
   # TODO: .rc file needs its own arena.
-  c_parser = parse_ctx.MakeOshParser(line_reader)
+  if line_reader is not None:
```
andychu, line 558: please remove this condition / change to assert as well

i.e. I think this code should have the invariant that `line_reader` is valid, so I want to avoid extra conditionals

 (actually if we turned on strict None checking in  MyPy it would do that)

core/shell.py
```diff
@@ -629,7 +654,8 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
       except util.UserExit as e:
         return e.status
 
-    line_reader.Reset()  # After sourcing startup file, render $PS1
+    if line_reader is not None:
```
andychu, line 657: ditto here

mycpp/gc_mylib.h
andychu: This code will change anyway -- see Zulip's Unix I/O vs. C I/O

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Unix.20I.2FO.20instead.20of.20C.20I.2FO

I have a good plan for that

andychu (commented):
Wow this is amazing, thank you!  Great to see the huge improvements in tests!

Now I'm thinking about unifying `bin/oil.py` and `bin/osh_eval.py`, which shouldn't be hard

There is a lot of experimental cruft in `bin/oil.py` that should simply go away

core/shell.py
melvinw: ah nice catch. forgot about these other checks

andychu:
Oh I just pushed a change that conflicted, let me merge it

PR #1445, opened 2023-01-07
[pyext/posix] add system(3)

melvinw:
`core/main_loop.py` uses this [here](https://github.com/oilshell/oil/blob/master/core/main_loop.py#L102). If we add `system()` to `posix_` we can use it instead of `os.system()`. Then we can add a specialization to `cpp/stdlib` to translate the main loop stuff.

Does this seem reasonable?


andychu:
Ah I actually think it's OK to leave that whole ShowDescriptorState function out of the translation ... it is purely for debugging purposes (and the implementation is a bit hacky)

i.e. I think we can just have an `if mylib.PYTHON` around the whole function

melvinw:
Ah great. Will close this

PR #1447, opened 2023-01-10
WIP [mycpp] translate yields

melvinw:


melvinw:
still a draft. just opening to see how CI feels so far.

it seems to work though! needs some cleanup. roots for the accumulators used by the yield statements is a little awkward and the new state in `Generator` can possibly be simplified

melvinw:
also sitll needs support for `it = returns_iterator()`

mycpp/examples/iterators.py
```diff
@@ -0,0 +1,76 @@
...
+  log('--- nested iterators')
+  for i in f(3):
+    log("f() gave %d", i)
+
+  for i in f(3):
+    for j in f(3):
+      log("f() gave %d, %d", i, j)
+
+  log('--- iterator assignment')
+  it = f(5) # type: Iterator[int]
```
melvinw, line 58: we can probably infer the type here

PR #1448, opened 2023-01-11
[mycpp] emit code for generators

melvinw:
This commit is a first attempt at translating generator functions. If mycpp sees a function that returns `typing.Iterator[T]` it will generate a C++ function with a type signature that differs in the following ways:

  1) it returns `void` instead of `Iterator[T]`
  2) it has an extra ouput param of type `List<T>*`

The defintion of the function emitted by mycpp will be identical to its pythonic ancestor, except all `yield ...` statements will be translated into calls to `append(...)` on the output param from (2).

At the call sites for the function mycpp will allocate a temporary `List<T>` on the stack and pass it as an ouput param to the modified function as an accumulator the generator's output, which will be computed in one big batch. All subsequent statements that refer to the generator will operate, unmodified, on an `ListIter<T>` that wraps the accumulator.

This approach described by @andychu here:
https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Solutions.20to.20yield.20problem/near/316873052

### DIFF
--- a/mycpp/cppgen_pass.py
+++ b/mycpp/cppgen_pass.py
@@ -196,6 +196,12 @@ def get_c_type(t, param=False, local=False):
       c_type = 'void'
       is_pointer = True
 
+    elif type_name == 'typing.Iterator':
+      assert len(t.args) == 1, t.args
+      type_param = t.args[0]
+      inner_c_type = get_c_type(type_param)
+      c_type = 'ListIter<%s>' % inner_c_type
+
     else:
       # note: fullname => 'parse.Lexer'; name => 'Lexer'
       base_class_names = [b.type.fullname for b in t.type.bases]
@@ -268,7 +274,7 @@ def get_c_type(t, param=False, local=False):
   return c_type
 
 
-def get_c_return_type(t) -> Tuple[str, bool]:
+def get_c_return_type(t) -> Tuple[str, bool, Optional[str]]:
   """
   Returns a C string, and whether the tuple-by-value optimization was applied
   """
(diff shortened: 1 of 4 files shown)


mycpp/cppgen_pass.py
```diff
@@ -268,7 +274,7 @@ def get_c_type(t, param=False, local=False):
   return c_type
 
 
-def get_c_return_type(t) -> Tuple[str, bool]:
+def get_c_return_type(t) -> Tuple[str, bool, Optional[str]]:
   """
   Returns a C string, and whether the tuple-by-value optimization was applied
```
andychu, line 279: please update this comment

mycpp/cppgen_pass.py
```diff
@@ -534,14 +547,28 @@ def visit_yield_from_expr(self, o: 'mypy.nodes.YieldFromExpr') -> T:
...
       for i, arg in enumerate(o.args):
         if i != 0:
           self.write(', ')
         self.accept(arg)
+
+      # Will be set if we're:
+      # a) accumulating the output of an iterator
+      # b) constructing an iterator with the result of (a)
+      if self.current_stmt_node in self.yield_accumulators:
+          if len(o.args) > 0:
```
andychu, line 566: Indentation looks a little inconsistent here

I know it's already messed up because MyPy uses 4 and I use 2 ... at some point we should run a formatter over it

mycpp/cppgen_pass.py
```diff
@@ -697,6 +724,12 @@ def visit_call_expr(self, o: 'mypy.nodes.CallExpr') -> T:
 
         callee_name = o.callee.name
 
+        if isinstance(o.callee, MemberExpr) and callee_name == 'next':
+          self.accept(o.callee.expr)
+          self.write('.NextValue')
```
andychu, line 729: nitpick: how about IterNext() ?    to make it clear it's for iterators

andychu (approved):
Thanks, looks great!  The tests seem comprehensive

I didn't follow all the logic but it seems like it's pretty localized

andychu:
Merged, thanks!

PR #1451, opened 2023-01-14
[cpp/pylib] add path_stat::isdir()

melvinw:
We'll need this to translate the completion spec.

### DIFF
--- a/cpp/pylib_test.cc
+++ b/cpp/pylib_test.cc
@@ -26,6 +26,12 @@ TEST os_path_test() {
   PASS();
 }
 
+TEST isdir_test() {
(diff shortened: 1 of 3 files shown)


cpp/pylib_test.cc
```diff
@@ -26,6 +26,12 @@ TEST os_path_test() {
   PASS();
 }
 
+TEST isdir_test() {
+  ASSERT(path_stat::isdir(StrFromC(".")));
+  ASSERT(path_stat::isdir(StrFromC("/")));
```
andychu, line 31: I sometimes add stuff like `!isdir("/nonexistent_ZZZ")` , which is very cheesy, but I think acceptable ...

We do run our tests in containers except on local machines ...

PR #1452, opened 2023-01-14
[cpp/core] add PasswdEntry, GetAllUsers()

melvinw:
We'll use this in place of python's pwd.getpwall() in core/completion.py


andychu:
Cool thanks!   (What a weird API in C! )

PR #1453, opened 2023-01-14
[cpp/stdlib] add stat() and listdir() bindings

melvinw:


cpp/stdlib.cc
```diff
@@ -139,10 +240,10 @@ time_t time() {
...
 // NOTE(Jesse): time_t is specified to be an arithmetic type by C++. On most
 // systems it's a 64-bit integer.  64 bits is used because 32 will overflow in
-// 2038.  Someone on a comittee somewhere thought of that when moving to 64-bit
-// architectures to prevent breaking ABI again; on 32-bit systems it's usually
-// 32 bits.  Point being, using anything but the time_t typedef here could
-// (unlikely, but possible) produce weird behavior.
+// 2038.  Someone on a comittee somewhere thought of that when moving to
+// 64-bit architectures to prevent breaking ABI again; on 32-bit systems it's
+// usually 32 bits.  Point being, using anything but the time_t typedef here
+// could (unlikely, but possible) produce weird behavior.
```
melvinw, line 246: Huh. I've noticed `clang-format` can be a little non-deterministic at times... (either that or this wasn't formatted before...)

cpp/stdlib.h
```diff
@@ -118,6 +120,44 @@ void execve(Str* argv0, List<Str*>* argv, Dict<Str*, Str*>* environ);
...
+
+// sys/stat.h defines these macros for backwards compatability reasons.  We
+// undefine them here to avoid collisions with stat_result's members. We
+// shouldn't have any code that depends on these macros, so this should be
+// safe...
+#ifdef st_atime
+  #undef st_atime
+  #undef st_mtime
+  #undef st_ctime
+#endif
```
melvinw, line 133: this bit is awkward. open to alternatives

cpp/stdlib.h
```diff
@@ -118,6 +120,44 @@ void execve(Str* argv0, List<Str*>* argv, Dict<Str*, Str*>* environ);
...
+  int st_mode;
+  int st_ino;
+  int st_dev;
+  int st_nlink;
+  int st_uid;
+  int st_gid;
+  int st_size;
+  int st_atime;
+  int st_mtime;
+  int st_ctime;
```
melvinw, line 156: these _should_ be  `unsigned long`, but i'm not exactly sure how to honor that without being inconsistent with the `stat_results` type defined in `pyext/posix_.pyi`... any ideas?

cpp/stdlib.cc
```diff
@@ -120,6 +142,85 @@ void kill(int pid, int sig) {
...
+  n_fields = 0;
+  n_sequence_fields = 0;
+  n_unnamed_fields = 0;
+
+  st_mode = st->st_mode;
+  st_ino = st->st_ino;
+  st_dev = st->st_dev;
+  st_nlink = st->st_nlink;
+
+#if defined(MS_WINDOWS)
```
melvinw, line 186: maybe not necessary...?

andychu:
Hmmm do we actually use all those fields in stat ?  I don't see them?

I would think we could use standard interfaces with no `./configure`, but I don't remember all the code

I think using the "pyos" style of stub could save some code here?  

That is how we do `bool_stat.py`  -- we do not translate it 1-for-1


```
$ egrep --color '\.st_.*' */*.py
core/completion.py:      key = (d, st.st_mtime)
osh/bool_stat.py:      mode = posix.lstat(s).st_mode
osh/bool_stat.py:  mode = st.st_mode
osh/bool_stat.py:    return st.st_size != 0
osh/bool_stat.py:    return st.st_uid == posix.geteuid()
osh/bool_stat.py:    return st.st_gid == posix.getegid()
osh/bool_stat.py:    m1 = 0 if st1 is None else st1.st_mtime
osh/bool_stat.py:    m2 = 0 if st2 is None else st2.st_mtime
osh/bool_stat.py:    return st1.st_dev == st2.st_dev and st1.st_ino == st2.st_ino
pylib/path_stat.py:    return stat.S_ISDIR(st.st_mode)

```




```
$ egrep 'stat\.' */*.py

core/completion.py:          if path_stat.isdir(path):
core/completion.py:        if self.add_slash and path_stat.isdir(path):
core/completion.py:        if path_stat.isdir(candidate):  # TODO: test coverage
core/state.py:      if path_stat.exists(name):
core/state.py:        found = path_stat.exists(full_path)  # for 'source'
osh/bool_stat.py:bool_stat.py
osh/bool_stat.py:    return stat.S_ISLNK(mode)
osh/bool_stat.py:    return stat.S_ISREG(mode)
osh/bool_stat.py:    return stat.S_ISDIR(mode)
osh/bool_stat.py:    return stat.S_ISBLK(mode)
osh/bool_stat.py:    return stat.S_ISCHR(mode)
osh/bool_stat.py:    return bool(stat.S_IMODE(mode) & stat.S_ISVTX)
osh/bool_stat.py:    return stat.S_ISFIFO(mode)
osh/bool_stat.py:    return stat.S_ISSOCK(mode)
osh/bool_stat.py:    return bool(stat.S_IMODE(mode) & stat.S_ISUID)
osh/bool_stat.py:    return bool(stat.S_IMODE(mode) & stat.S_ISGID)
osh/sh_expr_eval.py:          return bool_stat.DoUnaryOp(op_id, s)
osh/sh_expr_eval.py:            return bool_stat.isatty(s, node.child)
osh/sh_expr_eval.py:          return bool_stat.DoBinaryOp(op_id, s1, s2)
pylib/path_stat.py:path_stat.py - Functions from os.path that import 'stat'.
pylib/path_stat.py:    return stat.S_ISDIR(st.st_mode)
pylib/path_stat_test.py:path_stat_test.py: Tests for path_stat.py
pylib/path_stat_test.py:    self.assertEqual(True, path_stat.exists('/'))
pylib/path_stat_test.py:    self.assertEqual(False, path_stat.exists('/nonexistent__ZZZZ'))
pylib/path_stat_test.py:    self.assertEqual(True, path_stat.exists('/'))
pylib/path_stat_test.py:    self.assertEqual(False, path_stat.exists('/nonexistent__ZZZZ'))

```

melvinw:
> Hmmm do we actually use all those fields in stat ? I don't see them?

Completion only uses `st_mtime` from what I can tell and the `st_*time` fields are the ones that requires the `./configure` changes... The others are more straightforward.

The code that uses this is in one of the `Matches(...)` [generators](https://github.com/oilshell/oil/blob/master/core/completion.py#L576-L582) and I don't see a non-awkward way of factoring it out into `pyos` (I guess we could do `GetModifiedTime(path)`, but we'd end up writing most of `stat()` anyway...)

cpp/stdlib.cc
```diff
@@ -16,6 +16,28 @@
 #include "cpp/core_pyerror.h"  // e_die for strftime
 #include "mycpp/runtime.h"
 
+#ifdef HAVE_DIRENT_H
```
andychu, line 19: Hmm it seems like this block is from CPython?

I feel like it should be possible to just use POSIX, I would surprised if we can't list a directory with just POSIX, and ZERO `#ifdef`

Generally I want to minimize `#ifdef` as much as possible, since it explodes the testing matrix

cpp/stdlib.cc
```diff
@@ -120,6 +142,85 @@ void kill(int pid, int sig) {
...
+  st_uid = st->st_uid;
+  st_gid = st->st_gid;
+#else
+  st_uid = 0;
+  st_gid = 0;
+#endif
+
+  st_size = st->st_size;
+
+#if defined(HAVE_STAT_TV_NSEC)
```
andychu, line 196: Ditto here, I don't know this interface, but I think we should be able to do it without `#ifdef`

All we are doing is testing the mtime of a directory?  That should be doable in portable POSIX?  The resolution doesn't really matter AFAICT


I guess if this unblocks us, it could be worth putting in, but I would want to make another pass over it

andychu (commented):
Let me look into what the standard APIs are ...

I don't think we should need anything in `./configure` to port `core/completion.py`

Conceptually what it's doing is simple -- it's listing a directory, and then caching it based on the directory's mtime

So I think that should be doable in a portable way.  

We are trying to be portable like Lua (standards-based), not CPython (lots of `#ifdef`)  .  That was on the blog at some point, but maybe I should write it down in the `doc/` dir or something

andychu:
OK I looked it all up... It's always a crapshoot to find the right POSIX pages through Google, and also the POSIX PDF -> HTML conversions are wonky

However I ended up here, and then I search for readdir() and stat()

https://pubs.opengroup.org/onlinepubs/9699919799/


https://pubs.opengroup.org/onlinepubs/9699919799/functions/readdir.html

https://pubs.opengroup.org/onlinepubs/9699919799/functions/stat.html


So  from what I see, this matches Linux exactly, like

- `man stat`
- `man readdir`

on my machine.  So I think we should get rid of all  configure tests, and just make it work on Linux, and that will be POSIX.

For example, the filename is called `d_name`, so we shouldn't need any macros around it.

---

 CPython was created in  1990, and it also works on many non-Unix platforms, so it has a lot of "cruft" .  I remember that they were talking about phasing out DOS and OS/2 support and all that, because people generally don't test those platforms anymore !

(The GNU readline might be a little different, but we can research that when we get to it ... That is a single implementation, and not a standard)

andychu:
Also the POSIX page seems to have pretty good code examples!   It seems like would just work, although our error handling can be borrowed from CPython and so forth (OSError and so forth)

melvinw:
Ah great! I wasn't sure how much we cared about matching cpython compatibility-wise. I was actually pretty surprised by all the different places dirent could come from and all the different names for the time fields...

Will push a commit to simplify in a bit!

melvinw:
Ok, all set. Much simpler now. Pushed a couple changes to the tests too

cpp/stdlib.h
andychu: Oh sorry I didn't respond to this comment earlier.

Yeah the integer types are one case where our translation is a little weird.

I guess these can overflow pretty easily?  We probably don't want that bug

Actually it seems like Python gives a float!

We are hashing a float in the Python code, which is bad!  doh.  Let me think about this


```
>>> import os
>>> st=os.stat("README.md")
>>> st
posix.stat_result(st_mode=33204, st_ino=6301016, st_dev=66307, st_nlink=1, st_uid=1000, st_gid=1000, st_size=10657, st_atime=1673379783, st_mtime=1665704712, st_ctime=1665704712)
>>> st.st_mtime
1665704712.5696013
```

cpp/stdlib.cc
```diff
@@ -120,6 +121,46 @@ void kill(int pid, int sig) {
...
+
+  auto* ret = NewList<Str*>();
+  while (true) {
+    errno = 0;
+    struct dirent* ep = readdir(dirp);
+    if (ep == NULL) {
+      if (errno == 0) {
+        break;
+      } else {
+        closedir(dirp);
```
andychu, line 138: Nitpick: let's use the "early return" error style here, because we do above on line 126 (and throughout the codebase)

andychu (commented):
Thanks, looks great!

I just noticed we are hashing a float in Python, which is kind of bad.

And there is a fundamental problem that `int`  is not big enough for a precise Unix timestamp


It kind of argues for something like `pyos.GetDirCacheKey() -> str`, which could return the number of nanoseconds as an opaque **byte string** ?

We could also round the result to an `int` (seconds instead of nanoseconds), but not sure if that's a good idea

Open to other suggestions

-----

Python does have a standalone function

```
>>> import os
>>> os.path.getmtime('README.md')
1665704712.5696013
```

----

We could keep the stat() binding, because that will probably be useful for end users at some point (similar to rstrip/lstrip), or for writing a  `stat` builtin in Oil (some people  want more externalb binaries to be folded into Oil)


But it seems like there is still a correctness issue here, both in the  original code (hashing floats), and in the translation

andychu:
I guess one way to think of this problem is that mycpp is a Java-like language with only `int` types -- no unsigned, and no control over the size of the int type

So that makes it awkward to wrap certain Unix and network APIs.  They can't be wrapped "1 for 1" in general.

So whatever people do in Java is probably what we should do here.  I think since it's just a cache key, making it an opaque string is a decent idea

melvinw:
Meh yeah this int width stuff is a bummer.. A specialized helper like `pyos.GetDirCacheKey()`with opaque strings sounds like the best way to keep the completion translation moving for now. Will try it out tonight.

>We could keep the stat() binding, because that will probably be useful for end users at some point (similar to rstrip/lstrip), or for writing a stat builtin in Oil (some people want more externalb binaries to be folded into Oil)

I'm okay with dropping this branch until we have a better answer to the int width issue (we need to figure this out for the arithmetic spec test anyway, right?). I'm worried it will be confusing to have an implementation of a standard API with mismatched types. Maybe that's not a big deal, though?

andychu:
Yeah I'm researching this a bit ... I'm wondering if we can rely on some lowest common denominator behavior

Probably the mtime with a precision of seconds is fine ?

https://docs.python.org/2/library/os.html#module-os

> The exact meaning and resolution of the st_atime, st_mtime, and st_ctime attributes depend on the operating system and the file system. For example, on Windows systems using the FAT or FAT32 file systems, st_mtime has 2-second resolution, and st_atime has only 1-day resolution. See your operating system documentation for details.


----

Python 3 added an exact integer nanoseconds field, instead of float.

Though not all file systems would support it



https://docs.python.org/3/library/os.html

> Similarly, although [st_atime_ns](https://docs.python.org/3/library/os.html#os.stat_result.st_atime_ns), [st_mtime_ns](https://docs.python.org/3/library/os.html#os.stat_result.st_mtime_ns), and [st_ctime_ns](https://docs.python.org/3/library/os.html#os.stat_result.st_ctime_ns) are always expressed in nanoseconds, many systems do not provide nanosecond precision. On systems that do provide nanosecond precision, the floating-point object used to store [st_atime](https://docs.python.org/3/library/os.html#os.stat_result.st_atime), [st_mtime](https://docs.python.org/3/library/os.html#os.stat_result.st_mtime), and [st_ctime](https://docs.python.org/3/library/os.html#os.stat_result.st_ctime) cannot preserve all of it, and as such will be slightly inexact. If you need the exact timestamps you should always use [st_atime_ns](https://docs.python.org/3/library/os.html#os.stat_result.st_atime_ns), [st_mtime_ns](https://docs.python.org/3/library/os.html#os.stat_result.st_mtime_ns), and [st_ctime_ns](https://docs.python.org/3/library/os.html#os.stat_result.st_ctime_ns).

andychu:
OK I wrote a little demo of this API

https://github.com/oilshell/oil/commit/60cf09a60de1a91454fd7990254a54a055baa14c

It looks to me like the POSIX API gives you time in seconds with `st_mtime`, and then Linux has `st_mtim.tv_sec` and `st_mtim.tv_nsec` which can be more precise

CPython converts the nanoseconds to a float for you


```
$ ninja _bin/cxx-asan/cpp/stdlib_test  && $_ -t mtime
ninja: no work to do.
mtime        = 1665704712
mtim.tv_sec  = 1665704712
mtim.tv_nsec =  569601209
2022-10-13
INT_MAX      = 2147483647
diff         =  481778935
2038-01-18
.
Total: 1 test (516 ticks, 0.001 sec), 0 assertions
Pass: 1, fail: 0, skip: 0.

```

----

So one thing to do would be make pyos.GetDirCacheKey():

- in Python, it can just round the float mtime to an integer.  We could also write a proper C extension in `pyext`, although that is kind of busy-work
- in C++, it will just use the portable POSIX `st_mtime` field

But then I also tested that we have 15 years left before the year 2038 problem LOL

I guess it may not make the code incorrect if it overflows, because it's used as a cache value.  Could add a comment to that effect


---

(Honestly I wonder if this cache is even very important ... these kinds of operations should be pretty fast, but I think this translation  should be doable)

cpp/core_test.cc
```diff
@@ -262,6 +263,25 @@ TEST passwd_test() {
   PASS();
 }
 
+TEST dir_cache_key_test() {
+  struct stat st;
+  ASSERT(::stat("/", &st) == 0);
+
+  Tuple2<Str*, int> key = pyos::MakeDirCacheKey(StrFromC("/"));
+  ASSERT(str_equals(key.at0(), StrFromC("/")));
+  ASSERT(key.at1() == st.st_mtime);
```
melvinw, line 272: this feels a tad literal...

melvinw:
Huh. The `app_test` and `ovm-tarball` failures are kinda bizarre (lots of seemingly unrelated warnings in the latter)
```
Could not import bin.oil
  File "/home/uke/oil/bin/oil.py", line 37, in <module>
    from core import main_loop
  File "/home/uke/oil/core/main_loop.py", line 19, in <module>
    from core import process
  File "/home/uke/oil/core/process.py", line 28, in <module>
    from core import dev
  File "/home/uke/oil/core/dev.py", line 17, in <module>
    from core import state
  File "/home/uke/oil/core/state.py", line 25, in <module>
    from core import pyos
  File "/home/uke/oil/core/pyos.py", line 9, in <module>
    import os
  File "/home/uke/tmp/Python-2.7.13/Lib/os.py", line 117, in <module>
    raise ImportError, 'no os specific module found'
ImportError: no os specific module found
FATAL: couldn't import from app bundle '../../_bin/osh' (1)
```

melvinw:
ah DOH. I used `os` instead of `posix` in pyos for `stat()` -.-

core/completion.py
```diff
@@ -575,11 +576,10 @@ def Matches(self, comp):
     executables = []  # type: List[str]
     for d in path_dirs:
       try:
-        st = posix.stat(d)
+        key = pyos.MakeDirCacheKey(d)
```
andychu, line 579: I also wonder if Dict[Tuple2] will cause problems at runtime ... The issue is whether Tuple2 equality is implemented.

I think there was a bug about that in the past , but I guess we'll see

The problem is that we have some ad hoc C++ specializations, I think in `mycpp/comparators.h`

Yeah we might need `keys_equal()` for `Tuple2[str, int]` there?    I haven't looked at the code in awhile

andychu (commented):
Great, thanks for doing this!  I made one note about a possible issue that will come up at runtime ...

I think by default `Dict` compares if pointers are equal, but we want to know if the values are equal, which requires specialized functions

core/completion.py
melvinw: Oh good question. Wasn't sure what the state of affairs was here. I did do a manual check to be sure `Dict<Tuple<..>, ...>` compiled at least and things looked OK. I'll take a look at the comparators later today to be sure

PR #1462, opened 2023-01-24
[mycpp] add data structure builtins for completion

melvinw:
There are a few builtins and data structure methods used throughout the completion code that aren't yet supported in the runtime. This series of commits adds them.

### DIFF
--- a/mycpp/gc_builtins.cc
+++ b/mycpp/gc_builtins.cc
@@ -291,6 +291,18 @@ bool are_equal(Tuple2<Str*, int>* t1, Tuple2<Str*, int>* t2) {
   return result;
 }
 
+bool are_equal(Tuple2<int, int>* t1, Tuple2<int, int>* t2) {
+  return t1->at0() == t2->at0() && t1->at1() == t2->at1();
+}
+
+bool keys_equal(Tuple2<int, int>* t1, Tuple2<int, int>* t2) {
+  return are_equal(t1, t2);
+}
+
+bool keys_equal(Tuple2<Str*, int>* t1, Tuple2<Str*, int>* t2) {
+  return are_equal(t1, t2);
+}
+
 bool str_equals0(const char* c_string, Str* s) {
   int n = strlen(c_string);
   if (len(s) == n) {
--- a/mycpp/gc_builtins.h
+++ b/mycpp/gc_builtins.h
@@ -100,6 +100,13 @@ class OSError : public IOError_OSError {
   }
 };
 
+class SystemExit : public _ExceptionOpaque {
+ public:
+  explicit SystemExit(int code) : _ExceptionOpaque(), code(code) {
+  }
+  int code;
+};
+
 void print(Str* s);
 
 Str* repr(Str* s);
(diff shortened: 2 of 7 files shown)


melvinw:
Slightly easier to review commit-by-commit

mycpp/gc_list.h
```diff
@@ -295,17 +297,15 @@ T List<T>::pop() {
...
   DCHECK(len_ > 0);
-  DCHECK(i == 0);  // only support popping the first item
 
-  T result = index_(0);
+  T result = index_(i);
   len_--;
 
   // Shift everything by one
-  memmove(slab_->items_, slab_->items_ + 1, len_ * sizeof(T));
+  memmove(slab_->items_ + i, slab_->items_ + (i + 1), len_ * sizeof(T));
```
melvinw, line 308: I _think_ this should be fine in general... Perhaps we could add a few extra tests to convince ourselves if unsure

mycpp/gc_builtins.cc
```diff
@@ -291,6 +291,18 @@ bool are_equal(Tuple2<Str*, int>* t1, Tuple2<Str*, int>* t2) {
   return result;
 }
 
+bool are_equal(Tuple2<int, int>* t1, Tuple2<int, int>* t2) {
+  return t1->at0() == t2->at0() && t1->at1() == t2->at1();
+}
+
+bool keys_equal(Tuple2<int, int>* t1, Tuple2<int, int>* t2) {
```
andychu, line 298: I feel like there is some template refactoring to simplify the keys_equal / are_equal stuff, but let's worry about that separately ...

mycpp/gc_builtins.h
```diff
@@ -100,6 +100,13 @@ class OSError : public IOError_OSError {
   }
 };
 
+class SystemExit : public _ExceptionOpaque {
```
andychu, line 103: I think this is probably unused given the comment in `core/completion.py`, but it's probably hard to `if mylib.PYTHON` out ...

mycpp/gc_list_test.cc
```diff
@@ -408,6 +408,41 @@ TEST contains_test() {
...
+
+  bool caught = false;
+  try {
+    l->index(3);
+  } catch (ValueError* e) {
+    caught = true;
+  }
+  ASSERT(caught);
+
+  ASSERT_EQ(l->index_(0), 1);
```
andychu, line 440: Generally the style for these types of frameworks is to put the expected value FIRST

If you do `ASSERT_EQ_FMT()` then you'll see `greatest.h` does respect that convention

i.e. it prints a message "Expected: arg1, Actual: arg2"

For these cases it's not a big deal

andychu (commented):
Great thanks!  Tests look pretty good

mycpp/gc_list_test.cc
melvinw: Ah whoops. I'm used to gtest which encourages the opposite order (At least in recent years. I think it used to be the other way around). Will send a followup to make these consistent with the rest

PR #1463, opened 2023-01-24
[frontend] add const code gen and move flag setup for completion

melvinw:
These commits make sure we have access to some constants needed for completion from C++.

There's also a commit to move the flag setup for `osh/builtin_comp` into `frontend/flag_gen`. This will make it easier to translate the completion code. Looks like it was left out in the previous migration. 

Probably easiest to review this one commit-by-commit.


andychu:
Cool thanks!

PR #1466, opened 2023-01-26
[translation] make iterator state in ReadlineCallback translatable

melvinw:
This is the last non-mechanical thing we need to figure out before completion can be made translatable.

### DIFF
--- a/core/completion.py
+++ b/core/completion.py
@@ -50,6 +50,7 @@ from core import ui
 from core import util
 from frontend import consts
 from frontend import reader
+from mycpp import mylib
 from pylib import os_path
 from pylib import path_stat
 from osh import word_
@@ -1194,7 +1195,11 @@ class ReadlineCallback(object):
     self.root_comp = root_comp
     self.debug_f = debug_f
 
-    self.comp_iter = None  # current completion being processed
+    # current completion being processed
+    if mylib.PYTHON:
+      self.comp_iter = None # type: Iterator[str]
+    else:
+      self.comp_iter = None # type: List[str]
 
   def _GetNextCompletion(self, state):
     # type: (int) -> Optional[str]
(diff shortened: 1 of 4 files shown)


mycpp/gc_list.h
```diff
@@ -434,11 +434,20 @@ class ListIter {
...
+
  private:
   List<T>* L_;
   int i_;
 };
 
+// list(it) returns the iterator's backing list
+template <typename T>
+List<T>* list(ListIter<T> it) {
+  return list(it.list());
```
melvinw, line 448: potential optimization: This could just be `return it.list()`, but the backing lists for the iterators used for generators live on the stack. If we change `mycpp` to allocate one we could avoid this copy, but it's probably not worth the hassle... Thoughts?

core/completion.py
```diff
@@ -1212,14 +1217,23 @@ def _GetNextCompletion(self, state):
...
-    except StopIteration:
-      next_completion = None  # signals the end
+    if mylib.PYTHON:
+      try:
+        next_completion = self.comp_iter.next()
+      except StopIteration:
+        next_completion = None  # signals the end
+    else:
+      try:
+        next_completion = self.comp_iter.pop()
```
andychu, line 1234: I feel like this will have the reverse order of next()?   It pops from the end

But I guess we'll find out through tests

mycpp/gc_list.h
andychu: Hm isn't it just the pointer that lives on the stack?    Does anything break if you remove `list()`?

Although remember we have to be careful about rooting when we call back into the interpreter.  RootCompleter eventually does that

----

Nit: maybe call it `GetList()` or something, since `list()` is hard to search for

andychu (commented):
Cool -- If this makes the whole thing translate, it seems like not too much cost!

core/completion.py
melvinw: oh good question. i'll do a quick test to check and update this if needed

mycpp/gc_list.h
melvinw: > Hm isn't it just the pointer that lives on the stack? Does anything break if you remove list()?

Yeah, in most cases it's just the pointer that's on the stack, but for some reason (buried somewhere in zulip I think...) we decided to put the lists for generators on the stack in #1448

> Nit: maybe call it GetList() or something, since list() is hard to search for

👍

core/completion.py
melvinw: you were right! this should be `pop(0)`

andychu: I think it's better to do `L.reverse()` and then `pop()`  

Because pop(0) does the memcpy() every time I think, which would give a quadratic iteration

And then reverse() might allow us to copy it from the stack to a GC list ??

mycpp/gc_list.h
andychu: Ah right :)  Yeah I suggested that to save GC pressure

But I think the reverse() can put it on the heap?

melvinw: `List::reverse()` looks like it operates on the list in place. so, we'll still need `list()`

melvinw:
squashed all those fixups. ready for another round

andychu:
Thanks!

PR #1470, opened 2023-01-27
[translation] fix type checker errors in completion code

melvinw:


melvinw:
this depends on #1466. base should automatically change to `soil-staging` or `master` once that is merged

core/completion.py
```diff
@@ -548,14 +553,14 @@ def __init__(self, mem):
     # (dir, timestamp) -> list of entries perhaps?  And then every time you hit
     # tab, do you have to check the timestamp?  It should be cached by the
     # kernel, so yes.
-    self.ext = []
+    # XXX(unused?) self.ext = []
```
melvinw, line 556: Ah right, forgot about this. `git grep "\.ext\."` suggests there aren't any users of this member. Should we drop it? If not, I'm not sure what the type should be... The comment above suggests it might be a list of some sort of dirent-like thing

build/dynamic-deps.sh
```diff
@@ -65,6 +63,7 @@ _gen/
 .*_def\.py
 .*_spec\.py
 asdl/py.*
+core/completion.py
```
melvinw, line 66: still some work left to get things translating. splitting out the type checking changes to make reviewing easier

core/completion.py
andychu: I don't remember what this is now, but yes let's just leave it out of this PR

osh/builtin_comp.py
```diff
@@ -26,15 +27,17 @@
...
 
+
+HELP_TOPICS = [] # type: List[str]
+
 from mycpp import mylib
 if mylib.PYTHON:
   # Hack because we don't want libcmark.so dependency for build/dev.sh minimal
   try:
     from _devbuild.gen import help_
+    HELP_TOPICS = help_.TOPICS
```
andychu, line 38: BTW I want to rewrite the help system so it doesn't depend on code gen

melvinw, line 38: Ah, this is gonna require some more work. CI doesn't produce `_devbuild.gen.help_`. I forgot I manually generated this in my local tree when I first wrote this...

andychu: Ah OK let me know if you want me to look at the `help` stuff tomorrow ... I can probably remember how it works :)

melvinw: That would be great!

andychu:
Ahh this is a CI problem, the PR is good! 

The problem is that we have `build/py.sh minimal` to reduce dependencies over `build/py.sh all`

The theory behind that was that people could clone the Oil repo and get a working `bin/osh` in like 30 seconds.  They don't need to install `re2c` or CommonMark

But in practice I'm not sure if anyone is really taking advantage of that.  We could probably get rid of `minimal` to make our lives easier

The other thing I mentioned is I want to remove CommonMark as a dependency for the `help` builtin.  I think the real docs should be HTML , and `help echo` should just print one line, and then link to real HTML docs.  (The problem I was trying to solve is that bash duplicates info in `man bash` / `info bash`, AND `help`)

PR #1471, opened 2023-01-28
[translation] make completion and related builtins translateable

melvinw:


andychu:
```
--------------------------------------------------------------------------------
osh/builtin_comp.py:40: error: unused 'type: ignore' comment
--------------------------------------------------------------------------------

mv: cannot stat '_build/tmp/oils_cpp_raw.cc.tmp': No such file or directory

```

Hm this is weird, I'm not getting it on my machine, or on other CI tasks ... maybe that CI container is somehow out of whack?


http://travis-ci.oilshell.org/github-jobs/2023-01-28__21-55-07.wwz/_tmp/soil/logs/osh-parser.txt

andychu:
Oh nevermind, I'm getting it now ... I had to rebuild somehow (?)

Let me see what's causing that

andychu:
Ah OK cool I got it building locally, by changing the MyPy config in the same way as `devtools/types.sh` (which I just changed because of a `type: ignore` we need)

core/completion.py
```diff
@@ -293,6 +313,16 @@ def Matches(self, comp):
...
 
+  # mycpp: for rewrites of isinstance()
+  def IsFileSystemAction(self):
+    # type: () -> bool
+    return False
+
+  # mycpp: for rewrites of isinstance()
+  def IsShellFuncAction(self):
+    # type: () -> bool
+    return False
```
melvinw, line 324: these are a little awkward. open to ideas

melvinw:
Ok this is ready to go. Diff is large, but the changes are all pretty mechanical:
- getting rid of kwargs
- replacing `log(msg, ...)` with `log(msg % ...)`
- massaging multiple-assignments/returns
- replacing closures with regular functions
- replacing `try/except/else` blocks

melvinw:
We've got a non-zero number of tests passing `builtin-completion` for `oils-cpp` now :) The next PR will finish the readline bindings and we should pass the whole suite

osh/builtin_comp.py
```diff
@@ -129,22 +143,22 @@ def Build(self, argv, attrs, base_opts):
...
-        a = completion.FileSystemAction(dirs_only=True)
+        a = completion.FileSystemAction(True, False, False)
 
       elif name == 'file':
-        a = completion.FileSystemAction()
+        a = completion.FileSystemAction(False, False, False)
 
       elif name == 'function':
-        a = _DynamicDictAction(cmd_ev.procs)
+        a = _FixedWordsAction(cmd_ev.procs.keys())
```
andychu, line 161: Oh I think you can remove `.keys()` here --  this is going to have the bug where when `procs`  is updated with new procs, they won't be completed

osh/builtin_comp.py
```diff
@@ -335,7 +358,7 @@ def __init__(self, comp_state, errfmt):
   def Run(self, cmd_val):
     # type: (cmd_value__Argv) -> int
     argv = cmd_val.argv[1:]
-    arg_r = args.Reader(argv)
+    arg_r = args.Reader(argv, [])
```
andychu, line 361: I think the second arg here can be `cmd_val.arg_spids[1:]`

The idea there is to be able to "blame" arguments in errors.  The "spids" are location info that allow that (though they're also the thing I'm in the middle of refactoring!)

osh/builtin_comp.py
```diff
@@ -364,7 +387,7 @@ def __init__(self, mem):
...
```
andychu, line 390: can also be `cmd_val.arg_spids[1:]`

osh/builtin_comp.py
```diff
@@ -387,7 +410,11 @@ def Run(self, cmd_val):
     if arg.s:  # implied
       break_chars.remove('=')
     # NOTE: The syntax is -n := and not -n : -n =.
-    omit_chars = arg.n or ''
+    # mycpp: rewrite of or
+    omit_chars = arg.n
+    if omit_chars is None:
+        omit_chars = ''
```
andychu, line 416: indent 2 rather than 4

osh/builtin_comp.py
```diff
@@ -406,7 +433,10 @@ def Run(self, cmd_val):
 
     if arg.s:
       if cur.startswith('--') and '=' in cur:  # Split into flag name and value
-        prev, cur = cur.split('=', 1)
+        # mycpp: rewrite of multiple-assignment
+        scur = cur.split('=', 1)
```
andychu, line 437: please use mylib.split_once(cur, '=') here

This will avoid a bit of allocation and is shorter :)

andychu (commented):
This looks great, thank you!

osh/builtin_comp.py
melvinw: Ah nice catch! This was leftover from before we talked about the difference between this and FixedWordsAction. Will fix

melvinw:
pushed some fixes

PR #1475, opened 2023-01-30
Fix python2 package dependency for voguish ubuntu

AALEKH:
Since `python-dev` is no longer available in Ubuntu 20 and beyond, I added this additional check to install correct python package. Without this change, upon executing: `build/py.sh ubuntu-deps` it throws `E: Package 'python-dev' has no installation candidate` error. Having conditional check for legacy ubuntu release ensures existing systems do not break.

### DIFF
--- a/build/py.sh
+++ b/build/py.sh
@@ -25,7 +25,6 @@ export PYTHONPATH='.:vendor/'
 
 ubuntu-deps() {
   # python-dev: for all the extension modules
-  #   TODO: upgrade Ubuntu and change to python2-dev
   # gawk: used by spec-runner.sh for the special match() function.
   # time: used to collect the exit code and timing of a test
   # libreadline-dev: needed for the build/prepare.sh Python build.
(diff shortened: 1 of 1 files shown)


build/py.sh
```diff
@@ -25,16 +25,24 @@ export PYTHONPATH='.:vendor/'
...
   # time: used to collect the exit code and timing of a test
   # libreadline-dev: needed for the build/prepare.sh Python build.
   # cmake: for build/py.sh yajl-release
   set -x  # show what needs sudo
 
   # pass -y for say gitpod
+
+  ubuntu_release_year=$(lsb_release -rs | cut -d. -f1)
+
+  if $(expr $ubuntu_release_year > 20); then
```
andychu, line 38: Please try 

```
if apt-get show python2-dev >/dev/null; then
 ...

```


So this will   prefer the python2-dev package if it exists


See this page for background

https://github.com/oilshell/oil/wiki/Feature-Detection-Is-Better-than-Version-Detection


(Also note the `expr` usage and `if` usage aren't correct here  -- you're doing a redirect with `>` , not greater than, and if tests the exit code and not the arg value)



```
$ apt-cache show python2-dev; echo $?
N: Unable to locate package python2-dev
E: No packages found
100
andy@lenny:~/git/oilshell/oil$ apt-cache show python-dev; echo $?
Package: python-dev
Architecture: amd64
```

andychu (commented):
Thanks for the patch!

This is bug https://github.com/oilshell/oil/issues/1172

so that's helpful

andychu:
Does the suggestion make sense?  I'd like to merge this, thanks!

andychu:
I fixed this here, thank you for the PR

https://github.com/oilshell/oil/commit/0319b8416a688430729212bd897f41125a1d08c0

PR #1477, opened 2023-02-03
[osh] replace print in compgen builtin

melvinw:
For some reason mycpp removes calls to `print` in this file.

This fix reduces the `spec/builtin-completion` delta to 3 (with c++ passing some tests that python doesn't...)!


andychu:
Ah sorry there was a bug in mycpp, I think this will fix it, and prevent the problem in other files:

https://github.com/oilshell/oil/commit/8461a42c83e8a006f79bba0a7727847d0f39dc1d

melvinw:
Ah great! Nice catch

PR #1479, opened 2023-02-03
[core] refactor main loop to avoid ambiguous break

melvinw:
Another case of `break`ing inside of a type switch inside of a loop causing chaos


melvinw:
oh boo this seems to have broken `spec/stateful`

andychu:
Thanks!  I think mycpp could detect this... will file a bug

PR #1480, opened 2023-02-03
[mycpp] a few small changes for C++ readline

melvinw:
A few things to match python's behavior for some builtins. The negative width format string one is probably the most interesting change

### DIFF
--- a/mycpp/gc_builtins.cc
+++ b/mycpp/gc_builtins.cc
@@ -1,6 +1,12 @@
 #include <ctype.h>  // isspace()
 #include <errno.h>  // errno
 
+#include "_build/detected-cpp-config.h"
+
+#ifdef HAVE_READLINE
+  #include <readline/readline.h>
+#endif
+
 #include "mycpp/runtime.h"
 
 // Translation of Python's print().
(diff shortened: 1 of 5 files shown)


mycpp/gc_builtins.cc
```diff
@@ -343,3 +353,15 @@ int max(List<int>* elems) {
...
+
+Str* raw_input(Str* prompt) {
+#ifdef HAVE_READLINE
+  char* ret = readline(prompt->data());
+  if (ret == nullptr) {
+    throw Alloc<EOFError>();
+  }
+  return StrFromC(ret, strlen(ret));
+#else
+  assert(0);  // not implemented
```
andychu, line 365: I think we can fill this in pretty easily to be something like `LineReader.readline()`, but we can do it in another PR

Maybe we need a global `CFileLineReader(mylib.Stdin())`  not sure?

Or there is also `ReadLine()` for the `read` builtin

It's complex because of buffering -- shells do some weird things there

mycpp/gc_builtins.cc
```diff
@@ -243,6 +249,10 @@ bool str_equals(Str* left, Str* right) {
     return true;
   }
 
+  if (left == nullptr || right == nullptr) {
```
andychu, line 252: OK I guess the problem here is that we don't always call `maybe_str_equals()`

This is related to the null safety issue

If we're not Null safe, then arguably we don't need both methods.  I'll record that as  a code smell

andychu:
Thanks!

PR #1481, opened 2023-02-03
[frontend] unify py_reader and reader

melvinw:
`InteractiveLineReader` used to have a bunch of `Any`s and dependencies on types that couldn't be translated to C++, but those have all been addressed recently. So, we can move `InteractiveLineReader` into `frontend/reader.py` and drop `frontend/py_reader.py` all-together now.

There are also some small type-signature tweaks to the readline pseudo-module to improve clarity.


melvinw:
Ah this one depends on #1480 for `raw_input()` so the `cpp-*` CI steps are failing

PR #1483, opened 2023-02-03
[translation] implement c++ readline stubs

melvinw:
This is the last big PR of the interactive shell series! This targets readline >= 4.0 on linux for now. I figure we can graft on the cross-platform stuff from cpython in smaller PRs when they become more pressing

### DIFF
--- a/NINJA-config.sh
+++ b/NINJA-config.sh
@@ -94,8 +94,7 @@ main() {
   echo DEPS prebuilt/ninja/*/deps.txt
 
   echo
-  # TODO: remove --without-readline once frontend_pyreadline is complete.
-  _OIL_DEV=1 ./configure --without-readline
+  ./configure
 
   # Reads the deps.txt files above
   PYTHONPATH=. build/ninja_main.py
--- a/core/shell.py
+++ b/core/shell.py
@@ -509,11 +509,8 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
 
   elif flag.i:  # force interactive
     src = source.Stdin(' -i')
-    if mylib.PYTHON:
-      line_reader = reader.InteractiveLineReader(
-          arena, prompt_ev, hist_ev, readline, prompt_state)
-    else:
-      line_reader = None
+    line_reader = reader.InteractiveLineReader(
+        arena, prompt_ev, hist_ev, readline, prompt_state)
     mutable_opts.set_interactive()
 
   else:
@@ -526,11 +523,8 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
         stdin = mylib.Stdin()
         if stdin.isatty():
           src = source.Interactive()
-          if mylib.PYTHON:
-            line_reader = reader.InteractiveLineReader(
-                arena, prompt_ev, hist_ev, readline, prompt_state)
-          else:
-            line_reader = None
+          line_reader = reader.InteractiveLineReader(
+              arena, prompt_ev, hist_ev, readline, prompt_state)
           mutable_opts.set_interactive()
         else:
           src = source.Stdin('')
@@ -602,6 +596,7 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
 
     return status
 
+  history_filename = os_path.join(home_dir, '.config/oil/history_%s' % lang)
   if exec_opts.interactive():
     state.InitInteractive(mem)
     # bash: 'set -o emacs' is the default only in the interactive shell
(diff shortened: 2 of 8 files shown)


melvinw:
If I'm not mistaken, this also seems to have brought `delta_cpp` down to ZERO!

cpp/frontend_pyreadline.cc
```diff
@@ -4,73 +4,236 @@
...
 void Readline::read_history_file(Str* path) {
+#if HAVE_READLINE
+  char* p = nullptr;
+  if (path != nullptr) {
+    p = path->data();
+  }
+  read_history(p);
+#else
   assert(0);  // not implemented
+#endif
```
melvinw, line 117: An alternative to this `#if HAVE_READLINE ... #else ...` pattern could be to just keep two implementations in this file, separated by one big if-else. Not sure which is uglier...

cpp/frontend_pyreadline.cc
```diff
@@ -4,73 +4,236 @@
...
+    return nullptr;
+  }
+
+  // According to https://web.mit.edu/gnu/doc/html/rlman_2.html#SEC37, readline
+  // will free any memory we return to it.
+  int n = len(result) + 1;
+  auto* ret = static_cast<char*>(malloc(n));
+  DCHECK(ret != nullptr);
+  memcpy(ret, result->data(), n);
+  return ret;
```
melvinw, line 39: might be better as
```suggestion
  return strdup(rest->data());
```

cpp/frontend_pyreadline.cc
```diff
@@ -4,73 +4,235 @@
...
+
+static void display_matches_hook(char** matches, int num_matches,
+                                 int max_length) {
+  if (gReadline->display_ == nullptr) {
+    return;
+  }
+  auto* gc_matches = Alloc<List<Str*>>();
+  // It isn't clear from the readline documentation, but matches[0] is the
+  // completion text and the matches returned by any callbacks start at index 1.
+  for (int i = 1; i <= num_matches; i++) {
```
melvinw, line 54: This was the reason for that weird indentation error you saw in this screencast:
[asciicast](https://asciinema.org/a/557003)

You can see the correct behavior in this new one:
[asciicast](https://asciinema.org/a/557335)

NINJA-config.sh
```diff
@@ -94,8 +94,7 @@ main() {
   echo DEPS prebuilt/ninja/*/deps.txt
 
   echo
-  # TODO: remove --without-readline once frontend_pyreadline is complete.
-  _OIL_DEV=1 ./configure --without-readline
+  ./configure
```
andychu, line 97: OK it looks like we need `_build/detected-config.sh`

I might add back the guard to remove the `.h` file, because we have a separate CPP `.h` file that is tiny

The other `.h` file annoys me because it takes awhile to generate, and is part of the CPython stuff

core/shell.py
```diff
@@ -668,6 +661,12 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
     cmd_ev.MaybeRunExitTrap(box)
     status = box[0]
 
+    if readline:
```
andychu, line 664: One thing to think about is if IOError and OSError that bubble to the top prevent this from being reached

I believe the Interactive() and MaybeRunExitTrap() functions catch MOST exceptions, but not all

But I think we can see if it comes later

cpp/frontend_pyreadline.cc
andychu: Awesome!  I'm excited to try this out!

cpp/frontend_pyreadline.cc
```diff
@@ -4,73 +4,235 @@
...
+#endif
+
 namespace py_readline {
 
+static Readline* gReadline = nullptr;
+
+// Assuming readline 4.0+
+#if HAVE_READLINE
+
+  #define completion_matches(x, y) \
```
andychu, line 19: Can we just remove the macro?  It seems like it's only used in one place or so



I have been trying to use CAPS() for macros consistently too

cpp/frontend_pyreadline.cc
andychu: I'm not too bothered by it ... it would be nice to conditionally compile and link this one file, but we can look at that later

cpp/frontend_pyreadline.cc
```diff
@@ -4,73 +4,235 @@
...
 void Readline::resize_terminal() {
+#if HAVE_READLINE
+  rl_resize_terminal();
+#else
   assert(0);  // not implemented
+#endif
 }
 
 Readline* MaybeGetReadline() {
   // TODO: incorporate OIL_READLINE into the build config
```
andychu, line 231: looks like this TODO can be removed

andychu (commented):

The code looks pretty short too, which I like

andychu:
Some things I can do after this lands:

- Fix up the `spec-cpp` test harness.  Some of the delta cpp are NEGATIVE now haha!  In theory there is  no divergence between `oil.py` and `oils_cpp.py`, so I think clearing that up will make it less confusing
- I'm interested if `test/stateful` can run now on osh-native

melvinw:
> I'm interested if test/stateful can run now on osh-native

I tried this manually earlier! Would be cool to get it into CI, though. `stateful/interactive` matches python! `stateful/signals` just needs the `SIGINT` fix. `stateful/job-control` needs some attention, though

cpp/frontend_pyreadline.cc
melvinw: ah yeah. this isn't really necessary. i guess `line_input` had it because of the cross-platform stuff

PR #1487, opened 2023-02-05
[cpp] set rl_completion_append_character to NUL

melvinw:
Else we end up with awkward spaces after <TAB>-completion.


andychu:
OK I assume CPython also does this?  It changes some of the defaults

melvinw:
Yup. I had originally copied it over, but it got lost while I was doing some refactoring

PR #1488, opened 2023-02-05
[wip] readline sigint

melvinw:


melvinw:
This makes `osh-cpp` pass the entire `spec/stateful/signals` suite!

cpp/core.cc
```diff
@@ -256,6 +268,15 @@ int LastSignal() {
   return gSignalHandler->last_sig_num_;
 }
 
+int SigintCount() {
+  if (gSignalHandler == nullptr) {
```
andychu, line 272: Comment about when it can be nullptr?  I guess it depends on the order of InitShell() vs. ReadByte() and so forth?

I am wondering if it should be an DCHECK(), but I guess we have not really settled on the order of those things ...

cpp/frontend_pyreadline.cc
```diff
@@ -235,4 +242,50 @@ Readline* MaybeGetReadline() {
...
+    gReadline->latest_line_ = nullptr;
+  } else {
+    gReadline->latest_line_ = StrFromC(line);
+    free(line);
+  }
+  gReadline->ready_ = true;
+  rl_callback_handler_remove();
+}
+
+Str* readline(Str* prompt) {
```
andychu, line 256: Probably should point to these docs here:

https://tiswww.case.edu/php/chet/readline/readline.html#Alternate-Interface-Example

cpp/frontend_pyreadline.cc
```diff
@@ -235,4 +242,50 @@ Readline* MaybeGetReadline() {
...
+  FD_ZERO(&fds);
+  rl_callback_handler_install(prompt->data(), readline_cb);
+
+  gReadline->latest_line_ = nullptr;
+  gReadline->ready_ = false;
+  while (!gReadline->ready_) {
+    FD_SET(fileno(rl_instream), &fds);
+    int ec = select(FD_SETSIZE, &fds, NULL, NULL, NULL);
+    if (ec == -1) {
+      if (errno == EINTR && pyos::LastSignal() == SIGINT) {
```
andychu, line 267: What's the difference between `SigintCount() > 0` and `pyos::LastSignal() == SIGINT` ?

cpp/frontend_pyreadline.h
```diff
@@ -43,18 +43,23 @@ class Readline {
...
+           maskbit(offsetof(Readline, display_)) |
+           maskbit(offsetof(Readline, latest_line_));
   }
 
   int begidx_;
   int endidx_;
   Str* completer_delims_;
   completion::ReadlineCallback* completer_;
   comp_ui::_IDisplay* display_;
+  Str* latest_line_;
```
andychu, line 55: This could use some comments about what the state means, i.e. latest_line_ != nullptr vs. ready_ vs EOF, etc.

cpp/frontend_pyreadline.cc
```diff
@@ -235,4 +242,50 @@ Readline* MaybeGetReadline() {
...
+  gReadline->ready_ = true;
+  rl_callback_handler_remove();
+}
+
+Str* readline(Str* prompt) {
+  fd_set fds;
+  FD_ZERO(&fds);
+  rl_callback_handler_install(prompt->data(), readline_cb);
+
+  gReadline->latest_line_ = nullptr;
```
andychu, line 261: This loop could use some comments.  How does it work?

spec/stateful/signals.py
```diff
@@ -105,6 +105,7 @@ def sigmultiple_trapped_prompt(sh):
   expect_prompt(sh)
 
   sh.kill(signal.SIGUSR1)
+  time.sleep(0.1)
```
andychu, line 108: Add a comment here about this

I guess the shell does the right thing, but it's hard to test the non-deterministic output?

andychu (commented):
Thanks for getting this working!  By and large it looks very good, and pretty short, though I would like more comments since I find it hard to wrap my head around this stuff :)

The GNU readline interface is particularly confusing because it uses globals everywhere -- it would be nice to document our assumptions

andychu:
Also it seems like we need to do something for when the shell is trapped in pure computation

```
for i in {1....100000}; do echo $i; done > /dev/null
```

i.e. in that case it will not hit `ReadByte()` or anything like that 

So will KeyboardInterrupt not get called?

My thought was to put it in `gHeap.Allocate()`, although there is also the place in the main loop where we do:

```
mylib.MaybeCollect()
RunPendingTraps()
```

We can do this in another change of course -- just wondering if you've thought about this

And do we have a test case that exposes it?

If it's easier we can have a call about this, especially since I don't remember all the concerns

cpp/core.cc
melvinw: Oh this is a holdover from a branch where I was trying to throw from `Alloc()`, which might have been allocating `gSignalHandler`. I'll change this to a `DCHECK`

melvinw:
> Also it seems like we need to do something for when the shell is trapped in pure computation

Yeah, I was gonna send a separate change for this. Might as well tack it onto this PR, though

cpp/frontend_pyreadline.cc
melvinw: Oh whoops these should all be `SigintCount() > 0` (it's more robust to multiple signals arriving in quick succession). Nice catch!

spec/stateful/signals.py
melvinw: Yeah I think so. You always see the output of both traps, just sometimes out of order

melvinw:
OK updated this a bit. Let me know what you think. Should we do anything else in the new test in `spec/stateful`?

cpp/core.cc
andychu: Hm should this be DCHECK or not?  Either way adding a comment would help

cpp/frontend_pyreadline.cc
```diff
@@ -9,6 +12,8 @@
   #include <readline/readline.h>
 #endif
 
+#include "core.h"
```
andychu, line 15: Should be `cpp/core.h` to follow the style

Also in `cpp/NINJA_subgraph.py`, it means you need a `//cpp/core` dependency (wish this would fail if you forget it -- I recently fixed other missing deps)

cpp/frontend_pyreadline.cc
```diff
@@ -235,4 +242,59 @@ Readline* MaybeGetReadline() {
...
+    // Remove this check if we start calling select() with a timeout above.
+    DCHECK(ec > 0);
+    if (FD_ISSET(fileno(rl_instream), &fds)) {
+      // Feed readline.
+      rl_callback_read_char();
+    }
+  }
+
+  if (gReadline->latest_line_ != nullptr) {
+    Str* s = StrFromC(gReadline->latest_line_);
```
andychu, line 291: Yeah I like this new style where we do the `StrFromC()` allocation outside the callback

Even though it's a "synchronous callback" I like keeping all the GC stuff "on the edges" of the code, and having plain C in the  middle

andychu (commented):
Looks great!  Thank you

Just a few small things

andychu:
And I mentioned on Zulip that some Internet people are eagerly awaiting this work :)

cpp/core.cc
melvinw: Ah doh. Missed this one

melvinw:
All fixed up!

andychu:
Oh can you add the dep changes in cpp/NINJA_subgraph.py ?  I don't see that file

(Also I get a little confused by the force push, I think you can just do a normal push and then it's easy to review the incremental diffs? )

melvinw:
That change should in in the second-to-last commit.

And good to know! I wasn't sure if you got much use out of those unsquashed `fixup!` commits. I'll leave them unsquashed from now on

andychu:
Great thanks!  Yeah maybe I missed the diff

PR #1490, opened 2023-02-05
[refactor] _PrintWithSpanId -> _PrintWithLocation

abathur:
This bite is a little bigger (and more complex) than I really wanted to take, so my confidence is only moderate. Be skeptical :)

### DIFF
--- a/core/executor.py
+++ b/core/executor.py
@@ -139,7 +139,7 @@ class ShellExecutor(vm._Executor):
           if e.span_id == runtime.NO_SPID:
             e.span_id = self.errfmt.CurrentLocation()
           # e.g. 'type' doesn't accept flag '-x'
-          self.errfmt.PrefixPrint(e.msg, prefix='%r ' % arg0, span_id=e.span_id)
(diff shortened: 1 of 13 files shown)


andychu:
Thanks for looking at this!

Now that I think about it, I said we were going to remove loc.Span().  That's true long term, but short term we're going to INTRODUCE loc.Span(), because it's a trivial wrapper around a span_id that lets us do the type-driven refactoring.

It would be nice to "pull the band-aid off all at once", but now that I look at it, we have to do it two steps:

1. Change APIs to take `loc_t`, and pass `loc.Span(span_id)` **mechanically**
2. Later get rid of `loc.Span()

The reason is that there are simply a lot of usages!  This command shows there are 677 !!  Location info is all over the codebase, because that's how we get fine-grained error messages.

```
$ metrics/source-code.sh osh-files | xargs egrep 'span_id|spid'  | wc -l
677
```

----

So I think I found a tinier piece to bite off.  Take a look at this PR:

https://github.com/oilshell/oil/pull/1491

And maybe try the same thing for one of the other methods `Print_()` or `PrintMessage()`

I think those will be smaller bites and we can lean on the type system more.  (Although leaning on the CI is also OK!)

Thanks

andychu:
closing old PR

PR #1491, opened 2023-02-06
[refactor] Type-driven refactoring: remove span_id in PrefixPrint()

andychu:
The same can be done for Print_() and PrintMessage(), which now take a span_id, but should take a location.

### DIFF
--- a/core/executor.py
+++ b/core/executor.py
@@ -139,7 +139,7 @@ class ShellExecutor(vm._Executor):
           if e.span_id == runtime.NO_SPID:
             e.span_id = self.errfmt.CurrentLocation()
           # e.g. 'type' doesn't accept flag '-x'
-          self.errfmt.PrefixPrint(e.msg, prefix='%r ' % arg0, span_id=e.span_id)
+          self.errfmt.PrefixPrint(e.msg, '%r ' % arg0, loc.Span(e.span_id))
           status = 2  # consistent error code for usage error
 
     return status
--- a/core/ui.py
+++ b/core/ui.py
@@ -11,7 +11,7 @@ from __future__ import print_function
 
 from _devbuild.gen.id_kind_asdl import Id, Id_t, Id_str
 from _devbuild.gen.syntax_asdl import (
-    Token, SourceLine, command_t, command_str,
+    loc_t, Token, SourceLine, command_t, command_str,
     source_e, source__Stdin, source__MainFile, source__SourcedFile,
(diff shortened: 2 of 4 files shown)


core/executor.py
```diff
@@ -139,7 +139,7 @@ def RunBuiltin(self, builtin_id, cmd_val):
           if e.span_id == runtime.NO_SPID:
             e.span_id = self.errfmt.CurrentLocation()
           # e.g. 'type' doesn't accept flag '-x'
-          self.errfmt.PrefixPrint(e.msg, prefix='%r ' % arg0, span_id=e.span_id)
+          self.errfmt.PrefixPrint(e.msg, '%r ' % arg0, loc.Span(e.span_id))
```
andychu, line 142: This is the common case: simply wrap span ID in `loc.Span()`

core/ui.py
```diff
@@ -304,9 +304,10 @@ def CurrentLocation(self):
     else:
       return runtime.NO_SPID
 
-  def PrefixPrint(self, msg, prefix, span_id=runtime.NO_SPID):
-    # type: (str, str, int) -> None
+  def PrefixPrint(self, msg, prefix, blame_loc):
+    # type: (str, str, loc_t) -> None
     """Print a hard-coded message with a prefix, and quote code."""
+    span_id = location.GetSpanId(blame_loc)
```
andychu, line 310: we can use our function to convert a loc_t to a span_id

Later will remove it from this internal function, but for now it lets us make a small diff

osh/cmd_eval.py
```diff
@@ -1046,7 +1046,7 @@ def _Dispatch(self, node, cmd_st):
           else:
             # Only print warnings, never fatal.
             # Bash oddly only exits 1 for 'return', but no other shell does.
-            self.errfmt.PrefixPrint(msg, prefix='warning: ', span_id=tok.span_id)
+            self.errfmt.PrefixPrint(msg, 'warning: ', tok)
```
andychu, line 1049: Here we do NOT pass `loc.Span()` -- we just pass the token itself.

Remember that a Token **is a** location

Anywhere you would se `loc.Span(tok.span_id)` -- that's wrong and should just be `tok`

That's like doing a "double conversion" and is not necessary

andychu (commented):
I added a few comments which I think will help with similar changes

Also notice that we are mostly getting rid of keyword arguments, which means getting rid of names at the call site.  I think it's OK to have  3 args in any function ... the type system helps  prevent errors.

PR #1493, opened 2023-02-07
[completion] fix file system action trailing slash regression

melvinw:
Introduced in 78ca6b96f9e84e86899b3d6aeb908c39304dea95


andychu:
OK I think it's better the first way ... I'll see if we can get some test coverage there too

I'll take another look tomorrow, thanks :)

PR #1497, opened 2023-02-10
[cpp] don't use readline if stdio is not tty

melvinw:
This removes the stutter that happens from readline redisplaying to
stdout when osh-cpp is running with -i as part of a pipeline.

This should fix `spec/interactive`

### DIFF
--- a/cpp/frontend_pyreadline.cc
+++ b/cpp/frontend_pyreadline.cc
@@ -252,9 +252,7 @@ static void readline_cb(char* line) {
   rl_callback_handler_remove();
 }
 
-// See the following for some loose documentation on the approach here:
-// https://tiswww.case.edu/php/chet/readline/readline.html#Alternate-Interface-Example
-Str* readline(Str* prompt) {
+static Str* readline_tty(Str* prompt) {
   fd_set fds;
   FD_ZERO(&fds);
   rl_callback_handler_install(prompt->data(), readline_cb);
@@ -295,6 +293,29 @@ Str* readline(Str* prompt) {
   }
 
   return nullptr;
+}
+
+static Str* readline_no_tty(Str* prompt) {
+  mylib::Writer* w = mylib::Stderr();
+  w->write(prompt);
+  w->flush();
+
+  mylib::LineReader* r = mylib::Stdin();
+  Str* s = r->readline();
+  if (s != kEmptyString) {
+    return s;
+  }
+
+  return nullptr;
+}
+
+// See the following for some loose documentation on the approach here:
+// https://tiswww.case.edu/php/chet/readline/readline.html#Alternate-Interface-Example
+Str* readline(Str* prompt) {
+  if (!mylib::Stdin()->isatty() || !mylib::Stdout()->isatty()) {
+    return readline_no_tty(prompt);
+  }
+  return readline_tty(prompt);
 }
 
 }  // namespace py_readline
--- a/cpp/frontend_pyreadline.cc
+++ b/cpp/frontend_pyreadline.cc
@@ -252,7 +252,9 @@ static void readline_cb(char* line) {
   rl_callback_handler_remove();
 }
 
-static Str* readline_tty(Str* prompt) {
+// See the following for some loose documentation on the approach here:
+// https://tiswww.case.edu/php/chet/readline/readline.html#Alternate-Interface-Example
+Str* readline(Str* prompt) {
   fd_set fds;
   FD_ZERO(&fds);
   rl_callback_handler_install(prompt->data(), readline_cb);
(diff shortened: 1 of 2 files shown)


andychu:
I wonder if there is bug in the Python level of OSH?     e.g. in `core/shell.py` we are initializing `InteractiveLineReader` in two places -- so I guess it's overly aggressive ?

I looked and found this in the bash source code -- maybe the difference is that they check if `stderr` is a terminal?

( In general to keep our "executable spec" property, I want to aim for fixes in Python, and minimize conditionals / code paths in C++ )


```
  /* First, let the outside world know about our interactive status.                                                                                                                                                                                                                                                                                  
     A shell is interactive if the `-i' flag was given, or if all of                                                                                                                                                                                                                                                                                  
     the following conditions are met:                                                                                                                                                                                                                                                                                                                
  no -c command                                                                                                                                                                                                                                                                                                                                       
  no arguments remaining or the -s flag given                                                                                                                                                                                                                                                                                                         
  standard input is a terminal                                                                                                                                                                                                                                                                                                                        
  standard error is a terminal                                                                                                                                                                                                                                                                                                                        
     Refer to Posix.2, the description of the `sh' utility. */                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                      
  if (forced_interactive ||   /* -i flag */                                                                                                                                                                                                                                                                                                           
      (!command_execution_string && /* No -c command and ... */                                                                                                                                                                                                                                                                                       
       wordexp_only == 0 &&   /* No --wordexp and ... */                                                                                                                                                                                                                                                                                              
       ((arg_index == argc) ||    /*   no remaining args or... */                                                                                                                                                                                                                                                                                     
  read_from_stdin) &&   /*   -s flag with args, and */                                                                                                                                                                                                                                                                                                
       isatty (fileno (stdin)) && /* Input is a terminal and */                                                                                                                                                                                                                                                                                       
       isatty (fileno (stderr)))) /* error output is a terminal. */                                                                                                                                                                                                                                                                                   
    init_interactive ();   
```

andychu:
Well I guess there is an `||` statement there in bash, so `-i` should be enough

Oil does that too, hm let me see what Python does

```
  elif flag.i:  # force interactive                                                                                                                                                                                                                                                                                                                   
    src = source.Stdin(' -i')                                                                                                                                                                                                                                                                                                                         
    line_reader = reader.InteractiveLineReader(                                                                                                                                                                                                                                                                                                       
        arena, prompt_ev, hist_ev, readline, prompt_state)                                                                                                                                                                                                                                                                                            
    mutable_opts.set_interactive()                                                                                                                                                                                                                                                                                                                    
                                   
```

andychu:
OK I found the difference in `Python-2.7.13/Parser/myreadline.c`

```
    /* This is needed to handle the unlikely case that the                                                                                                                                                                                                                                                                                            
     * interpreter is in interactive mode *and* stdin/out are not                                                                                                                                                                                                                                                                                     
     * a tty.  This can happen, for example if python is run like                                                                                                                                                                                                                                                                                     
     * this: python -i < test1.py                                                                                                                                                                                                                                                                                                                     
     */                                                                                                                                                                                                                                                                                                                                               
    if (!isatty (fileno (sys_stdin)) || !isatty (fileno (sys_stdout)))                                                                                                                                                                                                                                                                                
        rv = PyOS_StdioReadline (sys_stdin, sys_stdout, prompt);                                                                                                                                                                                                                                                                                      
    else                                                                                                                                                                                                                                                                                                                                              
        rv = (*PyOS_ReadlineFunctionPointer)(sys_stdin, sys_stdout,                                                                                                                                                                                                                                                                                   
                                             prompt);   
```

Can you add a pointer to that?  In case we want to reduce our C++ code later?  :) 

For now this makes sense

cpp/frontend_pyreadline.cc
```diff
@@ -297,4 +295,27 @@ Str* readline(Str* prompt) {
...
+  Str* s = r->readline();
+  if (s != kEmptyString) {
+    return s;
+  }
+
+  return nullptr;
+}
+
+// See the following for some loose documentation on the approach here:
+// https://tiswww.case.edu/php/chet/readline/readline.html#Alternate-Interface-Example
```
andychu, line 313: Also this comment with bash should now be moved to `readline_tty()`

And then add the comment pointing to CPython for the stdin/stdout conditional can go here

melvinw:
Ah, I should have pointed out that `osh` depends on the python runtime for this. I do like the idea of making this switch an explicit thing in the spec. It should be trivial to move this decision into |InteractiveLineReader::_GetLine()](https://github.com/oilshell/oil/blob/master/frontend/reader.py#L190). I'll give that a try tonight probably

andychu:
Oh yeah, we can either move it there, or maybe even in the constructor to InteractiveLineReader ...

I guess the problem with putting it in the constructor is if the shell does a redirect and `dup()`, then the result can be different for each `raw_input()` call

Yeah I guess Python does it every time, and those isatty() syscalls should be very fast.   (If you strace bash it does a ton of syscalls every time you hit ENTER)

andychu:
BTW I think in the distant future I want to replace all the Python extensions with our working C++ bindings, then wrapped in Python ... 

I want to brag about how little native C/C++ code we have :)

melvinw:
Alrighty pushed a fixup with the new approach

andychu:
Excellent, thank you!

PR #1498, opened 2023-02-11
[cpp] implement pyos::InputAvailable()

melvinw:
this should reduce the delta on `spec/builtin-io`.


andychu:
Awesome!  How about adding a test that just passes `0`  for stdin ?  

In our harnesses I think it should reliably return false, but it's also OK to simply call the function and not assert anything.  I found several bugs doing that when making the coverage numbers go up :)

andychu:
Actually the bugs I found were GC bugs, and this function doesn't allocate

And there is a spec test for this, so this is good :)

PR #1502, opened 2023-02-14
[asdl] Reorder member variables of generated c++ types

watk:
Hey Andy, this seems to work fine.. I ran a few tests under asan and didn't see any issues but the CI run will be interesting.

I haven't tried to benchmark it. Did you want to try?

### DIFF
--- a/asdl/gen_cpp.py
+++ b/asdl/gen_cpp.py
@@ -372,23 +372,18 @@ class ClassDefVisitor(visitor.AsdlVisitor):
       self.Emit("class %s {" % class_name, depth)
     self.Emit(" public:", depth)
 
-    all_fields = ast_node.fields + attributes
-
-    bits = []
-    if all_fields:
-      for field in all_fields:
-        if _IsManagedType(field.typ):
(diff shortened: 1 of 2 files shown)


andychu:

Sorry I wonder if you can reopen this PR from a branch of `oilshell/oil`?  I would like it if contributors work in forks, but Github doesn't appear to support that with our CI secrets.

I just ran into this issue again with @abathur the last week -- our CI requires a "secret" to publish HTML, which isn't available in CI jobs running from forks.  



----

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Allow.20Contributors.20to.20Publish.20CI.20Results


The problem is Github only supports this for private repos:

https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

> enable these users to run workflows on fork pull requests, we’ve introduced three new settings at the enterprise, organization, and repository level for private repositories only.


(I think it should be user-based, not fork-based, but it's not.  I already added you as a contributor back in May)

watk:
Ohh, I forgot about this. Done!

PR #1503, opened 2023-02-14
[asdl] Reorder member variables of generated c++ types

watk:
PTAL

### DIFF
--- a/asdl/gen_cpp.py
+++ b/asdl/gen_cpp.py
@@ -372,23 +372,18 @@ class ClassDefVisitor(visitor.AsdlVisitor):
       self.Emit("class %s {" % class_name, depth)
     self.Emit(" public:", depth)
 
-    all_fields = ast_node.fields + attributes
-
-    bits = []
-    if all_fields:
-      for field in all_fields:
-        if _IsManagedType(field.typ):
-          bits.append('maskbit(offsetof(%s, %s))' % (class_name, field.name))
-
-    if bits:
-      mask_init = 'field_mask()'
-      #mask_init = 'field_mask_(kZeroMask)'
-    else:
-      mask_init = 'kZeroMask'
+    # Ensure that the member variables are ordered such that GC managed objects
+    # come before any unmanaged ones because we use `HeapTag::Scanned`.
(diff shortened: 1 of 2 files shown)


andychu:
Looks great, thank you!

I have this script to profile some workloads, which isn't in the CI yet:

https://github.com/oilshell/oil/blob/master/benchmarks/perf.sh

BEFORE this change:

http://www.oilshell.org/git-branch/dev-andy-21/97197b1e/perf/osh-parse.report.txt


```
# Overhead       Samples  Command  Shared Object        Symbol                                                                                                                                                                                         
# ........  ............  .......  ...................  ...............................................................................................................................................................................................
#
    15.93%           141  osh      oils-for-unix        [.] MarkSweepHeap::MaybeMarkAndPush
     8.04%            72  osh      oils-for-unix        [.] MarkSweepHeap::TraceChildren
     7.98%            72  osh      libc-2.27.so         [.] _int_malloc
     5.45%            50  osh      oils-for-unix        [.] MarkSweepHeap::Allocate
     5.05%            46  osh      oils-for-unix        [.] std::__detail::_Executor<char const*, std::allocator<std::__cxx11::sub_match<char const*> >, std::__cxx11::regex_traits<char>, true>::_M_dfs
     4.36%            40  osh      libc-2.27.so         [.] calloc
```

I will try pulling and running it with this change.  I'm not sure if it will show up in this particular workload (`profile-osh-parse`)

I'm pretty sure the other one will show up -- removing `FindObjHeader()` which works around the vtable problem, and which we no longer need

andychu:
I think this probably did something because `TraceChildren` seems faster, but the measurements with the sampling profiler are a bit noisy.  Notes here:

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Profile.20after.20HeapTag.3A.3AScanned.20change


I think removing `FindObjHeader()` will show up even more, I saw that in `perf report` interactively

I wasn't quite sure how to interpret it, but it definitely seemed hot

We can chat about that whenever you have time.  It's a bigger change, but pretty mechnical, and it touches the core parts of the GC

watk:
Sweet!  Thanks

Issue #1450, opened 2023-01-13
autoconf `$(( $* ))` requires shopt -s eval_unsafe_arith


zackw:
Here's another fun word-splitting problem exposed by Autoconf's test suite.

```sh
as_fn_arith ()                     
{
    as_val=$(( $* ))
}
as_fn_arith 1 + 1
echo $as_val
```

prints "2" when executed by ksh, dash, or bash, but when executed by osh it prints

```
 as_val=$(( $* ))
            ^~
test.sh:3: fatal: Invalid integer constant '1 + 1'
```

Since this is a use of `$*`, not within double quotes, I _think_ IFS should be irrelevant.

andychu:
Oof, I reproduced this.  So this one can actually be fixed by setting  `shopt -s eval_unsafe_arith` first

The issue is that this kind of dynamic arithmetic can be unsafe in certain contexts with shells like ksh / bash that have array references.

i.e. the  `1 + 1` can also be `a[$(echo 42 >  PWNED)]`

https://github.com/oilshell/blog-code/tree/master/crazy-old-bug

I reported this to OpenBSD ksh awhile ago, and they disallowed it as well.    But OSH is stricter than what they did.  

- I wonder how many autoconf scripts rely on it?  Is it mainly a test suite thing, or is it semantics autoconf relies on commonly ?   

I guess we should test more besides CPython's.  CPython `./configure` apparently does not rely on it.  (Although there is a slight possibility we could be running it incorrectly because of an issue like this.)

---

I also reproduced the IFS bug, so let's look at that and then circle back

andychu:
I guess the minimum we should do is actually put `shopt -s eval_unsafe_arith` in the error message.  We've done that for a few other strict options

But yeah I would say autoconf compatibility without extra flags is pretty clearly essential

zackw:
Well, what autoconf (technically, the [M4sh layer](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Programming-in-M4sh.html)) is trying to do with this construct, is use `$(( ... ))` for arithmetic if possible, falling back to `expr` if not.  So, early in a script, it emits

```
if (eval "test \$(( 1 + 1 )) = 2") 2>/dev/null
then :
  eval 'as_fn_arith ()
  {
    as_val=$(( $* ))
  }'
else case e in #(
  e) as_fn_arith ()
  {
    as_val=`expr "$@" || test $? -eq 1`
  } ;;
[...]

Issue #746, opened 2020-05-10
Support `>&/filename` as a synonym for `&>`


o11c:
This is used by Debian's default `.bashrc` (excerpted below), which is notable enough that people *will* copy from it and expect it to work.

If nothing else, it should be a more specific error message.

```
if [ -n "$force_color_prompt" ]; then
    if [ -x /usr/bin/tput ] && tput setaf 1 >&/dev/null; then
	# We have color support; assume it's compliant with Ecma-48
	# (ISO/IEC-6429). (Lack of such support is extremely rare, and such
	# a case would tend to support setf rather than setaf.)
	color_prompt=yes
    else
	color_prompt=
    fi
fi
```

andychu:
Hm I'm not seeing the issue.  What happens now, and what do you expect to happen?

o11c:
Ah, this only happens when `force_color_prompt=yes` is uncommented. I'd forgotten that some people haven't done that.

```
osh$ . .bashrc
      if [ -x /usr/bin/tput ] && tput setaf 1 >&/dev/null; then
                                                ^~~~~~~~~
.bashrc:51: Invalid descriptor '/dev/null'.  Expected D, -, or D- where D is an integer
```

andychu:
Oh geez, it looks like zsh supports both of them too... 

This is very lame and confusing... it's like if you made a typo of `**` as `^` and they changed the meaning based on the type of the argument.

andychu:
This is the kind of thing I really wish people would patch upstream... it's utterly trivial, and Oil will likely not run such rc files unmodified anyway.

See the analogy to Clang and GCC here.

http://www.oilshell.org/blog/2020/01/ambitions.html#appendix-comments-that-add-color

In other words, Clang has been around for 15+ years, and it builds Android and Chrome and Apple's stuff, but every distro still has to patch tons of things to build with Clang.  I think the situation will be similar with Oil.

I'm prioritizing having everything be "possible" and having Oil be fast first.  Once there are users they will be motivated to either patch Oil or upstream these changes (which IMO make the code clearer).

andychu:
This also reminds me I want to redesign the redirect syntax, but that will be left out of Oil v1.

Oil will have consistent meanings for `<  >  <<  >>`, rather than the weird asymmetry where `>>` means append and `<<` means here doc.  Here docs will be more like here strings without the operator `<<<`.

I haven't thought about what "stderr and stdout" looks like, but it has to be better than `&>` vs. `>&` ...

[...]

PR #1505, opened 2023-02-16
[spec/redirect] Add a test for bash's handling of >&file

watk:
Add a spec test for this bash only behaviour. From the man page:
```
There are two formats for redirecting standard output and standard error:
    &>word
and
    >&word
```

Should osh be matching this behaviour do you think? It's in ubuntu's default bashrc (which is how I came across this):
```
λ grep '>&' /etc/skel/.bashrc
    if [ -x /usr/bin/tput ] && tput setaf 1 >&/dev/null; then
```

### DIFF
--- a/spec/redirect.test.sh
+++ b/spec/redirect.test.sh
@@ -356,31 +356,42 @@ ZZ
 ## END
 
 #### &> redirects stdout and stderr
-stdout_stderr.py &> $TMP/f.txt
+tmp="$(basename $SH)-$$.txt"  # unique name for shell and test case
+#echo $tmp
+
+stdout_stderr.py &> $tmp
+
 # order is indeterminate
-grep STDOUT $TMP/f.txt >/dev/null && echo 'ok'
-grep STDERR $TMP/f.txt >/dev/null && echo 'ok'
+grep STDOUT $tmp
+grep STDERR $tmp
+
 ## STDOUT:
-ok
-ok
+STDOUT
+STDERR
 ## END
 ## N-I dash stdout: STDOUT
 ## N-I dash stderr: STDERR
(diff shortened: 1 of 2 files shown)


andychu:
The philosophical issue is basically that we have to balance old and new ... i.e. is the effort/reward of implementing bash more going to grow the project?  Or should we concentrate more on the Oil language?   It's a difficult balance

Maybe this is a small change and it's worth it (seems like it should be, although I would want to hide it behind `shopt --set strict_redirect` in Oil/YSH)

I'm on the fence here, but I know we have more  compatibility stuff to do, like #1450 


I can share my `bashrc` and `oshrc` right now, and I think that's a very useful milestone ... but yeah there will probably be a lot of people that run into this.  

One thing to try is to simply change it to `&>` and see what the next problem in `bashrc` is.  Probably `bind` ?

----

But we should have this spec test EITHER way, to document the difference.

andychu:
Oh BTW this is bug #746

watk:
> I was wondering why this test didn't fail though!
Oops, I completely forgot to mention I was running with `--rm-tmp`! I was planning to suggest making this the default and having a flag to `keep-tmp`. But whatever you prefer.


Yep, agreed on all those thoughts. I understand the tension and reluctance to implement stuff like this. I figured that since osh is being described as "bash compatible", that you might want to make this work. 

But if you want to invert this spec test and call bash behaviour a BUG lmk.

watk:
Given that it's low priority, we can just add this spec test for documentation and decide later on whether to implement.

andychu:
Thanks, merged!

Wow good point about `--rm-tmp` ... I will remove that flag and make it the default

It might expose some other bugs

Thanks!

PR #1509, opened 2023-02-19
[core optimize] SignalSafe::ReuseEmptyList() avoids many allocations

andychu:
- AFTER: http://travis-ci.oilshell.org/github-jobs/2023-02-19__21-36-31.wwz/_tmp/gc/index.html
  - max RSS MB of compute-fib and parse-abuild-help is 8.0 and 9.0 MB
  - allocated MB is 35.0 and 49.2
- BEFORE: http://travis-ci.oilshell.org/github-jobs/2023-02-19__21-18-05.wwz/_tmp/gc/index.html
  - max RSS MB is 11.2 and 12.5 MB
  - allocated MB is 166.0 and 162.0

### DIFF
--- a/cpp/core.h
+++ b/cpp/core.h
@@ -111,6 +111,7 @@ class SignalSafe {
   SignalSafe()
       : GC_CLASS_FIXED(header_, field_mask(), sizeof(SignalSafe)),
         pending_signals_(AllocSignalList()),
+        empty_list_(AllocSignalList()),  // to avoid repeated allocation
         last_sig_num_(0),
         received_sigint_(false),
         received_sigwinch_(false),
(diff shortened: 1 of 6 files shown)


andychu:
We'll probably have to do some stuff like this elsewhere in the interpreter

There is A LOT more garbage being created!  

This one happened to stand out because the allocation is large, e.g. 4 KB as opposed to 32 or 64 bytes

cpp/core.h
```diff
@@ -147,13 +148,23 @@ class SignalSafe {
...
+    // Make sure we have a distinct list to reuse.
+    DCHECK(empty_list_ != pending_signals_);
+    pending_signals_ = empty_list_;
+
     return ret;
   }
 
+  // Main thread returns the same list as an optimization to avoid allocation.
+  void ReuseEmptyList(List<int>* empty_list) {
+    DCHECK(len(empty_list) == 0);            // main thread clears
```
melvinw, line 162: To be extra clear about assumptions here, should we add `DCHECK(empty_list->capacity_ >= kMaxPendingSignals)`? I don't know if there's a world where we decide `List<T>::clear()` should shrink the underlying slab for some reason, though (it appears we leave it the same size now). The fact that we can drop signals now means we won't accidentally allocate if the list has a tiny slab, but this might still be nice to call out? Thoughts?

melvinw (approved):
Nice! One note inline

cpp/core.h
andychu: Good point, done!

andychu:
Also notice that this shrank the CPython configure max RSS MB from 45.2 MB to 36.5!


old: http://travis-ci.oilshell.org/github-jobs/2023-02-19__21-18-05.wwz/_tmp/osh-runtime/index.html

new: http://travis-ci.oilshell.org/github-jobs/2023-02-19__21-36-31.wwz/_tmp/osh-runtime/index.html


Again that is probably a weakness in our GC policy which doesn't track sizes ... 

But still I'm getting more confident we can perform as well as bash, across every workload, with enough effort

melvinw:
Ah whoops! Sorry I deleted that branch. Muscle memory kicked in when I was looking at this on my phone...

PR #1510, opened 2023-02-20
[interactive shell] some prep for job control

melvinw:
These commits add a few libc bindings and some plumbing that will be used by another PR to add basic job control.

### DIFF
--- a/build/oil-defs/Python-2.7.13/Modules/posixmodule.c/posix_methods.def
+++ b/build/oil-defs/Python-2.7.13/Modules/posixmodule.c/posix_methods.def
@@ -23,6 +23,7 @@ static PyMethodDef posix_methods[] = {
   {"getpid", posix_getpid, METH_NOARGS},
   {"getuid", posix_getuid, METH_NOARGS},
   {"getppid", posix_getppid, METH_NOARGS},
+  {"getpgid", posix_getpgid, METH_VARARGS},
 
   {"open", posix_open, METH_VARARGS},
   {"close", posix_close_, METH_VARARGS},
@@ -39,6 +40,7 @@ static PyMethodDef posix_methods[] = {
   /* job control stuff */
   {"setpgid", posix_setpgid, METH_VARARGS},
   {"tcsetpgrp", posix_tcsetpgrp, METH_VARARGS},
+  {"tcgetpgrp", posix_tcgetpgrp, METH_VARARGS},
 
   /* note: replaced wait() call with waitpid() */
   {"wait", posix_wait, METH_NOARGS},
--- a/core/executor.py
+++ b/core/executor.py
@@ -250,7 +250,7 @@ class ShellExecutor(vm._Executor):
       p = process.Process(thunk, self.job_state, self.tracer)
       if pipeline is not None:
         p.Init_ParentPipeline(pipeline)
-        p.Start(trace.PipelinePart())
+        p.Start(trace.External(cmd_val.argv))
         pipeline.Add(p)
         status = p.Wait(self.waiter)
       else:
--- a/core/executor.py
+++ b/core/executor.py
@@ -250,9 +250,11 @@ class ShellExecutor(vm._Executor):
       p = process.Process(thunk, self.job_state, self.tracer)
       if pipeline is not None:
         p.Init_ParentPipeline(pipeline)
+        p.Start(trace.PipelinePart())
         pipeline.Add(p)
-
-      status = p.RunWait(self.waiter, trace.External(cmd_val.argv))
+        status = p.Wait(self.waiter)
+      else:
+        status = p.RunWait(self.waiter, trace.External(cmd_val.argv))
 
       # this is close to a "leaf" for errors
       # problem: permission denied EACCESS prints duplicate messages
--- a/core/executor.py
+++ b/core/executor.py
@@ -26,7 +26,7 @@ from frontend import location
 
 import posix_ as posix
 
-from typing import cast, Dict, List, TYPE_CHECKING
+from typing import cast, Dict, List, Optional, TYPE_CHECKING
 if TYPE_CHECKING:
   from _devbuild.gen.runtime_asdl import (
       cmd_value__Argv, CommandStatus, StatusArray, Proc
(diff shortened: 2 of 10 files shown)


cpp/stdlib.h
```diff
@@ -107,6 +107,32 @@ inline void write(int fd, Str* s) {
...
+  }
+}
+
+inline int tcgetpgrp(int fd) {
+  pid_t ret = ::tcgetpgrp(fd);
+  if (ret < 0) {
+    throw Alloc<OSError>(errno);
+  }
+  return ret;
+}
```
melvinw, line 134: should we unit test these? :thinking:

build/oil-defs/Python-2.7.13/Modules/posixmodule.c/posix_methods.def
```diff
@@ -23,6 +23,7 @@ static PyMethodDef posix_methods[] = {
   {"getpid", posix_getpid, METH_NOARGS},
   {"getuid", posix_getuid, METH_NOARGS},
   {"getppid", posix_getppid, METH_NOARGS},
+  {"getpgid", posix_getpgid, METH_VARARGS},
```
melvinw, line 26: probably a copy/paste bug from `getppid` above? `posix_getpgid()` in `pyext/posixmodule.c` takes a a single int arg

andychu: Hm I'm not sure what you mean? 

These files should actually be auto-generated, so you can leave them out.  There is a weird procedure documented at the top of  `build/cpython-defs.sh`.

What it does is actually lex all the source code, and then if it sees the token `getpgid`, that function will automatically be copied be copied from `Python-2.7.13/` to this file.

So when you actually add those calls, it should "work automatically".


(This was a thing I did a long time ago to "strip down" the build, which ended up being a bit of a dead end.  Long term I want to get rid of the whole "OVM tarball" build.)

core/executor.py
```diff
@@ -247,7 +248,13 @@ def RunSimpleCommand(self, cmd_val, cmd_st, do_fork, call_procs=True):
     if do_fork:
       thunk = process.ExternalThunk(self.ext_prog, argv0_path, cmd_val, environ)
       p = process.Process(thunk, self.job_state, self.tracer)
-      status = p.RunWait(self.waiter, trace.External(cmd_val.argv))
+      if pipeline is not None:
```
andychu, line 251: OK I think I see why you did this, let me think about whether there is a structure we can use that avoids threading this `parent_pipeline` param through a lot of places ...

cpp/stdlib.h
andychu: I think it's OK to leave them off at first ... In some tests I have `for_test_coverage` that does trivial tests

build/oil-defs/Python-2.7.13/Modules/posixmodule.c/posix_methods.def
melvinw: Oh weird. `Python-2.7.13/Modules/posixmodule.c` (line 9000) uses `METH_VAARGS` for this. I wonder what went wrong.

My local build was failing because of a mismatched type signature without this change, but maybe I missed a step... I did the usual `build/py.sh all` thing

andychu:
OK actually now I see this comment, which makes it likely I probably messed something up awhile ago:

https://github.com/oilshell/oil/blob/master/build/oil-defs/Python-2.7.13/Modules/posixmodule.c/posix_methods.def#L3

> // Note: This file has been manually edited, unlike the other ones generated by
// build/cpython-defs.sh.

Yeah I really want to get rid of all this stuff ...

---

I was thinking we can attach the parent_pipeline to command nodes, but now I see it's hard because you can have say a for loop as the last element, or while loop, or anything

```

l$ bin/osh -n -c 'echo hi | for i in 1; do read x; echo x=$x; done'
(command.Pipeline
  children: [
    (C {<echo>} {<hi>})
    (command.ForEach
      iter_names: [i]
      iterable: (for_iter.Words words:[{<1>}])
      body: 
        (command.DoGroup
          children: [
            (command.Sentence child:(C {<read>} {<x>}) terminator:<Id.Op_Semi _>)
            (command.Sentence
              child: (C {<echo>} {<Id.Lit_VarLike 'x='> ($ Id.VSub_DollarName x)})
              terminator: <Id.Op_Semi _>
            )
          ]
        )
      redirects: []
    )
  ]
  negated: F
  stderr_indices: []
)

```

Let me think about it

andychu:
Oh I remembered the thing I found about zsh -- `lastpipe` is INHERENTLY incompatible with job control.   

I tested zsh, and it won't suspend a pipeline where the last element is running in the CURRENT shell process, like `echo`, not forked like `wc -l`.


( I made OSH use implement `lastpipe` before I realized this conflict, but  I think `lastpipe` is nice anyway. )



```
 
$ sleep 5
^Z
zsh: suspended  sleep 5
$ fg
[1]  + continued  sleep 5
$ 
$ 
$ sleep 5 | wc -l
^Z
zsh: suspended  sleep 5 | wc -l
$ fg
[1]  + continued  sleep 5 | wc -l
0
$ 
$ 
$ 
$ sleep 5 | echo hi
hi
^Zzsh: job can't be suspended
^Zzsh: job can't be suspended
^Zzsh: job can't be suspended

```


Some comments here: https://unix.stackexchange.com/questions/476834/read-from-stdin-works-differently-in-bash-and-zsh

andychu:
Oh yeah bash documents this too

```
lastpipe

    If set, and **job control is not active**, the shell runs the last command of a pipeline not executed in the background in the current shell environment.
```

https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html

melvinw:
Nice! Thanks for digging up those notes. I'll add a test for this to `spec/stateful`

andychu:
Which behavior do we want though -- bash or zsh?  I think zsh is better but I don't know how it's implemented

I think we should meet and try to figure that out

zsh is probably doing something pretty clever

The issue with bash is that it makes `echo hi | read x` behave differently interactively and in batch scripts, which is confusing to users

melvinw:
A call sounds good btw. I'll catch you on Zulip to set up a time

PR #1511, opened 2023-02-21
[interactive shell] add job control

melvinw:


melvinw:
note for when i pick this up next: benchmark job is getting `osh I/O error: Permission denied`. probably a bad call to `tcsetpgrp()`. need to investigate...

core/process.py
```diff
@@ -868,23 +887,10 @@ def MaybeClosePipe(self):
       posix.close(self.close_r)
       posix.close(self.close_w)
 
-  def Start(self, why):
-    # type: (trace_t) -> int
+  def Start(self, why, start_in_bg=False):
```
melvinw, line 890: Idea: replace with `StartInFg()` and `StartInBg()` to make intent clearer at callsites

melvinw: nevermind.... this difference is already captured in `Run*()` vs `Start()`

core/process.py
```diff
@@ -75,6 +75,42 @@
 STYLE_PID_ONLY = 2
 
 
+class ctx_IgnoreTtySignals(object):
```
melvinw, line 78: only used in `GiveTerminal()` and might not be strictly necessary, but is good hygiene?

core/process.py
```diff
@@ -1049,6 +1110,15 @@ def Add(self, p):
       self.procs.append(p)
       return
 
+    if p.pid != -1:
```
andychu, line 1113: let's try try moving this to the caller

andychu:
Another thing I thought of with the executor state, it might break on cases like:

- `ls | eval 'cat | cat' | wc -l `  -- pipeline in eval
- `ls | { cat | cat; } | wc -l`  --  nested pipeline

Unless we have a stack

That is why there is `self.process_sub_stack` too I think

We might want to rethink that, as I'm having some problems with reducing allocations too

melvinw:
Ah yes I just ran into this while testing the following. Group membership is all messed up. Gonna add a spec test for this.
```
osh-0.14.1$ (sleep 5 | cat) | ps -e -o pid,ppid,pgid,sid,tpgid,comm | tail
 395738     439  395738  395738      -1 sshd
 395744  395738  395738  395738      -1 sshd
 395745  395744  395745  395745  395750 bash
 395750  395745  395750  395745  395750 tmux: client
 395751       2       0       0      -1 kworker/u16:1-events_unbound
 395757  387938  395757  387938  395763 python2
 395760  395757  395760  387938  395763 cat
 395761  395757  395760  387938  395763 ps
 395762  395760  395762  387938  395763 sleep
 395763  395757  395763  387938  395763 tail
```

I think we can avoid using a stack here if we just clear `fg_pipeline` before forking in `RunCommandSub()`, `RunSubshell()`, and `RunProcessSub()`, right?

core/process.py
melvinw: ended up being unnecessary

andychu:
Hm I think the most straightforward way under the current architecture is to add a stack.  We have a lot of those all over the place!

But yes I'm thinking about whether we could get rid of some of the stacks, mainly because those lists (including empty lists) are  some of our hottest allocations!  We're paying for process subs even when there are no process subs, for example.

So I'm open to any way of doing it, but adding a stack is OK and usually straightforward.  It's better to be correct and simple first, and then we have metrics to track any inefficiencies

andychu:
Basically if you search for `class ctx_` and look at what they do, most of them push and pop a stack!  It's very natural in a tree interpreter

melvinw:
Actually, we may not need this after all... bash and zsh you the SAME GROUP for all subprocesses in the foreground pipeline

```
melvin@debian:~$ { sleep 1 | cat; } | ps -e -o pid,ppid,pgid,sid,tpgid,comm | tail
 396765       2       0       0      -1 kworker/5:1-events
 396956       2       0       0      -1 kworker/u16:2-ext4-rsv-conversion
 397080       2       0       0      -1 kworker/u16:1-events_unbound
 397116  286993  397116  397116  397126 bash
 397122  397116  397122  397116  397126 bash
 397126  397122  397126  397116  397126 bash
 397127  397122  397126  397116  397126 ps
 397128  397126  397126  397116  397126 sleep
 397129  397122  397126  397116  397126 tail
 397130  397126  397126  397116  397126 cat
melvin@debian:~$ exit
melvin@debian:~$ oil_DEPS/spec-bin/zsh
debian% { sleep 1 | cat; } | ps -e -o pid,ppid,pgid,sid,tpgid,comm | tail
 396765       2       0       0      -1 kworker/5:1-events
 396956       2       0       0      -1 kworker/u16:2-events_unbound
 397080       2       0       0      -1 kworker/u16:1-events_unbound
 397116  286993  397116  397116  397132 bash
 397131  397116  397131  397116  397132 zsh
 397132  397131  397132  397116  397132 zsh
 397133  397131  397132  397116  397132 ps
 397134  397132  397132  397116  397132 sleep
 397135  397131  397132  397116  397132 tail
 397136  397132  397132  397116  397132 cat
 ```

**Comment by melvinw** [comment]
Actually I think this means we need two pieces of state here...
1. `fg_pipeline` so all subprocesses are placed in the same group
2. `pipeline_stack` so any sub-pipelines can get the PID state right for `WhenDone()`

We could probably use the bottom of (2) as (1), but it might be better to be explicit

**Comment by andychu** [comment]
Oh this is funny ... I was going to suggest writing a test that uses the `ps` command

And then I grepped the codebase, and there is one!  Apparently I wrote this in February 2022 and forgot about it :-P

However it doesn't actually assert anything ... I wonder if we can make it assert something, since we're using `ps` to debug all these problems

We could have a Python script that parses the output or something ?

I guess there may be some difficult race conditions



https://github.com/oilshell/oil/blob/master/test/group-session.sh


```
$ test/group-session.sh compare_shells subshell


-----
bin/osh
-----
NON-INTERACTIVE
[subshell]
  PID  PPID  PGID   SID COMMAND
16395  6548 16395 16395 bash
16434 16395 16434 16395 bash
16461 16434 16434 16395 python2
16464 16461 16434 16395 python2
16465 16464 16434 16395 ps
ALIVE

INTERACTIVE
[subshell]
  PID  PPID  PGID   SID COMMAND
16395  6548 16395 16395 bash
16434 16395 16434 16395 bash
16466 16434 16434 16395 python2
16469 16466 16434 16395 python2
16470 16469 16434 16395 ps
ALIVE

andy@lenny:~/git/oilshell/oil$ 
andy@lenny:~/git/oilshell/oil$ 


```

**Comment by melvinw** [comment]
gonna reopen this when the final version is ready

**Comment by melvinw** [comment]
eh, actually nevermind. it might still be useful to have this check ci...

File: test/group-session.py [hunk 4]
```diff
@@ -0,0 +1,204 @@
...
+
+def main():
+  interactive = False
+  if sys.argv[1] == "-i":
+    interactive = True
+    sys.argv.pop(1)
+
+  # CI doesn't have a TTY. Relax interactivity checks in that case.
+  if not sys.stderr.isatty():
+    interactive = False
```
melvinw, line 179: This is a little awkward. Seems to affects all of the shells, though

melvinw:
Ok, I think this is finally ready! Added a few tests and managed to remove some of the awkward conditionals from the last time we walked through this. Let me know what you think

melvinw:
`test/group-session.sh` only runs on osh for now. It would be nice to run it against osh-cpp too, though

melvinw:
hmmm. it looks like `test/group-session` is racey like you feared @andychu. `ps` might run before some parts of a pipeline have called `exec()`... Let me see if I can make the test depend less on `comm`...

andychu (commented):
Thanks for adding all the tests!  I took another look, it generally looks good, although I haven't wrapped my head entirely around it

I think what I should do is pull it and poke at the tests, and then write some comments about the assertions being made.  That will help me understand it

So you can hold off on doing anything there

There are a few places where I don't understand all the if statements, but generally speaking I think it should fit into an "event loop" model, unless I'm mistaken

It looks pretty localized, which is good

core/process.py
```diff
@@ -1002,13 +1066,21 @@ class Pipeline(Job):
   $(foo | bar)
   foo | bar | read v
```
andychu, line 1067: I just pulled and tested out

```
sleep | cat
sleep | echo hi
```

It seems like they both can't be suspended, and that behavior is OK with me

But we should document that somewhere ?  It doesn't seem obvious to me from this Pipeline class

andychu:
Oh I wonder if we can use the `script` command to run CI tests with a TTY

I'll try that out

https://stackoverflow.com/questions/32910661/pretend-to-be-a-tty-in-bash-for-any-command

core/process.py
melvinw: Ah yes this is intended. We should definitely document. I'll also spend some time later today seeing if we can pull off the "pipelines can't be suspended" log like zsh

test/group-session.py
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python2
+"""
+Utility for checking the output of group-session.sh
```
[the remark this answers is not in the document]
melvinw: Yup! CI only runs against osh at the moment, but you can run it by hand to check the other shells. We should definitely have some kind of table like the other tests at some point

andychu:
I got it running in CI without disabling `-i`:

http://travis-ci.oilshell.org/github-jobs/2023-03-11__03-34-32.wwz/

Changes needed:

- `docker -t` for that CI task
- weird special case to remove `timeout 900` (GNU tool) in our CI runner.    That tool apparently does stuff with TTYs

I might try to split out the testing changes, and merge them first

It's nice that it's working with bash/dash, so we can compare!

melvinw:
Great! Merging the test stuff first sounds like a good plan

core/process.py
melvinw: Agh I forgot that zsh does this by letting the children get stopped, printing the message when `wait()` returns with WIFSTOPPED, then sending SIGCONT to the children.

I'm not sure why exactly, but that feels awkward. Maybe it's less awkward than any convoluted group magic we'd have to do otherwise, since that stuff is hard enough to keep straight as it is... Maybe it's best to just document this

core/executor.py
```diff
@@ -370,7 +370,8 @@ def RunCommandSub(self, cs_part):
     r, w = posix.pipe()
     p.AddStateChange(process.StdoutToPipe(r, w))
 
-    p.Start(trace.CommandSub())
+    # Command subs run in the parent shell's group
+    p.Start(trace.CommandSub(), posix.getpgid(0))
```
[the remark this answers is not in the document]
melvinw: Got rid of this as part of a refactor over in `process.py`. See 0ad664e

core/process.py
```diff
@@ -868,22 +889,19 @@ def MaybeClosePipe(self):
...
+    Start this process with fork(), handling redirects.
+    If job control is enabled and pgrp is positive, the process will be placed in the provided group.
+    """
+    if self.job_state.fg_pipeline:
+      self.parent_pipeline = self.job_state.fg_pipeline
+      pgrp = self.parent_pipeline.group_id
+
+    # Command substitution should run in its parent shell's group.
+    if why.tag_() == trace_e.CommandSub:
+      pgrp = posix.getpgid(0)
```
melvinw, line 904: I can't shake the feeling that this special case should just "fall out" of the default behavior somehow... Gonna keep spinning my wheels on this

andychu:
OK I fixed up the CI a bit, and put these new tests and the spec/stateful tests next to each other, running against C++

 (the yak shaving last week paid off)

And this change makes tons of stuff pass!

BEFORE: http://travis-ci.oilshell.org/github-jobs/2023-03-12__03-41-32.wwz/_tmp/job-control/index.html (scroll down)

AFTER: http://travis-ci.oilshell.org/github-jobs/2023-03-12__03-41-08.wwz/_tmp/job-control/index.html (only dash and mksh failures)

BEFORE: http://travis-ci.oilshell.org/github-jobs/2023-03-12__03-41-32.wwz/_tmp/spec/stateful/job-control.results.txt

AFTER: http://travis-ci.oilshell.org/github-jobs/2023-03-12__03-41-08.wwz/_tmp/spec/stateful/job-control.results.txt

---

So this is looking great!

I will look at the code a bit tomorrow, but the tests are looking great now

melvinw:
Wow! Thanks for pulling together a nice report for these tests. It's so nice to be able to see all of this in one place

andychu:
I've been looking at this more, cleaning up/renaming  some related code, and thinking about the structure of the code.  

I wonder if we can remove some of the state mutation?  It seems like these members are pretty entangled, and changed from multiple places:

```
    JobState.fg_pipeline
    Process.parent_pipeline
    Pipeline.group_id
```

I added some questions here to investigate:

https://github.com/oilshell/oil/pull/1511/commits/623e2b8c67e5b32d6174ccf8955e259be81c3d24


I see some problems with the original structure of the code, which may have made this harder.

- `! true`  is technically a pipeline according to the shell grammar, which leads to the `len(self.procs) == 0` case.  I think we can remove that entirely from Pipeline, so reduce conditionals.  I will try to do that and merge it.
  -  I believe this is why you needed an extra `group_id == -1` check.  If there are no procs, then there's no PID to set group_id to.
- It seems like `self.last_pipe` should ALWAYS be set.  Right now we have another conditional there.

So basically I would like to reduce the test matrix of `Pipeline`, because it's getting crowded 

- Is there a pipe() or is there not?    Definitely need to get rid of this, there should always be at least 1 pipe.
- Is  last_pipe set or not?  As mentioned, should always be set.
- Is the process run with `&` or not?  synchronous or async.
- Should a given Process or Pipeline object be involved in job control (main shell, foreground process/pipeline) or not?  
- Is command.Simple.do_fork set or not?  I feel like this shouldn't matter, but I'm sure you ran into some issue


A few things I'm thinking of:

- writing some unit tests for Pipeline to cover these cases, in `core/process_test.py`
- I think we should be able to use the StateChange abstraction.  That's for stuff that happens between fork() and exec()
  - calling setpgid() is a process state change
  - conditionally ignoring Ctrl-Z for pipelines is a process state change
- Is our test coverage sufficient?  I suspect there are some things we can comment out and they won't break anything?  Or at least this is something I want to try

So I think it could be cleaner to factor stuff out into a public API

---

OK so this is probably a lot to look at, but I'll start with some cleanup of Pipeline, and merge it back into the `job-ctrl` branch like I've been doing

I think those cleanups will clarify the job control implementation.   At its heart I think it is simple -- it is calling setpgid() for pipelines and processes in certain cases.  So I think we can simplify the book-keeping

Comments are appreciated ... at some point we should sync up again and walk through the code

(I also have to send back the MoU and write a blog post, so this might take awhile)

andychu:
Some notes:

- 2 cleanups of Pipeline class are merged to master and to job-ctrl.  Now Pipeline::RunPipeline() can always assume that a pipeline has at least one real process AND it has the `self.last_pipe`.  So this should simplify reasoning a lot.
  -  background pipelines are handled in core/executor, with Pipeline::StartPipeline()

As an "ideal" to shoot for, I wonder if we can move the setpgid() logic outside of Process/Pipeline, and into `Executor`, which has  two cases:

- Executor::RunPipeline for the foreground -- can this do a single if JobControlEnabled() check, and "configure" the Pipeline?
- Executor::RunBackgroundJob -- does not need to deal with any process groups, unconditionally

---

- Also, As an experiment, I disabled the job_state.fg_pipeline member
  - `test/job-control.sh` fails the `fgpipe` case
  - `test/stateful` doesn't fail any more tests.

core/process.py
```diff
@@ -928,6 +949,10 @@ def StartProcess(self, why):
     # Program invariant: We keep track of every child process!
     self.job_state.AddChildProcess(pid, self)
 
+    # QUESTION: Why do we need to add the last PID?
```
melvinw, line 952: We only need this to make sure we set the correct exit status in [Pipeline.WhenDone()](https://github.com/oilshell/oil/blob/master/core/process.py#L1184-L1198). We could relax the check in that code to just always write to the last status slot if `Pipeline.WhenDone()` gets called with a `PID` that isn't part of `Pipeline.pids`

core/process.py
```diff
@@ -969,20 +995,42 @@ def WhenDone(self, pid, status):
...
+      self.job_state.MaybeTakeTerminal()
 
   def RunProcess(self, waiter, why):
     # type: (Waiter, trace_t) -> int
     """Run this process synchronously."""
     self.StartProcess(why)
+    # ShellExecutor might be calling this for the last part of a pipeline.
+    if self.parent_pipeline is None:
+      # QUESTION: Can the PGID of a single process just be the PID?  i.e. avoid
+      # calling getpgid()?
```
melvinw, line 1008: Definitely. I think I an old version of this patch actually had a conditional here, but I figured it was useless to have an extra one unless we're worried about the overhead of an extra syscall

core/process.py
```diff
@@ -969,20 +995,42 @@ def WhenDone(self, pid, status):
...
     self.fd_state = fd_state
+    self.job_state = job_state
+
+    if self.job_state.JobControlEnabled():
+      UP_node = last_node
+
+      # QUESTION: Why only set fg_pipeline when do_fork?
+      # do_fork=False ONLY when CommandEvaluator is called through
+      # main_loop.Batch().  Although I suppose main_loop.Batch() can happen in
+      # 'eval', not just in non-interactive 'osh -c'.
```
melvinw, line 1027: We need this for when the last stage of the pipeline is a brace group (or potentially any of the other cases when [CommandEvaluator._NoForkLast()](
https://github.com/oilshell/oil/blob/master/osh/cmd_eval.py#L1578-L1612)) would unset this flag.

melvinw: I'll add a comment

core/process.py
```diff
@@ -1158,6 +1225,13 @@ def RunPipeline(self, waiter, fd_state):
 
     self._Start()
 
+    # QUESTION: I think we can get rid of this if we make sure that a Pipeline
+    # always have a non-zero number of processes
```
melvinw, line 1229: Yup, this would help. I think this is only here because of the `! ...` case

andychu:
OK I just pushed a related refactoring -- `Pipeline::StartInBackground()` for `ls | wc -l &` is now separate

It doesn't share any code, and it never has a `self.last_thunk`

And then `RunPipeline()` now calls `_Start()`, which is the thing that we're adding job control to

So basically I'm trying to simplify the code so there is less "conditional explosion", i.e. smaller test matrix

core/process.py
melvinw: If we did this refactor and used the last part of the pipeline as the leader of the pipeline's group (instead of the first) we could probably get rid of `fg_pipeline`! Let me give that a try

andychu:
I think what I would like to do is move everything "up" to the executor.  If you look at the methods, it actually corresponds very tightly to the test cases we have in `test/job-control.sh` now.

```
class ShellExecutor(vm._Executor):
  def __init__(self,
...
  def _MakeProcess(self, node, inherit_errexit=True):
  def RunBuiltin(self, builtin_id, cmd_val):
  def RunSimpleCommand(self, cmd_val, cmd_st, do_fork, call_procs=True):
  def RunBackgroundJob(self, node):
  def RunPipeline(self, node, status_out):
  def RunSubshell(self, node):
  def RunCommandSub(self, cs_part):
  def RunProcessSub(self, cs_part):
  def MaybeWaitOnProcessSubs(self, frame, compound_st):
  def PushRedirects(self, redirects):
  def PopRedirects(self):
  def PushProcessSub(self):
  def PopProcessSub(self, compound_st):

```


There are only two ways we make a process object

1. `RunSimpleCommand()` creates `process.Process()` directly
2. CommandSub, Subshell, etc. call `self._MakeProcess()`

And there are two ways that we use pipeline:

1. `RunPipeline()` synchronously
2. `RunBackgroundJob()` can be a pipeline


So I want to say that we can just use the state changes at this level, and that `if job_state.JobControlEnabled()` should only be at this level too

But I'm NOT confident of that yet

I was poking at the code more to understand it

andychu:
Oops, sorry that last change actually broke the `bgpipe` test case

http://travis-ci.oilshell.org/github-jobs/2023-03-14__03-53-18.wwz/_tmp/job-control/index.html

For some reason I thought background pipelines don't do the PGID thing since they don't get signals

I will revert it

melvinw:
> For some reason I thought background pipelines don't do the PGID thing since they don't get signals

Ah yeah. I think it would be nice to keep those grouped even if they don't need to receive TTY signals. Users might want to do things like `kill -<group_id>`

core/process.py
melvinw: Hmm there's actually a trade-off here. By removing `fg_pipeline` we would muddy the logic in `RunPipeline()` somewhat. It would have to conditionally start the last part of the pipeline first if it would fork or run it after starting the other parts if it's just a builtin, etc... Awkward logic is probably better than extra state, though. Let's see how it ends up looking....

core/process.py
```diff
@@ -1202,23 +1198,16 @@ def JobWait(self, waiter):
...
-    self.StartPipeline(waiter)
+    assert len(self.pids) == len(self.procs)
 
-    # QUESTION: I think we can get rid of this if we make sure that a Pipeline
-    # always have a non-zero number of processes
-    if self.group_id == -1:
-      self.group_id = self.job_state.shell_pgrp
-
-    self.job_state.MaybeGiveTerminal(self.group_id)
+    self.job_state.MaybeGiveTerminal(posix.getpgid(self.pids[0]))
```
melvinw, line 1210: this is the cost of dropping `Pipeline.group_id`. I think that's fine, though

andychu:
Thank you, this is excellent!

That was definitely not going to get done unless you did it :)

melvinw:
No sweat! 👍

Issue #1513, opened 2023-02-22
DATAROOTDIR fallback to $FLAG_prefix/share happens before $FLAG_prefix is set by arguments


pdebelak:
This means that you need to specify both datarootdir and prefix if you don't like `/usr/local` as the prefix. The fallback happens at the top level of the script, but it should happen inside main after `parse_flags` is called.

andychu:
Oops,  I think what happened is that I factored the flag processing into a function and didn't realize there was an order dependency

Gah!

PR #1514, opened 2023-02-22
Fix --readline handling and datarootdir fallback

### LINKED ISSUE #1467: configure error when trying to specify readline w/ `--readline`
Went to update the Nix expression for Oil to try to build the oils-for-unix tarball and noticed this configure failure:

```console
configuring
configure flags: --prefix=/nix/store/9jl9anqvyp4nnjbvba703g8izlcqzsrf-oil-0.14.0 --datarootdir=/nix/store/9jl9anqvyp4nnjbvba703g8izlcqzsrf-oil-0.14.0 --with-readline --readline=/nix/store/5rvfmpyiic6w2gzw91gk10x4vr9yi9bv-readline-all
./configure ERROR: readline was not detected on the system (--with-readline passed).
```

It looks like it cropped up around 0.13.0, since the main nixpkgs update bot [started failing with a similar message on that update](https://r.ryantm.com/log/oil/2023-01-01.log)

I noticed that this section of the configure added in december is similar to the next one below, sans returns:
https://github.com/oilshell/oil/blob/623a893eeebcd124ca00ae5c84dbf9fadfaac8c8/configure#L364-L375

https://github.com/oilshell/oil/blob/623a893eeebcd124ca00ae5c84dbf9fadfaac8c8/configure#L203-L218

~patching it with `substituteInPlace configure --replace "echo '#define HAVE_READLINE 1'" "echo '#define HAVE_READLINE 1' && return 0"` produces a usable executable, at least. 

Associated Nixpkgs PR: https://github.com/NixOS/nixpkgs/pull/212891

pdebelak:
Fixes #1467 and #1513

### DIFF
--- a/configure
+++ b/configure
@@ -197,38 +197,13 @@ cc_header_file() {
   cc_statement "$pp_var" 'return 0;' "#include <$c_lib>"
 }
 
-# Write a shell script to standard out with variables, or fail.
-detect_readline() {
-  if test -n "$FLAG_readline"; then
-    if cc_quiet build/detect-readline.c \
-	-L "$FLAG_readline/lib" \
-	-I "$FLAG_readline/include" \
-	-l readline; then
-      echo "READLINE_DIR=$FLAG_readline"
-      echo 'HAVE_READLINE=1'
-      return 0
-    fi
-  else
-    if cc_quiet build/detect-readline.c -l readline; then
-      echo "READLINE_DIR="
-      echo 'HAVE_READLINE=1'
-      return 0
-    fi
-  fi
-
-  if test "$FLAG_with_readline" = 1; then
-    die 'readline was not detected on the system (--with-readline passed).'
-  fi
-  echo 'READLINE_DIR='
-  echo 'HAVE_READLINE='
-  return 1
-}
-
 detect_and_echo_vars() {
   if test "$FLAG_without_readline" = 1; then
     echo 'HAVE_READLINE='
   else
-    detect_readline
+    # readline already confirmed to exist by detect_cpp
+    echo "READLINE_DIR=$FLAG_readline"
+    echo 'HAVE_READLINE=1'
   fi
   echo "PREFIX=$FLAG_prefix"
   echo "DATAROOTDIR=$FLAG_datarootdir"
@@ -372,18 +347,22 @@ detect_cpp() {
           -I "$FLAG_readline/include" \
           -l readline; then
         echo '#define HAVE_READLINE 1'
-        return 0
+        return
       fi
     else
       if cc_quiet build/detect-readline.c -l readline; then
         echo '#define HAVE_READLINE 1'
-        return 0
+        return
       fi
     fi
 
     if test "$FLAG_with_readline" = 1; then
       die 'readline was not detected on the system (--with-readline passed).'
     fi
+
+    echo '#define HAVE_READLINE 0'
+    # set flag to indicate readline not found
+    FLAG_without_readline=1
   fi
 }
 
--- a/configure
+++ b/configure
@@ -149,11 +149,10 @@ parse_flags() {
     esac
     shift
   done
+
+  # If not set, fallback to --prefix
+  FLAG_datarootdir=${FLAG_datarootdir:-$FLAG_prefix/share}
 }
-
-# If not set, fallback to --prefix
-FLAG_datarootdir=${FLAG_datarootdir:-$FLAG_prefix/share}
-
 
 # No output file, no logging, no stderr.
 # TODO: Maybe send stdout/stderr to config.log?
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -367,23 +372,19 @@ detect_cpp() {
           -I "$FLAG_readline/include" \
           -l readline; then
         echo '#define HAVE_READLINE 1'
+        return 0
```
andychu, line 375: Thanks for fixing this!

Looks like there are a lot of problems in this script -- I took a look at the git history 

-----

Are you up for a bigger refactoring?  You mentioned the problem with the duplication between with `detect_readline` and `detect_cpp`, and I agree

I think the problem is that there should be a SINGLE detect_readine that uses the EARLY return style.  Right now the styles are mixed, whic is probably why the code is buggy

Also the exit code doesn't seem to mean anything, so it should just `return`

I think it should look like

```
if test $FLAG_without_readline ...
  echo ...
  return

if test $FLAG_readline ...
  if cc_quiet ...
     echo ...
  return 

if cc_quiet ...
  echo ...
  return

if test $FLAG_with_readline ...
  die ...

echo 'readline not detected'

```


It's kind of annoying to have all these flags, but AFAIK they were based on what GNU autoconf has ??  Open to suggestions


I'm OK with a bigger change because we're focused on  making the C++ tarball work nicely now

I can merge this as is since it looks like a good fix, but I think we should eventually have the refactoring

andychu: Or actually there is a slight bug in that logic, because it can return before printing the "not detected" case!  

I guess `detect_readline` could be defined to print success or die, and then another function could print the "not detected" case

There is a unit test in `./configure-test.sh` that could possibly help with this

pdebelak: I did some refactoring to not check for readline twice, but it required setting a flag inside `detect_cpp` that modifies how `detect_and_echo_vars` acts. The code is written in a way that it echos values to be written to a file and needs to handle multiple cases: `without-readline` passed, `with-readline` passed and no `readline` flag, `with-readline` passed with a `readline` flag, and `readline` passed without `with-readline`.

andychu:
Thanks for adding all the tests!  I pulled this and tried it out

For some reason they're not passing on my machine.  It looks like this

```
$ ./configure-test.sh soil_run
#define SIZEOF_INT 4
#define HAVE_INT 1
./configure-test.sh ERROR: Expected detect_cpp to fail, but succeeded
```


```
++ test '\'''\'' = 1
++ test -n /path/to/fake/readline
++ cc_quiet build/detect-readline.c -L /path/to/fake/readline/lib -I /path/to/fake/readline/include -l readline
++ cc build/detect-readline.c -L /path/to/fake/readline/lib -I /path/to/fake/readline/include -l readline -o /dev/null
++ echo '\''#define HAVE_READLINE 1'\''
#define HAVE_READLINE 1
++ return'
+ test 0 = 0
+ die 'Expected detect_cpp to fail, but succeeded'
+ echo './configure-test.sh ERROR: Expected detect_cpp to fail, but succeeded'
./configure-test.sh ERROR: Expected detect_cpp to fail, but succeeded
+ exit 1
andy@lenny:~/git/oilshell/oil$ 
andy@lenny:~/git/oilshell/oil$ 
andy@lenny:~/git/oilshell/oil$ cc build/detect-readline.c -L /path/to/fake/readline/lib -I /path/to/fake/readline/include -l readline -o /dev/null; echo $?
0

```


It appears that my `cc` compiler doesn't fail with the fake paths?   I think it might have some other way to find the library?   

Does yours fail?

I'm on Ubuntu Xenial 18.04 with gcc 7.5

```
$ cc --version
cc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```

andychu:
Thinking aloud: another way to remove the duplication that could be easier to test is

```
# mutable globals
have_readline=
readline_dir=

detect_readline() {
  # sets have_readline and readline_dir  
}
```

then

```
echo_readline_cpp  # reads those 2 variables

echo_readline_sh  # ditto
```

pdebelak:
@andychu - Yes, the tests pass on my machine, but I think the issue is that if `-lreadline` works by itself you can pass nonsense to the `-L` and `-I` flags because those just add to the existing search path rather than replacing it. I think it will be difficult to get reliable tests for those as a result since it depends on the state of your system. I don't see a way to tell gcc to not use the default search path, but maybe I'm just missing it. Regardless, I'll refactor to detect readline in one place and then use the variables set by that in the other functions.

pdebelak:
@andychu - Thanks for the refactoring suggestion, now `detect_readline` just detects and sets variables (or dies). Then the downstream function can read the values from the global variables and trust they are correct. Also added a pre-condition check to the `echo_*` functions that `detect_readline` has been called.

andychu:
Looks great, thank you!

I think the CI should be happy with this -- if not I'll massage it


(Unfortunately Github doesn't have a way to approve passing our secrets from PRs coming outside the main repo fork -- https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Allow.20Contributors.20to.20Publish.20CI.20Results )


I hope to make a release this week!

PR #1516, opened 2023-02-25
Move clang-format config to .clang-format

watk:
This makes it easier to use clang-format in other ways (e.g., `git clang-format`) or via an editor.


andychu:
Great thanks!

PR #1517, opened 2023-02-25
Refactor GC object header initialization

watk:
Refactor GC objects to each now contain a static `obj_header` function that returns the the ObjHeader it should be initialized with. Now, instead of directly intiializing `header_` via the GC_* macros, we initialize it from obj_header(). This should be no different semantically -- it's step one in the transformation to move the ObjHeader outside of objects and have `Alloc()` initialize them.

@andychu  LMK if this is too hard to review and I can try to split it up for you. The commits are pretty self contained so you can also look at the diff per commit maybe.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,12 @@
+BasedOnStyle: Google
+# We have a lot of switch statements, and the extra indent doesn't help.
+IndentCaseLabels: false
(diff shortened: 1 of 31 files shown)


andychu:
This is great!  Thank you

Glad you found the `prebuilt/` stuff !  That is a side effect of our decision to have one big translation unit for `oils_for_unix.mycpp.cc` :-/

PR #1518, opened 2023-02-26
Move object headers outside of objects

watk:
This refactors GC managed objects so that their headers are external objects
instead of member variables. The headers are placed directly before their GC
objects in memory, so it's trivial to jump between them. The physical layout is
maintained by Alloc<T>() (and related allocation functions) and the GcGlobal<T>
wrapper class. Alloc<T>(), NewSlab(), etc, now allocate space for the headers
and initialize them directly with T::obj_header(). And global GC objects are
wrapped in GcGlobal which has an embedded header.

Notable changes:
 * FindObjHeader() is gone: Use ObjHeader::FromObject(obj) and
   header->ObjectAddress() to get between header and object.
 * Derived classes no longer modify the mutate their field mask in the
   constructor. Now each class in a hierarchy is required to have a
   field_mask() function that returns the union of its base's field_mask() (if
   any) and its own.

Followup changes to be done:
 * Minor optimization: Change MarkSweepHeap to keep lists of ObjHeader* instead
   of RawObject* to reduce calls to ObjHeader::FromObj()l.
 * s/maskbit_v/maskbit/: maskbit_v does nothing now
 * Remove ObjHeader::is_header: it's no longer necessary

### DIFF
--- a/asdl/gen_cpp_test.cc
+++ b/asdl/gen_cpp_test.cc
@@ -179,11 +179,11 @@ using typed_demo_asdl::flag_type;
 using typed_demo_asdl::SetToArg_;
 
 // TODO: We should always use these, rather than 'new flag_type::Bool()'
-flag_type::Bool g_ft = {};
+InlineGcObj<flag_type::Bool> g_ft = {flag_type::Bool::obj_header()};
 
 // Use __ style
 using typed_demo_asdl::cflow__Return;
-cflow__Return g_ret = {5};
+InlineGcObj<cflow__Return> g_ret = {cflow__Return::obj_header(), {5}};
 
 int i0 = 7;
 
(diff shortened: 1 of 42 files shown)


asdl/gen_cpp_test.cc
```diff
@@ -179,11 +179,11 @@ using typed_demo_asdl::flag_type;
 using typed_demo_asdl::SetToArg_;
 
 // TODO: We should always use these, rather than 'new flag_type::Bool()'
-flag_type::Bool g_ft = {};
+InlineGcObj<flag_type::Bool> g_ft = {flag_type::Bool::obj_header()};
 
 // Use __ style
 using typed_demo_asdl::cflow__Return;
-cflow__Return g_ret = {5};
+InlineGcObj<cflow__Return> g_ret = {cflow__Return::obj_header(), {5}};
```
watk, line 186: TODO: make these HeapTag::Global

mycpp/gc_alloc.h
```diff
@@ -85,23 +85,17 @@ class StackRoots {
 // https://eli.thegreenplace.net/2014/variadic-templates-in-c/
 template <typename T, typename... Args>
 T* Alloc(Args&&... args) {
+  // The objects are placed directly after the headers (i.e., at an address
```
watk, line 88: TODO make this clearer

mycpp/gc_obj.h
```diff
@@ -58,6 +59,23 @@ struct ObjHeader {
   unsigned obj_len : 29;     // Cheney: number of bytes to copy
 #endif
 
+  // Returns the address of the GC managed object associated with this header.
+  void* ObjectAddress() {
+    return reinterpret_cast<void*>(reinterpret_cast<char*>(this) +
+                                   sizeof(ObjHeader));
+  }
+
+  // Returns the header for the given GC managed object.
```
watk, line 68: TODO comment that we guarantee this with static assert elsewhere

cpp/osh.cc
```diff
@@ -16,7 +16,8 @@ using syntax_asdl::loc;
 
 namespace arith_parse {
 
-tdop::ParserSpec kArithSpec;
+InlineGcObj<tdop::ParserSpec> kArithSpec_ = {tdop::ParserSpec::obj_header().SetIsGlobal()};
```
watk, line 19: TODO brace init this

mycpp/cppgen_pass.py
```diff
@@ -2262,26 +2262,25 @@ def visit_class_def(self, o: 'mypy.nodes.ClassDef') -> T:
             else:
                 # Has inheritance
 
+                # The field mask of a derived class extends that of its base.
```
watk, line 2265: union

PR #1526, opened 2023-03-17
spec: Don't use ## for non-metadata comments.

xiaq:
I noticed that there are some non-metadata comments starting with `##` when writing my own parser of Oil's spec tests [1]. There are just a few of them, so I assume the use of `##` is accidental; changing them to `#` makes the spec files just a little bit easier to parse.

There is another place that is slightly weird (not changed in this PR):

https://github.com/oilshell/oil/blob/ed3f591537ffd754dc5770a8ad310544147cd683/spec/arith.test.sh#L362

This is the only metadata line that starts directly with a shell name, without `OK` / `BUG` / `N-I`. Reading the comment for this test case, 9 is the expected output on all shells, so presumably the `osh` part can just be removed?

[1] It's in https://github.com/elves/posixsh/blob/master/pkg/spec/oil_specs_test.go if you're interested; the project name `posixsh` is a lie since you can't implement a fully POSIX-compliant shell in Go, it's just my experiment to see _how_ compliant I can get.

### DIFF
--- a/spec/append.test.sh
+++ b/spec/append.test.sh
@@ -34,7 +34,7 @@ t=foofoo
 
 #### Append to array to undefined variable
 
-## TODO: strict_array could get rid of this?
+# TODO: strict_array could get rid of this?
 y+=(c d)
 argv.py "${y[@]}"
 ## STDOUT:
--- a/spec/bugs.test.sh
+++ b/spec/bugs.test.sh
@@ -128,7 +128,7 @@ echo x=${x@P}
 x=\D{%H:%M
 ## END
 
-## bash just ignores the missing }
+# bash just ignores the missing }
 ## BUG bash stdout-json: ""
 
 # These shells don't understand @P
--- a/spec/if_.test.sh
+++ b/spec/if_.test.sh
@@ -42,7 +42,7 @@ fi
 
 #### if break corner case
 
-## This is analogous to the 'while' case in spec/loop
+# This is analogous to the 'while' case in spec/loop
 f() {
   if break; then
     echo hi


andychu:
Great thank you!

BTW we also have a huge array of other test cases, in somewhat sloppy format.  I just typed all the errors I could think of over the years, and made sure they kept passing :)

https://github.com/oilshell/oil/blob/master/test/parse-errors.sh

http://travis-ci.oilshell.org/github-jobs/2023-03-15__07-06-56.wwz/_tmp/soil/logs/parse-errors.txt

I brainstormed some common table-based testing framework, but so far have gotten by without it.

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Table-Based.20Testing.20and.20Benchmarks

---

There are many more comparisons like job control and syscalls:


http://travis-ci.oilshell.org/github-jobs/2023-03-15__07-09-17.wwz/_tmp/syscall/by-code.txt

http://travis-ci.oilshell.org/github-jobs/2023-03-15__07-10-07.wwz/_tmp/job-control/index.html


It's probably a bit further in the future, but it would be nice to unify it all into a big test suite

It should output TSV files and then make an HTML report probably.  The spec test framework is a little "hard-coded".

---

BTW you should be able to use them without any change, like this:

Just add your shell as an argument.

```
andy@lenny:~/git/oilshell/oil$ test/spec.sh smoke yash
smoke.test.sh
case    line    dash    bash    mksh    osh     yash
  0       4     pass    pass    pass    pass    pass    builtin
  1       8     pass    pass    pass    pass    pass    command sub
  2      12     pass    pass    pass    pass    pass    command sub with builtin
  3      16     pass    pass    pass    pass    pass    pipeline
  4      20     pass    pass    pass    pass    pass    pipeline with builtin
```


andychu:
Oh also if you want a shell parser, not just tests, I've been talking with @abathur about exporting the Oil AST in some format.  This is what you get when you do `osh -n myscript.sh`.   We want to stabilize this format, and make Oil a definitive shell parser.

He's been using it for awhile in the "resholve" tool for packaging shell.

----


A related option is to link C++ into Go, which I think is fine sine the parser doesn't any I/O.

We have a nascent Python main file, which translates to a PURE C++ file that ONLY includes the parser.  It really only depends on is the garbage collector `gHeap`, and that can probably be stubbed out to an arena allocator.

https://github.com/oilshell/oil/blob/master/bin/osh_parse.py

So you're definitely welcome to link it into Elvish or whatever ... It will be a bit of work to massage it, but probably less work than writing a shell parser from scratch :)

The C++ translation tools and build tools are much improved over the last few months

xiaq:
Yeah I've seen that Oil has much more tests, the jobs and syscalls ones are pretty cool :)

I want something tightly integrated with Go (notably generating test coverage for Go code) though and the spec files are easy enough to parse.

What do you think of https://github.com/oilshell/oil/blob/ed3f591537ffd754dc5770a8ad310544147cd683/spec/arith.test.sh#L362? Should the `osh` part be dropped?

andychu:
Also if the goal is Go coverage support, re2c now apparently has Go support.  (It didn't when Oil started using it!)

So at the very least, you can reuse our lexer and list of token IDs.  I published them here specifically for people to port Oil:

https://www.oilshell.org/release/0.14.2/quality.html#source-code

- https://www.oilshell.org/release/0.14.2/source-code.wwz/frontend/lexer_def.py
- https://www.oilshell.org/release/0.14.2/source-code.wwz/_build/tmp/frontend/match.re2c.txt  -- We can make a file like this with Go in it
- https://www.oilshell.org/release/0.14.2/source-code.wwz/_gen/frontend/match.re2c.h

-----

As a data point, Rob Landley wrote a good portion of busybox and was primary maintainer, so he knows more about Linux and shell than 99.99% of programmers  (Used on Alpine Linux, millions of embedded Linux devices, etc.)

I worked on the toybox project exactly 7 years ago in 2016, and he said then had been working on / figuring out a shell since 2007.

He's still working on parsing shell in 2023, e.g. with these posts with respect to here docs:

http://landley.net/notes.html

So that's why I say that reusing the lexer is a big leg up :)  One of the bugs in this last release was a simple lexer change.

(Landley is approaching it from an embedded systems C programmer mindset ... I advised him to think of it from a grammar mindset, to avoid combinatorial state space explosions, but he wasn't  receptive to that advice ... )

PR #1529, opened 2023-03-19
refactor int -> loc_t in core/ui.py

CoffeeTableEspresso:


core/ui.py
```diff
@@ -199,14 +199,15 @@ def GetLineSourceString(arena, line, quote_filename=False):
   return s
 
 
-def _PrintWithSpanId(prefix, msg, span_id, arena, show_code):
-  # type: (str, str, int, Arena, bool) -> None
+def _PrintWithSpanId(prefix, msg, blame_loc, arena, show_code):
```
andychu, line 202: Can you rename this `_PrintWithLocation()` ?

core/ui.py
```diff
@@ -307,22 +308,19 @@ def CurrentLocation(self):
...
-    # type: (str, int) -> None
+    _PrintWithSpanId(prefix, msg, blame_loc, self.arena, show_code=True)
+
+  def Print_(self, msg, blame_loc=None):
+    # type: (str, loc_t) -> None
+    if not blame_loc:
+        blame_loc = loc.Missing()
+    _PrintWithSpanId('', msg, blame_loc, self.arena, show_code=True)
+
+  def PrintMessage(self, msg):
```
andychu, line 319: Hm interesting, looks like we do not have any PrintMessage() calls with explicit locations. 

But I think we should make this the same as `Print_` -- it will accept a  location, and fall back  on the current one if missing

Another step, doesn't need to be in this PR: I want to make blame_loc REQUIRED, so then we can grep for `loc.Missing()` to find the actual missing locations

We should probably go "wide" before we go "deep" into doing that

andychu (commented):
Looks great, thank you very much!

2 small issues

core/ui.py
CoffeeTableEspresso: Done

PR #1531, opened 2023-03-22
[oil_lang,pgen2] emit c++ grammar tables

melvinw:
The next PR will fill out the runtime in `cpp/pgen2.{h,cc}` to use this.


pgen2/grammar.py
```diff
@@ -171,6 +171,72 @@ def dump_nonterminals_cpp(self, f):
...
+              src_f.write('                std::vector<arc>({\n')
+              for a, b in arcs:
+                src_f.write('                    {%d, %d},\n' % (a, b))
+
+              src_f.write('                }),\n')
+              src_f.write('                0, 0, nullptr, 0,\n')
+              src_f.write('            },\n')
+
+            src_f.write('        },\n')
+            src_f.write('        bitset_t{"%s"},\n' % make_bitset(first))
```
melvinw, line 224: `first_t` is basically a bitset, right? the corresponding field in cpython's grammar.h is

melvinw:
`_gen/oil_lang/grammar_tables.cc` looks like this:
```
grammar OilGrammar = {
  std::vector<dfa>{
    dfa{
        256,
        "augassign",
        0,
        std::vector<state>{
            state{
                std::vector<arc>({
                    {1, 1},
                    {2, 1},
                    {3, 1},
                    {4, 1},
                    {5, 1},
                    {6, 1},
                    {7, 1},
                    {8, 1},
                    {9, 1},
                    {10, 1},
                }),
                0, 0, nullptr, 0,
            },
            state{
                std::vector<arc>({
                    {0, 1},
                }),
                0, 0, nullptr, 0,
            },
        },
        bitset_t{"0111111111100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"},
      },
      ...
  },
  std::vector<label>{
      {98, "end_stmt"},
      {69, "dq_string"},
      {146, "tea_enum"},
      {171, "repeat_range"},
      {159, "proc_param"},
      {152, "proc_params"},
      {78, "tea_func"},
      ...
  },
  256,
  0,
};
```

pgen2/grammar.py
andychu: I actually don't know but it seems very plausible :)

andychu (commented):
This generally looks good

The only thing I wonder is if all the std::string and std::vector cause a lot of code to be run before main()?  That was my impression of how the STL containers work

Generally speaking I've avoided any constructors at the global level, e.g. the GLOBAL_STR and GLOBAL_LIST macros all end up `{ }` initialization of structs, which take up zero runtime.   It just becomes plain old data in the linker

I think there is probably a way to check but I don't know off hand ... I think there is a symbol for the code before `main()`, and we can maybe check how big it is

---

Anyway I'm fine with proceeding because the hard part is probably to get the runtime working against these tables

This looks pretty straightforward, so I wonder if it will "just work"  ... I thought it was going to be more complicated when we had the call, e.g. dealing with the lexer, but now it looks like it might ?

andychu:
LOL I asked chatgpt about it

gpt-string-vector

andychu:
I do NOT know if that's correct

But on the other hand Google feels a bit useless.  I feel like it could have done this 10 years ago



google-string-vector

andychu:
I followed ChatGPT instructions, not sure if this is the right spot, but definitely seems related

I guess what I'm going to do is merge this PR and see how this changes :)  An honest experiment



```
andy@lenny:~/git/oilshell/oil$ objdump -t _bin/cxx-dbg/osh  |grep __init_array
00000000003d1198 l       .init_array    0000000000000000              __init_array_end
00000000003d1168 l       .init_array    0000000000000000              __init_array_start

$ objdump -s --section=.init_array _bin/cxx-dbg/osh

_bin/cxx-dbg/osh:     file format elf64-x86-64

Contents of section .init_array:
 3d1168 d0920100 00000000 16c20e00 00000000  ................
 3d1178 80051000 00000000 09631000 00000000  .........c......
 3d1188 cb721400 00000000 f5b01400 00000000  .r..............


```

andychu:
OK well I looked around a little, and either:

- there are no global constructors being called before main()
- there are but I don't know how to find them and ChatGPT did not help :)


I used some commands it suggested (and also Googled) but I don't see a real difference

```
$ readelf -s _bin/cxx-dbg/osh | grep __cxx  | tee _tmp/before-grammar.txt

$ wc -l _tmp/before-grammar.txt 
309 _tmp/before-grammar.txt

$ wc -l _tmp/with-grammar.txt 
311 _tmp/with-grammar.txt

```

So I guess let's just move on with our lives for now :)

melvinw:
Ah OK 😅

If this becomes a problem we can always change the interface from
```
extern grammar OilGrammar;
```
to
```
grammar* GetOilGrammar();
```

PR #1534, opened 2023-03-24
[build] fix nix-shell 'clean' command

Melkor333:
A small fix to `shell.nix` :)


andychu:
Thank you!  Glad to know this works still.  I thought we might have broken it!

FYI I'm pretty sure @abathur added it initially

PR #1535, opened 2023-03-24
Draft: [core/shell] Load rc files from a {lang}rc.d folder

Melkor333:
Started some very basic hacking. Let's see where we get :)


andychu:
This is great, let's kick off this discussion!

The way I'm thinking about it, this relates to the "Neighborly Shell" idea from @abathur .  Or at least I like that name and I want to think about it that way

To me the problem is **GLOBAL** hooks and **ORDERING**

This patch does it in alphabetical order of `os.listdir()` -- is that enough?


What are a list of things that get stomped on?

- `$PATH`
- `$PS1`
- `$PROMPT_COMMAND`
- `trap DEBUG` -- executes before every command

I wonder if we can do something like `--append` and `--prepend` ?  And let the user control it?

```
trap --prepend DEBUG 'foo'
trap --prepend PROMPT_COMMAND 'foo'   # equivalent to $PROMPT_COMMAND
```

and maybe even

```
append (PATH, 'foo')
append (PS1, 'bar')
```

Not sure if there's any use for that

In any case I want to kick off a design dicussion about this

andychu:
### Could we do this at the "user level"?

I wonder if we can ahve

```
# ~/.config/oils/yshrc

for file in $_this_dir/*; do
  source $file
done


```


and then maybe something like

```
```
# ~/.config/oils/yshrc

for file in $_this_dir/*; do
  source --protect-global-traps $file
done
```

If you do `--protect-global-traps`, then it tries to prevent things from stepping on each other

Which MIGHT be very unrealistic

I'm not sure if anyone wants it

But it's just an idea

**Comment by andychu** [comment]
I guess the problem is that this relies on the user to understand and change shell

I have a feeling all the `$(curl | bash)` and mutating `bashrc` magic is just because most people don't know how to do that (including me for a long time !)

File: core/shell.py [hunk 1]
```diff
@@ -548,8 +548,28 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
   assert home_dir is not None
   rc_path = flag.rcfile
   # mycpp: rewrite of or
+
+  rc_paths = []
+  # TODO: Decide if this should be first or last rc_path (as in: should it overwrite modular includes or not)
```
andychu, line 553: I'm thinking it should be the last one, so the user can REMOVE stuff that  nix and starship added if they want?  Without editing their code

This is related to the "Neighborly Namespacing Idea" and `trap --add` and `trap --remove`


https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Neighborly.20Namespacing.20Idea

core/shell.py
```diff
@@ -548,8 +548,28 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
...
   if rc_path is None:
-    rc_path = os_path.join(home_dir, '.config/oil/%src' % lang)
+    rc_paths.append(os_path.join(home_dir, '.config/oil/%src' % lang))
+  else:
+    rc_paths.append(rc_path)
+
+  # Experimentally try to load all files in ~/.config/oil/oshrc.d | ~/.config/oil/oilrc.d
+  # TODO: Decide on proper directory, `oil/oshrcd.d` seems ugly
```
andychu, line 560: Yeah I agree it's a little odd, but it's consistent

I think it will be

```
~/config/oils/
  oshrc
  oshrc.d
```

`--rcfile`

`--rcdir`

?

core/shell.py
```diff
@@ -548,8 +548,28 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
...
+  # TODO: Add a flag/parameter to manually pass/disable reading the rc_dir
```
andychu, line 561: Yes, see `frontend/flag_def.py` 

Please add `--rcdir`, which is like `--rcfile`

This will help us test it!

andychu (commented):
This looks great!  Thank you

I did a bunch of brainstorming on Zulip, and I'm pretty convinced about this now

I think these features will work together to provide something "neighborly"

- `--rcdir`
- `trap --add` and `trap --remove`
  - instead of "clobber" behavior, we will have multiple traps like DEBUG and PROMPT_COMMAND

andychu:
Actually I'm thinking the user RC file should be the first one, and then we add an EVENT `after-rcdir` that lets the user run code there?  LIke this

```
trap --event after-rcdir {
  echo 'do other stuff'
}
```

core/shell.py
Melkor333: > Actually I'm thinking the user RC file should be the first one, and then we add an EVENT `after-rcdir` that lets the user run code there? LIke this
> 
> ```
> trap --event after-rcdir {
>   echo 'do other stuff'
> }
> ```
> 
> That can be done later, but I think giving the user the "first shot" makes sense

Yeah exactly. That would also allow easy "preconfiguration" for plugins. Plus it would theoretically be possible to change things at the end with a file like `ZZZ_replace_xyz.sh` (though a `trap --event after-rcdir` as you suggested is much cleaner!).

core/shell.py
Melkor333: It was the first thing I came up with. We can also brainstorm on a better naming (for the flag and the directory) over on zulip :)

Melkor333:
> # This flag has is named like bash's equivalent.  We got rid of --norc because

Found this in `frontend/flag_def.py`. It might make sense to "revive" the `--norc` option, what do you think?

andychu:
This looks great, including tests and docs, thank you!

About `--norc`, I'm not opposed to it ...  I guess it's a little cleaner than `--rcdir='' --rcfile=/dev/null` which is perhaps a little verbose

----

Unfortuantely we can't run the CI  with Github Actions secrets in your fork, it has to be from the `oilshell/oil` repo

That's a limitation of Github itself (and something to document)

I wonder if there's  a way to make a new PR based on this one, let me check it out

andychu:
Well good news is that Github is letting me push to your repo and branch :)

I noticed that the output of `os.listdir()` isn't deterministic

That's just how the Unix file system works

I think we should change is to use `glob()`, which is also good for C++ translation.  Let me try that



```
case: 3
[osh stdout] Expected '1\nRCFILE1\nRCFILE2\n2\n', got '1\nRCFILE2\nRCFILE1\n2\n'
osh stdout:
1
RCFILE2
RCFILE1
2


```

andychu:
Cool I changed it to use glob(), and added a test for when `--rcdir` doesn't exist

Merged, thank you!

Maybe we can do another tiny PR for `--norc` ?

PR #1537, opened 2023-03-26
[cpp] Use uint32_t to represent field masks

watk:
In various places (mainly the field_mask() functions) we previously represented field masks with uint16_t's even though the ObjHeader allocates 24 bits for them. Now we use uint32_t instead.

I'm motivated to do this now because my upcoming change to move the object header outside of objects results in one type requiring 17 bits for its field mask (the field masks get one bit longer).

Note: The only manual change here was increasing `kFieldMaskBits` to 24. The rest is: `s/uint16_t field_mask/uint32_t field_mask/`

### DIFF
--- a/core/optview_gen.py
+++ b/core/optview_gen.py
@@ -54,7 +54,7 @@ class _View {
   List<bool>* opt0_array;
   List<List<bool>*>* opt_stacks;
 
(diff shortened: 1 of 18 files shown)


andychu:
Makes sense, thank you!

PR #1539, opened 2023-03-27
Add a .git-blame-ignore-revs

watk:
And initialize it with a python code formatting change.

To configure git locally:

    git config blame.ignoreRevsFile .git-blame-ignore-revs

For more info see:
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# [mycpp cleanup] Reformat code with yapf auto-formatter
+c2d68ec41ae1d6a5f9f40645c9bc3152e95d8afc


watk:
Hey Andy, not sure if you're the type to notice or care what the blame looks like, but I noticed it got less useful after the yapf change the other day so I thought this would be nice

andychu:
Great, thank you!  I didn't know about this, I'll be interested to see how this is handled

https://github.com/oilshell/oil/blame/master/mycpp/cppgen_pass.py

I've been sort of "shy" to autoformat the rest of the code because of this issue

I did `asdl/`, and a couple files in `core/`

So far I think yapf is working well, and uniformly improving things

----

So maybe it's better to "rip off the band-aid" with one commit?  We'll see

andychu:
Wow this works very well!  For example

https://github.com/oilshell/oil/blame/master/mycpp/cppgen_pass.py

points at this old commit

https://github.com/oilshell/oil/commit/8135878cf1dcb33fa795eee05fa2e40185af4191

Not the formatting commit!!  Very cool, thank you

----

So yes I guess maybe the right strategy is to "rip the band-aid off" and not do incremental reformatting 

I wanted to test if `yapf` made good decisions, and now I think it does

---

I wonder if we should also switch to 4 space indents everywhere ?  I used 2 out of habit from Google, but most Python open source is 4.

watk:
Awesome! Yeah it's a very neat little feature. I like auto formatted code
too :)

PR #1540, opened 2023-03-29
[pgen2] make parser translate-able

melvinw:
This gets osh-cpp to match python for nearly all of `oil_lang/testdata/*.osh`!
```shell
melvin@debian:~/oil$ _bin/cxx-dbg/osh -n oil_lang/testdata/sigil-pairs.sh > cpp.out
melvin@debian:~/oil$ bin/osh --ast-format text -n oil_lang/testdata/sigil-pairs.sh > py.out
melvin@debian:~/oil$ diff py.out cpp.out
melvin@debian:~/oil$
```

`oil_lang/testdata/assign.osh` fails with a syntax eror atm because we still need to get the expression evaluator translated (will send as a separate PR since it might be large). `test/parse-error.sh` should pass too after translating that and the oil builtins.


cpp/grammar.h
melvinw, line 1: Whoops. Meant to remove this file.

oil_lang/run.sh
```diff
@@ -34,7 +34,7 @@ DISABLED-osh-cpp() {
 
   ninja $osh
 
-  OSH=$osh test-parse-osh
+  ASAN_OPTIONS=detect_leaks=0 OIL_GC_ON_EXIT=1 OSH=$osh test-parse-osh
```
melvinw, line 37: Only need one of these...

cpp/pgen2.h
```diff
@@ -3,73 +3,59 @@
...
+typedef Tuple2<int, int> arc_t;
+typedef Dict<int, int> first_t;
+typedef List<List<arc_t*>*> states_t;
+typedef Tuple2<states_t*, first_t*> dfa_t;
+
+class Grammar {
+ public:
+  Grammar();
 
   GC_OBJ(header_);
```
andychu, line 41: Oh BTW Chris has a change out to put the header before the object

So depending on which one lands first, we'll have to resolve, which I can help with

pgen2/pnode.py
```diff
@@ -27,3 +27,7 @@ def __repr__(self):
     ch_str = 'with %d children' % len(self.children) \
         if self.children is not None else ''
     return '(PNode %s %s %s)' % (self.typ, tok_str, ch_str)
+
+def NewPNode(typ, tok, children):
```
andychu, line 31: Maybe write a comment about how this is rewritten in C++

I don't really understand that part

cpp/pgen2.cc
```diff
@@ -2,14 +2,23 @@
...
-  assert(0);
+PNode* NewPNode(int typ, syntax_asdl::Token* tok,
+                List<pnode::PNode*>* children) {
+  // This is only ever called with a null token from Parser::setup(). Init/reset
+  // the PNode arena in that case.
+  if (tok == nullptr) {
+    arena.reserve(1000000);
+    arena.clear();
+  }
+  DCHECK(arena.size() < arena.capacity());
```
andychu, line 19: This should be CHECK() until we make it dynamically sized, i.e. always on


I learned the hard way by running out of GC objects :)

andychu (commented):
OK cool I took a quick look!  Great that this works

I suspect we'll notice all the PNode GC objects in some cases, but we can deal with that as it comes up

One thing this might unlock is running all the YSH/Oil spec tests with C++

Even if the evaluator is stubbed out, some may run

oil_lang/run.sh
```diff
@@ -34,7 +34,7 @@ DISABLED-osh-cpp() {
 
   ninja $osh
```
andychu, line 35: This function can be renamed to `test-parse-osh-cpp` or something, so then it will run in the CI without the DISABLED prefix

pgen2/pnode.py
melvinw: Ah sure, I can add a note about this. This is here as an attempt to avoid GC pressure from all the PNodes. What confused you?

cpp/pgen2.cc
melvinw: Oof good catch

pgen2/parse.py
```diff
@@ -92,6 +92,8 @@ def __init__(self, grammar):
         up to the converter function.
         """
         self.grammar = grammar
+        self.rootnode = None  # type: Optional[PNode]
```
melvinw, line 95: oh hmmm. I just realized mycpp will generate a class that includes this in the field mask. I change the type in `cpp/pgen2.h` so it is NOT a GC type now. How will the GC react to this? 🤔 Maybe the right thing to do here is give the c++ type a header with `HeapTag::Opaque` so we never scan past the root referenced by the parser...

pgen2/pnode.py
melvinw: Ah actually I think I see it. The list of child pointers is still managed by the GC, so the arena for the PNodes themselves is kinda useless...

melvinw: I'll push a commit to just translate PNode with mycpp. Then we can profile and optimize

oil_lang/run.sh
melvinw: We'll need to filter out `oil_lang/testdata/assign.osh` I think

melvinw:
Wow, yeah `PNode` GC time is gnarly (though, I haven't measured to confirm this is actually where time is being spent). `oil_lang/test-parse-osh-cpp` takes TWICE as long to run `oil_lang/test-parse-osh`!

```shell
melvin@debian:~/oil$ time oil_lang/run.sh test-parse-osh-cpp > /dev/null
real    2.000
user    0.000
sys     0.000
melvin@debian:~/oil$ time oil_lang/run.sh test-parse-osh > /dev/null
real    1.000
user    0.000
sys     0.000
```

pgen2/pnode.py
melvinw, line 1: TODO: add some comments to explain this interface

cpp/pgen2.cc
```diff
@@ -2,14 +2,21 @@
...
-}
+namespace pnode {
+
+static std::vector<PNode> arena;
 
-bool Parser::addtoken(int typ, syntax_asdl::Token* opaque, int ilabel) {
-  assert(0);
+PNode* NewPNode(int typ, syntax_asdl::Token* tok,
+                List<pnode::PNode*>* children) {
+  if (tok == nullptr) {
```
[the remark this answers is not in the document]
andychu: One thing occurred to me -- I think it would be better to add a Python stub in `Parser::setup()` like `pgen2::InitArena()` or something.   That way it's explicit, and we're not mixing the logic with allocation info

Though another thing I thought of is that we should make `_ParseOil` free the memory

You can imagine a test like this at the INTERACTIVE shell:

```
$ var x = 42 + 1
$ var x = 42 + 1
...
... # do this a million times
```

It shouldn't leak PNodes, they can be cleaned up every time

One way to do this might be to leverage the `ctx` pattern, like

```
with ctx_PNode(...):
  e_parser.Parse(...)
```

But we could do this in a separate step, since it could be tricky

It doesn't have  to be done all at once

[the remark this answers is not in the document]
melvinw: Ah yes, pulling this out of `NewPNode()` would be much easier to understand.

We can probably pull off the freeing thing you mentioned by tracking the pnode allocator as a member of `Parser`. It looks like `ExprParser` calls `Parser::setup()` at the beginning of every call to `ExprParser::Parse()`. I'll try that out

andychu (commented):
Wow this looks great!  I like how you cleaned up the pnode.children[] interface!

andychu:
> oil_lang/testdata/assign.osh fails with a syntax eror atm because we still need to get the expression evaluator translated (will send as a separate PR since it might be large). 

OK nevermind, now I see this part

I'm not sure why it would fail just parsing but it's OK

andychu:
I made some minor comments ... So if the tests pass, and you can allow the stdout  to appear in the CI, I think this should be pretty much ready

melvinw:
Ah. I tried turning on `test/parse-errors.sh`, but it has the same issues as `oil_lang/testdata/assign.osh` (syntax errors and `KeyErrors` for disabled builtins)

melvinw:
Thanks for the comments! Pushed a few fixes. Looks like `soil-staging` might be broken? Seeing failures from CI like:
```
[1/11] cxx asan32 '' mycpp/bump_leak_heap.cc _build/obj/cxx-asan32/mycpp/bump_leak_heap.o
FAILED: _build/obj/cxx-asan32/mycpp/bump_leak_heap.o 
build/ninja-rules-cpp.sh compile_one cxx asan32 '' mycpp/bump_leak_heap.cc _build/obj/cxx-asan32/mycpp/bump_leak_heap.o _build/obj/cxx-asan32/mycpp/bump_leak_heap.o.d
In file included from /usr/lib/gcc/x86_64-linux-gnu/8/include/stdint.h:9,
                 from /home/uke/oil/mycpp/bump_leak_heap.h:6,
                 from mycpp/bump_leak_heap.cc:3:
/usr/include/stdint.h:26:10: fatal error: bits/libc-header-start.h: No such file or directory
 #include <bits/libc-header-start.h>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
[2/11] cxx asan32 '' mycpp/cheney_heap.cc _build/obj/cxx-asan32/mycpp/cheney_heap.o
FAILED: _build/obj/cxx-asan32/mycpp/cheney_heap.o 
build/ninja-rules-cpp.sh compile_one cxx asan32 '' mycpp/cheney_heap.cc _build/obj/cxx-asan32/mycpp/cheney_heap.o _build/obj/cxx-asan32/mycpp/cheney_heap.o.d
In file included from /home/uke/oil/mycpp/cheney_heap.h:9,
                 from mycpp/cheney_heap.cc:1:
/usr/include/c++/8/cassert:43:10: fatal error: bits/c++config.h: No such file or directory
 #include <bits/c++config.h>
          ^~~~~~~~~~~~~~~~~~
compilation terminated.
[3/11] cxx asan32 '' mycpp/gc_builtins.cc _build/obj/cxx-asan32/mycpp/gc_builtins.o
FAILED: _build/obj/cxx-asan32/mycpp/gc_builtins.o 
build/ninja-rules-cpp.sh compile_one cxx asan32 '' mycpp/gc_builtins.cc _build/obj/cxx-asan32/mycpp/gc_builtins.o _build/obj/cxx-asan32/mycpp/gc_builtins.o.d
In file included from /usr/include/ctype.h:25,
                 from mycpp/gc_builtins.cc:1:
/usr/include/features.h:424:12: fatal error: sys/cdefs.h: No such file or directory
 #  include <sys/cdefs.h>
            ^~~~~~~~~~~~~
compilation terminated.
Wrote _test/cpp-unit.html
```

andychu:
Excellent, thank  you!

PR #1544, opened 2023-04-01
[cpp] MarkSweepHeap stores lists of ObjHeader*

watk:
Previously MarkSweepHeap's live_objs_ and to_free_ were vectors of RawObject*, but now they contain ObjHeader* instead. It's slightly less code this way.


watk:
Note: I'll change the target branch of this PR to soil-staging after dev-watk-1 is merged

andychu:
Excellent, thank you !!

PR #1560, opened 2023-04-08
[cpp] Remove ObjHeader::is_header

watk:
is_header no longer used so remove it. We now have a spare bit!


andychu:
Great thank you!

PR #1563, opened 2023-04-08
[benchmarks] Fix time-helper bug

watk:
I debugged this issue while trying to run the benchmarks and noticing they were silently failing! 

time-helper was hitting this "unreachable" abort() on my system because getopt() returns an int but we were assigning it to a char. This breaks when 'char' is 'unsigned char' (because getopt() can return -1).

### DIFF
--- a/benchmarks/time-helper.c
+++ b/benchmarks/time-helper.c
@@ -129,7 +129,7 @@ int main(int argc, char **argv) {
 
   // http://www.gnu.org/software/libc/manual/html_node/Example-of-Getopt.html
   // + means to be strict about flag parsing.
-  char c;
+  int c;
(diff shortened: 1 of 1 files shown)


andychu:
Oh wow, so `char` is `unsigned char` on your machine?  Is this aarch64 ?

andychu:
OK, I googled and found this ...  TIL !!

https://news.ycombinator.com/item?id=18269886

This is a good bug, it would be nice if we can automatically test these kinds of issues

I guess we can pass -funsigned-char and so forth

https://stackoverflow.com/questions/20518527/gcc-force-compiler-to-use-unsigned-char-by-default

PR #1565, opened 2023-04-10
[oil_lang] translate some builtins, split out json

melvinw:


melvinw:
delta for ysh-cpp is down to 295 with this change (from 314)

andychu (commented):
Looks great, thanks!

It's marked as Draft still?

melvinw:
Oh whoops!

PR #1567, opened 2023-04-16
[refactor] Replace span_id -> loc_t in oil_lang/expr_eval

PossiblyAShrub:


oil_lang/expr_eval.py
```diff
@@ -66,15 +66,15 @@
 _ = log
 
 
-def LookupVar(mem, var_name, which_scopes, span_id=runtime.NO_SPID):
-  # type: (Mem, str, scope_t, int) -> Any
+def LookupVar(mem, var_name, which_scopes, var_loc=loc.Missing()):
```
andychu, line 69: Can you try removing the default argument and see if it still type checks?  In all the functions

The style we're moving toward is to explicitly pass loc.Missing() , rather than relying on default args

This way we can 'grep' for places in the code that are missing location info

-----

Also when there are no defaults, you can eliminate the name at the call site too

like just `var_loc` instead of `var_loc=var_loc`

So we will use required positional arguments


----

If this doesn't work it's OK, but I think it likely will

andychu (commented):
Looks great, thanks for doing this!

PossiblyAShrub:
I've moved `loc.Missing()` to all of the call-sites and reduced `var_loc=var_loc` to `var_loc`. This should be ready to be looked at again.

andychu:
Awesome, thank you!

Wow there were many more loc.Missing() than I thought ... hm something to think about

---

BTW if you get sick of pushing and waiting, the failures tell you what command to run, like


http://travis-ci.oilshell.org/github-jobs/2023-04-16__05-00-45.wwz/


`test/spec.sh oil-all-serial`

And then it's not obvious but the way to run a specific test is like `test/spec.sh oil-scope -r 25 -v` , also documented on the Wiki

PR #1568, opened 2023-04-16
[mycpp] Add a pool allocator for small objects

watk:
Add a Pool allocator for allocating objects sized 32 bytes and under. This gives us a small speedup on most workloads. Though not as much as we had hoped, given that in many benchmarks 75+% of objects allocated are 32 bytes or under.

On my machine the benchmarks show a 10-25% performance improvement.

### DIFF
--- a/cpp/osh.cc
+++ b/cpp/osh.cc
@@ -17,7 +17,7 @@ using syntax_asdl::loc;
 namespace arith_parse {
 
 GcGlobal<tdop::ParserSpec> kArithSpec_ = {
-    {TypeTag::OtherClass, kZeroMask, HeapTag::Global, kIsGlobal}};
+    {kNotInPool, TypeTag::OtherClass, kZeroMask, HeapTag::Global, kIsGlobal}};
 tdop::ParserSpec* kArithSpec = &kArithSpec_.obj;
 
 }  // namespace arith_parse
--- a/mycpp/gc_alloc.h
+++ b/mycpp/gc_alloc.h
@@ -111,7 +111,9 @@ T* Alloc(Args&&... args) {
   ObjHeader* header = new (place) ObjHeader(T::obj_header());
 #if MARK_SWEEP
   header->obj_id = obj_id;
+  #if POOL_ALLOC
   header->in_pool = in_pool;
+  #endif
 #endif
   void* obj = header->ObjectAddress();
   // mycpp doesn't generated constructors that initialize every field
@@ -152,7 +154,9 @@ inline Str* NewStr(int len) {
 
 #if MARK_SWEEP
   header->obj_id = obj_id;
+  #if POOL_ALLOC
   header->in_pool = in_pool;
+  #endif
 #endif
   return s;
 }
@@ -174,7 +178,9 @@ inline Str* OverAllocatedStr(int len) {
   auto s = new (header->ObjectAddress()) Str();
 #if MARK_SWEEP
   header->obj_id = obj_id;
+  #if POOL_ALLOC
   header->in_pool = in_pool;
+  #endif
 #endif
   return s;
 }
@@ -218,7 +224,9 @@ inline Slab<T>* NewSlab(int len) {
   auto slab = new (obj) Slab<T>(len);
 #if MARK_SWEEP
   header->obj_id = obj_id;
+  #if POOL_ALLOC
   header->in_pool = in_pool;
+  #endif
 #endif
   return slab;
 }
--- a/mycpp/gc_alloc.h
+++ b/mycpp/gc_alloc.h
@@ -113,7 +113,10 @@ T* Alloc(Args&&... args) {
   header->obj_id = obj_id;
   header->in_pool = in_pool;
 #endif
-  return new (header->ObjectAddress()) T(std::forward<Args>(args)...);
+  void* obj = header->ObjectAddress();
+  // mycpp doesn't generated constructors that initialize every field
+  memset(obj, 0, sizeof(T));
+  return new (obj) T(std::forward<Args>(args)...);
 }
 
 //
@@ -140,6 +143,7 @@ inline Str* NewStr(int len) {
 
   auto s = new (header->ObjectAddress()) Str();
 #if defined(MARK_SWEEP) || defined(BUMP_LEAK)
+  s->data_[len] = '\0';  // NUL terminate
   s->len_ = len;
 #else
   // reversed in len() to derive string length
@@ -207,7 +211,11 @@ inline Slab<T>* NewSlab(int len) {
   void* place = gHeap.Allocate(num_bytes);
 #endif
   ObjHeader* header = new (place) ObjHeader(Slab<T>::obj_header(len));
-  auto slab = new (header->ObjectAddress()) Slab<T>(len);
+  void* obj = header->ObjectAddress();
+  if (std::is_pointer<T>()) {
+    memset(obj, 0, obj_len);
+  }
+  auto slab = new (obj) Slab<T>(len);
 #if MARK_SWEEP
   header->obj_id = obj_id;
   header->in_pool = in_pool;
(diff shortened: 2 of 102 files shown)


cpp/osh.cc
```diff
@@ -17,7 +17,7 @@ using syntax_asdl::loc;
 namespace arith_parse {
 
 GcGlobal<tdop::ParserSpec> kArithSpec_ = {
-    {TypeTag::OtherClass, kZeroMask, HeapTag::Global, kIsGlobal}};
+    {kNotInPool, TypeTag::OtherClass, kZeroMask, HeapTag::Global, kIsGlobal}};
```
watk, line 20: Aside: Should we add change ObjHeader to be constructed via a constructor instead of aggregate init so we don't have to change all these places when changing the header?

mycpp/gc_alloc.h
```diff
@@ -100,10 +100,18 @@ T* Alloc(Args&&... args) {
 
   DCHECK(gHeap.is_initialized_);
 
-  void* place = gHeap.Allocate(sizeof(ObjHeader) + sizeof(T));
+  constexpr size_t num_bytes = sizeof(ObjHeader) + sizeof(T);
+#if MARK_SWEEP
+  int obj_id;
+  bool in_pool;
+  void* place = gHeap.Allocate(num_bytes, &obj_id, &in_pool);
```
watk, line 107: I know you said you wanted to unify the Allocate interface but it felt like we should wait until we remove cheney to do that?

mycpp/mark_sweep_heap.cc
```diff
@@ -57,12 +57,18 @@ int MarkSweepHeap::MaybeCollect() {
 }
 
 // Allocate and update stats
-void* MarkSweepHeap::Allocate(size_t num_bytes) {
+// TODO: Make this interface nicer.
+void* MarkSweepHeap::Allocate(size_t num_bytes, int* obj_id, bool* in_pool) {
   // log("Allocate %d", num_bytes);
+  if (num_bytes <= pool_.kMaxObjSize) {
```
andychu, line 63: Can you add #ifdef POOL_ALLOC around this and any other relevant parts?

I'm going to optimize StackRoots, and then I want to do some measurements with this

andychu (commented):
I will try optimizing StackRoots, and add some measurements, and then we can take another look at this

The code largely looks good, but I'm still puzzled about performance

mycpp/mark_sweep_heap.cc
watk: Yep sure, done

andychu:
I just pushed a change to measure the pool also with GC on, and it actually looks pretty good

- It's not that far off bumpsmall
- It's reduces the total time with GC on as well


http://travis-ci.oilshell.org/github-jobs/2023-04-22__19-18-35.wwz/_tmp/gc/index.html

http://travis-ci.oilshell.org/github-jobs/2023-04-22__19-14-42.wwz/_tmp/gc-cachegrind/index.html

On Fib:

```


54.5 | _bin/cxx-opt+bumpsmall/osh | mut+alloc

58.1 | _bin/cxx-opt+pool/osh | mut+alloc
58.0 | _bin/cxx-opt+pool/osh | mut+alloc+free+gc

80.4 | _bin/cxx-opt/osh | mut+alloc
75.6 | _bin/cxx-opt/osh | mut+alloc+free+gc

```

watk:
> I just pushed a change to measure the pool also with GC on, and it actually looks pretty good
> 
> * It's not that far off bumpsmall

Nooo sorry, I temporarily pushed a hack to this branch to make the pool backed by a bump allocator, so these measurements are of the wrong thing. I've reverted it now. Should've pushed it to another PR to avoid confusion in the first place.

andychu:
I'm still seeing pretty good timings ?

(and now I agree we really need a connection from the PR to the CI run)

http://travis-ci.oilshell.org/github-jobs/2023-04-22__23-17-55.wwz/_tmp/gc-cachegrind/index.html


I wouldn't expect changing the Block allocation to be that big a deal?  Because Blocks are big and and cells are small

There should be a 128x fewer Block mallocs than cell allocations, right?



-----

parse.abuild

```

66.9 | _bin/cxx-opt+bumpsmall/osh | mut+alloc

69.8 | _bin/cxx-opt+pool/osh | mut+alloc
87.5 | _bin/cxx-opt+pool/osh | mut+alloc+free+gc

84.9 | _bin/cxx-opt/osh | mut+alloc

102.1 | _bin/cxx-opt/osh | mut+alloc+free+gc


```

andychu:
Yeah it seems the stable timings are still almost identical

watk:
Oh, yeah you're right.. ok, great! I thought I saw a bigger difference.

But yes, I definitely agree that we shouldn't see much difference by adding a bump allocator to the pool and was surprised to find (I thought) that it did make a big difference

andychu:
Yes, I think the numbers are looking pretty good now

Not 100% why it looked different to start, but I think a lot of it was not splitting out rooting and GC

Once you eliminate those two things, and concentrate on allocations (`mut+alloc`), I think the theory ended up being correct -- the pool allocator approximates the bump allocator

And actually I see we might want 2 or 3 pools now?  Let's save that for after this PR though

---



Now let me actually look in detail at the code!

mycpp/gc_alloc.h
```diff
@@ -147,11 +166,21 @@ inline Str* NewStr(int len) {
 // s->MaybeShrink() afterward!
 inline Str* OverAllocatedStr(int len) {
   int obj_len = kStrHeaderSize + len + 1;  // NUL terminator
-  void* place = gHeap.Allocate(sizeof(ObjHeader) + obj_len);
+  const size_t num_bytes = sizeof(ObjHeader) + obj_len;
+#if MARK_SWEEP
+  int obj_id;
+  bool in_pool;
+  void* place = gHeap.Allocate(num_bytes, &obj_id, &in_pool);
```
andychu, line 173: I wonder if there is some way for Allocate() to fill in the header directly, and then `T::obj_header()` only fills in the changed ones ...

Can save that for later ...

mycpp/mark_sweep_heap.h
```diff
@@ -62,6 +65,134 @@ class MarkSet {
...
+
+ private:
+  using Cell = std::array<uint8_t, CellSize>;
+
+  struct Block {
+    std::array<Cell, CellsPerBlock> cells;
+  };
+
+  // Unused/free cells are tracked via a linked list of FreeCells. The FreeCells
+  // are stored in the unused Cells, so it takes no extra memory to track them.
```
andychu, line 173: OK this is pretty interesting

One thing I thought of was using the complement of the MarkSet as the free list.  As long as mutating/allocating and mark/sweep are distinct phases, and they are, that should work.



But I think this is probably even better!

andychu (commented):
This looks great, thank you!

andychu:
OK I merged this -- it's still off by default

But we can turn it on in another PR, maybe just in the build system

And maybe look at doing 2 pools?  Looking at the numbers, it seems like that could help.  I started a thread on Zulip -- https://oilshell.zulipchat.com/#narrow/stream/384167-performance/topic/Are.202.20or.20more.20Pools.20justified.3F

Also I might want to get the pool stats in `benchmarks/gc`

andychu:
And just to complete the conversation, I think we were initially confused because

- not measuring rooting separately
- GC mixed into some metrics, like parsing.  GC is more expensive there relative to allocation, as opposed to runtime metrics.

But also

- I think the FreeCell change must have improved performance a bit?  Or something else

mycpp/mark_sweep_heap.h
watk: Oh right, and on allocation you do a linear scan of the mark set? I considered that, worth a try for sure. Might be faster

PR #1569, opened 2023-04-17
[refactor] Replace span_id with loc_t in error.Usage

PossiblyAShrub:


osh/builtin_assign.py
```diff
@@ -247,7 +248,7 @@ def Run(self, cmd_val):
       for pair in cmd_val.pairs:
         if pair.rval is not None:
           e_usage("doesn't accept RHS with -n",
-                  span_id=word_.LeftMostSpanForWord(pair.blame_word))
+                  loc.Span(word_.LeftMostSpanForWord(pair.blame_word)))
```
andychu, line 251: loc.Word(pair.blame_word) should work here

The exception handler should be smart enough (and we should be able to improve it )

osh/builtin_assign.py
```diff
@@ -269,7 +270,7 @@ def _ReconcileTypes(rval, flag_a, flag_A, blame_word):
   """
   if flag_a and rval is not None and rval.tag_() != value_e.MaybeStrArray:
     e_usage("Got -a but RHS isn't an array",
-            span_id=word_.LeftMostSpanForWord(blame_word))
+            loc.Span(word_.LeftMostSpanForWord(blame_word)))
```
andychu, line 273: ditto please see if loc.Word() works

osh/builtin_assign.py
```diff
@@ -282,7 +283,7 @@ def _ReconcileTypes(rval, flag_a, flag_A, blame_word):
 
     if rval.tag_() != value_e.AssocArray:
       e_usage("Got -A but RHS isn't an associative array",
-              span_id=word_.LeftMostSpanForWord(blame_word))
+              loc.Span(word_.LeftMostSpanForWord(blame_word)))
```
andychu, line 286: ditto

andychu (commented):
Wow!  This is great thank you

I see you got the idea that you can remove `tok.span_id` and just use `tok` -- a Token is a location.  There is a Zulip thread about that I will ping


Likewise for words.

BTW a word is something like `${dir}/file.c`  and a token is more like `${` and `dir` and `}`

----

Not sure if you've seen it yet, but you can do

`bin/osh -n -c 'echo hi'`

`bin/osh --ast-format text -n -c 'echo hi'`

to pretty print the AST

PossiblyAShrub:
> Not sure if you've seen it yet, but you can do
>
> `bin/osh -n -c 'echo hi'`
>
> `bin/osh --ast-format text -n -c 'echo hi'`
>
> to pretty print the AST

I've seen the `-n` flag but did not know about the `--ast-format`. Thanks!

---

I've addressed the feedback you gave in your review. This should be good for another look.

andychu:
Great thank you!  

The verbose format is in a "transitional state" now -- notice how it duplicates lines, i.e. every token points to a line.

That's intended, but the pretty printer should be smarter ...

PR #1573, opened 2023-04-20
[refactor] Replace spids -> List[loc_t] in args.Reader

PossiblyAShrub:
This was a big one, but `args.Reader` seems pretty core so I think this is good progress!

### DIFF
--- a/bin/oils_for_unix.py
+++ b/bin/oils_for_unix.py
@@ -21,7 +21,6 @@ import posix_ as posix
 import sys
 
 from _devbuild.gen.syntax_asdl import loc
-from asdl import runtime
 from core import error
 from core import shell
 from core import pyos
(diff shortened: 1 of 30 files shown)


bin/oils_for_unix.py
```diff
@@ -96,7 +95,7 @@ def AppBundleMain(argv):
   main_name, ext = os_path.splitext(b)
 
   # TODO: Do we need span IDs here?
-  arg_r = args.Reader(argv, spids=[runtime.NO_SPID] * len(argv))
+  arg_r = args.Reader(argv, locs=[loc.Missing()] * len(argv))
```
PossiblyAShrub, line 98: If we aren't using the span ids here, can we just pass in `None`?

andychu (commented):
This is fantastic, thank you!

PR #1574, opened 2023-04-20
[mycpp/runtime] Remove std::array due to big template expansions

andychu:


andychu:
This removes over 15% of the lines fed into the compiler!

After this change:


```
   55567 _build/preprocessed/cxx-opt/cpp/frontend_match.cc
   67386 _build/preprocessed/cxx-opt/mycpp/gc_str.cc
   84340 _build/preprocessed/cxx-opt/_gen/bin/oils_for_unix.mycpp.cc
  843515 total
```

Before


```
   ...
   45320 _build/preprocessed/cxx-opt/_gen/frontend/syntax.asdl.cc
   50030 _build/preprocessed/cxx-opt/cpp/core.cc
   63928 _build/preprocessed/cxx-opt/cpp/frontend_match.cc
   67387 _build/preprocessed/cxx-opt/mycpp/gc_str.cc
   92654 _build/preprocessed/cxx-opt/_gen/bin/oils_for_unix.mycpp.cc
 1025414 total
```

Kept track on Zulip over the last year:


https://oilshell.zulipchat.com/#narrow/stream/384167-performance/topic/Preprocessed.20Output.20.28ninja.20is.20fun.29

andychu:
@watk I noticed that the preprocessed metric went up, and as usual it's some STL stuff

Eventually I want to get rid of `std::regex` and more, but this one is easy right now

watk:
Oh :( that makes sense thanks. Gotta get used to this C++ dialect we're
targeting

andychu:
Yeah although people frown on it, I think "C with classes" makes sense for us

Everything is data-driven here, no religiosity about styles :)

Actually I believe fish shell did something similar for a long time -- C++ but without STL

PR #1577, opened 2023-04-22
[soil] Add a link to the PR (if any) to the build index

watk:
I'm not sure if this even works, how do you test this stuff?

### DIFF
--- a/soil/web.py
+++ b/soil/web.py
@@ -190,6 +190,7 @@ def ParseJobs(stdin):
     github_branch = meta.get('GITHUB_REF') 
     branch_str = github_branch or '?'  # no data for sr.ht
 
+    # Show the branch ref/heads/soil-staging or ref/pull/1577/merge (linkified)
     pr_number = meta.get('GITHUB_PR_NUMBER')
     if pr_number and github_branch:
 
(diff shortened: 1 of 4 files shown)


soil/web.py
```diff
@@ -210,9 +211,12 @@ def ParseJobs(stdin):
     &nbsp;
     <code><a href="https://github.com/oilshell/oil/commit/%(commit-hash)s">%(commit_hash_short)s</a></code>
   </td>
-  <td class="commit-line" colspan=4>
+  <td class="commit-line" colspan=3>
     <code>%(commit-line)s</code>
   </td>
+  <td colspan=1>
+    <code><a href="https://github.com/oilshell/oil/pull/%(pr-number)s">#%(pr-number)s</a></code>
```
andychu, line 218: This will display a blank `#` when there's no PR

But OK let's just run it and see what it looks like

We need a template language with conditionals ...

andychu:
Oh this is still a draft, I was going to just merge it

Maybe the easier way to fix the `#` is to call it `pr_link_html` 

And put the whole table cell in there

(note: I think there are some HTML escaping bugs in this code, which is embarrassing)

watk:
Oh, sorry, yeah I just pushed this half-baked thinking I could test it :p

watk:
@andychu if you want to take over from here, go for it. How would you test it?

andychu:
The way I test it is with SSH access to travis-ci.oilshell.org

So yes I'll take it over from here, thanks!  I agree this needs to be fixed

andychu:
Thanks for getting this started!  This definitely needed to be improved


----

BTW I documented the  fast way of testing this, without pushing

But it only works if you have SSH access to the server:

```

  $ soil/web-init.sh deploy-code
  $ soil/github-actions.sh remote-rewrite-jobs-index github-

```

watk:
Oh great, I checked it out and it looks good :)

PR #1580, opened 2023-04-23
FOR CI: Try enabling LTO

watk:


andychu:
BTW the CPython build has PGO, and I wondered about how to do to that as well

Though my feeling after seeing the pool allocator results is that we will get BIG WINS from tuning the sizes

i.e. taking over 90% of allocations is A LOT better than taking over 70% or 50%

( But as mentioned we can do that after the Tokens and such settle down )


I think LTO and PGO can only add a few percent on top of that (?)

---

One thing to consider is that we already have a HUGE translation unit `_gen/bin/oils_for_unix.mycpp.cc`, and most of the mycpp/runtime is templated, so in theory we have something approximating LTO


But still it would be nice to see how what the results are here, can't hurt to add!

watk:
Oh right, yeah I wanted to try PGO too. The tricky part is, I imagine,
figuring out a representative workload.

The weird thing is that LTO locally made like 20% difference on parsing.
But almost no difference on the CI.

I'll have to double check it

PR #1581, opened 2023-04-23
[build] Make the link step respect OIL_NINJA_VERBOSE

watk:
Just wanted this while doing something else.

Tested by running both:
```
TIME_TSV_OUT=/tmp/t OIL_NINJA_VERBOSE=1 ninja
OIL_NINJA_VERBOSE=1 ninja
```
and verifying the command line output and timing output file look as expected

### DIFF
--- a/build/ninja-rules-cpp.sh
+++ b/build/ninja-rules-cpp.sh
@@ -287,13 +287,12 @@ link() {
     prefix="benchmarks/time_.py --tsv --out $TIME_TSV_OUT --append --rusage --field link --field $out --"
   fi
 
+  if test -n "${OIL_NINJA_VERBOSE:-}"; then
+    echo "__ $prefix $cxx -o $out $@ $link_flags" >&2
+  fi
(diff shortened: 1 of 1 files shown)


build/ninja-rules-cpp.sh
```diff
@@ -289,7 +289,11 @@ link() {
 
   # IMPORTANT: Flags like -ltcmalloc have to come AFTER objects!  Weird but
   # true.
-  $prefix "$cxx" -o "$out" "$@" $link_flags
+  local cmdline=($prefix "$cxx" -o "$out" "$@" $link_flags)
```
andychu, line 292: This file unfortunately can't use arrays because it's sourced by the end user build, `_build/oils.sh` , which runs under `#!/bin/sh`

See the comments the analogous place in compile_one

So I guess copy and paste is the only real solution ...

andychu (commented):
Thanks, good idea

andychu:
That is also why all the functions are `snake_case` and not `kebab-case` BTW

There is a comment at the top of the file, but it's a bit confusing -- you could make that more explicit

PR #1583, opened 2023-04-24
[scripts] Give common scripts usage errors on no args

PossiblyAShrub:
Previously invoking scripts like `devtools/types.sh` with no args would silently do nothing. I remember being caught a few times trying to figure out why the script wasn't giving me any output.

I've only changed the scripts I've used. I know some scripts are sourced by others and are expected to do "nothing" when run and didn't want to break those. I figure that just these more common scripts should be enough to have the most impact.

### DIFF
--- a/build/py.sh
+++ b/build/py.sh
@@ -14,7 +14,8 @@ shopt -s strict:all 2>/dev/null || true  # dogfood for OSH
 REPO_ROOT=$(cd "$(dirname $0)/.."; pwd)
 readonly REPO_ROOT
 
-source build/common.sh  # for log, $CLANGXX
+source build/common.sh     # log, $CLANGXX
+source devtools/common.sh  # main
 # TODO: We could have the user run deps/from-apt.sh directly
 
 if test -z "${IN_NIX_SHELL:-}"; then
@@ -478,9 +479,4 @@ gitpod-minimal() {
   bin/osh -c 'echo hi'
 }
 
-if [ $# -eq 0 ]; then
-  echo "usage: $0 <function name>"
-  exit 1
-fi
-
-"$@"
+main "$@"
--- a/build/deps.sh
+++ b/build/deps.sh
@@ -34,7 +34,8 @@ set -o nounset
 set -o pipefail
 set -o errexit
 
-source deps/from-apt.sh   # PY3_BUILD_DEPS
+source deps/from-apt.sh    # PY3_BUILD_DEPS
+source devtools/common.sh  # main
 
 # Also in build/dev-shell.sh
 USER_WEDGE_DIR=~/wedge/oils-for-unix.org
@@ -309,4 +310,4 @@ container-wedges() {
   #deps/wedge.sh build deps/source.medo/R-libs/
 }
 
-"$@"
+main "$@"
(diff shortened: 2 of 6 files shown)


andychu:
Thanks!  Yes this is one of the things on "dev friction", so I'm glad to see something concrete

I think this deserves its own file and a name other than `main`, how about:

```
source devtools/run-task.sh

...

run-task "$@"
```

( BTW `devtools/completion.sh` does completion of the tasks with `"$@"`.  I think it can support `run-task "$@"`.  Ask me to demo this for you next time !!  It helps a lot)


---

BTW @watk did something similar for `test/spec.sh`

I think we can just do it piecemeal, we don't have to consolidate them right now, but we should aim for it

https://github.com/oilshell/oil/blob/master/test/spec.sh#L1176

---

Good thread on "Task files":


https://oilshell.zulipchat.com/#narrow/stream/264891-oil-help/topic/Nice.20Task.20File.20Template.20.28Why.20.22Array.20foo.20can't.20be.20referred.2E.2E.2E

Lots of people use this pattern, and we should make it better in OSH and YSH !

PossiblyAShrub:
I moved this to a separate file and changed the name. I also changed the `run-task` helper to behave like @watk's changes in `test/spec.sh`. So now we have function listings in all the scripts!

> I think we can just do it piecemeal, we don't have to consolidate them right now, but we should aim for it.

Do you mean we shouldn't have a common `run-task` utility? Or is this for more specialized help info as we develop these tools further?

> Good thread on "Task files":
>
> https://oilshell.zulipchat.com/#narrow/stream/264891-oil-help/topic/Nice.20Task.20File.20Template.20.28Why.20.22Array.20foo.20can't.20be.20referred.2E.2E.2E
>
> Many people use this pattern, and we should make it better in OSH and YSH !

Thanks for linking this. There's some good ideas in there. I agree that we need better support for patterns like this.

devtools/run-task.sh
```diff
@@ -0,0 +1,31 @@
...
+#    ...  # define task functions
+#   run-task "$@"
+
+# List all functions defined in this file (and not in sourced files).
+_list-funcs() {
+    local funcs=($(compgen -A function))
+    # extdebug makes `declare -F` print the file path, but, annoyingly, only
+    # if you pass the function names as arguments.
+    shopt -s extdebug
+    declare -F "${funcs[@]}" | grep --fixed-strings " $0" | awk '{print $1}'
```
watk:
```suggestion
    declare -F "${funcs[@]}" | grep --fixed-strings " $0" | awk '{print "  "$1}'
```
To list the tasks indented (if we don't pipe into column)

build/py.sh
```diff
@@ -14,7 +14,8 @@ shopt -s strict:all 2>/dev/null || true  # dogfood for OSH
 REPO_ROOT=$(cd "$(dirname $0)/.."; pwd)
 readonly REPO_ROOT
 
-source build/common.sh  # for log, $CLANGXX
+source build/common.sh       # log, $CLANGXX
+source devtools/run-task.sh  # main
```
watk:
```suggestion
source devtools/run-task.sh  # run-task
```

watk (commented):
Nice!

andychu (approved):
Great thank you!

andychu:
OK I addressed most of the comments, and I think this is pretty good

Keeping  `column` for now, but we can iterate on this more

As mentioned I want to fix the completion too

And maybe write a blog post with a asciinema screencast ... *How I've Used Bash as a "Task Runner" for 15 Years* :-)

I guess people started using the "Task Runner" name recently

PR #1586, opened 2023-04-25
[refactor] Replace span_id -> loc_t all instances in syntax.asdl

PossiblyAShrub:
We no longer have any span ids in `syntax.asdl`!

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -22,7 +22,7 @@ from _devbuild.gen.runtime_asdl import (
     redirect_arg__MoveFd, redirect_arg__HereDoc, value, value_e, value__Str,
     trace, trace_t)
 from _devbuild.gen.syntax_asdl import (
-    loc, loc_t,
+    loc_t,
     redir_loc,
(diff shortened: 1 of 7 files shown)


core/process.py
```diff
@@ -368,7 +368,7 @@ def _ApplyRedirect(self, r):
         except OSError as e:
           self.errfmt.Print_("Can't open %r: %s" %
                              (arg.filename, pyutil.strerror(e)),
-                             blame_loc=loc.Span(r.op_spid))
+                             blame_loc=r.op_loc)
```
andychu, line 371: BTW I want to get rid of all the keyword arguments like blame_loc=

As mentioned they should be required arguments, which means they don't need keywords

The code is inconsistent now, but we can gradually change that

andychu:
Great thank you!  (btw this was runtime.asdl not syntax.asdl -- we should talk a bit about some portions of syntax.asdl, maybe I should do a few after the 0.15.0 release)

PR #1587, opened 2023-04-25
[builtin/history] Add support for '-a' (append) flag

PossiblyAShrub:
Now we can do something like this:

```sh
$ echo 'start of osh session'
start of osh session
$ cat ~/.config/oil/history_osh
$ history -a
$ cat ~/.config/oil/history_osh
echo 'start of osh session'
cat ~/.config/oil/history_osh
history -a
```

### DIFF
--- a/core/shell.py
+++ b/core/shell.py
@@ -9,6 +9,7 @@ import time
 from _devbuild.gen import arg_types
 from _devbuild.gen.option_asdl import option_i, builtin_i
 from _devbuild.gen.syntax_asdl import loc, source, source_t, IntParamBox
+from _devbuild.gen.runtime_asdl import value, scope_e
 
 from asdl import runtime
 
@@ -36,6 +37,7 @@ _ = flag_def
 from frontend import flag_spec
 from frontend import reader
 from frontend import parse_lib
+from frontend import location
 
 from oil_lang import expr_eval
 from oil_lang import builtin_json
@@ -383,7 +385,9 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
   home_dir = pyos.GetMyHomeDir()
   assert home_dir is not None
 
+  # init the HISTFILE variable to our default history file
   history_filename = os_path.join(home_dir, '.config/oil/history_%s' % lang)
+  mem.SetValue(location.LName('HISTFILE'), value.Str(history_filename), scope_e.GlobalOnly)
 
   #
   # Initialize builtins that don't depend on evaluators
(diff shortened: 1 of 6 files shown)


andychu:
This looks great!  I'll demo adding a spec test during our meeting tomorrow

andychu:
BTW the `test/spec-bin.sh` installation might help with the bash issue

It makes sure we have a specific version of bash to test

on my laptop, I think we were falling back to the system bash and getting weird results, but our own compiled version is probably more predictable

PossiblyAShrub:
This should be ready for review. I am not sure why the ovm-tarball test is failing. The logs say "all tests passed". Any ideas?

andychu:
Hm I merged master and it passed

I thought that only soil-staging was broken last night, not master, not sure why this branch got broken ?

Anyway I will look at this now

core/shell.py
```diff
@@ -378,6 +380,15 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
...
   dir_stack = state.DirStack()
 
+  # Find common directories, these are consumed by some builtins and then much
+  # later used while setting up an interactive shell.
+  home_dir = pyos.GetMyHomeDir()
+  assert home_dir is not None
+
+  # init the HISTFILE variable to our default history file
+  history_filename = os_path.join(home_dir, '.config/oil/history_%s' % lang)
+  mem.SetValue(location.LName('HISTFILE'), value.Str(history_filename), scope_e.GlobalOnly)
```
andychu, line 390: There's a helper `state.SetGlobalString('HISTFILE', history_filename)` you can use here

spec/builtins2.test.sh
```diff
@@ -121,6 +121,101 @@ command command -v seq
 ## N-I zsh stdout-json: ""
 ## N-I zsh status: 127
 
+#### history -a
+case $SH in dash|mksh|zsh) exit 0 ;; esac
```
andychu, line 125: After this PR we should probably split this into  `spec/builtins-history.test.sh`

If a feature is only implemented in bash, and not other shells, then that's a good reason to have another file

I just did that with `spec/builtin-trap.test.sh` and `spec/builtin-trap-bash.test.sh`

spec/builtins2.test.sh
```diff
@@ -121,6 +121,101 @@ command command -v seq
...
+echo 1
+history -a
+cat tmp
+
+echo 2
+
+cat tmp
+' | $SH -i
+
+# match osh's behaviour of echoing ^D for EDF
```
andychu: typo EDF -> EOF

andychu (commented):
Thanks, this looks great!

Just one small improvement

PossiblyAShrub:
All done resolving your review feedback.

PR #1588, opened 2023-04-26
[oil_lang] more refactor prep, refactor unary expression to value_t

melvinw:


oil_lang/expr_eval.py
```diff
@@ -459,6 +490,351 @@ def _EvalConst(self, node):
...
+
+  def _EvalUnary(self, node):
+    # type: (expr__Unary) -> value_t
+    child = _PyObjToValue(self._EvalExpr(node.child)) # XXX
+    if node.op.id == Id.Arith_Minus:
+      UP_child = child
+      with tagswitch(child) as case:
+        if case(value_e.Bool):
+          child = cast(value__Bool, UP_child)
+          return value.Int(-child.b) # gets promoted...
```
andychu, line 548: Hmm I think we should leave this out?    This would make `-true` equal to `-1`

Which is true in Python for historical reasons, but I don't think we want than in YSH

Does a test fail if we leave this out?  If so we should take this opportunity to "tighten up the spec".  It's sorta borrowed from Python, and there are cases where we should deviate

oil_lang/expr_eval.py
```diff
@@ -459,6 +490,351 @@ def _EvalConst(self, node):
...
+
+        elif case(value_e.Int):
+          child = cast(value__Int, UP_child)
+          return value.Int(~child.i)
+
+    if node.op.id == Id.Expr_Not:
+      UP_child = child
+      with tagswitch(child) as case:
+        if case(value_e.Undef):
+          return value.Bool(True) # "not None"
```
andychu, line 573: yeah likewise I think `not` could only apply to booleans

And this could actually make the interpreter a little faster

I think YSH will be more explicit

oil_lang/expr_eval.py
```diff
@@ -459,6 +490,351 @@ def _EvalConst(self, node):
...
+          child = cast(value__Bool, UP_child)
+          return value.Int(-child.b) # gets promoted...
+
+        elif case(value_e.Int):
+          child = cast(value__Int, UP_child)
+          return value.Int(-child.i)
+
+        elif case(value_e.Float):
+          child = cast(value__Float, UP_child)
+          return value.Float(-child.f)
```
andychu, line 556: can we put `else: raise AssertionError()` here and in all the `with tagswitch()`?

Or it could be something like `error.Expr(...)`, or maybe `error.InvalidType` which derives from `error.Expr()`

andychu (commented):
This looks great!

I added some comments about the semantics

Maybe one strategy is to do the minimal thing to get the tests to pass, and then make another pass to settle on the behavior of things like `-True` and `~True` and `not None`, adding those as spec test cases?

Right now YSH is underspecified

oil_lang/expr_eval.py
melvinw: Ah this is great to read! This is such a counterintuitive quirk. I'll add some spec tests for these operators if there aren't some already

melvinw:
Ok pushed a few new commits. Have a look!

andychu:
Looks great, thanks!

PR #1592, opened 2023-04-30
[refactor] Add `Token keyword` to the start of for and do command nodes

PossiblyAShrub:


frontend/syntax.asdl
```diff
@@ -309,21 +309,21 @@ module syntax
   | Pipeline(List[command] children, bool negated, List[int] stderr_indices)
   | AndOr(List[id] ops, List[command] children)
     # Part of for, while, until (but not if, case, ShFunction).  No redirects.
-  | DoGroup(List[command] children)
+  | DoGroup(Token left, Token right, List[command] children)
```
andychu, line 312: How about putting `Token right` last?  That is how I did it for single_quoted, double_quoted, command_sub, etc.

(which we should really rename to SingleQuoted, DoubleQuoted, CommandSub, since they are real concrete types, not sum types)

andychu (commented):
This is great!  exactly the right idea

Then in another pass we can remove spids and replace their usage in osh2oil (which I should also rename to ysh-prettify)

PR #1593, opened 2023-04-30
[test/spec] Split history builtin tests from builtins2 tests

PossiblyAShrub:
As per https://github.com/oilshell/oil/pull/1587#discussion_r1181092501.

---

While working on this I realized we were missing the case where `HISTFILE` does not point to a file (even if it's a valid string). This should return an error code if a user tries to then `history -r`.

### DIFF
--- a/osh/builtin_lib.py
+++ b/osh/builtin_lib.py
@@ -87,7 +87,7 @@ class History(vm._Builtin):
     if arg.r:
       history_filename = self.GetHistoryFilename()
       if not path_stat.exists(history_filename):
-        self.errfmt.Print_("The file '%s' ($HISTFILE) does not exist" % history_filename, loc.Missing())
+        self.errfmt.Print_("HISTFILE %r doesn't exist" % history_filename, loc.Missing())
         return 1
 
       readline.read_history_file(history_filename)
--- a/osh/builtin_lib.py
+++ b/osh/builtin_lib.py
@@ -13,6 +13,7 @@ from core import vm
 from core.pyerror import e_usage
 from frontend import flag_spec
 from mycpp import mylib
+from pylib import path_stat
 
 from typing import Optional, cast, TYPE_CHECKING
 if TYPE_CHECKING:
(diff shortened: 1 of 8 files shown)


osh/builtin_lib.py
```diff
@@ -83,7 +84,12 @@ def Run(self, cmd_val):
       return 0
 
     if arg.r:
-      readline.read_history_file(self.GetHistoryFilename())
+      history_filename = self.GetHistoryFilename()
+      history_file_exists = path_stat.exists(history_filename)
+      if not history_file_exists:
+        raise error.ErrExit(1, "The file '%s' ($HISTFILE) does not exist" % history_filename, loc.Missing())
```
andychu, line 90: hm actually `ErrExit` is intended for use with `set -o errexit`

I think simply `return 1` will do what you want here?

Also I would just say `if not path_stat.exists()` instead of the extra var

---

Is this intended to be friendlier than bash?  I think bash just ignores the error?

I tried `bash -i -c 'HISTFILE=a; history -r`

PossiblyAShrub: Ok. I still want to provide some information about the failure. I see that we use a `self.errfmt.Print_` pattern in other builtins. Would that be appropriate here?

> Is this intended to be friendlier than bash? I think bash just ignores the error?
>
> I tried bash -i -c 'HISTFILE=a; history -r

I do get an error code on my machine (no message though)

```sh
$ HISTFILE=_tmp/dne
$ history -r
$ echo $?
1
```

andychu:
Ah OK , makes sense

I changed the error message to use `%r` because that will quote the filename properly (e.g. in the rare case that the filename contains a single quote)


In the C++ runtime we have a format string implementation that supports `%s %r %d` and that's it basically

- `%s` and `%r` must be strings (unfortunately you get a runtime error if not, not compile time
- `%d` must be an integer

andychu:
Merged, thanks for doing this!

PR #1596, opened 2023-04-30
[oil_lang] statically type binary expressions

melvinw:


melvinw:
Whoops. Broke a few spec tests. Will push a patch

oil_lang/expr_eval.py
```diff
@@ -33,8 +33,9 @@
     scope_e, scope_t,
     part_value, part_value_t,
     lvalue,
-    value, value_e, value_t, value__Bool, value__Int, value__Float,
-    value__Str, value__MaybeStrArray, value__AssocArray, value__Obj
+    value, value_e, value_t, value__AssocArray, value__Bool, value__Dict,
+    value__Int, value__Float, value__List, value__MaybeStrArray, value__Str,
```
andychu, line 37: just FYI, I made a change to allow removing all these silly imports

Instead you just import `value, value_e, value_t`, and that's it


https://oilshell.zulipchat.com/#narrow/stream/208950-zephyr-asdl/topic/Remove.20value__Foo.20types.2C.20etc.2E

We have to migrate gradually though, separately from this change

oil_lang/expr_eval.py
```diff
@@ -441,6 +460,41 @@ def _ToInteger(self, val):
...
+      elif case(value_e.Str):
+        val = cast(value__Str, UP_val)
+        if match.LooksLikeInteger(val.s):
+          return int(val.s)
+        else:
+          raise ValueError("%r doesn't look like an integer" % val.s)
+
+    raise error.InvalidType('Expected Int', loc.Missing())
+
+  def _MaybeConvertStr(self, val):
```
andychu, line 481: I think the names were ToInteger and ToNumber (float, int)?  How about preserving that, like `_ValueToNumber()` ?

oil_lang/expr_eval.py
```diff
@@ -579,64 +633,243 @@ def _EvalUnary(self, node):
...
+                raise ZeroDivisionError()
+
+              return value.Float(left.i / float(right.i))
+            else:
+              raise NotImplementedError(op)
+
+          elif rcase(value_e.Float):
+            right = cast(value__Float, UP_right)
+            if op == Id.Arith_Plus:
+              return value.Float(left.i + right.f)
```
andychu, line 668: I think we might want to revisit if Int + Float and Int - Float work

Python allows it, and maybe the spec tests did, but we can think about it more, and tighten up the behavior later

Another thing to remember is that we want to do the overflow detection for integers, so it will turn into something like `oils_number_lib.Add(left_int, right_int)`

oil_lang/expr_eval.py
```diff
@@ -579,64 +633,243 @@ def _EvalUnary(self, node):
...
+
+    raise NotImplementedError()
+
+  def _ArithLogical(self, left, right, op):
+    # type: (value_t, value_t, Id_t) -> value__Bool
+    UP_left = left
+    UP_right = right
+
+    with tagswitch(left) as lcase:
+      if lcase(value_e.Bool):
```
andychu, line 774: I wonder if we can somehow switch left and right at the same time, I'll think about that ...

The tags are small integers so technically we could shift them into a single integer like (16 bit left tag, 16 bit right tag)

oil_lang/expr_eval.py
```diff
@@ -579,64 +633,243 @@ def _EvalUnary(self, node):
...
+      else:
+        raise error.InvalidType('Expected List or String', loc.Missing())
+
+  def _EvalBinary(self, node):
+    # type: (expr__Binary) -> value_t
+
+    left = _PyObjToValue(self._EvalExpr(node.left))
+    right = _PyObjToValue(self._EvalExpr(node.right))
+
+    if node.op.id == Id.Arith_Plus \
```
andychu, line 840: btw as a shortcut, mycpp supports `node.op.id in (Id.Arith_Plus, IdArith_Minus, ...)`

It expands to basically this chain


----

Another thing we can think about is a 2-level dispatching -- we have those for tokens -- Id and Kind

andychu (commented):
Looks great, thank you!

andychu:
I made a bunch of general comments, but the only one that was specific to this PR was _ValueToNumber() name ... I kind of like that symmetry, i.e. a Number is either an Int or Float

But I just merged anyway, as I think we'll be making many iterations on this file

Thanks!

oil_lang/expr_eval.py
melvinw: ah sure. i can send a follow up for this

oil_lang/expr_eval.py
melvinw: ah yes that would back following the logic here much easier

oil_lang/expr_eval.py
melvinw: ah great! i'll switch to that in the followup with the rename

PR #1597, opened 2023-05-02
[oil_lang] some binary expression tweaks

melvinw:
Didn't get to these before the last patch merged.


andychu:
Cool thanks

andychu:
I was also thinking last night we could have a syntax like

```
with tagswitch2(node1, node2) as case:
  if case(value_e.Int, value.e.Int):
    ...
  elif case(value_e.Int, value.e.Float):
    ...
  

```

which probably wouldn't be hard to implement, and   would reduce nesting

Though not sure how much code that would actually save

PR #1598, opened 2023-05-02
[refactor] Add keyword tokens to If nodes

PossiblyAShrub:


frontend/syntax.asdl
```diff
@@ -263,9 +263,9 @@ module syntax
 
   # Each arm tests one word against multiple words
   CaseArm = (List[word] pat_list, List[command] action, List[int] spids)
-  IfArm = (condition cond, List[command] action, List[int] spids)
+  IfArm = (Token keyword, condition cond, Token? then_kw, List[command] action, List[int] spids)
```
andychu, line 266: How about adding comments that

- keyword is either `if` or `elif` 
- `then` is optional for YSH / Oil ?

andychu (commented):
Looks great!

Glad the tests caught an issue ... they are kind of rough, but still useful

If any of the changes seem like they need more tests, feel free to dump them in that file.  And then we can clean it up separately.

PossiblyAShrub:
Good point on adding those comments. I've done so in https://github.com/oilshell/oil/pull/1598/commits/6880ce0ba83b9d556a4dec94a191256eab89fc41.

andychu:
Great thanks!  I think the frontend will be way cleaner after we get rid of the span IDs

and ysh-prettify can start to be more "real"

I was wondering whether it should be part of the project, but it seems possible to preserve and extend.

Although maybe a formatter/linter is equally or more important than the translator

andychu:
I also wonder if we should call it `rough-osh2ysh` or `ysh-ify` :)

Since as mentioned it's not really possible to do a 100% precise translation

PossiblyAShrub:
I think that a translator is just a composition of a formatter and linter. It rewrites code to fit a style and produces diagnostics where something isn't idiomatic. So any work on a formatter/linter/translator mutually benefits each other. Once `spids` are gone, I think that we will have a great base to build this tooling on.

I like the name `ysh-ify` quite a bit actually!

andychu:
Yeah it would be nice to have all the functionality

Though I honestly wonder if it should all be done in a single pass?

Does the single pass have options like `--ysh-ify` and `--format-whitespace` and `--produce-lint-errors` ?

The design is actually a bit nontrivial

I think there was a good discussion somewhere about `go fmt` vs. `rustfmt`.  

- Someone said that `go fmt` actually does it right -- it formats your code without touching a lot of other stuff.
- `rust fmt` parses the code and then pretty prints it -- this ends up messing up your code a lot

The architecture of `osh2oil` / `ysh-ify`  is more similar to `go fmt` as far as I understand, but it's not done yet

I will try to find that thread

PR #1600, opened 2023-05-04
[refactor] Replace `spids` with explicit `Token` fields

PossiblyAShrub:


osh/cmd_parse.py
```diff
@@ -85,6 +85,16 @@ def _KeywordToken(UP_w):
   return cast(Token, part)
 
 
+def _DelimToken(word, arena):
+    # type: (word_t, Arena) -> Token
+    """
+    TODO: get token directly
+    For cases where we'd want to use `word_.LeftMostSpanForWord` but also want
```
andychu, line 92: I feel like this pattern is a little confusing

One possibility for `((` and `(` is to call it `_OperatorToken(w)` and then assert that `w.tag() == word_e.Token`

Which should really be called `word_e.Operator`, it could be

```
word =
  Operator %Token
  Compound %CompoundWord
```

This doesn't work  for `[[`, because in shell `[` is not an operator token while `(` is, but we could deal with that separately

andychu (commented):
Hmm 

- ShFunction and TimeBlock look good
- I think we could simplify Subshell and Dparen by nothing that they are operator tokens 
  -    that means `Kind.Op`, which is the same as `Id.Op_*`
  - note in `_ReadWord` in `osh/word_parse.py`, we return a Token directly for `Kind.{Op,Redir,Arith}`
- DBracket is more complicated, and we could punt on it for now

I added this debugging tool to show the `Id` and `Kind` of the tokens that aren't attached to the tree:


https://github.com/oilshell/oil/commit/b0f3b386921d157e4059c00b910e0dfb08bf4f3b






It could probably be improved/renamed

andychu:
On second thought we could keep the  code you have for `[[` because it works, but used a more optimized function when we have `word_e.Token`, which `(` and `((` are

(and again at some point I think renaming it to `word_e.Operator` makes sense, I think that term is in the POSIX shell spec)

PossiblyAShrub:
Thanks for that tool. It's quite helpful to see all of the tokens listed out while I am figuring out all the details of our parser.

I've replaced the `_DelimToken` function in favour of a `_OperatorToken` function which is much simpler.

This is ready to be looked at again.

PossiblyAShrub:
Now we're good!

I found that the right terminating tokens were being consumed in other parsers before they terminated. This meant that we have to extract those tokens in those other parsers and return them.

osh/word_.py
```diff
@@ -620,6 +620,32 @@ def LiteralToken(UP_w):
...
+  Given a word that IS A keyword, return the single token at the start.
+
+  In C++, this casts without checking, so BE CAREFUL to call it in the right context.
+  """
+  assert UP_w.tag() == word_e.Compound, UP_w
+  w = cast(CompoundWord, UP_w)
+
+  part = w.parts[0]
+  assert part.tag() == word_part_e.Literal, part
+  return cast(Token, part)
```
andychu, line 635: I wonder if we should `assert consts.GetKind(tok.id) == Kind.KW` here?   I don't think there's any case where that's false?

That is sort of what the comment is saying -- "given that the word is a keyword"

----

Also how about shortening the names like

```
word_.AsKeywordToken()
word_.AsOperatorToken()
```

?

Since we always import and use `module.Func()`, I think this shortens the code without affecting readability

andychu (commented):
This looks great, thank you!  Good call on adding the right tokens, IIRC we do that for the expression parser too (frontend/parse_lib.py)

And good call on moving them to `word_.py` as well

I made a couple minor comments, but if you think it's too fiddly to change now, we could them in a different PR

osh/word_.py
PossiblyAShrub: > I wonder if we should assert consts.GetKind(tok.id) == Kind.KW here? I don't think there's any case where that's false?

I'll check to make sure that this assumption actually is correct. If so, I'll add it.

Re: changing the names. That seems reasonable, especially if we adhere to the `module.Func()` pattern everywhere.

PossiblyAShrub:
Good call on adding the assert. There were a few cases where `word_.AsKeywordToken` was called on literal tokens! I've changed those places to use `word_.LiteralToken`.

This is ready for a final check.

osh/cmd_parse.py
```diff
@@ -1863,49 +1855,57 @@ def ParseSubshell(self):
...
+    anode, right = self.w_parser.ReadDParen()
     assert anode is not None
 
     self._Peek()
+    # The right span id must lie directly after the Op_DRightParen token for
+    # `Tracer` to print properly.
+    # (( a = 42 )) <...>
+    #              ^^^^^-- span id points here
+    # NOTE: we have a simular situation in ParseDBracket
+    # TODO: fix `Tracer.PrintSourceCode` so we don't need this span id
```
andychu, line 1904: Oh yes, good point, we should fix this!

andychu (commented):
Good point about the bad token IDs

I think maybe the Cursor needs two methods

- cursor.PrintUntil()   -- stop before this token
- cursor.PrintPast()  -- include this token

We can talk about that on Zulip

andychu:
Merged, thank you!

PR #1601, opened 2023-05-04
[test/ysh-ify] Add test for non-brace function bodies

PossiblyAShrub:


test/ysh-prettify.sh
```diff
@@ -431,7 +444,7 @@ test-posix-func() {
   proc f
   {
     echo "hi"
-  }'
+  }' \
```
andychu, line 447: does this do anything?  looks a little weird

andychu (commented):
OK great, glad you found the INVALID pattern!

test/ysh-prettify.sh
PossiblyAShrub: No it doesn't :P. I'll remove that.

PR #1602, opened 2023-05-05
[oil_lang] statically type comparison expressions

melvinw:


oil_lang/expr_eval.py
```diff
@@ -176,10 +183,156 @@ def _ValueToPyObj(val):
...
+      return True
+
+    elif case(value_e.AssocArray):
+      left = cast(value.Dict, UP_left)
+      right = cast(value.Dict, UP_right)
+      if len(left.d) != len(right.d):
+        return False
+
+      for k in left.d.keys():
+        if not k in right.d or right.d[k] != left.d[k]:
```
andychu, line 256: style nitpick: in Python you can use `k not in right.d`

oil_lang/expr_eval.py
```diff
@@ -176,10 +183,156 @@ def _ValueToPyObj(val):
...
+
+    else:
+      raise error.InvalidType('Expected List or Dict', loc.Missing())
+
+  return False
   
 
+def AsPosixEre(eggex):
+  # type: (value.Eggex) -> str
+  if len(eggex.as_ere):
```
andychu, line 327: I think this is clearer as `eggex.as_ere is not None`, i.e. it's set if it's not None

And `core/runtime.asdl` should have `str? as_ere`

Similar to oil_lang/objects.py (can we delete that?)

Not sure if it actually comes up, but  in theory you can have an empty eggex that translates to an empty regex ?  (probably should have a test case for that)

oil_lang/expr_eval.py
```diff
@@ -244,24 +397,37 @@ def EvalLHS(self, node):
...
-    else:
-      raise RuntimeError(
-          "RHS of ~ should be string or Regex (got %s)" % right.__class__.__name__)
+    UP_right = right
+    right_s = None # type: str
+    with tagswitch(right) as case:
+      if case(value_e.Str):
+        right = cast(value.Str, UP_right)
+        right_s = right.s
+        pass
```
andychu, line 411: can remove `pass` here

oil_lang/expr_eval.py
```diff
@@ -126,10 +126,17 @@ def _PyObjToValue(val):
...
   elif isinstance(val, list):
-    return value.List(val)
+    return value.List([_PyObjToValue(elem) for elem in val])
 
   elif isinstance(val, dict):
-    return value.Dict(val)
+      return value.Dict({k: _PyObjToValue(v) for k, v in val.items()})
+
+  elif isinstance(val, objects.Regex):
+    eggex = value.Eggex(val.regex, '')
```
andychu, line 135: can make second arg  `None`, see below

oil_lang/expr_eval.py
```diff
@@ -1196,7 +1437,7 @@ def _EvalExpr(self, node):
         node = cast(expr.RegexLiteral, UP_node)
 
         # TODO: Should this just be an object that ~ calls?
-        return objects.Regex(self.EvalRegex(node.regex))
+        return _ValueToPyObj(value.Eggex(self.EvalRegex(node.regex), ''))
```
andychu, line 1440: also `None` initialization I think

oil_lang/expr_eval.py
```diff
@@ -866,70 +1032,145 @@ def _EvalSlice(self, node):
...
-            left2 = True
-          elif left in ('false', '0'):
-            left2 = False
-          else:
-            return False
+          elif case(value_e.Bool):
+            right = cast(value.Bool, UP_right)
+            left2 = left2.lower()
+            lb = False
+            if left2 in ('true', '1'):
```
andychu, line 1158: random note that doesn't relate to this PR -- I wonder if we should not equate `true` and `1`, like in C

Because it's actually the opposite in shell!   e.g. `if grep foo bar` takes the branch if grep returns 0!

andychu (commented):
This looks great, thank you!  Just some small notes

Mainly about the `''` vs. `None` initialization

oil_lang/expr_eval.py
melvinw: ah great. wasn't sure if mycpp could deal with `not in`

melvinw:
@andychu pushed some fixes for you comments and removed 0/1-to-bool conversion stuff

andychu:
Looks great, thank you!

PR #1603, opened 2023-05-07
[osh refactor] Replace spids with explicit token fields

PossiblyAShrub:
Nodes changed:
 - `command.AndOr`
 - `command.ShAssignment`
 - `command.Pipeline`
 - `command.Case`

### DIFF
--- a/frontend/syntax.asdl
+++ b/frontend/syntax.asdl
@@ -298,12 +298,12 @@ module syntax
   | ExpandedAlias(command child, List[Redir] redirects, List[EnvPair] more_env)
   | Sentence(command child, Token terminator)
     # Note: Only represents "bare assignment"
-  | ShAssignment(List[AssignPair] pairs, List[Redir] redirects)
+  | ShAssignment(Token var, List[AssignPair] pairs, List[Redir] redirects)
   | ControlFlow(Token keyword, word? arg_word)
     # Note: There are spids for every pipeline operator, parallel to
     # stderr_indices
-  | Pipeline(List[command] children, bool negated, List[int] stderr_indices)
-  | AndOr(List[id] ops, List[command] children)
+  | Pipeline(Token pipline_op, List[command] children, bool negated, List[int] stderr_indices)
+  | AndOr(Token op, List[id] ops, List[Token] op_toks, List[command] children)
     # Part of for, while, until (but not if, case, ShFunction).  No redirects.
   | DoGroup(Token left, List[command] children, Token right)
     # A brace group is a compound command, with redirects.
@@ -323,7 +323,8 @@ module syntax
   | WhileUntil(Token keyword, condition cond, command body, List[Redir] redirects)
   | If(Token if_kw, List[IfArm] arms, Token? else_kw, List[command] else_action,
        Token? fi_kw, List[Redir] redirects)
-  | Case(word to_match, List[CaseArm] arms, List[Redir] redirects)
+  | Case(Token case_kw, word to_match, Token? in_kw, List[CaseArm] arms,
+         Token? esac_kw, List[Redir] redirects)
     # The keyword is optional in the case of bash-style functions
     # (ie. "foo() { ... }") which do not have one.
   | ShFunction(Token? keyword, loc name_loc, str name, command body)
--- a/frontend/syntax.asdl
+++ b/frontend/syntax.asdl
@@ -302,7 +302,7 @@ module syntax
   | ControlFlow(Token keyword, word? arg_word)
     # Note: There are spids for every pipeline operator, parallel to
     # stderr_indices
-  | Pipeline(Token pipline_op, List[command] children, bool negated, List[int] stderr_indices)
+  | Pipeline(Token pipeline_op, List[command] children, bool negated, List[int] stderr_indices)
   | AndOr(Token op, List[id] ops, List[Token] op_toks, List[command] children)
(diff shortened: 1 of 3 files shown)


frontend/syntax.asdl
```diff
@@ -298,12 +298,12 @@ module syntax
...
   | Sentence(command child, Token terminator)
     # Note: Only represents "bare assignment"
-  | ShAssignment(List[AssignPair] pairs, List[Redir] redirects)
+  | ShAssignment(Token var, List[AssignPair] pairs, List[Redir] redirects)
   | ControlFlow(Token keyword, word? arg_word)
     # Note: There are spids for every pipeline operator, parallel to
     # stderr_indices
-  | Pipeline(List[command] children, bool negated, List[int] stderr_indices)
-  | AndOr(List[id] ops, List[command] children)
+  | Pipeline(Token pipline_op, List[command] children, bool negated, List[int] stderr_indices)
```
andychu, line 305: typo pipeline_op

frontend/syntax.asdl
```diff
@@ -298,12 +298,12 @@ module syntax
...
+  | AndOr(Token op, List[id] ops, List[Token] op_toks, List[command] children)
```
andychu, line 306: I think we can remove `List[id]` and just have `List[Token] ops`

A token contains an Id

osh/cmd_parse.py
```diff
@@ -1422,31 +1427,38 @@ def ParseCase(self):
...
+      in_kw = word_.AsKeywordToken(self.cur_word)
     self._NewlineOk()
 
     if self.c_id != Id.KW_Esac:  # empty case list
-      self.ParseCaseList(case_node.arms)
+      esac_kw = self.ParseCaseList(case_node.arms)
       # TODO: should it return a list of nodes, and extend?
       self._Peek()
+    else:
+        esac_kw = word_.AsKeywordToken(self.cur_word)
```
andychu, line 1446: indentation looks messed up here

(We have an open issue to autoformat the code)

osh/cmd_parse.py
```diff
@@ -1422,31 +1427,38 @@ def ParseCase(self):
...
-    case_node.spids.append(case_spid)
-    case_node.spids.append(in_spid)
-    case_node.spids.append(esac_spid)
+    case_node.case_kw = case_kw
+    case_node.in_kw = in_kw
+    case_node.esac_kw = esac_kw
+
+    case_node.spids.append(case_kw.span_id)
+    case_node.spids.append(in_kw.span_id if in_kw else runtime.NO_SPID)
+    case_node.spids.append(esac_kw.span_id if in_kw else runtime.NO_SPID)
```
andychu, line 1461: looks like a typo -- `esac_kw` ?

I'm surprised that didn't cause any tests to fail?  It looks like `test/ysh-prettify` is doing something with `esac`

osh/cmd_parse.py
```diff
@@ -2182,7 +2198,8 @@ def _ParseAndOr(self):
       if self.c_id not in (Id.Op_DPipe, Id.Op_DAmp):
         break
 
-    node = command.AndOr(ops, children)
+    assert len(op_toks) >= 1, "There must be at least 1 operator token"
+    node = command.AndOr(op_toks[0], ops, op_toks, children)
     node.spids = op_spids
```
andychu, line 2203: It seems like we could remove op_spids and do something like:

```
node.spids = [tok.span_id for tok in op_toks]
```

Although since we're going to get rid of all the `node.spids` soon anyway, maybe it doesn't matter

osh/cmd_parse.py
```diff
@@ -2080,11 +2092,11 @@ def ParsePipeline(self):
     negated = False
 
     # For blaming failures
-    pipeline_spid = runtime.NO_SPID
+    pipeline_op = None  # type: Token
```
andychu, line 2095: Hmm I agree that `!` should be the `Token left_tok` if present

But I'm not sure about `|` otherwise  -- I think it should be the left-most location of the command that the pipeline starts with

andychu (commented):
Thanks!  looks good overall, there a few minor issues we can discuss on the call tomorrow

PossiblyAShrub:
This is ready for another look.

osh/cmd_parse.py
andychu: shouldn't it be `esac_kw.span_id if esac_kw else ...` ?

osh/cmd_parse.py
```diff
@@ -2183,7 +2193,9 @@ def _ParseAndOr(self):
         break
 
     node = command.AndOr(ops, children)
-    node.spids = op_spids
+    for token in ops:
+        # we cannot use a list comprehension as those are not supported by mycpp
```
andychu, line 2197: oh yeah there is a special case for something like

```
tmp = [x for x in mylist if x != 0]
node.spids = tmp
```

Where we turn it into a loop that appends to `tmp`

But yeah the original suggestion might not have worked

This looks fine though

osh/cmd_parse.py
PossiblyAShrub: Ah, sorry. I've fixed this.

The reason why the tests always passed is that `in_kw` and `esac_kw` are either both always defined or both always `None`. So technically the above code is correct, despite it being deceivingly so.

I've pushed a fix anyways.

PR #1606, opened 2023-05-08
[ASDL] Implement the zero_arg_singleton generate option

PossiblyAShrub:


andychu:
Hmm I just pulled and looked at the generated code, looks pretty good

Let me look at the allocation stats

andychu:
OK the reason the allocation stats didn't change is because NoOp and Break and Continue are very rare

We should get some movement on `loc::Missing`, which  we can do separately.  It's actually 4 or 5% of allocations

http://travis-ci.oilshell.org/github-jobs/3934/benchmarks2.wwz/_tmp/uftrace/index.html


```
3 syntax_asdl::loc__Missing* Alloc<syntax_asdl::loc__Missing>()     4119    5.24
```

-----

I looked into the reason why no tests failed when the type tag was incorrect.  This is very subtle!

Shell ignores errors in command subs!  So we fork a process, and try to execute `command.NoOp`, which hits an assertion  failure.

And that failure is **ignored**!   I'm adding at test that will fail

```

$ _bin/cxx-dbg/osh  -c 'echo -$()- ".$()."'; echo $?
osh: _gen/bin/oils_for_unix.mycpp.cc:27079: int cmd_eval::CommandEvaluator::_Dispatch(syntax_asdl::command_t*, runtime_asdl::CommandStatus*): Assertion `false' failed.
osh: _gen/bin/oils_for_unix.mycpp.cc:27079: int cmd_eval::CommandEvaluator::_Dispatch(syntax_asdl::command_t*, runtime_asdl::CommandStatus*): Assertion `false' failed.
-- ..
0
```

andychu:
This is great, thanks!  I think the next step is to do `loc.Missing()`, and then check `benchmarks2/uftrace` and see if the numbers changed


I usually use `sed`, something like `sed -i 's/Missing()/Missing/g' might work

PR #1609, opened 2023-05-11
[core] take TTY before doing I/O from main shell

melvinw:
Also, call tcsetpgrp() in a loop until we are positive we own it. Otherwise we risk getting SIGTTOU or SIGTTIN even though we _should_ have control of the terminal.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -1436,7 +1436,8 @@ class JobControl(object):
       return
 
     try:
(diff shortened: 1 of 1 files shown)


melvinw:
This might be ineffectual! The main shell should be ignoring SIGTTIN and SIGTTOU

PR #1613, opened 2023-05-14
[oil_lang] statically type some more expression evaluators

melvinw:
This patch adds ranges, slices, and tuples to the runtime and 
statically types the expression evaluators for slices, ranges, tuples,
lists, dictionaries, and if-statements. 

It also adds limited stride support to slices and updates a few spec
tests to reflect the latest dictionary semantics, namely: only strings
can be used as keys now (tuples, ints, etc... have been dropped).

### DIFF
--- a/core/runtime.asdl
+++ b/core/runtime.asdl
@@ -4,8 +4,10 @@ module runtime
 {
   # import these types from syntax.asdl
   use frontend syntax {
-    word CompoundWord ArgList command re redir_loc Token proc_sig loc IntParamBox
+    word CompoundWord ArgList command re redir_loc Token proc_sig loc
   }
+
+  IntBox = (int i)
 
   # Evaluating SimpleCommand results in either an argv array or an assignment.
   # in 'local foo', rval is None.
@@ -72,8 +74,8 @@ module runtime
 
     # TODO: we might be able to combine these types when the refactor is done
     # and we don't need to be able to recover their pyobj representations. 
-  | Slice(IntParamBox? lower, IntParamBox? upper, IntParamBox? step)
-  | Range(IntParamBox? lower, IntParamBox? upper, IntParamBox? step)
+  | Slice(IntBox? lower, IntBox? upper, IntBox? step)
+  | Range(IntBox? lower, IntBox? upper, IntBox? step)
 
     # Hack for a PyObject.  # TODO: Remove
   | Obj(any obj)
--- a/core/runtime.asdl
+++ b/core/runtime.asdl
@@ -4,7 +4,7 @@ module runtime
 {
   # import these types from syntax.asdl
   use frontend syntax {
-    word CompoundWord ArgList command re redir_loc Token proc_sig loc
+    word CompoundWord ArgList command re redir_loc Token proc_sig loc IntParamBox
   }
 
   # Evaluating SimpleCommand results in either an argv array or an assignment.
@@ -54,6 +54,8 @@ module runtime
 
     -- TODO: remove MaybeStrArray in favor of this
   | List(List[value] items)
+    # Tuples are like lists, but immutable.
+  | Tuple(List[value] items)
 
     -- TODO: remove AssocArray in favor of this
   | Dict(Dict[str, value] d)
@@ -67,6 +69,11 @@ module runtime
     # - Expr for ^[1 + 2] 
     # - Template for ^"$1 and $2"
     # - Proc, Func
+
+    # TODO: we might be able to combine these types when the refactor is done
+    # and we don't need to be able to recover their pyobj representations. 
+  | Slice(IntParamBox? lower, IntParamBox? upper, IntParamBox? step)
+  | Range(IntParamBox? lower, IntParamBox? upper, IntParamBox? step)
 
     # Hack for a PyObject.  # TODO: Remove
   | Obj(any obj)
--- a/oil_lang/expr_eval.py
+++ b/oil_lang/expr_eval.py
@@ -8,7 +8,7 @@ from _devbuild.gen.id_kind_asdl import Id, Id_t, Kind
 from _devbuild.gen.syntax_asdl import (
     loc, loc_t, re, re_t, Token, word_part, word_part_t,
     SingleQuoted, DoubleQuoted, BracedVarSub, SimpleVarSub, ShArrayLiteral,
-    CommandSub,
+    CommandSub, IntParamBox,
 
     expr, expr_e, expr_t, place_expr, place_expr_e, place_expr_t,
     Attribute, Subscript,
@@ -88,6 +88,11 @@ def Stringify(py_val, word_part=None):
 
   We don't want to tie our sematnics to the Python interpreter too much.
   """
+  # XXX: once verything is typed this should go away and this function should
+  # use tagswitch
+  if isinstance(py_val, value_t):
+    py_val = _ValueToPyObj(py_val)
+
   if isinstance(py_val, bool):
     return 'true' if py_val else 'false'  # Use JSON spelling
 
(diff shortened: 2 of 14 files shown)


andychu:
If anything looks questionable from a language design POV, let me know

Ranges and slices are separate in Python -- ranges have no syntax, but slices do

I think I wanted to unify them, but I'm not sure how that went, and we may want to review

I'm not sure what the spec tests "mandate" at the moment

I put in my best thoughts at the time, but they shouldn't be viewed as "set in stone"

melvinw:
Ah yeah @andychu, the tests are pretty sparse right now. So, now is a good time to make more disruptive changes.

The biggest design thing that will be an issue is subscripting using tuples as indexes/keys. See https://github.com/oilshell/oil/blob/master/oil_lang/expr_eval.py#L462-L468 and https://github.com/oilshell/oil/blob/master/frontend/syntax.asdl#L441. I don't think we'll be able to translate dynamically types and arbitrarily-sized tuples to C++. Judging by the comment in `_EvalIndices()` this was meant for dictionary indexing, but we only support string keys now IIRC. What if we update the subscript syntax to a singular expression instead of a list?

andychu:
Hmm yes I think it makes sense to allow either

```
a[1]
a[1:2]
```

but not

```
a[1, 2]
a[1:2, 3:4]
```

which I think Python allows, mostly for multi-dimensional arrays I think

(We could add it back later perhaps, but it's probably quite low priority ...)

melvinw:
Ok @andychu this is finally ready for a pass! Let me know what you think

oil_lang/expr_eval.py
```diff
@@ -125,10 +130,46 @@ def _PyObjToValue(val):
...
+
+      if val.step:
+        s.step = IntParamBox(val.step)
+
+      return s
+
+  elif isinstance(val, xrange):
+      r = value.Range(None, None, None)
+      # awkward, but should go away once everything is typed...
+      l = list(val)
```
andychu, line 164: maybe call this `li` or `L`, I think `l` looks too close to `1` which we're also using here :)

oil_lang/expr_eval.py
```diff
@@ -1236,19 +1338,143 @@ def _EvalFuncCall(self, node):
...
+              try:
+                return obj.items[index.i]
+              except IndexError:
+                # TODO: expr.Subscript has no error location
+                raise error.Expr('index out of range', loc.Missing)
+
+          else:
+            raise error.InvalidType('expected Slice or Int', loc.Missing)
+
+      elif case(value_e.Tuple):
```
andychu, line 1388: Language design note (I will copy these to Zulip) -- It's  a little weird we have Tuples, but we don't have tuples as `Dict` keys ...

I was wondering if maybe we use tuples mostly for object unpacking, like

```
return x, y

var x, y = Foo()

```

Tuples are kind of non-orthogonal in Python, I wonder if we can have named tuples , and immutable Dicts, etc.

oil_lang/expr_eval.py
```diff
@@ -1236,19 +1338,143 @@ def _EvalFuncCall(self, node):
...
+                # TODO: expr.Subscript has no error location
+                raise error.Expr('index out of range', loc.Missing)
+
+          else:
+            raise error.InvalidType('expected Slice or Int', loc.Missing)
+
+      elif case(value_e.Dict):
+        obj = cast(value.Dict, UP_obj)
+        if index.tag() != value_e.Str:
+            raise error.InvalidType('expected String', loc.Missing)
```
andychu, line 1467: How about 'Expected string index for Dict' to be more precise

And it does remind me we need to add tokens to all the YSH AST nodes, similar to what Aidan and I have been doing for OSH

oil_lang/expr_eval.py
```diff
@@ -1017,18 +1084,42 @@ def _EvalBinary(self, node):
...
+      lower = IntParamBox(cast(value.Int, UP_lower).i)
+
+    if node.upper:
+      UP_upper = _PyObjToValue(self._EvalExpr(node.upper))
+      if UP_upper.tag() != value_e.Int:
+        raise error.InvalidType('Expected Int', loc.Missing)
+
+      upper = IntParamBox(cast(value.Int, UP_upper).i)
+
+    return value.Slice(lower, upper, IntParamBox(1))
```
andychu, line 1122: is the  stride in the value.Slice()  always 1  ?

oil_lang/expr_eval.py
```diff
@@ -1017,18 +1084,42 @@ def _EvalBinary(self, node):
...
 
-    lower = self._EvalExpr(node.lower) if node.lower else None
-    upper = self._EvalExpr(node.upper) if node.upper else None
-    return slice(lower, upper)
+    lower = None # type: Optional[IntParamBox]
+    upper = None # type: Optional[IntParamBox]
+    if node.lower:
+      UP_lower = _PyObjToValue(self._EvalExpr(node.lower))
+      if UP_lower.tag() != value_e.Int:
+        raise error.InvalidType('Expected Int', loc.Missing)
```
andychu, line 1111: I checked and Python gives an error like this

```
>>> a[1:'a']
TypeError: slice indices must be integers or None or have an __index__ method
>>> 

```

oil_lang/expr_to_ast.py
```diff
@@ -140,10 +140,8 @@ def _Trailer(self, base, p_trailer):
       assert p_args.typ == grammar_nt.subscriptlist
       indices = []  # type: List[expr_t]
       n = p_args.NumChildren()
-      for i in xrange(0, n, 2):  # was children[::2]
-        a = p_args.GetChild(i)
-        indices.append(self._Subscript(a))
-      return Subscript(base, indices)
+      a = p_args.GetChild(0)
```
andychu, line 143: If n > 1, we should do something like `p_die("Only 1 subscript is accepted", p_args.GetChild(1).tok)`  which should produce the pretty error highlighting

There is a `test/parse-errors.sh` file that tickles all of these cases

core/runtime.asdl
```diff
@@ -68,6 +70,11 @@ module runtime
     # - Template for ^"$1 and $2"
     # - Proc, Func
 
+    # TODO: we might be able to combine these types when the refactor is done
+    # and we don't need to be able to recover their pyobj representations. 
+  | Slice(IntParamBox? lower, IntParamBox? upper, IntParamBox? step)
+  | Range(IntParamBox? lower, IntParamBox? upper, IntParamBox? step)
```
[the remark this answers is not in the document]
melvinw: Ah sure. I'll add `IntBox` (or `BoxedInt`...) for this

oil_lang/expr_eval.py
melvinw: If they're being constructed with `:`-syntax, yes. But, `slice(...)` might produce one with a different step. This is currently captured in `_PyObjToValue()`, but will eventually be a "first class" thing once we overhaul the builtins

spec/oil-slice-range.test.sh
```diff
@@ -89,21 +89,9 @@ write @x
 5
```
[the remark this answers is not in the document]
melvinw: Good question... It's definitely kinda counter-intuitive. Even in python, I didn't know this was a thing until very recently. If we expand the `:`-syntax for slices to include an optional step, this will become redundant, though

melvinw:
> Thanks, this looks great!
> 
> Looks like it was a lot of work to shuffle the representations back and forth ... I wonder if we still think this is the best strategy?
> 
> I feel like I should jump in and do an EvalX() method and see what it's like ...

Oh, this has actually been working pretty nicely for me. I think it just looks intimidating now because you see everything all at once now. It was slowly built up over the two weeks I've been working on this patch, though. We're also _basically_ done translating the expression evaluator after this! The last big things are builtin-function dispatch and attribute accessors. Once those are done we can start to shed the scaffolding

> Also how after landing this change, I rename `oil_lang/` -> `ysh/` ?

That sounds good to me!

andychu:
Oh you know I actually had this memory of removing `a[x:y:step]`  from the grammar quite awhile ago

Perhaps that was a bad idea

I believe one of the reasons was that `a[x::step]` is the only place in Python where `::` can occur

And I wanted to use `::` for namespacing -- but we're not using that now!

I figured that step is not very common, and could be done with the explicit slice() syntax.

---

One other quirk of Python I realized is that it's awkward for slices to be first class expressions because of then you'd have this issue with the colon

```
if 3:  # expression
  pass
if 3:4:  # hypothetical slice expression then colon
  pass
if 3:4:5:
  pass
```

However we don't have this issue because we don't use colons after statements

----

So anyway, I wouldn't be opposed to adding `a[1:2:3]` back if people want it

However I think we should probably "go wide" with the translation first, and then do another pass to clean up, perhaps also seeing if it makes sense  to unify slice/range

andychu:
OK looks like this is ready now, I'll merge ?

PR #1614, opened 2023-05-16
[oil_lang] New case command syntax tests and impl

PossiblyAShrub:


PossiblyAShrub:
This should be ready for a review.

andychu:
I took a quick look at the code, looks very good !

I will pull the branch and play with it

andychu:
This is great progress, thank you!  I made a note in the commit about what I changed, the main thing was allowing 3 tests to fail with `--osh-failures-allowed` (which I need to move into `test/spec_osh.py`, out of the `test/spec.sh` shell script)

The failures are sort of "TODOs"

Now you're the only other person who has been in the parser !  (as far as I remember, Koiche was pretty deep in the evaluator at one point)

andychu:
Oh I added one spec test case with 2 cases side by side

But I actually think we should have `shopt --unset parse_old_case` in `ysh:all`

Because the new case statement does everything the old one does

That way ysh-only code is consistent.  It's supposed to be a "clean slate" language without legacy

So we can have ParseCase calling

- ParseOldCase
- ParseYshCase (should start calling these YSH instead of Oil)


except that it's just a hard syntax error if not `parse_old_case`, and we end up taking the `ParseOldCase` branch

PR #1619, opened 2023-05-18
[soil/app-tests] Run the full tests of ble.sh

akinomyoga:
The current version 0.15.0 of osh can parse all the modules used in the full tests of ble.sh (with adjustments in branch `osh` of ble.sh), so we can now run the full tests.

### DIFF
--- a/test/ble.sh
+++ b/test/ble.sh
@@ -16,6 +16,8 @@ clone() {
   mkdir -p $out
   git clone --recursive --depth=50 --branch=osh \
     https://github.com/akinomyoga/ble.sh $out
+  git clone --depth=50 \
+    https://github.com/akinomyoga/contra.git "$out"/ext/contra.src
 }
 
 # TODO: What version of osh
@@ -28,6 +30,11 @@ build() {
   # make ble.sh
   cd $BASE_DIR
   make
+
+  # make contra for test
+  cd ext/contra.src
+  make
+  cp src/contra ..
 }
 
 # https://superuser.com/questions/380772/removing-ansi-color-codes-from-text-stream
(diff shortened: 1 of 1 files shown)


andychu:
OK thanks

I invited you to be a contributor, so that way Github Actions will allow your PRs to have the secret to publish results to a web server (mentioned on https://github.com/oilshell/oil/wiki/Contributing)

Then we can see if it passes BEFORE merge.  But I'll merge this one now and we'll see if it passes

If it fails it will  stay in `soil-staging` and won't make it to `master`

andychu:
Merged.   I will  reproduce the  perf issues, thanks for the detailed reports!

andychu:
Ah OK looks like we need libx11-dev in our containers

I will rebuild a container with  that dependency a bit later

```
tx11.cpp:6:10: fatal error: X11/Xlib.h: No such file or directory
 #include <X11/Xlib.h>
          ^~~~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:25: ../out/tx11.o] Error 1
```

http://travis-ci.oilshell.org/github-jobs/4022/#job-app-tests

akinomyoga:
Ah, thank you! So the commit is now dropped. What should I do? For example, can I create a PR on top of  `dev-andy-24`?

Also, `ble.sh --test` returns status `1` unless all the tests pass, which makes the CI fail. I think I'm going to suppress it by specifying `|| true`.

andychu:
Yes I wrote on Zulip I needed these 3 packages for the container:

```
libx11-dev
libxft-dev
libncursesw5-dev
```

The tests are working for me, thanks for adjusting them for OSH!   I want to compare `bash` vs. `osh-cpp` timing in the CI, but it might take some work

`osh-py` works, but it's slow

PR #1626, opened 2023-05-20
 [osh/cmd_parse refactor] Some simplifications to 'case' parsing

andychu:
- Get rid of command.Case.CreateNull() -- just return it instead
- Use separate functions
  - ParseCase calls
    - ParseOldCase -> ParseOldCaseList -> ParseOldCaseItem
    - ParseYshCase -> ParseYshCaseList -> ParseYshCaseItem
    - TODO: should we INLINE the CaseList functions?

- Document metalanguage of POSIX grammar
- Try to synchronize grammar comments and the method bodies
  - This didn't entirely work because of the recursive BNF -> iterative
  EBNF translation

Nonetheless, this method paid off:

- Got rid of extraneous self._Peek() before self._Eat()
  - _Eat() calls self._Peek() first, so it's a no-op
- Got rid of extraneous self._Next() in ParseOldCase
  - it's hard to follow why this doesn't change anything, but it
  doesn't!

### DIFF
--- a/osh/cmd_parse.py
+++ b/osh/cmd_parse.py
@@ -431,7 +431,40 @@ SECONDARY_KEYWORDS = [
 class CommandParser(object):
   """Recursive descent parser derived from POSIX shell grammar.
 
-  TODO: document the metalanguage.
+  This is a BNF grammar:
+  https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_10
+
+  - Augmented with both bash/OSH and YSH constructs.
+
+  - We use regex-like iteration rather than recursive references
+    ?  means  optional (0 or 1)
+    *  means  0 or more
+    +  means  1 or more
+
+  - Keywords are spelled in Caps:
+    If   Elif   Case
+
+  - Operator tokens are quoted:
+    '('   '|'
+
+    or can be spelled directly if it matters:
+
+    Op_LParen   Op_Pipe
+
+  - Non-terminals are snake_case:
+    brace_group   subshell
+
+  Methods in this class should ROUGHLY CORRESPOND to grammar productions, and
+  the production should be in the method docstrings, e.g.
+
+  def ParseSubshell():
+    "
+    subshell : '(' compound_list ')'
+
+    Looking at Op_LParen   # Comment to say how this method is called
+    "
+
+  The grammar may be factored to make parsing easier.
   """
   def __init__(self, parse_ctx, parse_opts, w_parser, lexer, line_reader,
       eof_id=Id.Eof_Real):
@@ -1312,6 +1345,8 @@ class CommandParser(object):
     """
     case_item: '('? pattern ('|' pattern)* ')'
                newline_ok command_term? trailer? ;
+
+    Looking at '(' or pattern
     """
     self.lexer.PushHint(Id.Op_RParen, Id.Right_CasePat)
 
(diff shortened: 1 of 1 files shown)


andychu:
This is basically where we left off on the call

I'd say it's an 80% success on maybe half the case parser, which is pretty decent

We could chat about details on Zulip, or another call whenever you have time

I want to try to do the rest of the typed case in a similar style -- i.e. do a grammar-based design with all the details we talked about


- where does `|` appear
- `(else)`
- composing the expression parser, which we didnt quite talk about

And then

- schema design for location info is consistent, across OSH and YSH, which is probably something we haven't quite worked out in general

andychu:
BTW it's probably easier to read the functions changed, rather than the diff, which is larger than it seems

andychu:
Yeah I was confused about the _Next thing, but all that method does is

```
self.next_lex_mode = lex_mode_e.ShCommand
```

and then `_Peek()` checks if it's not `lex_mode_e.ShCommand`

So you can see that 2 calls to `_Next()` are actually the same as one:

```
self._Next()
self._Next()  # no-op
```

i.e. you NEED a _Peek() to make _Next() actually do anything.  Also 2 peeks are the same as one:

```
self._Peek()
self._Peek()  # no-op
```

It's not always trivial to tell.

Usually what I see is that  _Peek() returns the token ID.   I think the only reason it doesn't is because we have 3 related state variables:

```
self.cur_word
self.c_id
self.c_kind
```

So _Peek() could return a 3-tuple, but I didn't do that.

Anyway this structure is slightly weird, and I'm open to any possible revisions (I've been thinking about it), but it seems like  it works pretty well.

Maybe the names need to be changed.   _Next() especially

----

The origin of _Peek() and _Next() is for the interactive shell and the nested CommandParser/WordParser structure.  The first attempt would "over-read" a newline, which caused problems.

andychu:
Also, I might clean up other parts of the parser to be more consistent  like this

There is definitely other mismatch between grammar comments and the parsing methods

This is partly due the weirdness of translating the POSIX grammar from recursive BNF to iterative EBNF style

That part is sort of an "art"

The POSIX grammar is very accurate but also "contorted" a bit to handle newlines and stuff like optional `;;` in the last `case`, etc.

PossiblyAShrub:
Ok. That makes a lot of sense. Usually I've seen `_Peek` as an idempotent operation like we have here. But the `_Next` behaviour certainly is weird. If it does a good job making interactive parsing easier, then I'll just have to get used to it.

andychu:
Oh yeah another thing I forgot to mention is that I changed ParseYshCaseArm to "cheat" and call `ParseBraceGroup()`, which originates with POSIX shell `{ echo hi; }`, but it also used in many other YSH constructs now

I throw away the resulting BraceGroup AST node and use its fields to populate CaseArm

This seems easier than handling `Lit_LBrace`  in another place

I figure it's better to keep the parser as short as possible to reduce inconsistency, with regard to newlines and all that

----

I think I'm going to inline the short ParseOldCaseItemList and ParseYshCaseItemList functions

Since they're only called from one place, I think it makes it easier to read , and also COMPARE the two functions

i.e. one thing to compare is if they allow empty things like

    case (x) { }
    case $x in esac

We have to allow the latter because of POSIX.  But it's arguably weird and useless ...

andychu:
OK this just landed on master!

Issue #1627, opened 2023-05-20
All types should have bool(x) ,    'foo' or 'bar' should work


metov:
I have a convenience script to make it easier to open VS Codium, which I do very often:
```
#!/usr/bin/env oil
#
# Launch codium.
#
# Without parameters, it will attempt to start with a sensible workspace.
#  * If you are in a git folder, it will open at git root
#  * Otherwise, it will open at the current folder

const GIT_ROOT=$(git rev-parse --show-toplevel 2>/dev/null || true)
const WORK_DIR=$(pwd)

const WS_DIR = GIT_ROOT or WORK_DIR

codium $WS_DIR
```

This script used to work great. However, after a recent update, it now dies with:
```
$ cod
  const WS_DIR = GIT_ROOT or WORK_DIR
        ^~~~~~
'.../cod':12: fatal: Expected Bool
```

Because expression syntax is not yet comprehensively documented, and because it so closely resembles Python, I had assumed that Python-style `may_be_null or fallback_value` would be valid oil syntax, which it was for quite some time. However, coincidentally around the time of the rename, it stopped working, so I guess it's not valid ysh syntax :)

I'm guessing that the issue here is that since I've used `or` the expression evaluates to a boolean. Is it possible to preserve the old behavior of returning the first truthy value like in Python?

andychu:
Thanks this is a good catch, I think it's due to some recent refactoring around divorcing us from the Python interpreter #636 

I'm not sure exactly what we will do -- I like the `or` idiom too, and I agree Python users will expect it

Though I also think there is a wart in Python because    empty string and None or often confused, and it can cause subtle bugs.  I think there were a few of those in Oils itself

For now I think you might have to use a ternary

```
const WS_DIR = GIT_ROOT if len(GIT_ROOT) else WORK_DIR
```

which is longer but at least clear

Although then you might also wonder why it's not:

```
const WS_DIR = GIT_ROOT if GIT_ROOT else WORK_DIR
```

I will look into why this changed

andychu:
I just tested JavaScript and its `||` behaves like Python, so yeah I think we need to retain this idiom

Thanks for the feedback!

in the meantime hopefully the ternary if will work

andychu:
@metov This is now fixed, thank you for the report

I also fixed

- #1605 

and documented

- #1625 

FAQ here - http://travis-ci.oilshell.org/github-jobs/4147/ovm-tarball.wwz/_release/VERSION/doc/ysh-faq.html

[...]

PR #1631, opened 2023-05-25
[ysh language] and / or operators do implicit bool(x) on operands

andychu:
Test that it matches Python.

This is #1627

Related discussion - https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Bool.28x.29.2C.20'foo'.20or.20'bar'

### DIFF
--- a/ysh/expr_eval.py
+++ b/ysh/expr_eval.py
@@ -408,6 +408,59 @@ def AsPosixEre(eggex):
   return eggex.as_ere
 
 
+def ToBool(val):
+  # type: (value_t) -> bool
+  """
+  Convert any value to a boolean.  TODO: expose this as Bool(x), like Python's
+  bool(x).
+  """
+  UP_val = val
+  with tagswitch(val) as case:
+    if case(value_e.Undef):
+      return False
+
+    elif case(value_e.Str):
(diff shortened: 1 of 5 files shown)


andychu:
Someone already noticed recent changes!  (that's good)

It seems that Python and JavaScript agree on all this behavior, so I think we need it for YSH

Comments welcome

A pretty common idiom is `var x = param or 'default'`

ysh/expr_eval.py
```diff
@@ -408,6 +408,59 @@ def AsPosixEre(eggex):
...
+      return len(val.s) != 0
+
+    # OLD TYPES
+    elif case(value_e.MaybeStrArray):
+      val = cast(value.MaybeStrArray, UP_val)
+      return len(val.strs) != 0
+
+    elif case(value_e.MaybeStrArray):
+      val = cast(value.MaybeStrArray, UP_val)
+      return len(val.strs) != 0
```
melvinw, line 433: was one of these supposed to be `AssocArray`?

melvinw (commented):
Thanks! Just one thing inline

andychu:
Ahh great catch, thanks!

andychu:
merged separately

PR #1632, opened 2023-05-25
[treewide] make python indentation consistent

melvinw:
Let's see if this breaks anything...


andychu:
OK great, seems like the docstrings worked!   

How about doing `core/*.py` or reverting those 2 files, just don't have a mixed style in that dir

I'll do the rest of them after that, and fix up the list of files to format

I noticed one Qtt docstring got a little mangled, but it's not a big deal

melvinw:
Ah yes, I thought the `core` thing was odd. Pushed a fixup

andychu:
cool thanks, that was pretty simple!  thanks @PossiblyAShrub for the tip on the docformatter


I will do the other dirs

PR #1633, opened 2023-05-25
[ysh language] Update case parser per new grammar

PossiblyAShrub:
Grammar discussed here: https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/YSH.20Case.20Grammer.


frontend/syntax.asdl
```diff
@@ -279,6 +279,11 @@ module syntax
     Word(word w)
   | YshExpr(expr e)
 
+  pat =
```
andychu, line 282: the types should be capitalized now, e.g. Pat or maybe Matcher

osh/cmd_parse.py
```diff
@@ -1342,21 +1343,26 @@ def ParseWhileUntil(self, keyword):
...
-    Looking at '(' or pattern
+  def _ParsePatWords(self, relaxed_newlines):
+    # type: (bool) -> List[word_t]
     """
-    self.lexer.PushHint(Id.Op_RParen, Id.Right_CasePat)
+    pat_words   : pat_word (newline_ok '|' newline_ok pat_word)*
+    pat_word    : WORD
 
-    left_tok = location.LeftTokenForWord(self.cur_word)  # ( or pat
+    If `relaxed_newlines` is set to False, no newlines are allowed in the
```
andychu, line 1352: It might be better to  duplicate the functions, I think it will only be a few lines, and then they can each have their own grammar

test/parse-errors.sh
```diff
@@ -1200,6 +1200,115 @@ ysh_case() {
...
+      en-US
+    | en-CA
+    | en-UK {
+      echo Hello
+    }
+    fr-FR |
+    fr-CA {
+      echo Bonjour
+    }
+    else {
```
andychu, line 1235: Although it's a bit unusual, I was thinking `(else)` because it's less likely to be confused with `'else'`

The bare `else` word is a bit ambiguous here I think, because say `if` or `elif` are bare words

`(else)` is not a valid expression too, so we can "take over" that syntax for something related


---

On Zulip I suggested the possibility of adding a grammar production

```
matcher = '(' 'else' | testlist ')'
```

andychu (commented):
I can see the difficulty here, we are trying to look at 2 lexer modes at once in some sense

- WORD -- in `lex_mode_e.ShCommand`
- `(` -- could be either `ShCommand` or `Expr`
- `/` -- probably should be parsed with `lex_mode_e.Expr`, because that's how Eggex is parsed


So there are kind of 2 conflicting lexer modes ... Let me think about that

andychu:
It might help to get `word|word2` working first, and then `(expr) | (expr2)`

And then later think about eggex ... I'm sure we can do it (because our model is powerful and flexible), but it does add some complication

andychu:
Hm thinking about this a bit more, an issue is that after

```
case (x) {
```

then we have three possibilities:

1. word
2. (
3. /

But `/` is also part of a word.  I think we might need something special on the WordParser to call from the CommandParser

which has some precedent with LookPastSpace() and LookAheadFuncParens

It might be a custom function that tells us which of those 3 things we're supposed to parse

Perhaps LookAheadCaseArm(...)

?

osh/cmd_parse.py
```diff
@@ -584,14 +585,21 @@ def _Eat(self, c_id, msg=None):
...
     if self.c_id == Id.Op_Newline:
       self._Next()
       self._Peek()
 
+  def _NewlineOk(self):
+    # type: () -> None
+    """Check for optional newline and consume it."""
+    if self.w_parser.LookPastSpace() == Id.Op_Newline:
+      self._Peek()
+      self._Next()
```
andychu, line 601: Hm I would have to go back to my branch (dev-andy-vb8), but I was leaning toward keeping this function like the old one, but removing the last _Next()

I thought that worked, but again we didn't test every case

osh/cmd_parse.py
```diff
@@ -1175,7 +1183,7 @@ def _ParseForExprLoop(self, for_kw):
     self._Peek()
     if self.c_id == Id.Op_Semi:
       self._Next()
-      self._NewlineOk()
+      self._NewlineOkLegacy()
```
andychu, line 1186: Since this change is likely to be big, I think it's better to do the reverse change, e.g. add _NewlineOkNew(), and then flip the names in a subsequent change

spec/ysh-case.test.sh
```diff
@@ -45,6 +49,9 @@ three
 ## END
 
 #### case syntax, complex expressions
+# TODO: make this execute
```
andychu, line 52: Please don't disable failures this way

I did that in the last commit in `test/spec.sh ysh-case` -- it has --osh-failures-allowed 3

That is better so we still see the red/green as a TODO

frontend/lexer.py
```diff
@@ -320,6 +326,16 @@ def PushHint(self, old_id, new_id):
     #log('   PushHint %s ==> %s', Id_str(old_id), Id_str(new_id))
     self.translation_stack.append((old_id, new_id))
 
+  def MoveToNextLine(self):
+    # type: () -> None
+    """For YSH case
+    For lookahead on the next line
```
andychu, line 332: Maybe clarify that LookAhead would return Id.Unknown_Tok without this

frontend/syntax.asdl
```diff
@@ -279,6 +279,12 @@ module syntax
...
+    Else
+  | Words(List[word] w)
+  | YshExprs(List[expr] e)
```
andychu, line 285: naming - I would use plural for the Lists

- w -> words
- e -> exprs

And then `r` could be `re eggex`

osh/cmd_parse.py
```diff
@@ -1395,55 +1437,65 @@ def ParseCaseArm(self):
 
     self._NewlineOk()
 
-    return CaseArm(left_tok, pat_words, middle_tok, action_children, dsemi_tok)
+    return CaseArm(left_tok, pat.Words(pat_words), middle_tok, action_children, dsemi_tok)
 
-  def ParseYshCaseArm(self):
-    # type: () -> CaseArm
+  def ParseYshCaseArm(self, discriminant):
+    # type: (Id_t) -> Optional[Tuple[CaseArm, Id_t]]
```
andychu, line 1443: Although I said the "out param" was a smell, I feel like the Optional[Tuple[...]] is the same issue, i.e. the logic is a bit convoluted

The real "smell" is that _NewlineOkForYshCase has to peek into the lexer, because the WordParser is not able to expose enough info to the CommandParser

ysh/grammar.pgen2
```diff
@@ -597,6 +597,35 @@ re_alt: (re_atom [repeat_op])+
...
+# Patterns are the start of a case arm. Ie,
+#
+# case (foo) {
+#   (40 + 2) | (0) { echo number }
+#   ^^^^^^^^-- This is pattern
+# }
+
+# TODO: remove regex_pat, expr_pat
+regex_pat: regex [re_flags] '/'
+expr_pat: expr ')'
```
andychu, line 609: can these be removed?

andychu (commented):
OK I pulled and tested this out, seems to work well!

The main things TODO are making `(3) | (4)` parse, and executing the rest, which depends on what Melvin is doing.

So we can do that in a followup PR

---

What do you think of the Optional[Tuple[...]] style?  Now that I see it, I think it's harder to follow than the out param

As mentioned, I think the real smell is the CommandParser <-> Lexer interactions

That leads to the out param, but removing the out param doesn't remove the smell

---

I am also going to finish moving all the --osh-allowed-failures to `test/spec_osh.py` , which will hopefully make it more obvious/consistent

osh/cmd_parse.py
PossiblyAShrub: You are right. If we handle the final `newline_ok` in `ParseYshCase` we can simplify this significantly.

PossiblyAShrub:
This PR should be ready for another look.

ysh/grammar.pgen2
```diff
@@ -597,6 +597,31 @@ re_alt: (re_atom [repeat_op])+
...
+# This odd factorization is present so pat_exprs+pat_else can be parsed
+# unambiguously. Furthermore, seperating it from `case_pat` makes it both
+# easier to read and transform in `expr_to_ast.py`
+pat_paren: '(' (pat_else | pat_exprs)
+
+pat_exprs: pat_expr # ([Op_Newline] '|' [Op_Newline] '(' pat_expr)*
+                    # ^^^-- adding all of this seems to break things
+pat_expr: expr ')'
+
+pat_else: 'else' ')'
```
andychu, line 621: I think it can be more readable to  just inline simple patterns like this

It also makes the expr_to_ast.py code a bit simpler too

Ditto for `pat_eggex`, it could be inlined since the `/` is unambiguous

andychu (commented):
Thanks for cleaning up the Id_t, looks good!

PR #1635, opened 2023-05-26
[mycpp] fix Callable -> func pointer translation

melvinw:
There's a bug in the syntax emitted for function pointers (not seemingly used until now). In all contexts, function pointer are declared like variables and given the same name, `f`. This 1) makes it impossible to have more than one function pointer on the stack, and 2) makes it impossible to use function pointers as template parameters. This patch tweaks GetCType() to return a decomposed representation of the type for function pointers. This way callers can reassemble the parts in a context-appropriate manner.

### DIFF
--- a/mycpp/cppgen_pass.py
+++ b/mycpp/cppgen_pass.py
@@ -22,7 +22,7 @@ from mycpp import format_strings
 from mycpp.crash import catch_errors
 from mycpp.util import log
 
-from typing import Tuple
+from typing import Tuple, Union
 
 T = None
 
@@ -176,12 +176,19 @@ def _CheckCondition(node, types):
 
 
 def CTypeIsManaged(c_type):
-    # type: (str) -> bool
+    # type: (Union[str, Tuple[str, str]]) -> bool
     """For rooting and field masks."""
+    if isinstance(c_type, tuple):
+        return False
+
     assert c_type != 'void'
 
     # int, double, bool, scope_t enums, etc. are not managed
     return c_type.endswith('*')
+
+def CTypeIsFuncPointer(c_type):
+    # type: (Union[str, Tuple[str, str]]) -> bool
+    return isinstance(c_type, tuple)
 
 
 def GetCType(t, param=False, local=False):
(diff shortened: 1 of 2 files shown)


melvinw:
@andychu, What do you think? The patch in it's current form is just enough to get the example I added working. So, there are probably a few more places I need to propagate this change to. If you think this makes sense, I'll take care of those before merge

melvinw:
This will come in handy for https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Builtin.20function.20dispatch

andychu:
Ah OK yeah I'm not too surprised function pointers aren't translated well, since we don't use them anywhere

There may have been some function pointers in the beginning, but now most of them are classes

Arguably the function interface should be a class too though

The reason is stuff like `split()` and `glob()` which have methods.  See `ysh/funcs_builtin.py`

I think the `lambda x:` stuff won't translate, we should make those some kind of thing like `_Builtin::Run()`

it could be `_Function::Call()` or something like that

andychu:
Well I should say I haven't thought about it that carefully, but it wouldn't surprise me if a bunch of them end up as classes and not strictly function pointers

But we can also take some gradual steps there, that is fine

I was going to say we may have pure vs. impure functions , e.g. `max()` and `Bool()` are pure, and others aren't

---

BUT I think I wanted to confine funcs to be pure in YSH, to discourage the "N process anti-pattern"

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/N.20process.20anti-pattern

But split() and glob() are pure, but they still have some other context like options

Like split() may respect IFS, and glob() may respect `shopt -s nullglob`, and that sort of thing

So yeah there may be some language design issues here

melvinw:
Yeah, I think many functions in `ysh/funcs_builtin.py` (almost everything in `Init()`) look pretty pure and should be compatible with the function pointer approach.

We'll definitely need to come up with some kind of wrapper for the lambdas and the things in `Init2()`/`Init3()`. We could maybe use `std::function` and the lambda expressions added in C++11, but those rely on destructors for assign/move and that might get hairy if we aren't careful (this bit me on an earlier version of this patch when I tried to ditch function pointers all-together for `std::function`...).

andychu:
Oops sorry I realized I didn't respond to this ... Let's talk about it tomorrow

PR #1644, opened 2023-06-02
xxx funcs

melvinw:


ysh/funcs_builtin.py
```diff
@@ -192,6 +195,39 @@ def Init3(mem, config_parser, eval_to_dict, block_as_str, hay_result):
...
+
+    def __init__(self):
+        # type: () -> None
+        vm._Func.__init__(self)
+
+    def Run(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+        assert len(pos_args) == 1
+        assert len(named_args) == 0
+        return value.Bool(expr_eval.ToBool(pos_args[0]))
```
andychu, line 208: Yeah actually this is a case where we need an internal ToBool() to implement `and` and `or` operators

But then we ALSO have a wrapper Bool(vm._Func)  to expose `Bool()` to the user

So it makes sense

ysh/funcs_builtin.py
```diff
@@ -192,6 +195,39 @@ def Init3(mem, config_parser, eval_to_dict, block_as_str, hay_result):
...
+        assert len(pos_args) == 2
+        assert len(named_args) == 0
+
+        UP_s = pos_args[0]
+        UP_prefix = pos_args[1]
+        assert UP_s.tag() == value_e.Str
+        assert UP_prefix.tag() == value_e.Str
+        s = cast(value.Str, UP_s)
+        prefix = cast(value.Str, UP_prefix)
+        return value.Bool(s.s[:len(prefix.s)] == prefix.s)
```
andychu, line 228: actually I realized here we could just use `s.startswith(prefix.s)`

that will translate fine, since our `Str` implementation has it

melvinw: ah right. good catch

ysh/funcs_builtin.py
```diff
@@ -192,6 +195,39 @@ def Init3(mem, config_parser, eval_to_dict, block_as_str, hay_result):
...
+        vm._Func.__init__(self)
+
+    def Run(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+        assert len(pos_args) == 2
+        assert len(named_args) == 0
+
+        UP_s = pos_args[0]
+        UP_prefix = pos_args[1]
+        assert UP_s.tag() == value_e.Str
```
andychu, line 224: Another issue is what to do with the argument errors

Python tends to raise `TypeError`, so I guess we can use `raise error.InvalidType()` ?

It is going to get pretty verbose so yeah we may end up with something like Python's ParseArgs()

Right now it does stuff like

```
$ bin/ysh -c '= 1->startswith("foo")'
  = 1->startswith("foo")
  ^
[ -c flag ]:1: fatal: Expression eval error: 'int' object has no attribute 'startswith'
andy@lenny:~/git/oilshell/oil$ bin/ysh -c '= 1->startswith("foo")'
  = 1->startswith("foo")
  ^
[ -c flag ]:1: fatal: Expression eval error: 'int' object has no attribute 'startswith'
andy@lenny:~/git/oilshell/oil$ bin/ysh -c '= "foo"->startswith("foo")'
(Bool)   True
andy@lenny:~/git/oilshell/oil$ bin/ysh -c '= "foo"->startswith()'
  = "foo"->startswith()
  ^
[ -c flag ]:1: fatal: Type error in expression: startswith() takes at least 1 argument (0 given)

```

core/util.py
```diff
@@ -75,3 +80,47 @@ def isatty(self):
         # type: () -> bool
         """Used by node.PrettyPrint()."""
         return self.f.isatty()
+
+
+def MustCastInt(val):
```
andychu, line 85: naming: how about MustBeInt(), MustBeStr() ?

ysh/expr_eval.py
```diff
@@ -1564,18 +1552,28 @@ def _EvalSubscript(self, node):
...
+        obj = _PyObjToValue(self._EvalExpr(node.obj))
+        UP_obj = obj
         id_ = node.op.id
         if id_ == Id.Expr_RArrow:
             # Used for s->startswith(x)
-            name = node.attr.tval
-            return getattr(o, name)
+            func_name = node.attr.tval
+            func = self.LookupVar(func_name, node.attr)
+            # If there are collisions on method names for different types, mem
```
andychu, line 1563: Yeah this is probably a good idea

we may need to change it for user-defined functions/methods too

ysh/funcs_builtin.py
```diff
@@ -21,121 +22,195 @@
...
-        return []
+    def Run(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+        assert len(pos_args) == 2
 
-    # TODO: Need proper span IDs
-    if not isinstance(obj, str):
-        raise error.Expr('maybe() passed arg of invalid type %r' %
-                         obj.__class__.__name__)
+        # XXX: python's join() accepts dict keys, strings, and other iterables.
```
andychu, line 48: Yeah this is a good question ... We don't have any concept of "iterators"

So it may be that users need to call `join(d->keys(), '')` and so forth

Not `''.join(d)` like in Python, which is shorter, but arguably a little implicit and hard to learn

andychu (commented):
This looks great overall

BTW Aidan expressed some interest in helping

It's probably a good idea if both he and I do some of the functions

Just so we all agree on the idioms and so forth

melvinw:
much of this is defunct now. gonna close

PR #1646, opened 2023-06-02
Fix extra newlines bug (case/while/until/if)

PossiblyAShrub:


PossiblyAShrub:
This is ready to be looked at.

osh/cmd_parse.py
```diff
@@ -1399,6 +1399,23 @@ def ParseFor(self):
             n2.redirects = self._ParseRedirectList()
             return n2
 
+    def _ParseConditionList(self):
+        # type: () -> condition_t
+        """
+        condition_list: command_list
+        Looking at `command_list`
```
andychu, line 1406: I think this "looking at" comment should name a `self.cur_word` if it exists

osh/cmd_parse.py
```diff
@@ -1590,6 +1604,11 @@ def ParseOldCase(self, case_kw):
                     "This is a constant string.  You may want a variable like $x (parse_bare_word)",
                     loc.Word(w))
 
+        if w.tag() == word_e.Token:
+            token = cast(Token, w)
+            if token.id == Id.Op_Newline:
```
andychu, line 1609: I think we should disallow ANY `word_e.Token` -- we can consider renaming it to `word_e.Operator` to make that clear

An operator is a character like `() | ;` roughly  where `echo |` doesn't echo the character

This is in contrast to the CompoundWord.  The case pattern should be a CompoundWord


I tested this 

```
case |
in
esac
```

and it is similarly wrong in OSH


Similarly with `case ;` etc.

Please add one of those in `test/parse-errors.sh` too

andychu (commented):
Nice find!  Funny thing is I tested bash, dash, mksh, zsh on the

```
while
do
done
```

case, and bash and dash give syntax errors.  But mksh and zsh run in infinite loops (and can't be aborted!)

I actually wonder if this is a problem with the POSIX grammar, though I might not have time to look

andychu:
FWIW the "operator" term comes from the POSIX shell

andychu:
I checked and it appears the grammar doesn't allow an empty condition list, so we transcribed the grammar incorrectly ( as well as zsh and mksh)

I think this fix makes sense


```



if_clause        : If compound_list Then compound_list else_part Fi
                 | If compound_list Then compound_list           Fi


compound_list    : linebreak term
                 | linebreak term separator


term             : term separator and_or
                 |                and_or

and_or           :                         pipeline


pipe_sequence    :                             command


command          : simple_command
                 | compound_command
                 | compound_command redirect_list
                 | function_definition

```

osh/cmd_parse.py
PossiblyAShrub: Good catch! I wouldn't be opposed to renaming `word_e.Token` -> `word_e.Operator` in another PR.

osh/cmd_parse.py
PossiblyAShrub: Not quite sure what I should put here then. The `command_list` could start with a variety of different words.

Should we just not have a "looking at" here?

PossiblyAShrub:
> I checked and it appears the grammar doesn't allow an empty condition list, so we transcribed the grammar incorrectly ( as well as zsh and mksh)
> 
> I think this fix makes sense
> 
> ```
> 
> 
> 
> if_clause        : If compound_list Then compound_list else_part Fi
>                  | If compound_list Then compound_list           Fi
> 
> 
> compound_list    : linebreak term
>                  | linebreak term separator
> 
> 
> term             : term separator and_or
>                  |                and_or
> 
> and_or           :                         pipeline
> 
> 
> pipe_sequence    :                             command
> 
> 
> command          : simple_command
>                  | compound_command
>                  | compound_command redirect_list
>                  | function_definition
> ```

Yes, it seems to agree with the POSIX grammar.

I thought the leading newline in the `compound_list` seemed a little weird, but from what I can tell that is actually correct. Testing this in other shells and it seems that only zsh handles this correctly! 

```
bash$ if
> true
> then
> fi
bash: syntax error near unexpected token `fi'
```

*Note the newline after `if`*.

osh/cmd_parse.py
andychu: Sure you can remove it if it's not helpful

osh/cmd_parse.py
```diff
@@ -1590,6 +1606,9 @@ def ParseOldCase(self, case_kw):
                     "This is a constant string.  You may want a variable like $x (parse_bare_word)",
                     loc.Word(w))
 
+        if w.tag() == word_e.Token:
```
andychu, line 1609: Let's write  != word_e.Compound, because we're expecting the CompoundWord for the case pattern


Those things are equivalent now, but they might not always be

andychu:
Great thanks!

PR #1651, opened 2023-06-10
[ysh language] Parse multiple expression case patterns

PossiblyAShrub:
Ie. this can now be parsed:

```
case (x) {
  (1) | (2) {
    echo number
  }
}
```

### DIFF
--- a/osh/cmd_parse.py
+++ b/osh/cmd_parse.py
@@ -1161,6 +1161,11 @@ class CommandParser(object):
         ate = self._Eat(Id.Lit_LBrace)
         left = word_.BraceToken(ate)
 
+        return self._ParseBraceGroupAfterLBrace(left)
+
+    def _ParseBraceGroupAfterLBrace(self, left):
+        # type: (Token) -> BraceGroup
+
         # PROBLEM: can't use 'ate' here because osh/word_parse.py has a hack for
         # proc { } to translate Token Id.Op_RBrace -> Id.Lit_LBrace.
 
@@ -1515,6 +1520,9 @@ class CommandParser(object):
         if discriminant in (Id.Op_LParen, Id.Arith_Slash):
             # pat_exprs, pat_else or pat_eggex
             pattern = self.parse_ctx.ParseYshCasePattern(self.lexer)
+
+            # TODO: how to get this?
+            left = self.w_parser.cur_token
         else:
             # pat_words
             pat_words = []  # type: List[word_t]
@@ -1535,8 +1543,11 @@ class CommandParser(object):
                     break
             pattern = pat.Words(pat_words)
 
+            ate = self._Eat(Id.Lit_LBrace)
+            left = word_.BraceToken(ate)
+
         self._NewlineOk()
-        action = self.ParseBraceGroup()
+        action = self._ParseBraceGroupAfterLBrace(left)
 
         # The left token of the action is our "middle" token
         return CaseArm(left_tok, pattern, action.left, action.children,
--- a/osh/cmd_parse.py
+++ b/osh/cmd_parse.py
@@ -1161,11 +1161,6 @@ class CommandParser(object):
         ate = self._Eat(Id.Lit_LBrace)
         left = word_.BraceToken(ate)
 
-        return self._ParseBraceGroupAfterLBrace(left)
-
-    def _ParseBraceGroupAfterLBrace(self, left):
-        # type: (Token) -> BraceGroup
-
         doc_token = None  # type: Token
         self._GetWord()
         if self.c_id == Id.Op_Newline:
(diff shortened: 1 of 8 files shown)


ysh/grammar.pgen2
```diff
@@ -597,19 +597,15 @@ regex: [re_alt] (('|'|'or') re_alt)*
 case_pat: (
```
andychu, line 597: The comment above this with ^^^^^^ seems a bit confusing, since `(0)` is is also a pattern

osh/cmd_parse.py
```diff
@@ -1161,6 +1161,11 @@ def ParseBraceGroup(self):
         ate = self._Eat(Id.Lit_LBrace)
         left = word_.BraceToken(ate)
 
+        return self._ParseBraceGroupAfterLBrace(left)
+
+    def _ParseBraceGroupAfterLBrace(self, left):
```
andychu, line 1166: Another way to factor this is to use have ParseBraceGroup take an Optional[Token] left

And then if it's passed, it will skip the first 2 lines

What do you think?

( Aside: I noticed  we parse doc comments ### in any brace group, whereas they're only for function bodies  ... we might want to clean that up)

osh/cmd_parse.py
```diff
@@ -1161,6 +1161,11 @@ def ParseBraceGroup(self):
...
+        # type: (Token) -> BraceGroup
+
         # PROBLEM: can't use 'ate' here because osh/word_parse.py has a hack for
```
andychu, line 1169: this comment is obsolete and can be deleted

ysh/grammar_gen.py
```diff
@@ -191,24 +192,26 @@ def main(argv):
 
         is_expr = grammar_name in ('calc', 'grammar')
 
-        parse_opts = optview.Parse()
+        parse_opts = optview.Parse([], [])
         parse_ctx = parse_lib.ParseContext(arena, parse_opts, {}, gr)
-        p = expr_parse.ExprParser(parse_ctx, gr)
+        p = expr_parse.ExprParser(parse_ctx, gr, False)
```
andychu, line 197: Thanks for fixing this, I think I used it when starting the pgen2 integration, and then stopped using it

I probably will use it when parsing func and proc signatures

ysh/grammar.pgen2
```diff
@@ -597,19 +597,15 @@ regex: [re_alt] (('|'|'or') re_alt)*
...
     pat_paren
```
andychu, line 598: I think it would be more readable to inline pat_paren, like

```
ysh_case_pat =
  '(' ( 'else' ')' | pat_exprs )
| pat_eggex
```

So we see the 3 alternatives in one place

andychu (commented):
Sorry I didn't realize this was ready to look at yesterday!

I put a bunch of suggestions about style, but you can use your discretion if you don't think they are improvements

I think some of the comments should be cleaned up though

ysh/expr_to_ast.py
```diff
@@ -791,11 +791,13 @@ def YshCasePattern(self, pnode):
             if typ == grammar_nt.pat_else:
                 return pat.Else
             elif typ == grammar_nt.pat_exprs:
-                # TODO: recursively extract all expressions
-                #       but this is fine for now as we only parse one expression
-                e = pattern.GetChild(0).GetChild(0)
-                expr = self.Expr(e)
-                return pat.YshExprs([expr])
+                exprs = []  # type: List[expr_t]
```
andychu, line 794: > We need to descriminate against

This comment is incomplete, could just remove it

andychu:
Oh another thing I noticed is that `oil_proc` with the trailing `{` uses `self.w_parser.ParseProc`, which sets `self.buffered_word` for the `{`.   (That's actually what causes the BraceToken() issue, which is solved now)


And a bunch of the other ones do too

This avoids messing up `ParseBraceGroup()`

I think using that style would be more consistent

There is a lot of indirection, from CommandParser -> WordParser -> parse_ctx, but it does keep the parser cleaner

ysh/grammar.pgen2
PossiblyAShrub: Sure. I decided to leave the separate `pat_else` because I don't think it reduces readability while making the translation more explicit. Now we have:

```
ysh_case_pat: (
    '(' (pat_else | pat_exprs)
  | pat_eggex
)
```

If you still think it would be better to embed the `pat_else`, then we can change that.

andychu:
FWIW I was brainstorming some cleanups to the parser, and ~5 problems we ran into, mainly related to transitioning in and out of CommandParser  <-> ExprParser, lookahead, and the shared lexer

Long term I think there is a possible solution to "pass tokens as positions" between parsers

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Smells.20in.20osh.2Fcmd_parse.2Epy


But for this change I think we should follow the ParseProc style to be consistent, which means `ParseBraceGroup()` will be unaffected

The one token of fudge is handled in the WordParser with self.buffered_word

PossiblyAShrub:
I agree that it would be nice to find a more principled approach to solving all of these issues. The "tokens as positions" idea looks interesting, we will have to chat about it more on zulip or maybe over a call.

But for now I went with the `buffered_word` approach like we do with `ParseProc`.

I think I have resolved all of the issues you brought up. So this should be good for another look over.

andychu (commented):
Looks great, thanks!

andychu:
BTW I'm finishing up a good blog post now, after it's published we can review the initial design, with Melvin and anyone else interested

PR #1659, opened 2023-06-22
Use READLINE_DIR value in flags in ninja-rules-cpp

### LINKED ISSUE #1657: READLINE_DIR is not used in build/ninja-rules-cpp.sh
If you pass `--readline=/path/to/readline` to `configure`, then `build/ninja-rules.cpp.sh` should add `-I/path/to/readline/include` to `flags` and `-L/path/to/readline/lib` to `link_flags`. This can be accomplished by checking for the `READLINE_DIR` variable from `_build/detected-config.sh`.

pdebelak:
This should fix #1657

### DIFF
--- a/build/ninja-rules-cpp.sh
+++ b/build/ninja-rules-cpp.sh
@@ -24,7 +24,7 @@ REPO_ROOT=$(cd "$(dirname $0)/.."; pwd)
 
 . build/common.sh  # for $BASE_CXXFLAGS
 
-# for HAVE_READLINE
+# for HAVE_READLINE and READLINE_DIR
 if ! . _build/detected-config.sh; then
   die "Can't find _build/detected-config.sh.  Run './configure'"
 fi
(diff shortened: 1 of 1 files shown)


andychu:
Great thank you!  I think this is respected by the Python tarball, but never made it over to the C++ tarball

andychu:
BTW we make a tarball at every commit, so you can go here

http://travis-ci.oilshell.org/github-jobs/4250/

1. click "cpp-small"
2. make-tar -> RESULTS -- this is a link to an uncompressed tarball

http://travis-ci.oilshell.org/github-jobs/4250/cpp-small.wwz/_release/oils-for-unix.tar

Also I will invite you as a collaborator, so if you send a PR from `oilshell/oil`, our CI will run, and you can see tarballs and results BEFORE merge

Unfortunately this is  a Github limitation -- it doesn't allow CI secrets for forks, which we use to publish  (details on Zulip, I would like to fix this, but it seems impossible with github)

PR #1660, opened 2023-06-23
Properly Evaulate Case Expressions Patterns and Else Patterns

PossiblyAShrub:
Most of the ysh-case tests are passing! Now I just need to figure out how to evaluate and then match against eggexs.

Unfortunately I wasn't able to find a way to evaluate the expressions into a `value_t`. So I am not sure if the expression evaluator has progressed far enough in that respect.


andychu:
Oh is this ready to look at?  sorry looks like I introduced some conflicts

PossiblyAShrub:
> Oh is this ready to look at? sorry looks like I introduced some conflicts

Not quite ready to look at.

Right now the following doesn't work:

```
case (4) {
  (4) { echo should print }
}
```

This is due to `value_t.==` not holding `x == y <==> value.X(x) == value.X(y)`. I need to find where we implement comparison in ysh/expr_eval and use that instead.

I don't think the diff will change much if you want to look over it. But I hope to have this PR done later today.

andychu:
Oh I'm working in that area now... not sure if this is the problem but we have value.Obj(42) vs. value.Int(42) and so forth

That's what I'm trying to get rid of with the "PyObject -> value_t rewrite", thread on Zulip

PossiblyAShrub:
> Oh I'm working in that area now... not sure if this is the problem but we have value.Obj(42) vs. value.Int(42) and so forth
> 
> That's what I'm trying to get rid of with the "PyObject -> value_t rewrite", thread on Zulip

Ok, great! I suppose we could merge this right away then, just need to resolve the conflicts.

This PR is still progress as it:
 - Evaluates case arguments and expression patterns to `value_t` objects
 - Implements `(else)` pattern evaluation

We can finish `YshExprs` and `Eggex` pattern evaluation once "PyObject -> value_t rewrite" has progressed enough.

andychu:
I just landed this cleanup, which affects this PR

https://github.com/oilshell/oil/commit/7696693413d427b3e9f9c99c80c56ebfd066a81f

There is no val_ops.ExactlyEqual() and val_ops.RegexMatch()

One thing we should think about is the global match variable, which is accessed through `_match(1)` 

`_start(1)` etc.

PossiblyAShrub:
Awesome! That helps a lot.

I might as well implement the `YshExprs` and `Eggex` pattern evaluation then.

PossiblyAShrub:
Oops! I accidentally clicked closed.

PossiblyAShrub:
This is now complete and should be ready for review.

Thanks to your work on the expression evaluator, I was also able to implement evaluation for `pat.YshExprs` and `pat.Eggex`.

andychu:
Looks good!  I updated a comment in `spec/ysh-case`

I noticed we don't have test coverage for

- `(else)` 
- You should be able to use `_match(0)` with eggex now

e.g.

```
var x = 'foo.py'
case (x) in
  / <dot*> '.py' / { echo "match is $[_match(0)]" }
  / <dot*> '.cc' / { echo "match is $[_match(0)]" }
}
```

How about adding some test cases for those?

andychu:
Also I'm working on translating `ysh/expr_eval.py`, so we can remove all the `mylib.PYTHON` .   That will be a relief!

PossiblyAShrub:
Good point on tests. I added some more over the cases that you mentioned, plus some others over the execution semantics that I discussed on Zulip.

andychu:
Great thanks!

Issue #1272, opened 2022-08-12
doesn't build on netBSD


isomorphisms:
```
isomorphisms@sdf:~/oil-0.12.3$ ./configure --without-readline  --prefix ~ --datarootdir ~/.local/share
./configure: Wrote _build/detected-config.sh and _build/detected-config.h
isomorphisms@sdf:~/oil-0.12.3$ make
make: "/sdf/arpa/gm/i/isomorphisms/oil-0.12.3/Makefile" line 75: Need an operator
make: "/sdf/arpa/gm/i/isomorphisms/oil-0.12.3/Makefile" line 145: Need an operator
make: "/sdf/arpa/gm/i/isomorphisms/oil-0.12.3/Makefile" line 162: Need an operator
make: "/sdf/arpa/gm/i/isomorphisms/oil-0.12.3/Makefile" line 172: warning: duplicate script for target "_build/%/ovm-opt.stripped" ignored
make: "/sdf/arpa/gm/i/isomorphisms/oil-0.12.3/Makefile" line 160: warning: using previous script for "_build/%/ovm-opt.stripped" defined here
make: "/sdf/arpa/gm/i/isomorphisms/oil-0.12.3/Makefile" line 174: Need an operator
make: Fatal errors encountered -- cannot continue
make: stopped in /sdf/arpa/gm/i/isomorphisms/oil-0.12.3
```

andychu:
This is probably because it's not GNU make?

I mention GNU make and GNU bash here: https://www.oilshell.org/release/0.12.3/doc/INSTALL.html

---

But those should go away eventually with `oil-native`, which is written so you don't need ANY Make, just POSIX `sh` (although this could use more testing).   If you care about that you can try out the `oil-native` tarball, which is not yet a complete shell.


Thanks for trying it, and for the report!

isomorphisms:
I actually tried `native` first. Just re-ran it with:

```
isomorphisms@sdf:~/oil-native-0.12.3$ build/native.sh tarball-demo

_build/oil-native.sh: Building oil-native: _bin/cxx-opt-sh/osh_eval

CXX cpp/leaky_core.cc
cpp/leaky_core.cc: In function 'Dict<Str*, Str*>* pyos::Environ()':
cpp/leaky_core.cc:74:21: error: 'environ' was not declared in this scope
   for (char** env = environ; *env; ++env) {
                     ^
cpp/leaky_core.cc: In function 'void pyos::PrintTimes()':
[...]

PR #1661, opened 2023-06-23
Compatibility changes for building oil-native on macos

### LINKED ISSUE #1658: --gc-sections not supported by ld on macos
This is set as part of `link_flags` in `build/ninja-rules-cpp.sh` but throws an error on macos. Apparently the equivalent option is `-dead_strip`.

pdebelak:
This includes a change to check for `unistd.h` defining `environ` and one for the correct `ld` flags to use. This has currently only been tested on macos, so confirmation that this doesn't break linux would be welcome.

This should fix #1658 and at least the `environ` part of #1272.

### DIFF
--- a/build/ninja-rules-cpp.sh
+++ b/build/ninja-rules-cpp.sh
@@ -24,7 +24,7 @@ REPO_ROOT=$(cd "$(dirname $0)/.."; pwd)
 
 . build/common.sh  # for $BASE_CXXFLAGS
 
-# for HAVE_READLINE and READLINE_DIR
+# for HAVE_READLINE, READLINE_DIR, and STRIP_FLAGS
 if ! . _build/detected-config.sh; then
   die "Can't find _build/detected-config.sh.  Run './configure'"
 fi
@@ -50,7 +50,7 @@ link_flags=''  # link flags
 setglobal_cxx() {
   local compiler=$1
 
-  case $compiler in 
+  case $compiler in
     (cxx)   cxx='c++'    ;;
     (clang) cxx=$CLANGXX ;;
     (*)     die "Invalid compiler $compiler" ;;
@@ -226,7 +226,9 @@ setglobal_link_flags() {
       ;;
   esac
 
-  link_flags="$link_flags -Wl,--gc-sections"
+  if test -n "$STRIP_FLAGS"; then
+    link_flags="$link_flags -Wl,$STRIP_FLAGS"
+  fi
 }
 
 compile_one() {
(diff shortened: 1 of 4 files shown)


pdebelak:
I was successfully able to build from the file at http://travis-ci.oilshell.org/github-jobs/4258/cpp-small.wwz/_release/oils-for-unix.tar so this appears to fix the issues I found on macos.

build/ninja-rules-cpp.sh
```diff
@@ -226,7 +226,9 @@ setglobal_link_flags() {
       ;;
   esac
 
-  link_flags="$link_flags -Wl,--gc-sections"
+  if test -n "$STRIP_FLAGS"; then
```
andychu, line 229: I think we have to use `${STRIP_FLAGS:-}`, similar to `${OIL_NINJA_VERBOSE:-}`, because we `set -u` in this file

so this will probably fail if neither of the  STRIP_FLAGS is detected

configure
```diff
@@ -249,6 +249,16 @@ echo_shell_vars() {
   fi
   echo "PREFIX=$FLAG_prefix"
   echo "DATAROOTDIR=$FLAG_datarootdir"
+  cat >$TMP/ld_flags.c <<EOF
```
andychu, line 252: how about using `build/detect-cc.c` in place of this tmp file?  

(There are a lot of $TMP files I want to get rid of, which are mainly for the CPython build, which will eventually be obsolete)

cpp/core.cc
```diff
@@ -20,6 +20,10 @@
 
 #include "_gen/frontend/consts.h"  // gVersion
 
+#ifndef HAVE_ENVIRON // handle systems that don't include environ in unistd
+extern char **environ;
```
andychu, line 24: It is weird that OS X doesn't have this

I think it should be OK to unconditionally declare `extern char **environ`

andychu (commented):
Thanks for doing this, and glad you got the CI environment working!

build/ninja-rules-cpp.sh
pdebelak: Makes sense, updated.

configure
pdebelak: Yeah, that works since I don't care about the contents of the file being compiled. Updated.

cpp/core.cc
pdebelak: Okay, that let's us remove the check for its definition as well. As a note, it does appear that the `configure-coreutils` file included in this repo as testdata ends up setting a [`HAVE_ENVIRON_DECL` variable in it](https://github.com/oilshell/oil/blob/master/benchmarks/testdata/configure-coreutils#L10315) but I think you're right that setting an external var there should do no harm.

PR #1668, opened 2023-07-03
[ysh] Builtin Methods

PossiblyAShrub:


PossiblyAShrub:
This PR should be ready to be looked over. I might want to add some more tests, but I think the functionality is all good.

library/util.py
```diff
@@ -0,0 +1,47 @@
...
+from core import error
+
+from typing import TYPE_CHECKING, Dict, List
+
+class ArgsSpec(object):
+    """Utility to express argument specifications (runtime typechecking)."""
+
+    def __init__(self, pos_args, named_args):
+        # type: (List[int], Dict[str, int]) -> None
+        """Empty constructor for mycpp."""
```
andychu, line 18: this comment can be removed

library/util.py
```diff
@@ -0,0 +1,47 @@
...
+        expected = len(self.named_args)
+        if len(named_args) != 0:
+            raise error.InvalidType("%s() expects %d named arguments but %d were given" % (func_name, expected, nargs), loc.Missing)
+
+        for i in xrange(len(pos_args)):
+            expected = self.pos_args[i]
+            got = pos_args[i].tag()
+            if got != expected:
+                msg = "%s() expected %s but received %s" % (func_name, value_str(expected), value_str(got))
+                raise error.InvalidType(msg, loc.Missing)
```
andychu, line 40: I added `error.InvalidType2` for these kinds of "expected X but got Y" type errors

At some point we should consolidate them all, doesn't have to be now

library/func_misc.py
```diff
@@ -75,3 +76,80 @@ def Call(self, pos_args, named_args):
...
+
+    def __init__(self):
+        # type: () -> None
+        """Empty constructor for mycpp."""
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        spec = ArgsSpec([value_e.Str], {})
```
andychu, line 111: Now I see `{}` and see GC pressure in the common case :)    i.e. we could use None instead, even though that sorta uglifies the code

spec/ysh-methods.test.sh
```diff
@@ -0,0 +1,46 @@
+# spec/ysh-case
```
andychu, line 1: update to spec/ysh-methods

ysh/expr_eval.py
```diff
@@ -1124,16 +1143,14 @@ def _EvalAttribute(self, node):
...
-                    else:
-                        raise AssertionError()
-                else:
-                    raise error.InvalidType('Method %r' % name, loc.Missing)
+            ty = o.tag()
+
+            recv = self.methods.get(ty)
+            method = recv.get(name) if recv else None
+            if not method:
+                raise error.InvalidType('Unbound method %r' % name, node.attr)
```
andychu, line 1151: I would say something like "Method X doesn't exist on type Y"

andychu (commented):
Thanks, this looks great!  Sorry for the delay, somehow I didn't see it yesterday

library/func_misc.py
andychu: We can put this off for later, but I have a feeling we'll do it eventually

core/shell.py
```diff
@@ -287,6 +288,17 @@ def AddBlock(builtins, mem, mutable_opts, dir_stack, cmd_ev, shell_ex,
             hay_state, mem, cmd_ev)
 
 
+def AddMethods(methods):
+    # type: (Dict[int, Dict[str, vm._Callable]]) -> None
+    """Initialize methods table."""
+    methods[value_e.Str] = {
+        'startswith': func_misc.StartsWith(),
```
andychu, line 295: Another thing we should think about is whether it should be `s->startsWith("a")`

JavaScript uses that spelling

It's kind annoying to have it inconsistent with Python, but there will probably be more functions later which we'll want to be consistent with

I think right now it's `eval_hay()` and `parse_hay()`, and those should probably be `evalHay()` and `parseHay()`

We don't have to decide now, it's probably better to make a pass through everything, and then make these decisions when we **write the docs**

library/func_misc.py
PossiblyAShrub: That's a good point. I'll leave this for now; the `ArgSpec` utility will likely have to be updated later as it cannot, for example, express method signatures for variadic functions.

core/shell.py
PossiblyAShrub: Yes, I originally went with `startsWith` but found that `spec/ysh-user-feedback.test.sh` used `startswith` instead. So I decided to follow the precedent (and avoid a breaking change, if that even matters right now.)

Eventually, I imagine that we'd want to create our own style guide. Sort of like Python's PEP8.

PossiblyAShrub:
I believe I have addressed all of your feedback. This should be good for another look over.

andychu:
Great thanks!  I started a thread on Zulip about making `max()` work, or maybe `identity()` is a better start even

I think the first thing is that the `func` parsing is hidden behind `shopt -s  parse_tea` -- might be trivial to remove, but I forget what state it's in now

I guess ping me there if you have questions

library/func_misc.py
```diff
@@ -75,3 +74,76 @@ def Call(self, pos_args, named_args):
...
+class StartsWith(vm._Callable):
+
+    def __init__(self):
+        # type: () -> None
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        spec = ArgsSpec([value_e.Str, value_e.Str], {})
```
andychu, line 88: Actually one thing I noticed is that you have to do `value_e.Str` here, and then `cast(value.Str` later

I wonder if we can come up with an API without that

It might be more imperative, like

```
mystr = spec.Str(pos_args[0])
myint = spec.Int(pos_args[1])
spec.NoMorePositional()
str2 = spec.StrN(named_args, 'name2')
int3 = spec.IntN(named_args, 'name3')
spec.NoMoreNamed()
```

Of course this has to translate to C++, which makes it a bit trickier

We could also try a code gen solution

Something to think about

PossiblyAShrub: Yes, I also found that unsatisfying.

Handling signatures like `max()` (I know we are implementing it in YSH, but still) makes it even trickier. This is definitely something I plan to iterate on as we implement more builtin functions.

PR #1669, opened 2023-07-06
[ysh] Allow users to define functions

PossiblyAShrub:
- Adds a new parse option `parse_func` which allows the parsing (and execution) of YSH functions.
- Stores user-defined functions in a function namespace
  - We will have to move builtin functions to this namespace
- Allows the execution of functions called in expression contexts
- Adds a separate "mode" for `return` which allows typed returns from functions.

This PR makes following execute:

```
func identity(x) {
  return (x)
}

= identity("ysh")  # => "ysh"
```


osh/cmd_parse.py
```diff
@@ -2127,7 +2152,7 @@ def ParseCommand(self):
         # Top-level keywords to hide: func, data, enum, class/mod.  Not sure about
         # 'use'.
         if self.parse_opts.parse_tea():
-            if self.c_id == Id.KW_Func:
+            if self.c_id == Id.KW_Func:  # Duplicate kept for compatibility with tea
```
andychu: Hm yeah we should decide whether to maintain this code, but it looks like it doesn't cause a big problem right now ... that was sort of an experiment to give the mycpp runtime a nicer syntax :)

ysh/expr_to_ast.py
```diff
@@ -1124,17 +1124,17 @@ def _Suite(self, pnode):
...
         """Parse tree to LST
 
-        tea_func:
-          '(' [func_params] [';' func_params] ')' [type_expr_list] suite 
+        named_func: Expr_Name '(' [func_params] [';' func_params] ')'
         """
-        assert pnode.typ == grammar_nt.tea_func
-        assert pnode.GetChild(0).tok.id == Id.Op_LParen  # proc foo(
+        assert pnode.typ == grammar_nt.named_func
+
```
andychu, line 1134: FWIW I added self.p_printer in this file which can help visualize the `pnode`, which follows the grammar directly

PossiblyAShrub: Thanks for mentioning this! It made cleaning this function up much easier.

andychu:
One thing that was on Zulip (but not on the blog) was the issue of how to denote errors from functions

Errors are still integer status codes as in shell

So this is a similar test case to identity() and covers errors

```
func divide_42(x) {
  if (x === 0) {
    error 'divide by zero'  # uses same mechanism as `return 1`
  }
  return (42 / x)  # uses a different mechanism?
}

```

I could probably come up with some more realistic ones ...

I didn't think through this entirely, but I think making `return integer` equivalent to  "error" and then using ANOTHER mechanism for "typed return" could make things easier to implement

Generally speaking I like to disturb the existing interpreter as little as possible, so there's less to test.    

It maybe helps to think of `retval (42)` separate from `return 42`

andychu:
Maybe it helps to think of all the builtin errors

- `42 / 0`
- `d.nonexistent`
- `a[too_big]`

Those will generally raise `error.Expr`, which has status code 3

So we don't want to confuse that with `return 3`

There could be some test cases like

```
func divide_42(x) {
  ...
}

try {
  var y = divide_42(0)
}
if (_status === 3) {
}

```

vs.

```
try {
  var y = divide_42(14)  # should result in 3
}
echo $y
```

PossiblyAShrub:
I agree that `return <status>` and `return (<expr>)` are separate structures. We already expose this difference to users:

```
func foo() {
  return (bar)  # valid
  return bar  # invalid
}

proc baz() {  # or `function baz() {`
  return (bar)  # invalid
  return bar  # valid
}
```

Piggy-backing off of existing shell functions (at least for `error`) does make sense to me. But I do wonder if YSH should provide a mechanism to return/throw errors with more information than an error code. I don't plan on providing any support for errors in functions in this PR, but I think it would be worth thinking about as YSH develops.

> And we could have the explicit check for 0 in `divide_42()`, or just do `42 / x`, and I think those should behave the same ??

I feel that they should behave the same. I think the python style "better to ask for forgiveness than to ask for permission" works really well for shell. Especially when the main interface, the filesystem, is a big clump of global mutable state ripe for race conditions. So I would even be willing to say that, given `42 / x` will already throw for `x = 0`, the explicit check in `divide_42()` would be unidiomatic.

> There are some issues with respect to tracing as well

I am not sure what these would be. Could you elaborate on this?

andychu:
Yes exactly I think that func and proc have **distinct** `return` syntax, to keep those things clear

Originally it was `retval (expr)`, but as long as we keep it distinct, I think we can reuse the same name

----

I was thinking

- `error 'arbitrary string message' (status=3)`  , and maybe you can even add a keyword arguments dictionary

It might be another builtin "register" like `$?` , it could be `$_error_message` or something


----

As for tracing, I just mean that there is this `ctx_Call` thing in `core/state.py` that allows introspection for procs, inherited from bash, like the call stack `${BASH_SOURCE[@]}` etc.

We have to decide if we care about that for functions or not

PossiblyAShrub:
This is now ready for review. We are missing some more complex features like returning with `error` or accepting argument splats `...args`, but I think this is a good time to merge. The `identity` function works well, so that was my initial goal accomplished!

andychu:
I skimmed over this, it generally looks very good!

I'm going to pull it and test out a few things

osh/cmd_eval.py
```diff
@@ -1451,6 +1511,19 @@ def _Dispatch(self, node, cmd_st):
...
+            elif case(command_e.Func):
+                node = cast(command.Func, UP_node)
+
+                name = lexer.TokenVal(node.name)
+                if name in self.funcs and not self.exec_opts.redefine_proc():
+                    e_die(
+                        "Func %s was already defined (redefine_proc)" %
+                        node.name, node.name)
+
+                self.funcs[name] = Func(node, self.mem, self)
```
andychu, line 1523: minor thing (since  I have looked at GC pressure too much) -- probably should just pass in `name` to avoid calling lexer.TokenVal() twice

that will produce 2 separate strings when we only need 1

osh/cmd_eval.py
```diff
@@ -1451,6 +1511,19 @@ def _Dispatch(self, node, cmd_st):
 
                 status = 0
 
+            elif case(command_e.Func):
+                node = cast(command.Func, UP_node)
+
+                name = lexer.TokenVal(node.name)
+                if name in self.funcs and not self.exec_opts.redefine_proc():
```
andychu, line 1518: Probably should rename this `redefine_proc_func`

The idea there was to work around the lack of namespacing in shell

If you have `lib/foo.sh` and `lib/bar.sh` with the same function, it will alert you of the name clash at startup time

BUT in interactive mode, you want to turn that off so you can redefine!

osh/cmd_parse.py
```diff
@@ -1106,6 +1107,18 @@ def ParseSimpleCommand(self):
             return command.ShAssignment(left_tok, pairs, redirects)
 
         kind, kw_token = word_.IsControlFlow(suffix_words[0])
+
+        if (typed_args is not None and kw_token is not None and
+                kw_token.id == Id.ControlFlow_Return):
```
andychu, line 1112: These checks look good -- are they tested?

Maybe add some stuff to `test/ysh-runtime-error.sh`, or a new `test/ysh-parse-errors.sh` ?

osh/cmd_eval.py
```diff
@@ -225,6 +225,54 @@ def PlusEquals(old_val, val):
     return val
 
 
+class Func(vm._Callable):
+
+    def __init__(self, node, mem, cmd_ev):
+        # type: (command.Func, state.Mem, CommandEvaluator) -> None
+        self.name = lexer.TokenVal(node.name)
```
andychu, line 232: as mentioned below, should make this an explicit param

osh/cmd_eval.py
```diff
@@ -225,6 +225,54 @@ def PlusEquals(old_val, val):
...
+        if nargs != expected:
+            raise error.InvalidType("%s() expects %d arguments but %d were given" % (self.name, expected, nargs), self.node.keyword)
+
+        nargs = len(named_args)
+        expected = len(self.node.named_params)
+        if nargs != expected:
+            raise error.InvalidType("%s() expects %d named arguments but %d were given" % (self.name, expected, nargs), self.node.keyword)
+
+        with state.ctx_FuncCall(self.cmd_ev.mem, self):
+            for i in xrange(0, len(pos_args)):
```
andychu, line 250: I think technically the param binding can go outside `with` ?

Shouldn't break anything

osh/cmd_eval.py
```diff
@@ -225,6 +225,54 @@ def PlusEquals(old_val, val):
...
+                pos_arg = pos_args[i]
+                pos_param = self.node.pos_params[i]
+
+                arg_name = location.LName(lexer.TokenVal(pos_param.name))
+
+                self.mem.SetLocationToken(pos_param.name)
```
andychu, line 256: I think we don't need this, because SetLocationToken() is only for generating error messages when evaluating, and we're not evaluating things in between

(although I suppose when we add default arg passing we might)

osh/cmd_eval.py
```diff
@@ -225,6 +225,54 @@ def PlusEquals(old_val, val):
...
+                self.mem.SetValue(arg_name, pos_arg, scope_e.LocalOnly,
+                                  flags=_PackFlags(Id.KW_Const, state.SetReadOnly))
```
andychu, line 258: Hm should func params behave like `const`?  This means you can't re-assign?

Proc params don't behave like that, so we should probably be consistent either way

That is done in osh/cmd_eval.py RunProc()

osh/cmd_eval.py
```diff
@@ -225,6 +225,54 @@ def PlusEquals(old_val, val):
...
+                self.cmd_ev._Execute(self.node.body)
+
+                return value.Null  # implicit return
+            except vm.ControlFlow as e:
+                if e.IsRetval():
+                    return e.value
+                elif e.IsReturn():
+                    e_die("Unexpected proc return in func, use `return (expr)` instead", e.token)
+                else:
+                    e_die("Unexpected control flow in func: %s" % e.token, e.token)
```
andychu, line 272: Hm I wonder if this is an AssertionError (our code is wrong) or an actual runtime error?

andychu (commented):

I think we need a few more tests for

- VarChecker
- param binding semantics -- and decide what they are -- can you mutate a param?

```
func f(x) {
  setvar x = '42'
}
```

I guess I don't see why not -- what I usually do in these cases is refer to Python and JS, and try to choose the least surprising behavior

---

Also it might be nice to have

```
class ControlFlow(Execption)

class IntControlFlow(ControlFlow)
class ValueControlFlow(ControlFlow)
```

It may be slightly longer code, but think the type system will help us more that way

andychu:
Oh I also noticed we didn't make `proc/func` and `return x/return (x)` mutually exclusive

i.e. proc can only use one style, func can only use the other, and the difference is a parse error

I think this will help users keep the 2 things straight?  what do you think?

---

 I think it can be done with a either something like

- `self.allow_block` in CommandParser
- maybe a `ctx_ReturnStyle`, although we don't necessarily need the exception safety that provides (ensuring the __exit__()/dstructor  is called)

But we can do this in a follow-up if necessary

andychu:
Actually I do think this is good because it removes the runtime error  `e_die()` on line 270 of osh/cmd_eval.py 

It can be caught at parse time, so then that can just be an AssertionError

frontend/option_def.py
```diff
@@ -150,6 +150,7 @@ def DoneWithImplementedOptions(self):
 _BASIC_PARSE_OPTIONS = [
     'parse_at',  # @foo, @array(a, b)
     'parse_proc',  # proc p { ... }
+    'parse_func',  # func f(x) { ... }
```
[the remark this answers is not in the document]
PossiblyAShrub: Enabling `parse_func` in OSH is tricky because of the following:

```bash
func() { echo } # valid
func f { echo } # valid
func f() { echo } # invalid, this is where we add funcs into the grammar
```

So to parse YSH funcs, without breaking the first two cases, we'd have to do some fancier look-ahead.

Now I don't think the first two lines are a common case, which is why I didn't add the necessary lookahead in func parsing to support them. Then we can keep the parser fast and errors sane. But this comes at the cost of making `parse_func` a breaking change.

Btw, procs have the same issue:

```
$ bin/osh -c 'proc() { echo; }'; echo $?
0
$ bin/ysh -c 'proc() { echo; }'; echo $?
  proc() { echo; }
       ^
[ -c flag ]:1: Invalid proc name ')'
2
```

osh/cmd_eval.py
```diff
@@ -1113,6 +1163,16 @@ def _Dispatch(self, node, cmd_st):
...
 
+            elif case(command_e.Retval):
+                node = cast(command.Retval, UP_node)
+
+                if mylib.PYTHON:
+                    val = self.expr_ev.EvalExpr(node.val, node.keyword)
+                else:
+                    # FIXME: evaluate the argument once expr_ev is translated
+                    val = value.Null
+                raise vm.ControlFlow(node.keyword, -1, val)
```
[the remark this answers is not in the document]
PossiblyAShrub: This could be very ergonomic in ASDL. Maybe we could add a haskell-like `deriving` clause to the end of type definitions? Then we could write something like:

```asdl
control_flow =
    Break(int levels)
  | Contrinue(int levels)
  | Return(int code)
  | RetVal(value val)
  deriving [Exception]
```

But for now I'll do it manually, won't be too much work. I don't want to balloon the scope of this PR too much.

PossiblyAShrub:
I am pretty sure I have addressed all of your feedback. This is now ready for another look.

osh/builtin_meta.py
```diff
@@ -128,7 +128,7 @@ def Run(self, cmd_val):
                                     c_parser,
                                     self.errfmt,
                                     cmd_flags=cmd_eval.RaiseControlFlow)
-                            except vm.ControlFlow as e:
+                            except vm.IntControlFlow as e:
```
andychu, line 131: OK so `source` can only handle regular shell `return`, not `return ()`

In retrospect, I think that was a good reason to separate them

andychu (approved):

andychu:
Merged, thank you!  Great to see all the parse errors caught, I think people will really appreciate that

I'm  excited to have this in the next release!!

PR #1670, opened 2023-07-09
[ysh] remove slice(), drop step from value.Slice

melvinw:
With this change slices can only be created with the `:` syntax.

### DIFF
--- a/ysh/cpython.py
+++ b/ysh/cpython.py
@@ -98,15 +98,12 @@ def _PyObjToValue(val):
             return value.Dict(typed_dict)
 
     elif isinstance(val, slice):
-        s = value.Slice(None, None, None)
+        s = value.Slice(None, None)
         if val.start:
             s.lower = IntBox(val.start)
 
         if val.stop:
             s.upper = IntBox(val.stop)
-
-        if val.step:
-            s.step = IntBox(val.step)
 
         return s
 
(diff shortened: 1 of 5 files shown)


melvinw:
Surprisingly only one spec test was broken (the one removed) by this change...

ysh/expr_eval.py
```diff
@@ -975,34 +975,26 @@ def _EvalSubscript(self, node):
...
-
-                            try:
-                                if index.lower and index.upper:
-                                    return value.List(
-                                        obj.items[index.lower.i:index.upper.
-                                                  i:step])
+                        try:
+                            if index.lower and index.upper:
+                                return value.List(
+                                    obj.items[index.lower.i:index.upper.i])
```
andychu, line 981: Mostly a tangential style change, but this function could be shortened now to:

```
lower = index.lower.i if index.lower else 0
upper = index.upper.i if index.upper else len(obj.items)
```

And then a single slice operation, instead of 4 cases

ysh/cpython.py
```diff
@@ -176,18 +173,15 @@ def _ValueToPyObj(val):
 
         elif case(value_e.Slice):
             val = cast(value.Slice, UP_val)
-            step = 1
-            if val.step:
-                step = val.step.i
 
             if val.lower and val.upper:
-                return slice(val.lower.i, val.upper.i, step)
+                return slice(val.lower.i, val.upper.i)
```
andychu, line 178: It might be OK to remove convertions fo/from value.Slice,  because we are only converting at the "edges" now

and I don't think any functions actually take a slice, and none should in the future either


We could remove it later too

melvinw: Ah right. The evaluator already operates on `value.Slice` in the subscript code. Sweet

melvinw:
All set!

andychu:
Cool thanks!

I think `ysh/expr_eval.py` is pretty close to being translatable now

I will ping the thread on Zulip

PR #1672, opened 2023-07-10
[ysh] remove range() builtin, add RangeIterator

melvinw:
These commits replace the `range` builtin with `:` syntax for range creation. You can now do loops like the following:
```
$ bin/ysh
ysh$ for i in (0:10) { echo "$[i]" }
0
1
2
3
4
5
6
7
8
9
ysh$
```

In the current implementation, ranges must have both limits specified. So, ranges like `:10` that are otherwise equivalent to the one in the example, are disallowed. How do we feel about this?

### DIFF
--- a/library/func_cpython.py
+++ b/library/func_cpython.py
@@ -184,9 +184,6 @@ def Init(mem):
     # round()
     # divmod() - probably useful?  Look at the implementation
 
-    # Temporary: use xrange until we rewrite in YSH
-    SetGlobalFunc(mem, 'range', lambda *args: list(xrange(*args)))
-
     SetGlobalFunc(mem, 'any', any)  # IN-YSH with Bool
     SetGlobalFunc(mem, 'all', all)  # IN-YSH with Bool
     SetGlobalFunc(mem, 'sum', sum)  # IN-YSH with +
--- a/spec/ysh-funcs-builtin.test.sh
+++ b/spec/ysh-funcs-builtin.test.sh
@@ -104,9 +104,9 @@ false
 #### sum()
 var start = 42
 
-write $[sum( range(3) )]
-write $[sum( range(3), start)]
-write $[sum( range(0), start)]
+write $[sum( 0:3 )]
+write $[sum( 0:3, start)]
+write $[sum( 0:0, start)]
 ## STDOUT:
 3
 45
@@ -114,7 +114,7 @@ write $[sum( range(0), start)]
 ## END
 
 #### reversed()
-var x = reversed(range(3))
+var x = reversed(0:3)
 write @x
 ## STDOUT:
 2
--- a/spec/ysh-slice-range.test.sh
+++ b/spec/ysh-slice-range.test.sh
@@ -105,11 +105,17 @@ var b = a[:]
 for i in (1:5) {
     echo $[i]
 }
+for i, n in (1:4) {
+    echo "$[i], $[n]"
+}
 ## STDOUT:
 1
 2
 3
 4
+0, 1
+1, 2
+2, 3
 ## END
 
 #### Loops over bogus ranges terminate
--- a/spec/ysh-slice-range.test.sh
+++ b/spec/ysh-slice-range.test.sh
@@ -125,16 +125,6 @@ var t3 = mytable[:2, %(name age)]
 (Str)   'TODO: Table Slicing'
 ## END
 
-#### Index a list with a range, not a slice.  TODO: Figure out semantics
-shopt -s oil:all
-var mylist = [1,2,3,4,5]
-var r = 1:3
-var myslice = mylist[r]
-## status: 3
-## STDOUT:
-TODO
-## END
-
 #### List(0:3) should copy the list?
 shopt -s oil:all
 var mylist = List(0:3)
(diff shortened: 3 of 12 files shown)


library/func_cpython.py
```diff
@@ -184,9 +184,6 @@ def Init(mem):
     # round()
     # divmod() - probably useful?  Look at the implementation
 
-    # Temporary: use xrange until we rewrite in YSH
-    SetGlobalFunc(mem, 'range', lambda *args: list(xrange(*args)))
```
melvinw, line 188: Assuming this comment is out of date based on conversations on #oil-dev that suggest the only way to make ranges now is with `:`. Am I mistaken?

spec/ysh-word-eval.test.sh
```diff
@@ -92,18 +92,9 @@ true
...
 ## END
 
-#### @[range()]
-shopt -s oil:all
-write @[range(10, 15, 2)]
-## STDOUT:
-10
-12
-14
-## END
```
melvinw, line 102: Removed this since there is no more builtin to implicitly convert a range into a list. There was some code in `Stringify()` that explicitly rejects iterator-like things (like dict keys), so I figure ranges should be treated the same way

ysh/cpython.py
```diff
@@ -70,6 +70,15 @@ def _PyObjToValue(val):
...
         return value.List(typed_array)
 
+    elif isinstance(val, xrange):
+        # awkward, but should go away once everything is typed...
+        l = list(val)
+        if len(l) > 1:
+            return value.Range(l[0], l[-1])
+
+        # Empty range
+        return value.Range(0, 0)
```
melvinw, line 80: Added this only because of a compulsion for symmetry. Let me check that it's actually necessary...

melvinw: Looks like the spec tests don't depend on this, but we will need this for functions that return ranges

melvinw:
I think this is ready for review. I'm gonna push an extra commit with some spec tests for the `for i in n:m {` syntax

spec/ysh-slice-range.test.sh
```diff
@@ -101,6 +101,26 @@ var b = a[:]
...
+## STDOUT:
+1
+2
+3
+4
+## END
+
+#### Loops over bogus ranges terminate
+# Regression test for bug found during dev. Loops over backwards ranges should
+# terminate immediately.
```
melvinw, line 117: This is consistent with how python handles `for i in range(5, 1): print(i)`. We could also treat this as an error when we evaluate range expressions. Not sure which is better...

spec/ysh-funcs-builtin.test.sh
```diff
@@ -104,17 +104,17 @@ false
...
+write $[sum( 0:3 )]
+write $[sum( 0:3, start)]
+write $[sum( 0:0, start)]
 ## STDOUT:
 3
 45
 42
 ## END
 
 #### reversed()
```
andychu, line 116: Actually I think we should specify that reversed() only works on lists --  I don't see a reason to make it work on ranges

That was sort of a Python legacy

spec/ysh-slice-range.test.sh
andychu: OK great, I think Python's semantics makes sense

(could be convinced otherwise, but Python is a good default ... I think that range bounds can be computed, and empty ranges make sense, like empty slices)

ysh/expr_eval.py
```diff
@@ -643,6 +643,31 @@ def _EvalSlice(self, node):
 
         return value.Slice(lower, upper)
 
+    def _EvalRange(self, node):
+        # type: (expr.Range) -> value_t
+
+        if not node.lower:
+            raise error.Expr('Missing range lower bound', loc.Missing)
```
andychu, line 650: looks like this never happens, can change to assertion

ysh/expr_eval.py
```diff
@@ -643,6 +643,31 @@ def _EvalSlice(self, node):
...
+
+        UP_lower = self._EvalExpr(node.lower)
+        if UP_lower.tag() != value_e.Int:
+            raise error.InvalidType('Range indices must be Ints',
+                                    loc.Missing)
+
+        lower = cast(value.Int, UP_lower)
+
+        if not node.upper:
+            raise error.Expr('Missing range upper bound', loc.Missing)
```
andychu, line 660: ditto, expr_to_ast.py ensures this never happens

ysh/val_ops.py
```diff
@@ -221,6 +221,24 @@ def FirstValue(self):
...
 
 
+class RangeIterator(_ContainerIter):
+    """ for x in (m:n) { """
+
+    def __init__(self, val):
+        # type: (value.Range) -> None
+        _ContainerIter.__init__(self)
+        self.val = val
+        self.i = val.lower
```
andychu, line 231: Hm `self.i` in the base class is supposed to be the index, not the value being iterated over

i.e. I think we should make `for i, x in (3:5)` work analogously like in Lists, and then this part should be a different variable

andychu (commented):
Thanks, this looks great!

melvinw:
OK, think I got everything! Extended the spec tests to cover the `for i, n in (n:m) {` syntax

andychu:
Great thanks!

PR #1673, opened 2023-07-11
xxx [core] create jobs for procs when sending to bg...

melvinw:
and add basic job specs to `fg` builtin

Right now, only jobs started in the background get added to the jobs list. If some, say, starts `vim` in the foreground and sends it to the background with `Ctrl-z` the output of `jobs` will be empty. `fg` will still bring it back to the foreground as long as it's considered the last stopped process, but sending another process to the background will prevent the user from ever bringing the original process back to the foreground.

To make managing multiple background processes easier this commit adds processes stopped with `Ctrl-z` to the jobs list and adds support for the `%<digits>` flavor of job spec to the `fg` builtin.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -41,6 +41,7 @@ from osh import cmd_eval
 from mycpp import mylib
 from mycpp.mylib import log, print_stderr, tagswitch, iteritems, StrFromC
 
+import libc
 import posix_ as posix
 from posix_ import (
     # translated by mycpp and directly called!  No wrapper!
@@ -876,6 +877,7 @@ class Job(object):
         # type: () -> None
         # Initial state with & or Ctrl-Z is Running.
         self.state = job_state_e.Running
+        self.job_id = -1
 
     def DisplayJob(self, job_id, f, style):
         # type: (int, mylib.Writer, int) -> None
@@ -884,6 +886,11 @@ class Job(object):
     def State(self):
         # type: () -> job_state_t
         return self.state
+
+    def GroupId(self):
+        # type: () -> int
+        """Only supported for Process right now."""
+        raise NotImplementedError()
 
     def JobWait(self, waiter):
         # type: (Waiter) -> wait_status_t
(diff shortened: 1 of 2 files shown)


andychu:
Hm is this ready to look at?  It's still marked draft

Can it be tested with `spec/stateful`?  I think this code is fiddly and could use it

I'd like to try out some of the examples interactively

melvinw:
Oh, nah. I still want to add some tests and rework a few things. Just had this open for CI

melvinw:
I did kick the tires interactively a bit, though. And it seems like it does behave correctly. In the process I did discover an existing bug (can repro on master) where we get spurious SIGTTIN/SIGTTOU in some cases (e.g. try `cat &` followed by `cat` in quick succession). I also want to get to the bottom of that and fix it in a separate patch before merging this.

andychu:
Ah  yeah I can repro that pretty easily with one command

```
osh-0.16.0$ cat & cat

[PID 127286] Stopped with signal 21

[PID 127287] Stopped with signal 21

```

Hmm ...

andychu:
Oh I think I see a bug ?      `fg` should not call posix.killpg()` and then `WhenContinued()` directly?  

I think it should wait until it receives state change notification via `waitpid(-1)` to update the shell state

I think I didn't understand that when initially writing this code

----


 I think the flow of data should always be like this:

1. shell sends commands to kernel, e.g. `killpg()`
2. shell has a loop that polls for state changes from the kernel -- `waitpid(-1)`
3. shell updates its internal state based on those state changes

So the shell state reflects the kernel state

I'm not sure if it's causing this bug but it seems related

melvinw:
Yeah, it's odd... I was gonna investigate some once I get my office back in working order

melvinw:
Ah sure. That thing seems like it's worth looking into, but I doubt it's causing this bug (we aren't even using `fg` yet and we're seeing this odd hiccup)

melvinw:
Oh interesting. I can reproduce this with bash too. See https://asciinema.org/a/jqt4OXXo1JsMAS3Pzuzk5pxUC

Perhaps this makes sense and these aren't spurious SIGTTINs... The shell prompt is clearly still in the foreground and so the child `cat`s are getting SIGTTIN as expected when they try to read stdin. The only odd thing really then (in both shells) is that we see the "stopped" message twice once the second child gets SIGTTIN, but I don't think I'd call that a major bug

melvinw:
Figured this out! Fix is in #1676

melvinw:
going to combine this with https://github.com/oilshell/oil/pull/1675

PR #1675, opened 2023-07-14
job control overhaul

andychu:
[core] print notifications when bg jobs complete

To support notifying users about the completion of background jobs, this
commit updates the main loop for the interactive shell to pass the
WNOHANG to `WaitPid()`, adds state to `Job` about background/foreground
positioning. It also removes the `WhenDone()` method from `JobList` to
make reading the job completion codepath easier.

 [core] only take back tty on stop if necessary

When the main shell wakes up from wait() with WIFSTOPPED, it
unconditionally takes back the TTY from whichever process was just
stopped. This can cause problems when running background jobs, though.

Consider the following script:

  $ cat &
  $ cat

The first cat is forked and placed into its own process group. The shell
remains in the foreground. This means cat will immediately be stopped
with SIGTTIN when it tries to read from stdin. The main shell doesn't
know this yet, though! It's blocked waiting for input from readline. The
second cat is forked and placed into its own group, just like the first
one; and since it's running in the foreground, the main shell gives it
control of the TTY. After handing over the TTY the shell finally gets
notified about the first cat being stopped (as expected) via wait() and
does some post-stop bookkeeping, including taking back the TTY! This
causes the second cat to immediately get stopped by SIGTTIN, confusing
the user.

To avoid this awkward behavior, this commit updates the post-stop
bookkeeping to only take back the terminal from processes that were
originally running in the foreground.

[core] overhaul job control

- Remove JobList::WhenContinued(). It was only used in one place.
- Add support for POSIX job specs (%[0-9]+, %%, %-, %+, %) in `fg` builtin
- Add Job::ProcessGroupId() and refactor `fg` to deal with process groups and wait better
- Remove JobList::last_stopped_job. %+ and %- are computed by the builtins now
- Add a few more test to spec/stateful/job_control

[builtin/wait] Support "job spec" arguments

### DIFF
--- a/core/executor.py
+++ b/core/executor.py
@@ -364,6 +364,7 @@ class ShellExecutor(vm._Executor):
                 pi.Add(p)
 
             pi.StartPipeline(self.waiter)
+            pi.SetBackground()
             last_pid = pi.LastPid()
             self.mem.last_bg_pid = last_pid  # for $!
 
@@ -378,6 +379,7 @@ class ShellExecutor(vm._Executor):
             if self.job_control.Enabled():
                 p.AddStateChange(process.SetPgid(process.OWN_LEADER))
 
+            p.SetBackground()
             pid = p.StartProcess(trace.Fork)
             self.mem.last_bg_pid = pid  # for $!
             self.job_list.AddJob(p)  # show in 'jobs' list
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -217,9 +217,7 @@ def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, waiter, errfmt):
                         display.EraseLines()
                         # POSIX shell behavior: waitpid(-1) and show job "Done"
                         # messages
-                        r = waiter.WaitForOne(waitpid_options=WNOHANG)
-                        while r == process.W1_OK:
-                            r = waiter.WaitForOne(waitpid_options=WNOHANG)
+                        waiter.PollNotifications()
                         quit = True
                     elif case(parse_result_e.Eof):
                         display.EraseLines()
@@ -275,7 +273,7 @@ def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, waiter, errfmt):
 
             status = cmd_ev.LastStatus()
 
-            r = waiter.WaitForOne(waitpid_options=WNOHANG)
+            waiter.PollNotifications()
 
             if is_return:
                 done = True
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -24,7 +24,6 @@ from mycpp.mylib import log, print_stderr, tagswitch
 
 import fanos
 import posix_ as posix
-from posix_ import WNOHANG
 
 from typing import cast, Any, List, TYPE_CHECKING
 if TYPE_CHECKING:
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -218,6 +218,8 @@ def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, waiter, errfmt):
                         # POSIX shell behavior: waitpid(-1) and show job "Done"
                         # messages
                         r = waiter.WaitForOne(waitpid_options=WNOHANG)
+                        while r == process.W1_OK:
+                            r = waiter.WaitForOne(waitpid_options=WNOHANG)
                         quit = True
                     elif case(parse_result_e.Eof):
                         display.EraseLines()
--- a/core/main_loop.py
+++ b/core/main_loop.py
@@ -24,6 +24,7 @@ from mycpp.mylib import log, print_stderr, tagswitch
 
 import fanos
 import posix_ as posix
+from posix_ import WNOHANG
 
 from typing import cast, Any, List, TYPE_CHECKING
 if TYPE_CHECKING:
@@ -190,20 +191,19 @@ class Headless(object):
         return 0
 
 
-def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, errfmt):
-    # type: (arg_types.main, CommandEvaluator, CommandParser, _IDisplay, UserPlugin, ErrorFormatter) -> int
-
-    # TODO: Any could be _Attributes from frontend/args.py
+def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, waiter, errfmt):
+    # type: (arg_types.main, CommandEvaluator, CommandParser, _IDisplay, UserPlugin, process.Waiter, ErrorFormatter) -> int
 
     status = 0
     done = False
     while not done:
         mylib.MaybeCollect()  # manual GC point
 
-        # - This loop has a an odd structure because we want to do cleanup after
-        # every 'break'.  (The ones without 'done = True' were 'continue')
+        # - This loop has a an odd structure because we want to do cleanup
+        #   after every 'break'.  (The ones without 'done = True' were
+        #   'continue')
         # - display.EraseLines() needs to be called BEFORE displaying anything, so
-        # it appears in all branches.
+        #   it appears in all branches.
 
         while True:  # ONLY EXECUTES ONCE
             quit = False
@@ -215,6 +215,9 @@ def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, errfmt):
                 with tagswitch(result) as case:
                     if case(parse_result_e.EmptyLine):
                         display.EraseLines()
+                        # POSIX shell behavior: waitpid(-1) and show job "Done"
+                        # messages
+                        r = waiter.WaitForOne(waitpid_options=WNOHANG)
                         quit = True
                     elif case(parse_result_e.Eof):
                         display.EraseLines()
@@ -269,6 +272,9 @@ def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, errfmt):
                 break
 
             status = cmd_ev.LastStatus()
+
+            r = waiter.WaitForOne(waitpid_options=WNOHANG)
+
             if is_return:
                 done = True
                 break
--- a/core/process.py
+++ b/core/process.py
@@ -1749,6 +1749,7 @@ class JobList(object):
 # Some WaitForOne() return values
 W1_OK = -2  # waitpid(-1) returned
 W1_ECHILD = -3  # no processes to wait for
+W1_AGAIN = -4 # WNOHANG was passed and there were no state changes
 
 
 class Waiter(object):
@@ -1819,7 +1820,7 @@ class Waiter(object):
         """
         pid, status = pyos.WaitPid(waitpid_options)
         if pid == 0:  # WNOHANG passed, and no state changes
-            return W1_OK
+            return W1_AGAIN
         elif pid < 0:  # error case
             err_num = status
             #log('waitpid() error => %d %s', e.errno, pyutil.strerror(e))
--- a/core/process.py
+++ b/core/process.py
@@ -1069,8 +1069,12 @@ class Process(Job):
         # https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html
         self.status = 128 + stop_sig
         self.state = job_state_e.Stopped
-        self.SetBackground()
-        self.job_control.MaybeTakeTerminal()
+
+        self.job_list.last_stopped_pid = self.pid # for fg
+
+        if not self.in_background:
+            self.job_control.MaybeTakeTerminal()
+            self.SetBackground()
 
     def WhenDone(self, pid, status):
         # type: (int, int) -> None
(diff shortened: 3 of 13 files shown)


melvinw:
ok @andychu I think I'm done tweaking this branch. your `WNOHANG`/`WaitPid()` changes seemed good to me

core/process.py
```diff
@@ -890,6 +892,21 @@ def JobWait(self, waiter):
         """Wait for this process/pipeline to be stopped or finished."""
         raise NotImplementedError()
 
+    def InForeground(self):
+        # type: () -> bool
+        """Returns true if this job is running in the foreground."""
+        return self.in_foreground
```
melvinw, line 898: maybe this getter is a bit much and using `.in_foreground` directly might be OK? do you have a pattern/preference for this @andychu?

core/process.py
```diff
@@ -1499,7 +1519,13 @@ def WhenDone(self, pid):
         if pid == self.last_stopped_pid:
             self.last_stopped_pid = -1
 
-        mylib.dict_erase(self.jobs, pid)
+        assert pid in self.child_procs
+        proc = self.child_procs[pid]
+        if proc.job_id in self.jobs:
+            mylib.dict_erase(self.jobs, proc.job_id)
```
andychu, line 1525: the dict_erase() can go outside the `if`

The API of dict_erase() is that it ensures the key isn't present unconditionally

core/process.py
```diff
@@ -1499,7 +1519,13 @@ def WhenDone(self, pid):
...
+            # not all background processes (e.g. command subs, subshells) have jobs
+            if not proc.InForeground():
```
andychu, line 1527: Hm  is this the only thing we're using the flag for?

How about:

1. we add a flag to Process and Pipeline `self.in_background = False` 
2. There are 2 functions in the `_Executor` where we start processes and pipelines in the background.  In those cases we do `obj.SetBackground()`

I don't really like "MoveToForeground() MoveToBackground()" because it sounds like some syscalls are  happening

andychu (commented):
If this makes the test pass it seems good

I'm wondering about the Pipeline case though

Let me look at the other PRs too

core/process.py
melvinw: Ah yeah, `Move*()` is a little misleading. Will use the new naming

melvinw:
@andychu 

> Let me look at the other PRs too

Sounds good. On #1676 you can ignore the `POSITION_*` stuff. It will be replaced with the flag we added here

melvinw:
OK @andychu I renamed things. I also made it so pipelines only print the "Done" message once for the entire process group (added a test for this to `spec/stateful`). What do you think? Is there anything in particular you're wondering about the pipeline case?

core/main_loop.py
```diff
@@ -215,6 +215,9 @@ def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, errfmt):
                 with tagswitch(result) as case:
                     if case(parse_result_e.EmptyLine):
                         display.EraseLines()
+                        # POSIX shell behavior: waitpid(-1) and show job "Done"
+                        # messages
+                        r = waiter.WaitForOne(WNOHANG)
```
andychu, line 220: should add kwarg waitpid_options= here to be consistent

core/executor.py
```diff
@@ -422,6 +424,7 @@ def RunPipeline(self, node, status_out):
     def RunSubshell(self, node):
         # type: (command_t) -> int
         p = self._MakeProcess(node)
+        p.SetBackground()
```
andychu, line 427: Hm why is a subshell in the background?

Does "in background" mean "not part of the single PGID that has terminal control" ?   


Or something related but similar?  we should be precise about the terminology

core/executor.py
```diff
@@ -462,6 +465,7 @@ def RunCommandSub(self, cs_part):
         # don't use p.AddStateChange(process.SetPgid(...))
 
         r, w = posix.pipe()
+        p.SetBackground()
```
andychu, line 468: Ditto - why is a CommandSub in the background?

core/process.py
```diff
@@ -1499,7 +1513,21 @@ def WhenDone(self, pid):
         if pid == self.last_stopped_pid:
             self.last_stopped_pid = -1
 
-        mylib.dict_erase(self.jobs, pid)
+        assert pid in self.child_procs
+        proc = self.child_procs[pid]
+        pipe = proc.parent_pipeline
+        if pipe:
+            if pipe.AllDone():
```
andychu, line 1520: I wonder if we should put the print statements in `Process:WhenDone()` and `Pipeline::WhenDone()`

We will have to change them to print the actual code to match POSIX/bash, and that will probably be easier there -- i.e. they will have a connection the original source code, not just the PID

I probably wasn't thinking about that when I initially added this

(I also want to rename this to WhenJobDone() -- I realized it is confusing to have methods of the same name on different classes!  It makes searching harder)

andychu (commented):
OK glad that printing notification on pipelines wasn't too hard!

andychu:
Hm

> OK @andychu I renamed things. I also made it so pipelines only print the "Done" message once for the entire process group (added a test for this to `spec/stateful`). What do you think? Is there anything in particular you're wondering about the pipeline case?
> 
> One thing we'll need to be careful to specify and test around with #1673 is how pipelines behave when started in the background. Should a user be able to do something like the following?
> 
> ```
> $ sleep 1 | cat &
> $ fg %1
> ```
> 
> If yes, then should we prevent the user from returning the pipeline to the background with ctrl-z after using `fg`?

Hmm I'm not sure, but this is the way all shells work right?  Both processes and pipelines are jobs.  They have a job number like `%1`

I generally think we should 

- Focus on what we use, it's OK to not implement things
- But also listen to what other people use
- If we decide not to implement something, don't leave it unspecified.  Everything should have a DEFINED behavior, even if it differs.  For example, we documented the fact that you can't Ctrl-Z pipelines because of the lastpipe issue, in the "Known differences" doc.  (Although we don't print a message like zsh does, that would be nice.  Also on testing, we seem to disallow more than zsh ?    e.g. both `sleep |cat` and `sleep|read` -- only the latter has the lastpipe problem)

core/executor.py
melvinw: Ah yes. In my head "in the background" is synonymous with "cannot use the terminal"

Perhaps we should rename again from `in_background` to something like `has_tty` to avoid confusion

melvinw:
OK. Good

>  (Although we don't print a message like zsh does, that would be nice. Also on testing, we seem to disallow more than zsh ? e.g. both sleep |cat and sleep|read -- only the latter has the lastpipe problem)

Yeah, I forget the details of the discussions, but we decided to keep things simple (for us) and blanket disallow ctrl-z for all pipelines

core/executor.py
melvinw: This (and the above) actually aren't necessary. I was being overly cautious. These aren't tracked by the jobs list, so they shouldn't trigger the `Done` log

core/process.py
melvinw: I think we can actually remove `WhenDone()` from `JobList` entirely and replace it with `RemoveJob()`. All the logic other than updating `self.jobs` can live in the subclass' `WhenDone()`s

core/process.py
```diff
@@ -1653,8 +1779,14 @@ def __init__(self, job_list, exec_opts, signal_safe, tracer):
         self.tracer = tracer
         self.last_status = 127  # wait -n error code
 
-    def WaitForOne(self):
-        # type: () -> int
+    #def CheckForNotifications(self):
+    #    """
+    #    Called by main_loop::Interactive()
+    #    """
+    #    pid, status = pyos.WaitPid(WNOHANG)
```
melvinw, line 1786: @andychu do you have plans for this dead code or can i remove?

core/process.py
```diff
@@ -1542,6 +1586,88 @@ def JobFromPid(self, pid):
...
+        if len(running_jobs) > 0 and not current:
+            current = running_jobs.pop()
+
+        if len(running_jobs) > 0 and not previous:
+            previous = running_jobs.pop()
+
+        if not previous:
+            previous = current
+
+        return current, previous
```
melvinw, line 1643: there's probably a more compact way to do this... open to suggestions

core/process.py
andychu: you can remove this

andychu (commented):
Thanks for diving into this!

I took a quick look -- it looks like very good progress, and the code structure makes sense

There is one test failing in osh-cpp only -- can you reproduce locally?

http://travis-ci.oilshell.org/github-jobs/4594/interactive.wwz/_tmp/spec/stateful/index.html

The CI doesn't really assert spec/stateful very well, since it's a bit flaky

osh/builtin_process.py
```diff
@@ -69,20 +69,48 @@ def __init__(self, job_control, job_list, waiter):
...
         # suspsended again if it tries to read/write on the terminal.
-        pgid = posix.getpgid(pid)
         self.job_control.MaybeGiveTerminal(pgid)
+        job.SetForeground()
+        # needed for Wait() loop to work
+        job.state = job_state_e.Running
         posix.killpg(pgid, SIGCONT)
-        return self.job_list.WhenContinued(pid, self.waiter)
+
+        status = 1 # error
```
andychu, line 93: hm it seems like this should always be overwritten?

we can initialize it to -1 if we need that for mycpp ?

spec/stateful/interactive.py
```diff
@@ -55,6 +55,35 @@ def syntax_error(sh):
...
+    # not using trailing + because pexpect doc warns about that case
+    # dash doesn't print this
+    sh.expect(r'\[\d+\]')
+
+  expect_prompt(sh)
+
+  time.sleep(0.2)
+  sh.sendline('')
+  if 'osh' in sh.shell_label:
+      # need to wake up from wait() twice in osh
```
andychu, line 77: Oh is this because there are 2 processes in the pipeline?

if we put 5 of them do we have to check 5 times?

spec/stateful/job_control.py
```diff
@@ -266,6 +268,122 @@ def suspend_status(sh):
...
+  expect_prompt(sh)
+
+  sh.sendline('cat &') # stop
+  sh.sendline('sleep 0.1 &') # exit
+  expect_prompt(sh)
+
+  # background cat should have been stopped by SIGTTIN immediately, but we don't
+  # hear about it from wait() until the foreground process has been started because
+  # the shell was blocked in readline when the signal fired.
+  sh.sendline('python2 -c "import sys; print(sys.stdin.readline().strip() + \'bar\')"')
```
andychu, line 283: can use PYCAT ?

andychu:
Hm I think we should also test the case where

```
sleep 1 &
sleep 2 &
<wait 3 seconds before hitting enter>
now you should get 2 notifications
```

I think that is similar to the pipeline case

osh/builtin_process.py
melvinw: Ah yes looks like it. I copied this from the `Wait` builtin below where the logic is a bit more compilcated. Will set to `-1`

spec/stateful/interactive.py
melvinw: Yeah :/ We can coalesce these if we handle `SIGCHLD`. This will require some improvements to the signal handling stuff in `pyos`, through. `SIGCHLD` caries structured data that I'm sure will be handled differently between python's stdlib and libc

melvinw:
Ah sure. I'll add this to the existing `bg_proc_notify` test in `spec/interactive`

spec/stateful/interactive.py
andychu: Hmmm maybe we should chat about this, I'd like to understand what needs to be done -- is it a big deal?


Actually this is why I added that CheckForNotifications() method -- because I realized it wasn't enough to just  check for one

melvinw: I don't think it's a huge deal for now. And the fix might actually not even require and signal stuff. I didn't realize `main_loop::Interactive()` unconditionally waits for only one process when it gets an empty line. We could change that bit to "drain the queue". I'll push a quick patch to do that and we'll see if anything breaks...

core/main_loop.py
```diff
@@ -218,6 +218,8 @@ def Interactive(flag, cmd_ev, c_parser, display, prompt_plugin, waiter, errfmt):
                         # POSIX shell behavior: waitpid(-1) and show job "Done"
                         # messages
                         r = waiter.WaitForOne(waitpid_options=WNOHANG)
+                        while r == process.W1_OK:
+                            r = waiter.WaitForOne(waitpid_options=WNOHANG)
```
melvinw, line 222: Should we place a limit on this loop to avoid accidentally starving the outer loop? A storm of process state changes could make the shell feel sluggish for a moment (it would have to be a pretty big storm, though...)

melvinw:
Was able to reproduce that `osh-cpp` failure (it happened for `bin/osh` and `bash` too. just much less reliably). Was just some flakey test logic. Ran that suite a bunch locally and it looks more reliable. CI is happy too:
http://travis-ci.oilshell.org/github-jobs/4598/interactive.wwz/_tmp/spec/stateful/index.html

PR #1676, opened 2023-07-14
[core] only take back tty on stop if necessary

melvinw:
When the main shell wakes up from wait() with WIFSTOPPED, it unconditionally takes back the TTY from whichever process was just stopped. This can cause problems when running background jobs, though.

Consider the following script:
```bash
  $ cat &
  $ cat
```
The first cat is forked and placed into its own process group. The shell remains in the foreground. This means cat will immediately be stopped with SIGTTIN when it tries to read from stdin. The main shell doesn't know this yet, though! It's blocked waiting for input from readline. The second cat is forked and placed into its own group, just like the first one; and since it's running in the foreground, the main shell gives it control of the TTY. After handing over the TTY the shell finally gets notified about the first cat being stopped (as expected) via wait() and does some post-stop bookkeeping, including taking back the TTY! This causes the second cat to immediately get stopped by SIGTTIN, confusing the user.

To avoid this awkward behavior, this commit updates the post-stop bookkeeping to only take back the terminal from processes that were originally running in the foreground.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -87,6 +87,9 @@ STYLE_DEFAULT = 0
 STYLE_LONG = 1
 STYLE_PID_ONLY = 2
 
+POSITION_BACKGROUND = 0
+POSITION_FOREGROUND = 1
+
 
 class ctx_FileCloser(object):
     def __init__(self, f):
@@ -969,6 +972,8 @@ class Process(Job):
     def StartProcess(self, why):
         # type: (trace_t) -> int
         """Start this process with fork(), handling redirects."""
+        # Start in the background by default
+        self.SetPosition(POSITION_BACKGROUND)
         pid = posix.fork()
         if pid < 0:
             # When does this happen?
@@ -1060,7 +1065,9 @@ class Process(Job):
         # https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html
         self.status = 128 + stop_sig
         self.state = job_state_e.Stopped
-        self.job_control.MaybeTakeTerminal(self.pgid)
+        if self.position == POSITION_FOREGROUND:
+            self.SetPosition(POSITION_BACKGROUND)
+            self.job_control.MaybeTakeTerminal()
 
     def WhenDone(self, pid, status):
         # type: (int, int) -> None
@@ -1072,8 +1079,8 @@ class Process(Job):
         self.state = job_state_e.Done
         if self.parent_pipeline:
             self.parent_pipeline.WhenDone(pid, status)
-        else:
-            self.job_control.MaybeTakeTerminal(self.pgid)
+        elif self.position == POSITION_FOREGROUND:
+            self.job_control.MaybeTakeTerminal()
 
     def RunProcess(self, waiter, why):
         # type: (Waiter, trace_t) -> int
@@ -1081,10 +1088,16 @@ class Process(Job):
         self.StartProcess(why)
         # ShellExecutor might be calling this for the last part of a pipeline.
         if self.parent_pipeline is None:
+            self.SetPosition(POSITION_FOREGROUND)
             # QUESTION: Can the PGID of a single process just be the PID?  i.e. avoid
             # calling getpgid()?
             self.job_control.MaybeGiveTerminal(posix.getpgid(self.pid))
+
         return self.Wait(waiter)
+
+    def SetPosition(self, position):
+        # type: (int) -> None
+        self.position = position
 
 
 class ctx_Pipe(object):
@@ -1318,7 +1331,8 @@ class Pipeline(Job):
             # status of pipeline is status of last process
             self.status = self.pipe_status[-1]
             self.state = job_state_e.Done
-            self.job_control.MaybeTakeTerminal(self.pids[0])
+            # Pipelines are always in the foreground
+            self.job_control.MaybeTakeTerminal()
 
 
 def _JobStateStr(i):
@@ -1436,16 +1450,11 @@ class JobControl(object):
             e_die('osh: Failed to move process group %d to foreground: %s' %
                   (pgid, pyutil.strerror(e)))
 
-    def MaybeTakeTerminal(self, group):
-        # type: (int) -> None
+    def MaybeTakeTerminal(self):
+        # type: () -> None
         """If stdio is a TTY, return the main shell's process group to the
-        foreground only if the given group is running in the foreground."""
-        # Only take back the terminal if we gave it to this process in the first
-        # place. Otherwise we might accidentally take it back from another
-        # process running in the foreground, causing it to get stopped with
-        # SIGTTIN/SIGTTOU.
-        if self.Enabled() and posix.tcgetpgrp(self.shell_tty_fd) == group:
-            self.MaybeGiveTerminal(self.shell_pgid)
+        foreground."""
+        self.MaybeGiveTerminal(self.shell_pgid)
 
     def MaybeReturnTerminal(self):
         # type: () -> None
@@ -1495,6 +1504,7 @@ class JobList(object):
         if pid == self.last_stopped_pid:
             self.last_stopped_pid = -1
         job = self.JobFromPid(pid)
+        job.SetPosition(POSITION_FOREGROUND)
         # needed for Wait() loop to work
         job.state = job_state_e.Running
         return job.Wait(waiter)
--- a/core/process.py
+++ b/core/process.py
@@ -617,7 +617,6 @@ class SetPgid(ChildStateChange):
         # type: (Process) -> None
         try:
             posix.setpgid(proc.pid, self.pgid)
-            proc.pgid = self.pgid
         except (IOError, OSError) as e:
             print_stderr(
                 'osh: parent failed to set process group for PID %d to %d: %s' %
@@ -1021,8 +1020,7 @@ class Process(Job):
         for st in self.state_changes:
             st.ApplyFromParent(self)
 
-        if self.pgid == -1:
-            self.pgid = pid
+        self.pgid = posix.getpgid(pid)
 
         # Program invariant: We keep track of every child process!
         self.job_list.AddChildProcess(pid, self)
--- a/core/process.py
+++ b/core/process.py
@@ -617,6 +617,7 @@ class SetPgid(ChildStateChange):
         # type: (Process) -> None
         try:
             posix.setpgid(proc.pid, self.pgid)
+            proc.pgid = self.pgid
         except (IOError, OSError) as e:
             print_stderr(
                 'osh: parent failed to set process group for PID %d to %d: %s' %
@@ -921,6 +922,7 @@ class Process(Job):
         self.close_w = -1
 
         self.pid = -1
+        self.pgid = -1
         self.status = -1
 
     def Init_ParentPipeline(self, pi):
@@ -1019,6 +1021,9 @@ class Process(Job):
         for st in self.state_changes:
             st.ApplyFromParent(self)
 
+        if self.pgid == -1:
+            self.pgid = pid
+
         # Program invariant: We keep track of every child process!
         self.job_list.AddChildProcess(pid, self)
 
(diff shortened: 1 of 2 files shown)


melvinw:
@andychu Hopefully this makes sense. Let me know what you think

core/process.py
```diff
@@ -1057,7 +1057,12 @@ def WhenStopped(self, stop_sig):
         # https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html
         self.status = 128 + stop_sig
         self.state = job_state_e.Stopped
-        self.job_control.MaybeTakeTerminal()
+        # Only take back the terminal if we gave it to this process in the first
+        # place. Otherwise we might accidentally take it back from another
+        # process running in the foreground, causing it to get stopped with
+        # SIGTTIN/SIGTTOU.
```
andychu, line 1063: Hmm the bug makes sense

Does it make sense to set a boolean on this Process object if we called MaybeGiveTerminal in StartProcess()?

spec/stateful/job_control.py
```diff
@@ -266,6 +266,30 @@ def suspend_status(sh):
...
 
+@register(skip_shells=['zsh'])
+def no_spurious_tty_take(sh):
+  'A background job getting stopped by SIGTTIN should not disrupt foreground processes'
+  expect_prompt(sh)
+
+  sh.sendline('cat &')
+  time.sleep(0.1)  # seems necessary
+
+  # background cat should have been stopped by SIGTTIN immediately, but we don't
```
andychu, line 277: `cat` should only be stopped if we gave it input?    

But we didn't do that yet in this test case?

---

Hmm but it does in bash too ... not sure I get it

spec/stateful/job_control.py
```diff
@@ -266,6 +266,30 @@ def suspend_status(sh):
...
+  # hear about it from wait() until the foreground job has been started because
+  # the shell was blocked in readline when the signal fired.
+  sh.sendline('python2 -c "import sys; print(sys.stdin.readline().strip() + \'bar\')"')
+  if 'osh' in sh.shell_label:
+    # Quirk of osh. TODO: supress this print for background jobs?
+    sh.expect('.*Stopped.*')
+
+  # foreground cat should not have been stopped.
```
andychu, line 285: Do you mean python2, not cat ?

spec/stateful/job_control.py
```diff
@@ -266,6 +266,30 @@ def suspend_status(sh):
...
+def no_spurious_tty_take(sh):
+  'A background job getting stopped by SIGTTIN should not disrupt foreground processes'
+  expect_prompt(sh)
+
+  sh.sendline('cat &')
+  time.sleep(0.1)  # seems necessary
+
+  # background cat should have been stopped by SIGTTIN immediately, but we don't
+  # hear about it from wait() until the foreground job has been started because
+  # the shell was blocked in readline when the signal fired.
```
andychu, line 279: BTW `set -b`  in POSIX is a getting to give  the notifications immediately

Maybe we should implement that as it does make it a bit easier to reason about the tests

spec/stateful/job_control.py
melvinw: Ah yup. Will fix

spec/stateful/job_control.py
melvinw: If I understand correctly, SIGTTIN will get sent as soon as a child calls `read(STDIN,...)` (assuming they don't have the tty), not just when it has bytes to return

core/process.py
melvinw: Yeah, having a bit of state could be useful if you're worried about extra syscalls. I think it will need to be more dynamic than just at start, though (e.g. something repeatedly sent between the foreground and background might exhibit the same behavior)

spec/stateful/job_control.py
melvinw: Confirmed this with strace:

```bash
$ strace cat &
[1] 390091
melvin@oil-dev:~/oil$ execve("/usr/bin/cat", ["cat"], 0x7ffe6bdc2580 /* 28 vars */) = 0
brk(NULL)                               = 0x55edbf10e000
mmap(NULL, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f6ffc5d8000
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
newfstatat(3, "", {st_mode=S_IFREG|0644, st_size=22406, ...}, AT_EMPTY_PATH) = 0
mmap(NULL, 22406, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f6ffc5d2000
openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\3\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0Ps\2\0\0\0\0\0"..., 832) = 832
pread64(3, "\6\0\0\0\4\0\0\0@\0\0\0\0\0\0\0@\0\0\0\0\0\0\0@\0\0\0\0\0\0\0"..., 784, 64) = 784
newfstatat(3, "", {st_mode=S_IFREG|0755, st_size=1922136, ...}, AT_EMPTY_PATH) = 0
pread64(3, "\6\0\0\0\4\0\0\0@\0\0\0\0\0\0\0@\0\0\0\0\0\0\0@\0\0\0\0\0\0\0"..., 784, 64) = 784
mmap(NULL, 1970000, PROT_READ, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f6ffc3f1000
mmap(0x7f6ffc417000, 1396736, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x26000) = 0x7f6ffc417000
mmap(0x7f6ffc56c000, 339968, PROT_READ, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x17b000) = 0x7f6ffc56c000
mmap(0x7f6ffc5bf000, 24576, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x1ce000) = 0x7f6ffc5bf000
mmap(0x7f6ffc5c5000, 53072, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7f6ffc5c5000
mmap(NULL, 12288, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f6ffc3ee000
arch_prctl(ARCH_SET_FS, 0x7f6ffc3ee740) = 0
set_tid_address(0x7f6ffc3eea10)         = 390094
set_robust_list(0x7f6ffc3eea20, 24)     = 0
rseq(0x7f6ffc3ef060, 0x20, 0, 0x53053053) = 0
mprotect(0x7f6ffc5bf000, 16384, PROT_READ) = 0
mprotect(0x55edbe3cc000, 4096, PROT_READ) = 0
mprotect(0x7f6ffc60a000, 8192, PROT_READ) = 0
prlimit64(0, RLIMIT_STACK, NULL, {rlim_cur=8192*1024, rlim_max=RLIM64_INFINITY}) = 0
munmap(0x7f6ffc5d2000, 22406)           = 0
getrandom("\x03\xdb\x9a\xad\xf7\xc6\xf6\xc4", 8, GRND_NONBLOCK) = 8
brk(NULL)                               = 0x55edbf10e000
brk(0x55edbf12f000)                     = 0x55edbf12f000
openat(AT_FDCWD, "/usr/lib/locale/locale-archive", O_RDONLY|O_CLOEXEC) = 3
newfstatat(3, "", {st_mode=S_IFREG|0644, st_size=3048928, ...}, AT_EMPTY_PATH) = 0
mmap(NULL, 3048928, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f6ffc000000
newfstatat(1, "", {st_mode=S_IFCHR|0620, st_rdev=makedev(0x88, 0x4), ...}, AT_EMPTY_PATH) = 0
newfstatat(0, "", {st_mode=S_IFCHR|0620, st_rdev=makedev(0x88, 0x4), ...}, AT_EMPTY_PATH) = 0
fadvise64(0, 0, 0, POSIX_FADV_SEQUENTIAL) = 0
mmap(NULL, 139264, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f6ffc3cc000
read(0, 0x7f6ffc3cd000, 131072)         = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGTTIN {si_signo=SIGTTIN, si_code=SI_KERNEL} ---
--- stopped by SIGTTIN ---
```

spec/stateful/job_control.py
melvinw: Oh cool. I didn't know about that. Would be nice to add

andychu (commented):
The changes look good

I notice we call MaybeTakeTerminal() in 3 places

1. Process.WhenStopped() - this change
1. Process.WhenDone() 
1. Pipeline.WhenDone()

Are there any potential bugs in the other places?

I think the problem is that we are not clear about WhenDone() and WhenStopped are actually called

I think I will remove WhenContinued() from `Fg()` and see what happens -- that one seems  wrong

melvinw:
Ah, yes! We should definitely apply a similar fix in `WhenDone()`. You can reproduce this same issue by doing, say `sleep 5 &` then running `cat` in the foreground, waiting 5 seconds, then once you press enter you get stopped. We could potentially hide this inside `MaybeTakeTerminal()`, but we'd have to change the signature to take an arg for who to take it from

melvinw:
I think `WhenDone()` will require adding state like you suggested a few comments up. We can't use `getpgid()` in that case since the processes in question don't exist anymore. I'll try a few things for this tomorrow

melvinw:
Note to self: add test for `WhenDone()` case

core/process.py
```diff
@@ -1019,6 +1020,8 @@ def StartProcess(self, why):
         for st in self.state_changes:
             st.ApplyFromParent(self)
 
+        self.pgid = posix.getpgid(pid)
```
melvinw, line 1023: potential race here: if the child wakes up before the parent and exits before this call, this might not work...

melvinw: probably very unlikely and not worth worrying about?

andychu:
Hm yeah honestly I am having trouble thinking about how to take the terminal back safely when there are many concurrent processes

It might be useful to look at comments in other shells

core/process.py
melvinw: decided relying on kernel state here is dangerous and added some of our own state to track whether procs are in the fg or bg

melvinw:
> Hm yeah honestly I am having trouble thinking about how to take the terminal back correctly when there are many concurrent processes
> 
> It might be useful to look at comments in other shells

Oh, whoops. Missed this.

zsh's job control source is probably the most intelligible. They wrap `tcsetpgrp()` in this helper called `attachtty()` (see: https://github.com/zsh-users/zsh/blob/c4ec7442f1138f2c525f98febb0db6def0fbe142/Src/utils.c#L4710)

Like this patch, they guard calls to this `attachtty()` with checks on some per-job state. Like here: https://github.com/zsh-users/zsh/blob/c4ec7442f1138f2c525f98febb0db6def0fbe142/Src/jobs.c#L550-L555

melvinw:
I think this will end up sharing whatever state we use for determining if a process in the bg/fg with #1675. Let's take care of that there and wait for that to merge before this

spec/stateful/job_control.py
```diff
@@ -266,6 +266,31 @@ def suspend_status(sh):
...
+  sh.sendline('sleep 1 &') # exit
+  time.sleep(0.1)  # seems necessary
+
+  # background cat should have been stopped by SIGTTIN immediately, but we don't
+  # hear about it from wait() until the foreground process has been started because
+  # the shell was blocked in readline when the signal fired.
+  sh.sendline('python2 -c "import sys; print(sys.stdin.readline().strip() + \'bar\')"')
+  if 'osh' in sh.shell_label:
+    # Quirk of osh. TODO: supress this print for background jobs?
+    sh.expect('.*Stopped.*')
```
melvinw, line 284: Resolving this TODO should be easy with the changes from #1675

core/process.py
```diff
@@ -1069,7 +1078,7 @@ def WhenDone(self, pid, status):
         self.state = job_state_e.Done
         if self.parent_pipeline:
             self.parent_pipeline.WhenDone(pid, status)
-        else:
+        elif self.position == POSITION_FOREGROUND:
```
andychu, line 1081: Hm thinking about my comment on the other PR

Can this be `elif not self.in_background` , and we assume that processes / pipelines are in the foreground except when run with `&`?

And then it changes state on STOPPED I guess?

It seems like we don't need 3 states?  2 should be enough?

melvinw: Yup. All this position stuff will be replaced with the things from #1675

melvinw:
going to combine this with #1675

PR #1682, opened 2023-07-26
[osh] Create error builtin

PossiblyAShrub:


PossiblyAShrub:
This is now ready for review.

osh/builtin_meta.py
```diff
@@ -328,6 +332,46 @@ def Run(self, cmd_val):
...
+            e_usage('Expected 1 positional argument but received %d' % len(positional), blame)
+
+        message_val = self.expr_ev.EvalExpr(positional[0], blame)
+        if message_val.tag() != value_e.Str:
+            raise error.InvalidType2(message_val, 'expected a Str', blame)
+        message = cast(value.Str, message_val).s
+
+        named = cmd_val.typed_args.named
+        status = 1
+        for arg in named:
```
andychu, line 360: This checking looks good for now, I like all the tests, though we also have some code to unify

- I added `frontend/typed_args.py`, which is used in some of the builtins
- you added `library/util.py`
- some functions in `ysh/val_ops.py` too, some of which are unused

test/spec.sh
```diff
@@ -899,6 +899,10 @@ ysh-user-feedback() {
   run-file ysh-user-feedback "$@"
 }
 
+user-errors() {
```
andychu, line 902: To be consistent, how about renaming this file `spec/ysh-builtin-error` ?

osh/builtin_meta.py
```diff
@@ -328,6 +332,46 @@ def Run(self, cmd_val):
...
 
+class Error(vm._Builtin):
+
+    def __init__(self, expr_ev):
+        # type: (expr_eval.OilEvaluator) -> None
+        self.expr_ev = expr_ev
+
+    def Run(self, cmd_val):
+        # type: (cmd_value.Argv) -> int
+        blame = loc.Word(cmd_val.arg_locs[0])
```
andychu, line 343: I think you can just use arg_locs[0] directly right?  that is a CompoundWord, which is a `loc_t`

andychu (commented):
Thanks for doing this!  Looks great

andychu:
And I think we can wait for awhile to unify the arg checking ... I want to see what the new proc param binding looks like, and this will fall out of that

In YSH it would be

```
proc error (message ; status=1) {
}
```

Oh also I was thinking about making the message optional, so there would be no excuse not to use `error 1` instead of `return 1` in proc ?

you could mechanically replace all the non-zero `return` with `error`

PossiblyAShrub:
> And I think we can wait for awhile to unify the arg checking ... I want to see what the new proc param binding looks like, and the arg checking abstraction will fall out of that

Agreed. I think it will also help to have a few more builtins and real functions to drive the design of a unified argument checker.

> Well I guess the introduces the problem of status as positional vs. status as named, so maybe it's too much flexiblity for now
> 
> It's not hard to rewrite `return 2` to `error (status=2)`

We could change `error` so that the status and message params are swapped, so arguments should take the form: `proc error (status ; msg="<default message>")`. In my mind, the error message is more important, if not equally so, than the error code. So it feels wrong to place the arguments in that order. Put another way, how idiomatic do we want the presence/absence of an error message to be?

A related design question: should calling `error` with no arguments be supported and/or encouraged? A somewhat realistic example:

```
proc test-conn (; maxWaitS=1) {
  const remote = selectRemote()

  # ping -> 0  "all good"
  # ping -> 2  "sent, but no response" (pretend this means we have a conn)
  try { ping -w $maxWaitS $remote }
  if (_status === 0 or _status === 2) {
    return
  } else {
    error  # default error message and default status code
  }
}
```

Then if we call this and it fails:

```
ysh$ test-conn
      error  # default error message and default status code
      ^~~~~
[ interactive ]:12: fatal: command test-conn failed with status 1
```

Maybe this isn't so bad assuming the line information is preserved... hmm.

andychu:
Yeah I think shell should have "whipupitude" so just typing `error` seems valid

Agree that it seems more natural for the message to come first, than the status

Another option, possibly weird, is to have

    error
    error 2
    error "message" 2

I kind of like `(status=2)` though



We can try it out and iterate

bar-g:
Hi, there,
a question or idea I had was if it might make sense to let `error` allow stacking of multiple errors, and calling `error` without an argument to return the last error value (on the stack)?

Idea would be to directly support complete error checking and validation without erroring out on each error, e.g. continuing with the execution of further validation but skipping any further operations. Something like this:

```
proc validate # e.g. prints some help error messages lines

try validate $a # prints some help message on error
if not error
then
  # run $a stuff
fi

try validate $b # print some more, even after previous error
if not error
then
  # run $b stuff
fi
```

bar-g:
An example that might better show the feature, might be to not have valitate() print the error messages directly, but to print all error messages only at the end, say after checking everything and showing adequate help etc., i.e. with something like a final`while echo $(error)` loop.

andychu:
In my experience with Python and shell, having a single fatal error is the most common idiom, and the one we're optimizing for with the `error` builtin

If you want multiple errors, then you can create a list of such things and write "normal code" to do that

e.g.

```
proc p (a, b, error_out) {
  _ error_out->append('myerror')
}
```

i.e. as the Go designers say, error logic is just program logic -- it shouldn't be special 

Our `error` is arguably a bit special, but not that much (it's handled much like errexit and divide by zero -- with `try`), and it's optimizing the 99% case and also adding the concept of a user-friendly string to shell

bar-g:
Hm, I understand single fatal error it's the old traditional idiom (OSH). The typical "rinse-and-repeat until that command line progam finaly works" thing. And everything else would be optional. Let me still add the motivation, at least for competeness.

My thought was that an internal error stack (or list) may allow (and foster) more elegant idioms, say for things like parallel execution, and even allow for ad-hoc operations to be short enough to fit into an interactive "one-liner", while being "fault-tolerant", completing were possible and nicely reporting where errors have happened. When the modern software standards generally show all errors at once, say from some given input (e.g. a form or config) it could sure be nice if the shell could also support that in some neat way (i.e. for shell commands with no boilerplate function/proc stuff).

If I remember that correctly, an area that could be interesting for you, didn't you write something about wanting to use  ysh/hay to execute things in parallel on different machines, then joining results? Maybe a test/use case.

Concerning the practical error() ideas, from my examples ideas above, I now think it would need to differentiate between
* calling error with argument/expression -> adds new error to stack (i.e. would make value or even message mandatory?)
* calling only error -> returns last error message and value (even if called repeatedly)
* pop_error -> return last message/value and switches to previous (for a final `while pop_error`)

bar-g:
Maybe an even better way to handle/defer multiple errors could emerge after table handling landed in ysh?

bar-g:
Hm, as the oils encourage proper handling of all errors and making them non-fatal (e.g. try ...) an internal error stack could also be a natural step to gain new and more useful usage patterns for oil scripts (like summarized/delayed error presentation/piping/etc.), i.e. for regular (non-user) errors. Or, to list multiple occurrences during oil:upgrade migrations.

bar-g:
Oil:upgrade migration experience/pattern idea: The process may change from hunting from failure to failure of a failing script, to reducing a list non-fatal error messages (given it runs well under osh, and (a parallel running?) ysh:upgrade mode emmits the errors messages to fix).

bar-g:
Kind of like shellcheck output list for IDEs.

andychu:
Well, we should discuss this on Zulip, there is a new #language-design stream with dozens of related threads

Including about errors, and about formatting of errors, etc.

https://oilshell.zulipchat.com/#narrow/stream/384942-language-design

I'd say a meta-level we are avoiding adding "innovative" features that may or may not work ou, and which will be unfamiliar to users  ... Almost everything has a solid precedent in either shell or Python/JS/Ruby/Lisp/...

(Probably should add that to our Language Design Principles)

PR #1683, opened 2023-07-27
[ysh] Implement varadic arguments in funcs

PossiblyAShrub:


PossiblyAShrub:
This is ready for review.

I stuffed a test over `mymax` in `spec/ysh-funcs.test.sh`. I think that should be fine for this PR.

spec/ysh-funcs.test.sh
```diff
@@ -217,3 +217,80 @@ hit: 8
   ]
 ]
 ## END
+
+#### Varadic arguments, no other args
+func f(...args) {
+  = args
```
andychu, line 223: Note that we might have to go back and change most of these to `json write (args)`, for stability

As part of the PyObject -> value_t rewrite, this format may change

Although it seems like people like it, so maybe we should try to keep it the same

PR #1684, opened 2023-07-29
Parser Interaction Errors

PossiblyAShrub:


PossiblyAShrub:
This PR should be ready for a review.

I've fixed all the  issues I mentioned in [#oil-dev > Crash in MoveToNextLine()](https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Crash.20in.20MoveToNextLine.28.29). Some more details on what I settled on:

1) We had the issue with mixing the expression parser and command parser (`{ = x }`)

This was best fixed by `Unread()`-ing the `}` token. I was worried that unreading would create a mess where a more elegant solution could exist, but the change seems to be reasonable and quite localized.

2) There was an issue with trailing white space (`{ echo $x }.`)

This was probably the most complex issue and relates to `_NewlineOkForYshCase`. I found that implementing `_NewlineOkForYshCase` took a lot of puppeteering, which was annoying. Instead, I decided to move it from `osh/cmd_parse.py` to `osh/word_parse.py`. This simplified the problem significantly! The resulting function is a little simpler and much more robust in the word parser.

3) Finally we had that issue where the left token in the case arm was being read incorrectly

This was a bit of a silly mistake, so it ended up being a pretty easy fix. However that fix depended on (2), so I've decided to keep it bundled in with the PR. The issue was we were trying to read this from `self.cur_word` before we had actually advanced the word parser over that first token.

osh/word_parse.py
```diff
@@ -1130,9 +1130,12 @@ def ParseCommandExpr(self):
         """
         enode, last_token = self.parse_ctx.ParseYshExpr(self.lexer,
                                                         grammar_nt.command_expr)
-        if last_token.id == Id.Op_RBrace:
-            last_token.id = Id.Lit_RBrace
-        self.buffered_word = last_token
+
+        if last_token.id == Id.Op_RBrace or last_token.id == Id.Op_Semi:
```
andychu, line 1134: I would add a comment here -- what invariant are we trying to preserve?  I guess it's tested in the unit tests, but it's nice to have a description too

It looks like this is 2 of 4 cases in end_stmt -- } and ;

The other 2 are Op_Newline and Eof_Real

(I wonder if we need those as buffered words -- maybe we can Unread there too ?)

osh/word_parse.py
```diff
@@ -1157,14 +1160,52 @@ def ParseProc(self, node):
...
+
+    def NewlineOkForYshCase(self):
+        # type: () -> Id_t
+        """Check for optional newline and consume it.
+
+        This is a special case of `_NewlineOk` which fixed some "off-by-one" issues
+        which crop up while parsing Ysh Case Arms. For more details, see
+        #oil-dev > Progress On YSH Case Grammar on zulip.
+
+        Returns a token id, first_id_out, which is filled with the choice of
```
andychu, line 1180: this comment is slightly out of date, it just returns an Id_t

osh/cmd_parse.py
```diff
@@ -1606,18 +1567,19 @@ def ParseYshCase(self, case_kw):
...
         while discriminant != Id.Op_RBrace:
             arm = self.ParseYshCaseArm(discriminant)
             arms.append(arm)
 
-            discriminant = self._NewlineOkForYshCase()
+            discriminant = self.w_parser.NewlineOkForYshCase()
 
-        ate = self._Eat(Id.Lit_RBrace)
-        arms_end = word_.BraceToken(ate)
+        ate = self._Eat(Id.Op_RBrace)
```
andychu, line 1580: This is a bit subtle -- I guess it means we're always in lex_mode_e.Expr  coming out of NewlineOkForYshCase()

And that sort of "infects" the WordParser, and we have to change it the "normal" Lit_LBrace

The difference is that in `echo }` it's Kind.Lit,  while in ` var d = {}` it's Kind.Op

Maybe add a comment about this

andychu (commented):
This looks great, thank you for diving ind an adding all the tests!  

(It seems the original "brute force" was not exhaustive enough since there were a whole bunch of bugs lurking ! )

The code and strategy looks good, I just requested  comments  in some places

----

A good lesson here is to put more stuff in the WordParser, I will make some  notes on Zulip, and see if this applies to any other bugs

andychu:
Another thing I noticed is that WordParser.ParseBareDecl and ParseBareDecl do  the  Op_RBrace -> Lit_LBrace thing inside the WordParser

I don't think we have to disturb them now, but it's worth thinking about  if  the interaction is inherently different and why  -- I won't be surprised if that's true ... hm

andychu:
Oh yeah sorry  I was thinking about the `Eof_Real` case -- that comes AFTER another line, so it interacts BADLY with Unread()

( We should probably have  more asserts in the actual code to flag that problem )

Sorry if I mislead ... I was just brainstorming about how to make the code cleaner, but correctness comes before cleanliness!

So if you're not sure about a change it's OK to  leave it as is --- I think this change was probably OK as is, but I was just giving my initial thoughts

buffered_word is a bit weird but it's not strictly related to this change

PossiblyAShrub:
That's alright. You were correct that we didn't need `buffered_word` for the other tokens. After handling the `EOF_Real` case the code is now cleaner and correct.

I have added some comments so this should now be ready for another review.

andychu:
Great thank you!

PR #1686, opened 2023-08-02
[core] clean up child process state on exit

melvinw:
Right now this table grows unbounded.

This commit also removes the log generated by `wait` when a job doesn't exist. This isn't a fundamental thing, but it keeps the spec tests "clean". Happy to roll that change back if you think the log will be helpful, @andychu.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -1117,10 +1117,7 @@ class Process(Job):
                     print_stderr('[%d] Done PID %d' % (self.job_id, self.pid))
 
                 self.job_list.RemoveJob(self.job_id)
-
-                # TODO: The `wait` builtin depends on the process object hanging
-                # around after completion. Fix that then uncomment this line.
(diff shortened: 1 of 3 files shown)


melvinw:
> - Why is the log change related?

I thought the log may have been breaking some tests that were checking stderr (and adding that log felt clunky), but it may have been a red herring. I will double check.

> - Why is the sleep change related?

If one of the processes in the multi-wait test exits after the others, `wait` will exit w/ 127 because the others will have been removed from the `JobList` state by then.

Now that I'm thinking about this some more, the other shells don't seem to have this problem... There's definitely  something fishy going on. We're probably doing something different in the `wait` or `JobList` implementations

melvinw:
Ah, yep. Needed a small tweak in `wait`. See 115f95f

andychu:
Ah OK that was a subtle bug!  Thanks

PR #1688, opened 2023-08-05
Define Funcs in the Common Variable Namespace

PossiblyAShrub:
This enables us to create dicts of funcs or alias funcs with `var`/`const`.


PossiblyAShrub:
This should be ready for review. The red in CI seems to be from https://github.com/oilshell/oil/commit/bf37374435799876b95a064e9b38332e17126707?

andychu:
Oh sorry did you base this branch off `soil-staging`?

`master` is always green, the CI prevents broken commit from going from soil-staging -> master

So when you create new branches, they should generally be from `master`

PossiblyAShrub:
> Oh sorry did you base this branch off `soil-staging`?

Locally I've got it based on `master`. Maybe setting "merge into `soil-staging`" on github implicitly rebases? I guess I should leave PRs on "merge into `main`" in the future.

andychu:
Hmm yeah that is weird ... it does seem like the branch picked up commits from soil-staging even though you didn't push

That is a bit of a hole

But I guess it makes sense for github to do that, hm

spec/ysh-funcs.test.sh
```diff
@@ -46,13 +46,71 @@ proc t() { return (0) }
...
+var f = 0
+func f() { return (1) }
+## status: 0
+## STDOUT:
+## END
+
+#### Vars cannot redefine functions (even with shopt -s redefine_proc_func)
+shopt -s redefine_proc_func
+func f() { return (1) }
+const f = 0
```
andychu, line 83: Hmm this is interesting

Usually we try to survey Python/JavaScript for these kind of semantics

- Python doesn't have const at all
- Let's see what JS does

```
function f() { return 6 }


console.log(f)
VM175:1 ƒ f() { return 6 }


f=42
42

console.log(f)
VM194:1 42
```

OK it does let you

andychu (commented):
OK nice tests!   (I'm working on the doc strucutre so we should document all this stuff, but it's great to have it there)

Two things seem a bit unusual, or at least differ from  JS/Python

- closures -- I had always intended to leave this out.  However I wonder if it even makes sense to have inner functions that do something different but have the same syntax.  We could also disallow the syntax ?
- making functions const.  Is the idea behind this to prevent people from overwriting them with variables?  I think it might be OK to rely on redefine_func_proc

Interested in thoughts on these design issues

andychu:
One quirk of shell is that the nesting of "shell functions":

```
f() {
  g() { echo g; }

  g  # can call g here
}

g  # can't call g yet
f  # defines g globally, there is a single namespace for shell functions
g  # now you can call it

```

It's not clear to me that this is useful ... it's a bit confusing

Arguably we could disallow this in YSH, and then also disallow the analogous thing for `func` ?

To me, dynamically defining functions is weird

(I have this long term idea to "catch typos", which people always complain about for Python/JS, and it  kind of goes against this)

---

Returning a function from a function also seems like it's not very useful without closures

I think we WON'T do closures now, but I know somebody is gonna complain about that ... So I wonder if we just disallow "what would look like closures", and then if people complain, say "there is space to implement that if you want"

(I actually haven't thought about how to implement it -- Python notably makes some weird choices here, JS does better.)

spec/ysh-funcs.test.sh
PossiblyAShrub: Not if you use the `const` qualifier. You are just mutating `f` (I suppose we should have some tests over `setvar` on a func name). JS will complain if you try to *redeclare* `f`:

```
>> function f() { return 6 }
undefined
>> const f = 42
Uncaught SyntaxError: redeclaration of var f
    <anonymous> debugger eval code:1
debugger eval code:1:1
```

But I would argue that `func f` should behave like `const f = <...>`, with the exception of `redefined_proc_func` which adds the ergonomics of being able to write `func f` multiple times in a REPL session. So just as we cannot:

```
const f = 42
const f = 42  # error
```

We shouldn't be able to do the same in either order of `const f`, `func f`.

---

Maybe we want to change `redefine_proc_func` to `redeclare`? Then, in a REPL, we would let users redeclare variables (with `const`/`var`), funcs and procs.

PossiblyAShrub:
re: Nesting funcs

I think this behavior is actually useful despite the fact that we do not capture. Defining functions within a function allows us to create private helpers:

```
func someLargeFunc() {
  func helper() {
    echo I am helping
  }

  echo I need some help
  _ helper()

  echo Keep working
}

_ someLargeFunc()
_ helper()  # fatal: Undefined variable 'helper'
```

This is nice because then no other functions have access to `helper`. But I agree, the "all declarations become global" shell behaviour is confusing; that is why I decided to declare functions with `scope_e.LocalOnly`.

---

re: Returning funcs from funcs

This is also still useful. For example, maybe we want to select functions using some criteria:

```
func mysql__listTables() {
  return ("SHOW TABLES")
}

func sqlite__listTables() {
  return (".tables")
}

func listTables(db) {
  case (db) {
    mysql { return (mysql__listTables) }
    sqlite { return (sqlite__listTables) }
    (else) { error ("Unknown db: $db") }
  }
}
```

While this could be done with just a dictionary, if we want to select more complex functions based on more complex criteria, being able to return a function is necessary.

andychu:
Hmm I would say we should defer a decision until we have some more experience

The code for this PR looks ready, but I think the tests are actually almost TOO detailed -- they are spec'ing behavior that we haven't yet written real code for .  

Ironically the tests are almost TOO green :-)

I think we should keep in mind max() and other use cases, e.g. I marked these as good for YSH:

```
$ grep IN-YSH library/func_cpython.py
    # NOTE: IN-YSH means we can move it to YSH itself
    SetGlobalFunc(mem, 'identity', lambda x: x)  # IN-YSH
    SetGlobalFunc(mem, 'max', max)  # IN-YSH with <
    SetGlobalFunc(mem, 'min', min)  # IN-YSH
    SetGlobalFunc(mem, 'abs', abs)  # IN-YSH with <
    SetGlobalFunc(mem, 'any', any)  # IN-YSH with Bool
    SetGlobalFunc(mem, 'all', all)  # IN-YSH with Bool
    SetGlobalFunc(mem, 'sum', sum)  # IN-YSH with +
    SetGlobalFunc(mem, 'reversed', _Reversed)  # IN-YSH with for
```

And then we also need `describe` and `argparse`, which depends on proc param binding

We can see if we need nested functions in that code and anything else we want to include in YSH 

---

On the specific examples

- Agree with the second listTables one -- but this case doesn't require func within func, syntactically.    But a func is a regular value so it's definitely valid to return one (in this case  defined at the top level)
- I agree about helper functions, but I believe there is a problem with scope ... On `#language-design` I posted and idea about "polyglot" or "homonym" tests ... 

I want to make another little test framework to compare against Python and JavaScript in an executable way

I have done a lot of this in the past,  e.g. https://github.com/oilshell/oil/tree/master/benchmarks/javascript

Basically Python and JS (and awk and R) will count as a "vote" -- this is how I did the original spec tests -- bash and zsh were "votes" in favor of a behavior

---

I also updated this page - https://github.com/oilshell/oil/wiki/Language-Design-Principles

The most important principle  is **users don't read the manual** -- i.e. so if we look like JavaScript, and the scope is slightly different, I view that as bad.  People will just try stuff they expect to work, and we should be way/careful of  differences

So that is why I lean toward making func within func a SYNTAX error, so there's zero confusion about scope, but we don't have to do that NOW

----

So I would say we should make at least one test in `ysh-funcs` fail -- e.g. could just as TODO:specify to the output and `oils_failures_allowed:1`

As a reminder to ourselves that this isn't done yet -- we keep more of a "blank mind" and write code first, to test the language  (there are failing tests in many other `ysh-*` as reminders of TODOs as well)

We don't have to specify behavior the minute we implement it -- we can adjust based on usage (and I think the new Python/JS tests will help with this, although it may take me awhile before getting to it )

PossiblyAShrub:
Sorry, I forgot to mention: I've added a test marking nested funcs as an open design hole. This is ready for review now.

andychu:
Great thanks!

PR #1691, opened 2023-08-08
[fix] clarify some unclear comments

not-my-profile:
While fixing all the typos in this repository, I discovered that these comments don't actually make sense to me.

If we already touch these comments to fix the typos, we should make sure they actually make sense.

### DIFF
--- a/opy/_regtest/src/osh/cmd_parse.py
+++ b/opy/_regtest/src/osh/cmd_parse.py
@@ -511,7 +511,7 @@ class CommandParser(object):
         ls
         >out.txt
 
-        >out.txt FOO=bar   # this touches the file, and hten
+        >out.txt FOO=bar   # this touches the file
 
     Or any sequence:
         ls foo bar
--- a/spec/redirect.test.sh
+++ b/spec/redirect.test.sh
@@ -597,7 +597,7 @@ status=0
 
 #### can't mention big file descriptor
 echo hi 9>&1
-# 23 is the max descriptor fo rmksh
+# trivia: 23 is the max descriptor for mksh
 #echo hi 24>&1
 echo hi 99>&1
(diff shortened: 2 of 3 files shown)


opy/_regtest/src/osh/cmd_parse.py
```diff
@@ -511,7 +511,7 @@ def ParseSimpleCommand(self):
         ls
         >out.txt
 
-        >out.txt FOO=bar   # this touches the file, and hten
+        >out.txt FOO=bar   # this touches the file, and then
```
not-my-profile, line 514: and then ... what?

spec/redirect.test.sh
```diff
@@ -597,7 +597,7 @@ status=0
 
 #### can't mention big file descriptor
 echo hi 9>&1
-# 23 is the max descriptor fo rmksh
+# 23 is the max descriptor for mksh
 #echo hi 24>&1
 echo hi 99>&1
 echo hi 100>&1
```
not-my-profile, line 603: Doesn't make sense to me. If `echo hi 24>&1` is commented out because 23 is the max descriptor for mksh, then why does the next lines go up to 99 and 100?

andychu: This is a bit of mksh trivia that's not important to test, since the tests agree on a common behavior for all shells

It's OK to just fix the typo here, thanks

opy/_regtest/src/osh/cmd_parse.py
andychu: I would just remove "and then"

andychu (commented):
Thanks for taking a look!

not-my-profile:
Closing in favor of #1693 (since GitHub doesn't let you update the source branch of a PR).

PR #1697, opened 2023-08-14
[ysh] implement len() as vm.Callable

melvinw:
This patch implements `len()` over `value_t`. It also fixes some error presentation code in the command evaluator

### DIFF
--- a/library/func_misc.py
+++ b/library/func_misc.py
@@ -167,15 +167,7 @@ class Len(vm._Callable):
         UP_x = x
 
         with tagswitch(x) as case:
-            if case(value_e.BashArray):
-                x = cast(value.BashArray, UP_x)
-                return value.Int(len(x.strs))
-
-            elif case(value_e.BashAssoc):
-                x = cast(value.BashAssoc, UP_x)
-                return value.Int(len(x.d))
-
-            elif case(value_e.List):
+            if case(value_e.List):
                 x = cast(value.List, UP_x)
                 return value.Int(len(x.items))
 
--- a/library/func_misc.py
+++ b/library/func_misc.py
@@ -4,7 +4,7 @@ func_misc.py
 """
 from __future__ import print_function
 
-from _devbuild.gen.runtime_asdl import value, value_t, value_e
+from _devbuild.gen.runtime_asdl import value, value_str, value_t, value_e
 from _devbuild.gen.syntax_asdl import loc
 from core import error
 from core import vm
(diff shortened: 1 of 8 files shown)


library/func_misc.py
```diff
@@ -147,3 +147,44 @@ def Call(self, pos_args, named_args):
...
+        # type: () -> None
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        n_args = len(pos_args)
+        if n_args != 1:
+            raise error.InvalidType("len() expects exactly one argument but %d were given" %
+                          n_args, loc.Missing)
```
melvinw, line 164: Following the conventions in `frontend/typed_args.py` for consistency. I noticed python also raises `TypeError` in this case, but it feels like it should be something else...

melvinw:
Nice. Healthy jump in `delta-cpp` for the ysh spec tests from this. Down to 116 w/ 420/536 passing. Was 141 before

http://travis-ci.oilshell.org/github-jobs/4753/cpp-spec.wwz/_tmp/spec/ysh-cpp/compare.html

library/func_misc.py
```diff
@@ -147,3 +147,44 @@ def Call(self, pos_args, named_args):
...
+
+    def __init__(self):
+        # type: () -> None
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        n_args = len(pos_args)
+        if n_args != 1:
```
andychu, line 162: This check looks good for now, though I also sketched this typed_args.Reader thing that's not implemented

Should discuss on Zulip

When we do proc argument binding, this discussion of the style will probably be more relevant

library/func_misc.py
```diff
@@ -147,3 +147,44 @@ def Call(self, pos_args, named_args):
...
+            raise error.InvalidType("len() expects exactly one argument but %d were given" %
+                          n_args, loc.Missing)
+
+        x = pos_args[0]
+        UP_x = x
+
+        with tagswitch(x) as case:
+            if case(value_e.BashArray):
+                x = cast(value.BashArray, UP_x)
```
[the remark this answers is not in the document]
melvinw: Oh, hmm... A few spec tests rely on `len()` being defined for `BashArray` and `BashAssoc`. Will patch those

andychu:
Oh yeah sorry some of the tests like `var a = Dict()` are over-specified -- now I don't think we'll even have that synonym, it's just `var a = {}`

Remember a couple months ago I said the initial pass of YSH was "best first guess"

Looks like other tests are obsolete.  I did a lot of this pruning for the 0.16.0 release -- "breaking renames and YSH"

so yeah we should break these, let me make a pass right now

andychu:
OK in doing the spec tests, I think we may want

- Bool.fromObj() -- equivalent of Python's bool()
- Int.fromStr(s, base) -- like Python `int('ff', 16)`
- Float.fromStr(s) -- like Python `float()`,  is this different than JSON notation for float?  Not sure
- `read --lines :branches` instead of `readarray -t :branches` (bash syntax)

Let me know what you think, looking for feedback

We kinda have to "scrub" the language for consistency / design while porting to C++

The basic idea is to do the MINIMAL thing that can express say a flag parser `argparse` or a test framework `describe`

Other stuff can be added later, but we want to avoid writing a lot of code, and avoid a big test matrix (hence getting rid of length of BashArray etc.)

melvinw:
Ah great. Thanks for taking care of those tests

melvinw:
> Oh yeah sorry some of the tests like var a = Dict() are over-specified -- now I don't think we'll even have that synonym, it's just var a = {}

Agreed. All `Dict()` provided in my mind was a way to copy, but I think we should provide a generic way to do that over all `value_t`. Something like `obj.Copy()` or `obj.Clone()`

> OK in doing the spec tests, I think we may want
> 
>     * Bool.fromObj() -- equivalent of Python's bool()

`Bool()` feels cleaner to me. Why do we want the dot syntax here?

>     * Int.fromStr(s, base) -- like Python `int('ff', 16)`

`fromStr()` makes more sense here. Do we want to support creating ints from floats too? e.g. python has
```
$ python
Python 2.7.18 (default, Jul  9 2023, 16:30:32)
[GCC 12.2.0] on linux2
>>> int(1.2)
1
```
>     * Float.fromStr(s) -- like Python `float()`,  is this different than JSON notation for float?  Not sure

Similar questions here? Couldn't this be `Float()` and it can accept both `value.Str` and `value.Int`?

PR #1700, opened 2023-08-16
[ysh] Add --builtin flag to source

PossiblyAShrub:
One issue I have left to iron out is what happens when we call this from OSH?

Right now we get a syntax error, but I wonder if we should just outright ban `source --builtin` from OSH?

### DIFF
--- a/osh/builtin_meta.py
+++ b/osh/builtin_meta.py
@@ -102,6 +102,7 @@ class Source(vm._Builtin):
 
         if arg.builtin:
             try:
+                path = pyutil.os_path.join("stdlib", path)
                 contents = self.loader.Get(path)
             except IOError:
                 self.errfmt.Print_(
--- a/osh/builtin_meta.py
+++ b/osh/builtin_meta.py
@@ -22,6 +22,7 @@ from frontend import lexer
 from frontend import reader
 from frontend import typed_args
 from mycpp.mylib import log
+from pylib import os_path
 from osh import cmd_eval
 from ysh import expr_eval
 
@@ -102,9 +103,9 @@ class Source(vm._Builtin):
 
         if arg.builtin:
             try:
-                path = pyutil.os_path.join("stdlib", path)
+                path = os_path.join("stdlib", path)
                 contents = self.loader.Get(path)
-            except IOError:
+            except (IOError, OSError):
                 self.errfmt.Print_(
                     'source --builtin %r failed: No such builtin file' % path,
                     blame_loc=cmd_val.arg_locs[2])
--- a/osh/builtin_meta.py
+++ b/osh/builtin_meta.py
@@ -101,7 +101,14 @@ class Source(vm._Builtin):
         arg_r.Next()
 
         if arg.builtin:
-            contents = self.loader.Get(path)
+            try:
+                contents = self.loader.Get(path)
+            except IOError:
+                self.errfmt.Print_(
+                    'source --builtin %r failed: No such builtin file' % path,
+                    blame_loc=cmd_val.arg_locs[2])
+                return 2
+
             line_reader = reader.StringLineReader(contents, self.arena)
             c_parser = self.parse_ctx.MakeOshParser(line_reader)
             return self._Exec(cmd_val, arg_r, path, c_parser)
--- a/osh/builtin_meta.py
+++ b/osh/builtin_meta.py
@@ -31,10 +31,11 @@ from typing import Dict, List, Tuple, Optional, TYPE_CHECKING, cast
 if TYPE_CHECKING:
     from _devbuild.gen.runtime_asdl import Proc
     from frontend import args
-    from frontend.parse_lib import CommandParser, ParseContext
+    from frontend.parse_lib import ParseContext
     from core import optview
     from core import ui
     from osh.cmd_eval import CommandEvaluator
+    from osh.cmd_parse import CommandParser
 
 
 class Eval(vm._Builtin):
--- a/osh/builtin_meta.py
+++ b/osh/builtin_meta.py
@@ -30,7 +30,8 @@ _ = log
 from typing import Dict, List, Tuple, Optional, TYPE_CHECKING, cast
 if TYPE_CHECKING:
     from _devbuild.gen.runtime_asdl import Proc
-    from frontend.parse_lib import ParseContext
+    from frontend import args
+    from frontend.parse_lib import CommandParser, ParseContext
     from core import optview
     from core import ui
     from osh.cmd_eval import CommandEvaluator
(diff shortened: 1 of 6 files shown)


bar-g:
> to iron out is what happens when we call this from OSH?

I thought some posix shell code in stdlib/ is also ok. Currently,`oil-polifil.sh`. So maybe restrict OSH to `*.sh`?

For example, I could see it usefull to be able to `source --buildin` some centrally maintained settings from OSH.
It would only have to fail in other shells, and that's easily catched by adding `|| true`

spec/ysh-source.test.sh
```diff
@@ -0,0 +1,24 @@
+# spec/ysh-source
+
+## our_shell: ysh
+## oils_failures_allowed: 0
+
+#### --builtin flag
+source --builtin stdlib/math.ysh
```
andychu, line 7: I wonder if we should make `stdlib/` the implicit root?   So it's just `source --builtin math.ysh`

I think by definition, `stdlib/` is exactly the things that are built into the binary

osh/builtin_meta.py
```diff
@@ -75,66 +77,81 @@ def Run(self, cmd_val):
...
         path = arg_r.Peek()
         if path is None:
             e_usage('missing required argument', loc.Missing)
         arg_r.Next()
 
-        resolved = self.search_path.Lookup(path, exec_required=False)
-        if resolved is None:
-            resolved = path
+        if arg.builtin:
+            contents = self.loader.Get(path)
```
andychu, line 104: I think we should catch the IOError and print a message, and return error 2

Right now it looks like

```
$ bin/ysh -c 'source --builtin stdlib/math.ys'
oils I/O error (main): No such file or directory
```

( I actually think it would be nicer if loader.Get() returns None not IOError ... i.e. `Optional[str]`  ... that was left over from the file interface )

andychu (commented):
Yes good point about the syntax errors from OSH.  This is because `shopt -u parse_func` in OSH, which I think I agree with

It is a downside of writing functions in YSH -- if they were in Python, then you could reuse them from OSH

I think we should do something eventually, but for now I think it is not that bad that sourcing "math.ysh" from "osh" will not work

You can actually do

```
echo hi

shopt -s ysh:all {
  source math.ysh
}

# now back to OSH
echo hi

```

However that blows up our test matrix, so I would leave it out until someone wants it

---

Actually what we SHOULD test  is that `source --builtin math.ysh` works with `shopt --set ysh:upgrade` !

So I would remove the `our_shell:` line in the spec tests, and test with explicit `shopt --set`

---

Thanks, this looks great overall

andychu:
> I thought some posix shell code in stdlib/ is also ok. Currently,`oil-polyfill.sh`. So maybe restrict OSH to `*.sh`?
> 
> For example, I could see it usefull to be able to `source --buildin` some centrally maintained settings from OSH. It would only have to fail in other shells, and that's easily catched by adding `|| true`

Yes that's a good reminder, from osh you can do `source --builtin compat-stuff.sh` or `foo.osh`

So I think we have an obvious convention -- the extension should just match the shell you're using

We could hard-code that in now I suppose -- pass `lang` to `Source`, and then enforce that the file extension matches when sourcing (because the error message is better)

I don't see any downsides to this now I guess.  The main functions I want are `log()` and `die()` -- I use those everywhere

PossiblyAShrub:
I have addressed all of your feedback, this should be ready for another review.

I decided against checking `lang` in `Source` for now, as that doesn't account for `shopt --set ysh:upgrade` in scripts which are invoked with `bin/osh`. I figured we might want to think about that issue a bit more.

bar-g:
Something I have not understood yet is the different naming of the option `--buildin` vs. the source dir and internal reference `stdlib/`. Would using only one word or the other consistently not make sense?

PossiblyAShrub:
The idea behind the `—builtin` flag is that it allows us to source files which have been embedded into the binary. As I understand, we are anticipating other scripts to be embedded into the binary as well, but they may not necessarily be considered to be the “standard library”. So it’s not that “builtin” = “stdlib”, but rather “builtin” is more general than “stdlib”.

bar-g:
> `Source` for now, as that doesn't account for `shopt --set ysh:upgrade` in scripts which are invoked with `bin/osh`.

Hm, I no more or less nothing of the internals, but aren't the current `ysh:upgrade` and `ysh:all` states accessible from within `Source`?

andychu:
Hmm, that is an inconsistency

I guess I would lean toward renaming our internal dir  to `builtin/` in that case?

I liked `source --builtin math.ysh` because it means "math.ysh is builtin to the binary"

And also `builtin test` means "run the test that's in the binary" (although that's also the default)

I think that would be fine, then we have

```
library/
  func_foo.py  # builtin functions
  builtin_foo.py  # builtin commands
builtin/
  math.ysh  # builtin functions and commands implemented in YSH
```

It's not super consistent but that's only our internal organization ... what do you think?

andychu:
I suppose if we really want to be more consistent, we could do

```
library/
  func_foo.py
  proc_foo.py   # although this could be confusing, "builtin" is already accepted names for this
```

----

I'm OK with leaving out the `lang` check for now

We can add it later

Right now there is no introspection on `ysh:upgrade` -- it's just a bunch of options that can be set or unset individually

osh/builtin_meta.py
```diff
@@ -75,66 +78,89 @@ def Run(self, cmd_val):
...
-                                if e.IsReturn():
-                                    status = e.StatusCode()
-                                else:
-                                    raise
+
+        with dev.ctx_Tracer(self.tracer, 'source', cmd_val.argv):
+            source_argv = arg_r.Rest()
+            with state.ctx_Source(self.mem, path, source_argv):
+                with state.ctx_ThisDir(self.mem, path):
+                    src = source.SourcedFile(path, call_loc)
```
andychu, line 151: I think we should change this to 2 options

- `source.DiskFile()`
- `source.BuiltinFile()` 

or something.  And make the description in errors nicer.   It will be confusing for users if they try to find "math.ysh".


But I guess we can do that when we make a pass over error messages

andychu (commented):
Looks good, thanks!

bar-g:
>“builtin” is more general than “stdlib”.

Thanks, ok. In that way it's more understandable if all scripts need to be accessed through `source`. Or could that be different commands or command line options?

Is there any overlap of `--buildin` with shell buildins? If `--stdlib` isn't suitable, maybe something like `--internal` or `--embedded`, or `--blob` could be less confusing.

bar-g:
Wasn't there also some saying around "the stdlib of shell is" the unix/gnu coreutils or something, and we could avoid confusing that?

As compared to classic shell buildins, the adding of features with an internal source option is, well optional, maybe name it something in the direction of `--opt-in` and `opt-in/`?

bar-g:
Then?:
buildin: allways available shell commands
opt-in: optional shell commands (the fact that they are build into the binary is only a convenience/versatility implementation detail)
Opt-in scripts may be `source`ed into user scripts, and some even executed directly, e.g. with `osh --latest`.

bar-g:
`--loadin`

bar-g:
Could two separate flags be a solution to the fluent lang relation?
Instead making the storage and the loading explicit choices `--oshlib`, `--yshlib`?

bar-g:
On a second thought, having separate `-oshlib` and `--yshlib` is a bad idea, it would make source code incompatible.
And on that same thought,

1. couldn't the scripts themselfes be left responsible for being able to run in osh and ysh, and to return proper error messages, and that to be made a requirement for inclusion? 


But the larger inconsistency I see here may not be that easy to see from the developers view:

Users know shell has some buildin commands. And OSH and YSH (oils) says it's having some, e.g. math functions,  "buldin" . But they don't just work out of the box, they require to be explicitly "sourced" first. So this "buldin" doesn't really seem to work like a usual "buildin" at all. 

2. This (nice!) feature just seems to be more of a `--loadin` or `source --opt` " new thing" :-).

PR #1705, opened 2023-08-18
[frontend] Fill Out typed_args::Reader

melvinw:
This patch implements the positional and named arg parts. I left out the stuff for procs for now, since I haven't been following the discussions around what we need there.

I added some unit tests to serve as documentation of sorts. Let me know if you think we should add more comprehensive ones though.

### DIFF
--- a/frontend/typed_args.py
+++ b/frontend/typed_args.py
@@ -8,6 +8,7 @@ from _devbuild.gen.syntax_asdl import (loc, ArgList, BlockArg, command_t,
 from core import error
 from core.error import e_usage
 from mycpp.mylib import tagswitch
+from ysh import val_ops
 
 from typing import Optional, Dict, List, cast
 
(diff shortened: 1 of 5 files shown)


frontend/typed_args.py
```diff
@@ -113,32 +114,124 @@ def RestWords(self):
 
     ### Typed positional args
 
-    # TODO: may need location info
     def PosStr(self):
         # type: () -> str
-        return None  # TODO
+        try:
```
andychu, line 119: Hm I would like to see this API used in context.  We need to test it out before knowing if it's right

And particularly have a demo of the error message it produces e.g.


- `len()`  --  expected 1 argument, got 0
- len(1, 2)` -- expected 1 argument, got 2
- `_match('group')` and `_match(1)` -- I think that's our API for regexes, it takes either a string or int, like Python's `m.group(1)` and `m.group('name')`

"No arguments left" isn't a good error message

- I think it should use a counter internally, and also avoid IndexError
-  I think we can inline `MustBe*` into those functions, or at least put them in this file, and eliminate other usages

i.e. we should try to convert `args.Spec` and `MustBe*` and `DoesNotAccept` all into this single API

---

But if you want to do something more minimal, I would just start with `len()` and then see what the errors look like


i.e.  I try generalize from specific examples

melvinw: > "No arguments left" isn't a good error message
> - I think it should use a counter internally, and also avoid IndexError

Oh sure. This message could be better. I don't see how catching `IndexError` gets in the way of this or why we need extra state, though. What if we add a new method, like `ExpectNPos(n)` (and maybe `ExpextAtLeastN(n)`) and have the caller use that before any of the other `Pos*()`

> i.e. we should try to convert args.Spec and MustBe* and DoesNotAccept all into this single API

It won't be useful to have those helpers for use elsewhere? i.e., it will be awkward to have to pack an individual `value_t` into a list and construct a `Reader` in non-function-like/non-proc-like contexts. Or do you not think there will be many cases like that?

melvinw: Whoops. Missed this one.

> Hm I would like to see this API used in context. We need to test it out before knowing if it's right

The new unit tests don't capture this well enough? What else do you have in mind?

andychu: The unit tests are helpful, but synthetic usages ... At least we want to use it in len() and a few other places, and see what the API looks like

And I'll push and look at the error messages

At the very least it should replace the typed_args.Spec usages, which we all agreed about, and we can delete that API

 ExpectNPos() is redundant, so I would try to avoid that, similar to typed_args.Spec

---


As far exceptions, it may not matter here, but we should avoid using them internally -- i.e. when not caught at the top level and exiting the program.

It's the only case where C++ is slower than Python, see here it's 1.7x slower, and other cases are 100x faster

https://www.oilshell.org/release/0.17.0/benchmarks.wwz/mycpp-examples/

(C++ exceptions are optimized for the non-throw path, so the throw path is very slow)

---

The imperative nature of the API should also be hidden in the Done() error messages

melvinw: Ah damn! I didn't realize c++ exceptions were so much slower 👀 Will make a few tweaks

melvinw:
About to push a few fixes. Here are some samples of the different messages for different positional arg errors. These are also covered in the unit tests now (the behavior, not the message text)

```
ysh$ var s = 'foobar'

ysh$ = s->startswith(1)
  = s->startswith(1)
  ^
[ interactive ]:5: fatal: expected value.Str, but got value.Int

ysh$ = s->startswith()
  = s->startswith()
  ^
[ interactive ]:6: fatal: Expected at least 2 arguments, but only got 1

ysh$ = s->startswith('foo')
(Bool)   True
ysh$ = s->startswith('foo', 'blah')
  = s->startswith('foo', 'blah')
  ^
[ interactive ]:8: fatal: Expected 2 arguments, but got 3
```

andychu:
Thanks, the error messages look great!   (capitalize the first "expected" to be consistent)

Is this ready to look at?

andychu:
I took a look at the code, looks great!  

Only thing is that `->startswith()` says it takes 2 arguments, which is true including `self`, but could be confusing

Although I guess we already planned to do the thing where we allow `strip(s)` and `s->strip()`

So maybe we can defer that issue until then

Python also has that issue -- if you don't know that `self` counts as an arg, the errors can be a bit confusing

melvinw:
Ah yes this is ready! I'll fix that "expected" while you look at the rest

melvinw:
OK pushed!

andychu:
Thanks!

PR #1709, opened 2023-08-25
[ysh] Define identity, max, min and abs in YSH

PossiblyAShrub:


PossiblyAShrub:
This should be ready for a review.

stdlib/math.ysh
```diff
@@ -1,79 +1,79 @@
...
+  case (len(args)) {
+    (1) { return (_select(args[0], max)) }
+    (2) {
+      if (args[0] > args[1]) {
+        return (args[0])
+      } else {
+        return (args[1])
       }
     }
+    (else) { error ("max expects 1 or 2 args") }
```
andychu, line 42: It's kinda weird we get the status 1 here, but we have explicit status 3 elsewhere

Maybe we should change the default `error` status to 3 ?

Or `4` ?  

Just because it's different than say `ls` failures which are 1 or 2

I guess the issue is that the builtins in Python will `raise error.InvalidType()`, which gives a 3 status code

So then it becomes harder to get consistent behavior between builtins implemented in Python vs YSH

stdlib/math.ysh
```diff
@@ -1,79 +1,79 @@
...
 
-# This causes a name collision with `abs`. The rename is a work around for now.
-func abs_(x) {
-  if (x < 0) {
-    return (-x)
-  } else {
-    return (x)
+  var match = list[0]
+  for i in (1:len(list)) {
+    setvar match = cmp(list[i], match)
```
andychu, line 16: It's cool this higher order function works!


It reminds me that we may want a `strcmp()`, because `< >` are arithmetic only on purpose, unlike in Python

stdlib/funcs.ysh
```diff
@@ -0,0 +1,5 @@
+func identity(x) {
+  ## The identity function. Returns it's argument.
```
andychu, line 2: it's -> its   (since it's not short for "it is")

andychu (commented):
Thanks, looks great!

stdlib/math.ysh
```diff
@@ -1,79 +1,79 @@
+func _select(list, cmp) {
```
andychu, line 1: Hm I realized that this naming convention isn't YSH-like

Because we have public APIs like `_match()` and `_status` .  That convention came from PHP, where you have `_GET[]` and `_POST`.    I needed something that was easy to type but distinct from the user's variables

So we could maybe settle on `__select`, or even `__math_select` which is kind of Python style

( It is a bit annoying that we have a big global namespace, not sure how high a priority changing that is though )

stdlib/math.ysh
PossiblyAShrub: AFAIK error code `1` is used as a catch-all error code, so maybe it's fine left as the default?

But you are right, we should use a status `3` here to behave like the `raise error.InvalidType()` that we threw in the old builtins.

PossiblyAShrub:
I've updated my code in response to your feedback. This should be ready for another review.

andychu:
Thanks, merged!

PR #1712, opened 2023-08-30
[translation] translate a few more builtins

melvinw:


melvinw:
Nice. This knocks off a few more spec tests. Delta is down to 91

andychu:
OK awesome, is it ready to look at?  (still marked draft)

melvinw:
Oh, whoops. Yup it's ready

library/func_misc.py
```diff
@@ -153,3 +153,55 @@ def Call(self, pos_args, named_args):
...
+        # type: () -> None
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        r = typed_args.Reader(pos_args, named_args)
+        li = r.PosList()
+
+        delim = r.NamedStr('delim', '')
```
andychu, line 170: Hm not sure about this named arg?

- Do Python or JS have it?
- If we have it, we should document it (doc/ref is kinda ready), and have test coverage

I would be inclined to leave it out, it seems like `join(x, '+')`  is relatively clear

library/func_misc.py
```diff
@@ -153,3 +153,55 @@ def Call(self, pos_args, named_args):
...
+        if len(pos_args): # reader has a reference
+            delim = r.PosStr()
+
+        r.Done()
+
+        strs = []  # type: List[str]
+        for i, el in enumerate(li):
+            msg = 'join() items must be Str but item %d is %s' % (
+                i, value_str(el.tag()))
+            strs.append(val_ops.ToStr(el, msg, loc.Missing))
```
andychu, line 180: what do you think about val_ops.py.Stringify() ?  It is used for splicing like `@array` and also `$str`


I guess that doesn't match what Python or JS do, so it's up for debate

library/func_misc.py
```diff
@@ -153,3 +153,55 @@ def Call(self, pos_args, named_args):
...
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        r = typed_args.Reader(pos_args, named_args)
+        val = r.PosValue()
+        r.Done()
+
+        if val == value.Undef or val == value.Null:
```
andychu, line 198: Undef should only be used for signaling a variable is undefined, so it won't be passed to a function

You could add `assert val != value.Undef` to be sure

library/func_misc.py
```diff
@@ -153,3 +153,55 @@ def Call(self, pos_args, named_args):
...
+            return value.List([])
+
+        s = val_ops.ToStr(
```
andychu, line 201: Hmm looks like the original didn't convert to string, but it may not have been well thought out

Hm let me make sure there is a placeholder for documenting all this stuff

andychu: Actually the intended usage is `@[maybe(x)]`, and `@[]` already does the string conversion if necessary -- it calls `val_ops.Stringify()`, so I think we should leave this out

andychu:
Yeah I just tested Julia (with `demo/julia.sh`), and args must be EITHER positional or keyword, not both

The Python style causes subtle breakage when you rename params, and combined with varargs, mentioned here:

https://www.oilshell.org/blog/2023/06/ysh-sketches.html#julia-like-signatures-fix-python


The problems led to these fixes, which we don't want - https://peps.python.org/pep-0570/

---

I'm having trouble finding a good explanation, but it is documented here:

http://web.mit.edu/julia_v0.6.2/julia/share/doc/julia/html/en/manual/noteworthy-differences.html

> In Julia, keyword arguments must be passed using keywords, unlike Python in which it is usually possible to pass them positionally. Attempting to pass a keyword argument positionally alters the method signature leading to a MethodError or calling of the wrong method.

---

This is subtle and we still have to implement the param binding algorithm, but we should take that lesson from Python history, and Julia

library/func_misc.py
melvinw: Ah yeah. I'm OK to drop this. IIRC nothing depends on it right now, so it's just here because it was part of the original impl (https://github.com/oilshell/oil/pull/1712/files#diff-1b57cb0469c5f87781d4b8ac574803c6ca501351c865e4d8a10f04343d33700aL38)

library/func_misc.py
melvinw: Stringifying would actually be pretty nice. It's very jarring when I'm using python interactively and `','.join(foo)`  barfs because some member of `foo` is an int or something making you have to type out a whole `map(...)` to serialize things. For an interactive shell, I think it's appropriate to have some magic here

library/func_misc.py
melvinw: But, if it won't break anything, being less strict here seems fine. I'm not sure how the `if len(s)` guard around the "some" return path in the original would be generalized, though (e.g. to floats/ints)

andychu:
Oh sorry I may have actually misunderstood `@[maybe(x)]` 

What it is supposed to emulate is the "empty elision" behavior of shell, e.g. if you have `s=''` and then unquoted `ls one $s two`, then it gets removed

So it is a genuine design question whether to generalize it

Arguably in addition to `''` and `'non-empty'`, you might want `null` and some other kind of object you can stringify

Like a number?  

```
if (cond()) {
  var i = 3
} else {
  var i = null
}
mycommand first @[maybe(i)] second    # omit arg if it's not a number

```

Not even sure that makes sense ...


So I guess we can keep the original behavior, sorry for the confusion

melvinw:
Ah ok good. I was worried I was missing something re: `maybe()`. Leaving it as-is sounds good to me. I'll still fix that undef vs null check though

melvinw:
Whoops. Forgot to bump this. Is ready for another round

core/shell.py
```diff
@@ -677,6 +678,20 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
     func_init.SetGlobalFunc(mem, 'block_as_str', block_as_str)
     func_init.SetGlobalFunc(mem, '_hay', hay_func)
     func_init.SetGlobalFunc(mem, 'len', func_misc.Len())
+    func_init.SetGlobalFunc(mem, '_match', func_eggex.Match(mem))
+    func_init.SetGlobalFunc(mem, '_start', func_eggex.Start(mem))
+    func_init.SetGlobalFunc(mem, '_end', func_eggex.End(mem))
+    func_init.SetGlobalFunc(mem, 'join', func_misc.Join())
+    func_init.SetGlobalFunc(mem, 'maybe', func_misc.Maybe())
+
+    # Do we want to make these methods?
```
andychu, line 687: BTW I think we should make append and pop methods in a follow-on change

And we will should the append builtin to wrap the append method

It's a bit confusing/inconsistent now

library/func_misc.py
```diff
@@ -153,3 +153,53 @@ def Call(self, pos_args, named_args):
...
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        r = typed_args.Reader(pos_args, named_args)
+        li = r.PosList()
+
+        delim = ''
+        if len(pos_args): # reader has a reference
+            delim = r.PosStr()
```
andychu, line 172: OK actually this feels like we need to update our API for positional defaults

like `r.PosStr(default='')` or something

library/func_misc.py
```diff
@@ -153,3 +153,53 @@ def Call(self, pos_args, named_args):
...
+
+    def __init__(self):
+        # type: () -> None
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        r = typed_args.Reader(pos_args, named_args)
+        val = r.PosValue()
```
andychu, line 193: This should use val.PosStr() instead of PosValue(), no?

andychu (commented):
OK there is one minor thing, I think we should use PosStr() instead of PosValue() in maybe()

But I will just merge so the tests run

Also do you want to try turning append() and pop() back into methods?  I think the only reason they're free functions right now is because I wasn't sure if we would have methods in the language

We will also have an append builtin to make it easier to type, and for strings

```
:: mylist->append({key: 42})   # append a typed value

append :mylist foo bar *.py (42, 43)   # append strings and typed values

```

library/func_misc.py
melvinw: We use `PosValue()` here because we want to check for `value.Null` below

core/shell.py
melvinw: Ah sure. I can take a pass at this

PR #1713, opened 2023-08-30
[translation] Fix crash in oils-native from misuse of `error`

PossiblyAShrub:
This reduces the delta cpp by 2.

### DIFF
--- a/osh/builtin_meta.py
+++ b/osh/builtin_meta.py
@@ -386,7 +386,8 @@ class Error(vm._Builtin):
 
         status = 1
         for arg in cmd_val.typed_args.named_args:
-            if lexer.TokenVal(arg.name) == "status":
+            name = lexer.TokenVal(arg.name)
+            if name == "status":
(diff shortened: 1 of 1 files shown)


andychu:
OK cool!  still marked "draft" though

PossiblyAShrub:
Oops, sorry. This is ready for a review.

andychu:
Cool thanks 

On a related note, I just added `Token blame_tok, Str name` to params -- and we will do the same thing for `NamedArg`

In general I want to avoid calling `lexer.TokenVal()` at runtime.  The style should be to call it at parse time for only the things we need, and put it on the  AST in `frontend/syntax.asdl`

(There is that longstanding issue about reducing Token size, etc.  Not a top priority now)

---

Oh also we may be able to express this API with the new `typed_args.Reader`.   

There are also a couple more functions I tagged `IN-YSH` in `library/func_cpython` that can be moved

PR #1714, opened 2023-08-31
[ysh] Define any, all, sum, reversed in YSH

PossiblyAShrub:


PossiblyAShrub:
This should be ready for a review. I am not sure if we still want to keep `reversed()`, so I kept it in the PR for now.

I noticed that we used the `start` param of `sum()` in `spec/ysh-funcs-builtin`. We need to implement default function arguments before we can support `start`. I've let that test fail for now, but we can re-enable it when/if we add named params and want to support the `start` param.

stdlib/list.ysh
```diff
@@ -0,0 +1,55 @@
...
+      return (true)
+    }
+  }
+  return (false)
+}
+
+func all(list) {
+  ## Returns true if all values in the list are truthy.
+  ##
+  ## If the list is empty, return false.
```
andychu, line 17: This should be true!  I think it falls out of boolean algebra, and intuitively you can say "all elements in an empty list are truthy" -- it's vacously true

Funny thing is that I think this came up on Hacker News recently - https://news.ycombinator.com/item?id=37264149

Yeah I agree with the top answer, it's algebra -- 

- 0 is the identity element for addition -- (x + 0 == x) for all x
- 1 is the identity element for multiplication -- (x * 1 == x) for all x
- false is the identity element for AND  -- (x AND false == x) for all x
  - all() is a reduce() with AND
- true is the identity element for OR -- (x OR true == x) for all x
  - any() is a reduce() with OR

Although I guess it's backward.  alltrue([]) == true and anytrue([]) == false



```
>>> any([])
False
>>> all([])
True
>>> 
```

stdlib/list.ysh
```diff
@@ -0,0 +1,55 @@
...
+  ## Returns 0 for an empty list.
+
+  var sum = 0;
+  for item in (list) {
+    setvar sum += item
+  }
+  return (sum)
+}
+
+func reversed(list) {
```
andychu, line 43: Since we both agreed on Zulip, how about removing this, and changing the spec to have `mylist->reverse()` as a method?  for in-place reversing, which is more fundamental than this operation

stdlib/list.ysh
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func all(list) {
+  ## Returns true if all values in the list are truthy.
+  ##
+  ## If the list is empty, return false.
+
+  if (len(list) === 0) {
+    return (false)
+  }
```
andychu, line 21: So actually you can just remove this guard clause here

spec/ysh-stdlib.test.sh
```diff
@@ -97,3 +97,93 @@ echo status=$_status
...
+
+#### all
+source --builtin list.ysh
+
+json write (any([]))
+json write (any([true]))
+json write (any([false]))
+json write (any([true, true]))
+json write (any([true, false]))
+json write (any([false, true]))
```
andychu, line 131: could have some test cases with "foo" and "" and so forth

It will follow the `Bool()` builtin, which we don't have yet I think (or at least not in C++)

andychu (commented):
Thanks!

I guess this reminds me of the iterator discussion again ... in Python, any() and all() take iterables, so they have the short-circuit evaluation feature of the builtin `and` and `or`

e.g. if you wrote

```
f(x) and g(x) and h(x)
```

it would stop evaluating when false

But

```
all([f(x), g(x), h(x)])
```

does not do that.

Actually this relates to lazy evaluation -- should we have it on funcs?  It could be a use case later.

(We are going to have lazy evaluation for things like `assert (x == f(x))`, so that `assert` knows the expression passed to it, not just the value )

PossiblyAShrub:
Sorry, I am not sure why I added the special casing for `any`. Anyways, it's fixed now.

I removed `reversed()` and went ahead and implemented `List->reverse()`.

This should now be ready for another review.

library/func_misc.py
```diff
@@ -153,3 +153,21 @@ def Call(self, pos_args, named_args):
...
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        r = typed_args.Reader(pos_args, named_args)
+        list_ = r.PosList()
+        r.Done()
+
+        list_.reverse()
+
+        return value.List(list_)
```
andychu, line 173: Hm I actually think we should change this to value.Null to follow Python

```
>>> a=[1,2,3]
>>> print(a.reverse())
None
```

The reason is if people start using `a->reverse()` as a synonym for `reversed(a)`, they will get very confused by the resulting aliasing bugs (there's only one list, with two names, not two lists)


There is a spec test using that idiom

If we want that, we could add a reversed() wrapper that makes a copy, though I don't think it's 100% necessary now

i.e. I think Python's design of keeping mutation-in-place and transforming as separate APIs is a good one

andychu (commented):
Ah sorry I didn't look at this yesterday

library/func_misc.py
PossiblyAShrub: Gah, you're right. I see that I even made this error while updating a test. Returning `value.Null` is the way to go.

PossiblyAShrub:
Odd, the dummy test is failing to even start. That test aside, this should be ready for another review.

andychu:
Awesome thanks!

PR #1717, opened 2023-09-04
xxx constructors

melvinw:


andychu (commented):
Looks great so far!  Here are my initial comments

doc/ref/chap-builtin-func.md
```diff
@@ -32,8 +45,72 @@ This chapter in the [Oils Reference](index.html) describes builtin functions.
...
+Given a string, returns a list containing one element for each character in the source string.
+
+    $ = List('foo')
+    (List)   ['f', 'o', 'o']
+
+## Dict()
+
+Calling with no arguments returns a new, empty dictionary. Given a dictionary, returns a shallow copy of the original.
+
+Can also be called with a list of key-value pairs to construct a new dictionary seeded with the given mapping.
```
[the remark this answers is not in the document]
melvinw: Meh. I'm sure in some contexts this just "feels right", but it's arguably less useful without `zip()`, though. So, if we don't have any plans to add something like that, I'm fine dropping this-- especially if we will have dict comprehensions at some point

doc/ref/chap-builtin-func.md
```diff
@@ -32,8 +45,72 @@ This chapter in the [Oils Reference](index.html) describes builtin functions.
...
+
+    $ = Float('1.23')
+    (Float)   1.23
+
+    ysh$ = Float('bar')
+    # fails with an expression error
+
+## Str()
+
+Calling with no arguments returns a new, empty string. Given a string, returns a copy of the original.
```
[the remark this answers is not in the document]
melvinw: Ah, good to know re not copying!

melvinw:
Thanks for the comments! Updated those docs and tests. Implementation should match. Let me know if you think this needs anything else

doc/ref/chap-builtin-func.md
```diff
@@ -20,6 +20,19 @@ This chapter in the [Oils Reference](index.html) describes builtin functions.
...
 ### type()
 
+Given an arbitrary value, returns a string representing the value's runtime type.
+
+For example:
+
+    var d = {'foo': 'bar'}
+    var n = 1337
+
+    $ echo $[type(d)]
```
andychu, line 30: can you change these to the `=` style too ?   Also I guess it should be `$ var` on the first 2 lines to be consistent

doc/ref/chap-builtin-func.md
```diff
@@ -32,8 +45,66 @@ This chapter in the [Oils Reference](index.html) describes builtin functions.
...
+## Str()
+
+Returns its argument if it's a string.
+
+## List()
+
+Given a list, returns a shallow copy of the original.
+
+Given an interable value (e.g. a range or dictionary), returns a list containing one element for each item in the
+original colleciton.
```
andychu, line 97: typos - iterable, collection

library/func_misc.py
```diff
@@ -221,3 +222,200 @@ def Call(self, pos_args, named_args):
...
+        pass
+
+    def Call(self, pos_args, named_args):
+        # type: (List[value_t], Dict[str, value_t]) -> value_t
+
+        r = typed_args.Reader(pos_args, named_args)
+        val = r.PosValue()
+        r.Done()
+
+        tname = ValType(val)
```
andychu, line 240: We are a little inconsistent about this, but generally the default should be `import ui` and then `ui.ValType()`

library/func_misc.py
```diff
@@ -221,3 +222,200 @@ def Call(self, pos_args, named_args):
...
+
+        r = typed_args.Reader(pos_args, named_args)
+        val = r.PosValue()
+        r.Done()
+
+        UP_val = val
+        with tagswitch(val) as case:
+            if case(value_e.Int):
+                val = cast(value.Int, UP_val)
+                return value.Int(val.i)
```
andychu, line 277: can just return `val` here?

library/func_misc.py
```diff
@@ -221,3 +222,200 @@ def Call(self, pos_args, named_args):
...
+
+            elif case(value_e.Bool):
+                val = cast(value.Bool, UP_val)
+                return value.Int(int(val.b))
+
+            elif case(value_e.Float):
+                val = cast(value.Float, UP_val)
+                return value.Int(int(val.f))
```
andychu, line 285: OK it appears this does floor() in Python, wasn't quite sure about that

doc/ref/chap-builtin-func.md
```diff
@@ -32,8 +45,66 @@ This chapter in the [Oils Reference](index.html) describes builtin functions.
...
+## Bool()
+
+Returns the truth value of its argument. Similar to `bool()` in python, it returns `false` for `false`, `0`, `0.0`,
+`''`, `{}`, `[]`, and `null`.  Returns `true` for all other values.
+
+## Int()
+
+Given a float, returns the largest integer that is less than the given value.
+
+    $ = Int(1.23)
```
andychu, line 57: Just to be extra explicit,   maybe change this to 1.99 and mention floor(), since I think we also want the round() function, which has the "other" behavior

library/func_misc.py
```diff
@@ -221,3 +222,200 @@ def Call(self, pos_args, named_args):
...
+
+        UP_val = val
+        with tagswitch(val) as case:
+            if case(value_e.Int):
+                val = cast(value.Int, UP_val)
+                return value.Float(float(val.i))
+
+            elif case(value_e.Float):
+                val = cast(value.Float, UP_val)
+                return value.Float(val.f)
```
andychu, line 316: I would just return val again, it will save some GC objects :)

core/shell.py
```diff
@@ -684,6 +684,13 @@ def Main(lang, arg_r, environ, login_shell, loader, readline):
     func_init.SetGlobalFunc(mem, '_end', func_eggex.End(mem))
     func_init.SetGlobalFunc(mem, 'join', func_misc.Join())
     func_init.SetGlobalFunc(mem, 'maybe', func_misc.Maybe())
+    func_init.SetGlobalFunc(mem, 'type', func_misc._Type())
```
andychu, line 687: Generally speaking I use a leading `_` to mean "private"

And a trailing `_` to avoid name conflicts, for example `word_.py`, or `default_` to avoid the  C++ keyword



So I think we could have `List_` and `Dict_` ?   Not sure if we need to avoid name conflicts on the others

andychu (commented):
Looks great, thanks!

core/shell.py
melvinw: Ah, yeah. There were conflicts on the `value.*` names when compiling C++, so I just treated all of these to the same renaming. `Type` almost certainly didn't need it. I'm surprised namespacing didn't help prevent the others... I'll look a little deeper into that and switch to trailing `_` if there isn't some other way around this

doc/ref/chap-builtin-func.md
melvinw: Ah whoops. Thought I got all of these

melvinw:
OK all set!

andychu:
Great thanks!

PR #1718, opened 2023-09-05
[docs] Document all stdlib/{math,list}.ysh functions

PossiblyAShrub:
I don't like that I have to write "Note, you will need to `source --builtin <some-file>.ysh` to use this function" everywhere. Maybe we can add a link to the right of the function name? We might be able to fit a link to the source location there too.

But, this should be fine on it's own. I can always do the above changes in another PR.

EDIT: Here's the rendered docs for reference http://travis-ci.oilshell.org/github-jobs/4914/ovm-tarball.wwz/_release/VERSION/doc/ref/chap-builtin-func.html#abs.

### DIFF
--- a/devtools/release.sh
+++ b/devtools/release.sh
@@ -458,7 +458,7 @@ compress() {
   popd
 
   log "--- src-tree"
-  local out="$root/srec-tree.wwz"
+  local out="$root/src-tree.wwz"
   pushd _tmp/src-tree
   time zip -r -q $out .  # recursive, quiet
   popd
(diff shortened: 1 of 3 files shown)


andychu:
I wonder if at build time we should just concatenate all the files into one, like

```
source --builtin stdlib.ysh
```

It seems a little silly to make people remember what file each function is in.

---

And I think there might be some other things that are not stdlib that we might want to `source --builtin`, like some kind of bootstrap for a virtualenv or pip like thing.

----

I also think it would be nice to get rid of the `source` altogether ... but I think we can circle back to that.  That was one the "one way door" vs "two way door" thing -- we can remove the requirement without breaking any existing code

PR #1719, opened 2023-09-06
[core/vm] simplify _Callable.Call() signature

melvinw:
This will help us produce better errors in the future.

Figured I should send this before we add too many more functions

### DIFF
--- a/core/code.py
+++ b/core/code.py
@@ -14,7 +14,7 @@ from core import vm
 from frontend import typed_args
 from mycpp.mylib import log
 
-from typing import List, Dict, cast, TYPE_CHECKING
+from typing import List, cast, TYPE_CHECKING
 if TYPE_CHECKING:
     from _devbuild.gen.syntax_asdl import command, loc_t
     from _devbuild.gen.runtime_asdl import Proc
--- a/core/code.py
+++ b/core/code.py
@@ -11,6 +11,7 @@ from core import error
 from core.error import e_die
 from core import state
 from core import vm
+from frontend import typed_args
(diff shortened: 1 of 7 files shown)


core/code.py
```diff
@@ -33,8 +34,9 @@ def __init__(self, name, node, mem, cmd_ev):
         self.cmd_ev = cmd_ev
         self.mem = mem
 
-    def Call(self, pos_args, named_args):
-        # type: (List[value_t], Dict[str, value_t]) -> value_t
+    def Call(self, args):
+        # type: (typed_args.Reader) -> value_t
+        pos_args = args.RestPos()
```
andychu, line 39: Hm now I realized typed_args.Reader() was mainly for builtin functions, to bridge mycpp static types with YSH dynamic types

In this case they're both dynamic.  But yeah when we do the full param/arg binding for funcs we can look at this, I think it will work fine

andychu (commented):
CI is complaining about a lint error, but otherwise looks good, thanks!

melvinw:
Ah whoops. Fixed that litner error

melvinw:
Oh, hmm. Busted a unit test. Will look into taht

melvinw:
Interesting. It's `core/completion_test.py` that's failing. Haven't managed to reproduce locally yet. Is this a known flake? http://travis-ci.oilshell.org/github-jobs/4924/dev-minimal.wwz/_tmp/soil/logs/unit.txt

melvinw:
Ah looks like my branch was just out of date

andychu:
Cool thanks!

PR #1720, opened 2023-09-06
[translation] statically type extend(), glob(), and split()

melvinw:
Also, make some list functions into methods


andychu:
Oh BTW let's change extend() to a method, there is an AddMethods() function in core/shell.py

Same with append() and pop()

For awhile I thought we might not have methods at all, so that's why we have those functions

melvinw:
Pretty nice spec test bump with this one. Down to 74! http://travis-ci.oilshell.org/github-jobs/4935/cpp-spec.wwz/_tmp/spec/ysh-cpp/compare.html

doc/hay.md
```diff
@@ -612,7 +612,7 @@ This can be done with an evaluator that simply enumerates all files:
     for path in myconfig.d/*.hay {
       const code = parse_hay(path)
       const result = eval(hay)
-      _ append(results, result)
+      _ results->append(result)
```
andychu, line 615: Thanks for fixing these!

library/func_misc.py
```diff
@@ -401,3 +421,51 @@ def Call(self, args):
...
+        ifs = None  # type: str
+        if args.NumPos():
+            ifs = args.PosStr()
+
+        args.Done()
+
+        # mycpp: rewrite list comprehension over sum type
+        l = []  # type: List[value_t]
+        for elem in self.splitter.SplitForWordEval(s, ifs=ifs):
+            l.append(value.Str(elem))
```
andychu, line 446: FYI not all  list comprehensions work in mycpp, but basic ones do

I think this can be

```
strs = self.splitter...
items = [value.Str(s) for s in strs]
return value.List(items)
```

The limitation is that it has to look lke

```
x = [...]
```

not

```
x = value.List([ ... ])  # needs a temp variable
```

library/func_misc.py
```diff
@@ -401,3 +421,51 @@ def Call(self, args):
...
+
+        return value.List(l)
+
+
+class Glob(vm._Callable):
+
+    def __init__(self, globber):
+        # type: (glob_.Globber) -> None
+        vm._Callable.__init__(self)
+        self.globber = globber
```
andychu, line 456: Arguably we can put a Call() method directly on splitter and globber, but it doesn't matter much either way

library/func_misc.py
```diff
@@ -401,3 +421,51 @@ def Call(self, args):
...
+
+    def Call(self, args):
+        # type: (typed_args.Reader) -> value_t
+        s = args.PosStr()
+        args.Done()
+
+        out = []  # type: List[str]
+        self.globber._Glob(s, out)
+
+        # mycpp: rewrite list comprehension over sum type
```
andychu, line 466: oh yeah I think we can do a decent job as is, as long as there's a tmp var

andychu (commented):
Looks great, thanks!

library/func_misc.py
melvinw: Ah, yeah I had done this before. The issue was that the type checker was upset about trying to pass the temp var into the `value.List()` constructor because the tempvar was technically `List[value.Str]`. Adding a `type: List[value_t]` annotation to the tempvar fixed that, though!

library/func_misc.py
melvinw: Ah yeah, I thought about doing that, but it felt cleaner for the builtin code to be closer to the other builtins

melvinw:
OK addressed those comments. I also tacked on a few commits to translate `shvar_get` and removed `library/func_cpython.py` while I'm in here. Have a look

library/func_misc.py
```diff
@@ -37,6 +42,23 @@ def Call(self, args):
...
+
+    def Call(self, args):
+        # type: (typed_args.Reader) -> value_t
+
+        a = args.PosList()
+        b = args.PosList()
+        args.Done()
+
+        a.extend(b)
+        return value.List(a)
```
andychu, line 59: Should return `value.Null` here . 

```
>>> a=[]
>>> print(a.extend([1,2,3]))
None
```

(similar discussion with mylist.reverse(), Aidan agreed)

spec/ysh-builtins.test.sh
```diff
@@ -629,3 +629,13 @@ echo $[l2[1]] # NOTE: depends on dict being ordered
 3
 foo
 ## END
+
+#### List->extend()
+var l = List(1:3)
+echo $[len(l)]
+_ l->extend(List(3:6))
+echo $[len(l)]
```
andychu, line 637: maybe add a test for null return here, and for append(), e.g.

```
var x = l->extend(...)
if (x === null) { echo pass }
```

reminds me that we need the `assert` builtin

andychu (commented):
Looks awesome, thanks!

library/func_misc.py
melvinw: Ah, right! Nice catch

spec/ysh-builtins.test.sh
```diff
@@ -629,3 +629,31 @@ echo $[l2[1]] # NOTE: depends on dict being ordered
...
+shopt -s oil:all
+var l = List(1:3)
+
+var result = l->extend(List(3:6))
+if (result === null) { echo 'pass' }
+
+setvar result = l->append(6)
+if (result === null) { echo pass }
+
+setvar result = l->pop()
```
andychu, line 653: Actually pop() is the only one that returns a value!   

```
>>> ['a', 'b'].pop()
'b'
```

I guess the original code made that mistake too ..?

spec/ysh-builtins.test.sh
melvinw: Added `assert()` while I was in here. What do you think?

spec/ysh-builtins.test.sh
melvinw: Done

andychu:
OK thanks, I changed `assert` -> `assert_` since I think we're not quite ready to expose it

But we can use it in tests

Some threads on Zulip about the design -- proc vs. func, lazy eval, etc.

andychu:
Thanks, merged!

PR #1723, opened 2023-09-07
[frontend/typed_args] add positional arg error locations

melvinw:
Also add PosBool() and NamedBool().

An example:

**Before**
```shell
ysh$ = join(['foo', {}], ' ')
  = join(['foo', {}], ' ')
  ^
[ interactive ]:2: fatal: expected Null, Bool, Int, Float, Eggex, got value.Dict

ysh$ = join(['foo', 'bar'], 2)
  = join(['foo', 'bar'], 2)
  ^
[ interactive ]:1: fatal: Arg 2 should be a Str, got value.Int
```

**After**
```shell
ysh$ = join(['foo', {}], ' ')
  = join(['foo', {}], ' ')
         ^
[ interactive ]:2: fatal: expected Null, Bool, Int, Float, Eggex, got value.Dict

ysh$ = join(['foo', 'bar'], 2)
  = join(['foo', 'bar'], 2)
                         ^
[ interactive ]:1: fatal: Arg 2 should be a Str, got value.Int
```


library/func_misc.py
```diff
@@ -206,7 +207,9 @@ def Call(self, args):
 
         strs = []  # type: List[str]
         for i, el in enumerate(li):
-            strs.append(val_ops.Stringify(el, loc.Missing))
+            strs.append(
+                val_ops.Stringify(
+                    el, location.TokenForListElement(args.PosNode(0), i)))
```
melvinw, line 212: This is a good example of non-trivial use (from `join()`)

melvinw: Need to make this, and the other `val_ops` calls throughout less eager...

melvinw: Actually, just this one. The others are just calling `Reader::BlamePos()`, which is cheap

ysh/val_ops.py
```diff
@@ -79,51 +79,59 @@ def ToCommand(val, msg, blame_loc):
...
+    """
+    Used by
+
+    $[x]    stringify operator
+    @[x]    expression splice - each element is stringified
+    @x      splice value
+    """
+    s = _Stringify(val)
+    if s is None:
+        raise StringifyErr(val, blame_loc, prefix)
```
andychu, line 134: Hm why split this function up?

frontend/typed_args.py
```diff
@@ -74,52 +75,120 @@ def RestWords(self):
 
     ### Typed positional args
 
+    def BlamePos(self, pos=-1):
```
andychu, line 78: I would call this BlameLoc() since it returns loc_t, I think we're relatively consistent about that

library/func_misc.py
```diff
@@ -487,7 +494,7 @@ def Call(self, args):
         name = args.PosStr()
         args.Done()
         return expr_eval.LookupVar(self.mem, name, scope_e.Dynamic,
-                                   loc.Missing)
+                                   args.BlamePos())
```
andychu, line 497: For cases where BlamePos() is called on the non-error path, I wonder if we can just have a variant:

```
loc =
    ...
  | ArgList %ArgList
```

and then pass `args`, so it's essentially free

It may make the error message worse, but I feel like we could start with that, and then make the errors more precise when we encounter them

It seems like 90% of cases are only computing positions in error cases, which is good

library/func_misc.py
```diff
@@ -228,7 +234,7 @@ def Call(self, args):
 
         s = val_ops.ToStr(
             val, 'maybe() expected Str, but got %s' % value_str(val.tag()),
-            loc.Missing)
+            args.BlamePos())
```
andychu, line 237: I also wonder if we can have a style like

```
s = val_ops.TryStr(val)
if s is None:
   raise ...
```

Although that doesn't work for bool, int, float, which can't be null

andychu (commented):
This looks pretty reasonable, a few comments

andychu:
Thinking about this more, the issue is that the whole typed_args.Reader has the state of which argument is being processed

So it can give the most precise location ... I wish that could be an ASDL type and then it could be a `loc_t`

Although not sure that solves anything, hm

frontend/typed_args.py
melvinw: I originally wanted one `BlameLoc()`, but it felt awkward to have two kwargs for the positional and named arg cases. So, there are separate functions: `BlamePos()` and `BlameNamed()`

melvinw:
Yeah, we could add a type for this that captures the state tracked by `Reader`, but we would still have to pay to extra GC objs to pass it around :/

andychu:
Hm I was testing this out and got a crash


```
andy@hoover:~/git/oilshell/oil$ bin/osh -c '= "foo"->startswith(42)'
  File "/home/andy/git/oilshell/oil/bin/oils_for_unix.py", line 208, in <module>
    sys.exit(main(sys.argv))
  File "/home/andy/git/oilshell/oil/bin/oils_for_unix.py", line 183, in main
    return AppBundleMain(argv)
  File "/home/andy/git/oilshell/oil/bin/oils_for_unix.py", line 141, in AppBundleMain
    return shell.Main('osh', arg_r, environ, login_shell, loader, readline)
  File "/home/andy/git/oilshell/oil/core/shell.py", line 1002, in Main
    cmd_flags=cmd_eval.IsMainProgram)
  File "/home/andy/git/oilshell/oil/core/main_loop.py", line 353, in Batch
    is_return, is_fatal = cmd_ev.ExecuteAndCatch(node, cmd_flags=cmd_flags)
  File "/home/andy/git/oilshell/oil/osh/cmd_eval.py", line 1858, in ExecuteAndCatch
    status = self._Execute(node)
  File "/home/andy/git/oilshell/oil/osh/cmd_eval.py", line 1680, in _Execute
    status = self._Dispatch(node, cmd_st)
  File "/home/andy/git/oilshell/oil/osh/cmd_eval.py", line 1083, in _Dispatch
    val = self.expr_ev.EvalExpr(node.e, loc.Missing)
  File "/home/andy/git/oilshell/oil/ysh/expr_eval.py", line 303, in EvalExpr
    val = self._EvalExpr(node)
  File "/home/andy/git/oilshell/oil/ysh/expr_eval.py", line 1083, in _EvalExpr
    return self._EvalFuncCall(node)
  File "/home/andy/git/oilshell/oil/ysh/expr_eval.py", line 773, in _EvalFuncCall
    ret = f.Call(typed_args.Reader(pos_args, named_args, node.args))
  File "/home/andy/git/oilshell/oil/library/func_misc.py", line 87, in Call
    match = args.PosStr()
  File "/home/andy/git/oilshell/oil/frontend/typed_args.py", line 125, in PosStr
    self._BlamePos())
  File "/home/andy/git/oilshell/oil/frontend/typed_args.py", line 90, in _BlamePos
    l = location.TokenForExpr(self.args_node.pos_args[pos])
IndexError: list index out of range
```

andychu:
This one points to the second arg:

```
$ bin/osh -c '= join(["Y"], "x", 43)'
  = join(["Y"], "x", 43)
                ^
[ -c flag ]:1: fatal: Expected 2 arguments, but got 3
```

I think it could either point to:

- the `(`
- the third arg
- the comma after the second arg

What do you think?  

Adding some examples  in `test/ysh-runtime-errors.sh` will help

library/func_misc.py
```diff
@@ -228,7 +228,7 @@ def Call(self, args):
 
         s = val_ops.ToStr(
             val, 'maybe() expected Str, but got %s' % value_str(val.tag()),
-            loc.Missing)
+            args.BlameLoc())
```
andychu, line 231: Does it make sense to pass the `(` token directly here?  I am not sure if this allocates a new location

library/func_misc.py
```diff
@@ -206,7 +206,7 @@ def Call(self, args):
 
         strs = []  # type: List[str]
         for i, el in enumerate(li):
-            strs.append(val_ops.Stringify(el, loc.Missing))
+            strs.append(val_ops.Stringify(el, args.BlameLoc()))
```
andychu, line 209: ditto here, does it make sense to pass  `(` directly ?

Like maybe `args.ParenToken()`

Then it's clearer that we're not being more precise than that, which is fine

library/func_misc.py
melvinw: Ah. `BlameLoc()` just returns the left paren token and shouldn't allocate. I'll rename it for clarity, like you mentioned in the other comment, though

melvinw:
> This one points to the second arg:
> 
> ```
> $ bin/osh -c '= join(["Y"], "x", 43)'
>   = join(["Y"], "x", 43)
>                 ^
> [ -c flag ]:1: fatal: Expected 2 arguments, but got 3
> ```
> 
> I think it could either point to:
> 
>     * the `(`
> 
>     * the third arg
> 
>     * the comma after the second arg
> 
> 
> What do you think?
> 
> Adding some examples in `test/ysh-runtime-errors.sh` will help

Nice catch! Since this error is structured, I have tests for related things in the unit tests. I will patch and include a test for this

melvinw:
> Hm I was testing this out and got a crash
> 
> ```
> andy@hoover:~/git/oilshell/oil$ bin/osh -c '= "foo"->startswith(42)'
> Traceback (most recent call last):
>   File "/home/andy/git/oilshell/oil/bin/oils_for_unix.py", line 208, in <module>
>     sys.exit(main(sys.argv))
>   File "/home/andy/git/oilshell/oil/bin/oils_for_unix.py", line 183, in main
>     return AppBundleMain(argv)
>   File "/home/andy/git/oilshell/oil/bin/oils_for_unix.py", line 141, in AppBundleMain
>     return shell.Main('osh', arg_r, environ, login_shell, loader, readline)
>   File "/home/andy/git/oilshell/oil/core/shell.py", line 1002, in Main
>     cmd_flags=cmd_eval.IsMainProgram)
>   File "/home/andy/git/oilshell/oil/core/main_loop.py", line 353, in Batch
>     is_return, is_fatal = cmd_ev.ExecuteAndCatch(node, cmd_flags=cmd_flags)
>   File "/home/andy/git/oilshell/oil/osh/cmd_eval.py", line 1858, in ExecuteAndCatch
>     status = self._Execute(node)
>   File "/home/andy/git/oilshell/oil/osh/cmd_eval.py", line 1680, in _Execute
>     status = self._Dispatch(node, cmd_st)
>   File "/home/andy/git/oilshell/oil/osh/cmd_eval.py", line 1083, in _Dispatch
>     val = self.expr_ev.EvalExpr(node.e, loc.Missing)
>   File "/home/andy/git/oilshell/oil/ysh/expr_eval.py", line 303, in EvalExpr
>     val = self._EvalExpr(node)
>   File "/home/andy/git/oilshell/oil/ysh/expr_eval.py", line 1083, in _EvalExpr
>     return self._EvalFuncCall(node)
>   File "/home/andy/git/oilshell/oil/ysh/expr_eval.py", line 773, in _EvalFuncCall
>     ret = f.Call(typed_args.Reader(pos_args, named_args, node.args))
>   File "/home/andy/git/oilshell/oil/library/func_misc.py", line 87, in Call
>     match = args.PosStr()
>   File "/home/andy/git/oilshell/oil/frontend/typed_args.py", line 125, in PosStr
>     self._BlamePos())
>   File "/home/andy/git/oilshell/oil/frontend/typed_args.py", line 90, in _BlamePos
>     l = location.TokenForExpr(self.args_node.pos_args[pos])
> IndexError: list index out of range
> ```

Nice catch on this one too. The method code paths were completely untested! Doh! I'll add some spec tests for this

frontend/syntax.asdl
```diff
@@ -158,7 +158,7 @@ module syntax
   # Unevaluated, typed arguments for func and proc.
   # Note that ...arg is expr.Spread.
   ArgList = (
-     Token left, List[expr] pos_args, List[NamedArg] named_args, Token right
+     Token left, List[expr] pos_args, Token? named_delim, List[NamedArg] named_args, Token right
```
andychu, line 161: Add a comment about what named_delim points to?

frontend/typed_args.py
```diff
@@ -213,13 +344,18 @@ def Done(self):
...
-                                     len(self.pos_args)), loc.Missing)
+            self.pos_consumed += 1
+            raise error.TypeErrVerbose(
+                'Expected %d arguments, but got %d' %
+                (self.pos_consumed, self.pos_consumed + len(self.pos_args)),
+                self._BlamePos())
 
         if len(self.named_args):
+            assert self.args_node.named_delim is not None
             bad_args = ','.join(self.named_args.keys())
```
andychu, line 355: I would add a space like `', '.join()`

frontend/typed_args_test.py
```diff
@@ -28,16 +40,28 @@ def testReaderPosArgs(self):
...
+            _ = reader.PosStr()
+            reader.Done()
+
+        # Check that the error points to the first unconsummed argument
+        e = cm.exception
+        self.assertEqual(pos_exprs[2].c, e.location)
 
         # Arg is wrong type...
-        self.assertRaises(error.TypeErrVerbose, reader.PosStr)
+        with self.assertRaises(error.TypeErrVerbose) as cm:
```
andychu, line 56: hm didn't know you could do this, seems nicer than the old style

test/ysh-runtime-errors.sh
```diff
@@ -421,6 +421,10 @@ test-dict-construct() {
   _expr-error-case '= Dict([["too", "many", "parts"]])'
 }
 
+test-bound-func() {
+  _expr-error-case '= "foo"->startswith(1)' # shouldn't panic
```
andychu, line 425: can we add more of the errors in this file?  It's useful to be able to eyeball them all at one

Like `test-precise-error-locations() {`

andychu (commented):
OK thanks, I would like to see a shell function that tickles all the errors so we can eyeball them

I've found that to be very useful

We publish them, so we can see if they regress

melvinw:
Alrighty. Merged `master` and addressed this last batch of comments. Lemme know what you think

frontend/typed_args_test.py
melvinw: yeah, it's pretty handy

melvinw:
Yikes. I've made CI very angry... will check it out

melvinw:
Ok all patched up

andychu:
Merged -- thanks for doing all the iterations on this!  End result looks great!

PR #1726, opened 2023-09-13
[builtin/error] Update the error builtin to use typed_args.Reader

PossiblyAShrub:
This will cause error messages+codes to change for the error builtin, but now they are closer to what we have for the other builtin functions.

In order to make this work, we have to evaluate all arguments before constructing the `typed_args.Reader`. So previously something like `error (not_an_arg=42/0)` would fail on the `not_an_arg`, but now it fails on `42/0`. I think this may actually be a good thing as it better matches other languages such as python:

```py
>>> def myfunc(named_arg=2):
...   return named_arg
... 
>>> myfunc(different_arg=42/0)
ZeroDivisionError: division by zero
>>> myfunc(different_arg=42)
TypeError: myfunc() got an unexpected keyword argument 'different_arg'
```

### DIFF
--- a/frontend/typed_args.py
+++ b/frontend/typed_args.py
@@ -69,12 +69,11 @@ class Reader(object):
 
     def Word(self):
         # type: () -> str
-
-        return None
+        return None  # TODO
 
     def RestWords(self):
         # type: () -> List[str]
-        return None
+        return None  # TODO
 
     ### Typed positional args
 
--- a/frontend/typed_args.py
+++ b/frontend/typed_args.py
@@ -217,13 +217,15 @@ class Reader(object):
         """
         # Note: Python throws TypeError on mismatch
         if len(self.pos_args):
-            raise error.TypeErrVerbose('Expected %d arguments, but got %d' %
-                                    (self.pos_consumed, self.pos_consumed +
-                                     len(self.pos_args)), loc.Missing)
+            raise error.TypeErrVerbose(
+                'Expected %d arguments, but got %d' %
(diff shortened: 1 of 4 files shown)


osh/builtin_meta.py
```diff
@@ -366,39 +365,21 @@ def __init__(self, expr_ev):
...
+        message = t.PosStr()
+        status = t.NamedInt("status", 1)
+        t.Done()
+
+        if status == 0:
+            e_die("Status must be a non-zero integer", cmd_val.arg_locs[0])
+
+        if len(cmd_val.argv) > 1:
+            raise error.TypeErrVerbose(
+                'Expected 0 untyped arguments, but got %d' %
```
andychu, line 379: I think eventually we will have a variant of `Reader` for  procs that also checks this, but this is good for now

frontend/typed_args.py
```diff
@@ -213,13 +216,43 @@ def Done(self):
...
+    """
+    if typed_args is None:
+        raise error.TypeErrVerbose(
+            'Expected at least 1 typed argument, but got 0', loc.Missing)
+
+    pos_args = []  # type: List[value_t]
+    named_args = {}  # type: Dict[str, value_t]
+
+    for i, pos_arg in enumerate(typed_args.pos_args):
+        result = expr_ev.EvalExpr(pos_arg, loc.Missing)
```
andychu, line 247: Melvin has a PR open that makes it easier to add some of these locations

andychu (commented):
Looks good, thanks!

PR #1729, opened 2023-09-16
[mycpp] add hash table to Dict

melvinw:
This commit adds a proper hash table to Dict. The hash table uses open
addressing with linear probing (stride=1) and keys are hashed using
FNV1. Similar to CPython 3.X, keys can be recalled in the order they
were inserted; however deletions can disrupt this ordering.

GlobalDict still uses the linear scanning implementation for now.

This reduces the instruction count of several benchmarks by up to 50%!


melvinw:
note to self: tests that depend on dict iteration order may be fundamentally broken by this

andychu:
Wow looks like we're beating bash on a couple compute-bound benchmarks, and getting very close on the others  !!

http://travis-ci.oilshell.org/github-jobs/5071/benchmarks.wwz/_tmp/compute/index.html

mycpp/gc_dict.h
```diff
@@ -155,9 +144,7 @@ class Dict {
   int capacity_;  // number of entries before resizing
 
   // These 3 slabs are resized at the same time.
-  Slab<int>* entry_;  // NOW: kEmptyEntry, kDeletedEntry, or 0.
-                      // LATER: indices which are themselves indexed by // hash
-                      // value % capacity_
+  Slab<int>* entry_;  // kEmptyEntry, or a valid index into keys_/values_
```
andychu, line 147: Since we have a method name `find_key_in_index()`, should this be called `index_` ?  I think that's a better name anyway, not sure why I called it `entry_`

mycpp/gc_str.h
```diff
@@ -74,6 +75,10 @@ class Str {
     return ObjHeader::Str();
   }
 
+  int hash(HashFunc h);
+
+  static constexpr int kHashFlagMask = 0x08000000;
```
andychu, line 80: why this value?  Needs a comment

mycpp/gc_str.cc
```diff
@@ -525,6 +525,13 @@ List<Str*>* Str::split(Str* sep) {
   return this->split(sep, len(this));
 }
 
+int Str::hash(HashFunc h) {
+  if ((hash_value_ & Str::kHashFlagMask) == 0) {
+    hash_value_ = h(data_, len_) | Str::kHashFlagMask;
```
andychu, line 530: How about some unit tests for this function?

Maybe it's the same thing but it seems more logical / conventional to do `hash & 0x7fff_ffff` to mask off one bit

Also I would just inline the constant if it's only used in one place ... it makes it harder to read the formula if it's scattered

mycpp/gc_dict.h
```diff
@@ -307,49 +275,119 @@ void Dict<K, V>::clear() {
...
 template <typename K, typename V>
 void Dict<K, V>::set(K key, V val) {
-  int pos = position_of_key(key);
-  if (pos == -1) {  // new pair
-    reserve(len_ + 1);
+  DCHECK(obj_header().heap_tag != HeapTag::Global);
+  int pos = find_key_in_index(key);
+  if (pos == -1) {
+    // No room. Resize and see if we can find a slot.
+    for (int attempt = 0; attempt < 3; ++attempt) {
```
andychu, line 373: why 3 attempts?

mycpp/gc_dict.h
```diff
@@ -307,49 +275,119 @@ void Dict<K, V>::clear() {
...
+
+  // Hash the key onto a slot in the index. If the first slot is occupied, probe
+  // until an empty one is found.
+  unsigned h = hash_key(key);
+  int init_bucket = h % capacity_;
+
+  // If there's a vacancy because of deletion on the probe path for this key, we
+  // should fill it before consuming a slot that has never been used. This
+  // should help keep the index somewhat compact.
+  int tombstone = -1;
```
andychu, line 298: is this -1 different than kDeletedEntry?

mycpp/gc_dict.h
```diff
@@ -110,42 +98,43 @@ class Dict {
...
-  //   - Should we have wrappers like:
-  //   - V GetAndIntern<V>(D, &string_key)
-  //   - SetAndIntern<V>(D, &string_key, value)
-  //   This will enable duplicate copies of the string to be garbage collected
-  int position_of_key(K key);
+  // Returns -1 if the dictionary is full. The caller can use this as a cue to
+  // grow the table.
+  //
+  // Used by dict_contains(), index(), get(), and set().
+  int find_key_in_index(K key) const;
```
andychu, line 132: How about names like

- bucket_of(K key) -- well I guess it's not a bucket, since we're doing linear probing

- hash_and_probe(K key)
- probe(K key)

...

- index_of(K key) -- the `index_` member holds the index into the keys and values
- kv_index(K key)

mycpp/gc_dict.h
```diff
@@ -307,49 +275,119 @@ void Dict<K, V>::clear() {
...
+
+  // GlobalDict. Just scan.
+  for (int i = 0; i < len_; ++i) {
     if (keys_equal(keys_->items_[i], key)) {
       return i;
     }
   }
-  return -1;  // table is completely full?  Does this happen?
+
+  // Not found.
```
andychu, line 363: do we need a kNotFound ?

maybe it can be distinct from kDeletedEntry and kEmptyEntry, so it's easier to see the difference in a debugger

It could be -3 I think

mycpp/gc_dict.h
```diff
@@ -307,49 +275,119 @@ void Dict<K, V>::clear() {
...
 void Dict<K, V>::set(K key, V val) {
-  int pos = position_of_key(key);
-  if (pos == -1) {  // new pair
-    reserve(len_ + 1);
+  DCHECK(obj_header().heap_tag != HeapTag::Global);
+  int pos = find_key_in_index(key);
+  if (pos == -1) {
+    // No room. Resize and see if we can find a slot.
+    for (int attempt = 0; attempt < 3; ++attempt) {
+      reserve((capacity_ ?: 1) * 2);
```
andychu, line 374: comment on this formula?  It's doing something different the first time?

andychu (commented):
This seems reasonable, but there seem like a lot of if statements that can use unit test coverage

andychu:
Oh also do we have tests for the ordering properties?  That seems important

mycpp/gc_dict.h
melvinw: Ah yeah. I wanted to do this rename, but it collides with the `index_()` function :( I guess we could have mycpp generate a different function call for `[]` expressions. Not sure if worth... Maybe `indices_`?

mycpp/gc_str.h
melvinw: oh, it's just the most significant bit. gonna inline this, though

mycpp/gc_str.cc
melvinw: In this case we actually want to _set_ the bit. So, it's correct as-is. I agree that inlining will be less confusing, though. Will do.

The line immediately below this just shifts the hash to the right 1 bit to lop the flag bit off. Short and sweet.

melvinw:
> Oh also do we have tests for the ordering properties? That seems important

Oh yes! One (or several?) mycpp examples test this property, but I think it would be good to have a unit test, too. Will add

mycpp/gc_dict.h
melvinw: Ah sure. These ternary expressions are so terse, it may just be better to explode this into an `if`. Basically, `Dict` is initialized with `capacity_ = 0`. In that case we just want to pretend like it was size 1 so we don't resize to zero...

mycpp/gc_str.cc
melvinw: Will add a test 👍

mycpp/gc_dict.h
melvinw: Another option could be to initialize dicts to some minimum size. Then this just becomes `reserve(capacity_ * 2);`. Is there a reason we init to zero right now?

mycpp/gc_dict.h
melvinw: how about `hash_and_probe(K)` and `lookup_kv(K)`?

andychu: - I think `hash_and_probe()` or just `probe()` are good
  - or maybe `probe_for_bucket()`
- `lookup_kv()` is OK, I think I would prefer naming after something it returns, like `find_kv_index()`

(Although also our names are a bit inconsistent, maybe it should be ProbeForBucket() for non-user facing methods ... it has to be `.get()` for Python compat, but these don't have that constraint)

If we want the most descriptive names (which we don't always, but here it seems worth distinguishing), I think Verb-Noun is good

- probe is the verb, bucket is the noun
- find is the verb, kv_index is the noun

mycpp/gc_dict.h
andychu: Oh this is weird, I called that the `a[0]` function `a->index_(0)` because Python already has a different `a.index()` function

I think it's fine to leave it as is now, but I might rename things later

It could be `a.at(0)`, leaving room for the `a->index_` member

Tuples already have `t.at0()`, so this would make a lot of sense

mycpp/gc_dict.h
andychu: It feels like it should be logically impossible to grow it ONCE and not find space?

I guess we need to set out the contract for the growth factor and such ...

IIRC the `List<T>` roughly doubles, and I think `Dict<K, V>` can double too

After one doubling, it seems like you should always have room

mycpp/gc_dict.h
melvinw: Oh, yes! `at()` is also consistent with how the STL names this function (e.g. `std::map.at()` throws an exception if the key isn't found)

mycpp/gc_dict.h
melvinw: Yeah. Three times is excessive. Will simplify this

mycpp/gc_dict.h
melvinw: Let's go with `hash_and_probe()` and `find_kv_index()`. We don't really use the term bucket anywhere else in `Dict` right now so I think we should avoid introducing it, though I agree it feels natural.

mycpp/gc_dict.h
melvinw: this rename actually ended up being pretty easy. i'm gonna add it...

mycpp/gc_dict.h
melvinw: The existing `RoundCapacity()` method ended up being perfect for this

andychu:
Some notes, perf is looking good, even made the parser slightly faster (surprising since we don't really use dicts there):

http://travis-ci.oilshell.org/github-jobs/5122/benchmarks2.wwz/_tmp/gc-cachegrind/index.html

on fib before:

```


35.1 | _bin/cxx-opt+bumpleak/osh | mut
-- | -- | --
44.8 | _bin/cxx-opt+bumproot/osh | mut
47.0 | _bin/cxx-opt+bumpsmall/osh | mut+alloc
49.7 | _bin/cxx-opt/osh | mut+alloc

```

http://travis-ci.oilshell.org/github-jobs/5123/benchmarks2.wwz/_tmp/gc-cachegrind/index.html

on fib after: 

```


18.7 | _bin/cxx-opt+bumpleak/osh | mut
-- | -- | --
28.5 | _bin/cxx-opt+bumproot/osh | mut
30.4 | _bin/cxx-opt+bumpsmall/osh | mut+alloc
33.0 | _bin/cxx-opt/osh | mut+alloc

```

andychu:
Hmm there is a spec test difference, seems like a bug -

delta is 77 beforehand, and it's 78 afterward


http://travis-ci.oilshell.org/github-jobs/5122/cpp-spec.wwz/_tmp/spec/ysh-cpp/ysh-slice-range.html

http://travis-ci.oilshell.org/github-jobs/5123/cpp-spec.wwz/_tmp/spec/ysh-cpp/ysh-slice-range.html

(Also it is a bug that the CI passes nonetheless, I think this came up last time too ...)

mycpp/gc_dict.h
```diff
@@ -43,18 +40,12 @@ class GlobalDict {
  public:
   int len_;
   int capacity_;
-  GlobalSlab<int, N>* entry_;  // TODO: should be sized differently
+  GlobalSlab<int, N>* index_;  // TODO: should be sized differently
```
andychu, line 43: can remove this TODO

mycpp/gc_dict.h
```diff
@@ -110,42 +101,43 @@ class Dict {
...
-  // - hash functions, and linear probing.
-  // - resizing based on load factor
-  //   - which requires rehashing (re-insert all items)
-  // - Special case to intern Str* when it's hashed?  How?
-  //   - Should we have wrappers like:
-  //   - V GetAndIntern<V>(D, &string_key)
-  //   - SetAndIntern<V>(D, &string_key, value)
-  //   This will enable duplicate copies of the string to be garbage collected
-  int position_of_key(K key);
+  // Returns -1 if the key isn't in the table.
```
andychu, line 139: should be kNotFound now

mycpp/gc_dict.h
```diff
@@ -155,15 +147,13 @@ class Dict {
   int capacity_;  // number of entries before resizing
 
   // These 3 slabs are resized at the same time.
-  Slab<int>* entry_;  // NOW: kEmptyEntry, kDeletedEntry, or 0.
-                      // LATER: indices which are themselves indexed by // hash
-                      // value % capacity_
+  Slab<int>* index_;  // kEmptyEntry, or a valid index into keys_/values_
```
andychu, line 150: this can still be kDeletedEntry

mycpp/gc_mylib.h
```diff
@@ -41,14 +41,33 @@ Tuple2<Str*, Str*> split_once(Str* s, Str* delim);
...
-  if (pos == -1) {
+  DCHECK(haystack->obj_header().heap_tag != HeapTag::Global);
+  int pos = haystack->hash_and_probe(needle);
+  if (pos == kNotFound || haystack->index_->items_[pos] < 0) {
+    // Nothing to do.
     return;
   }
-  haystack->entry_->items_[pos] = kDeletedEntry;
+
+  int offset = haystack->index_->items_[pos];
```
andychu, line 51: I think it will make the code slightly more readable to have

```
int kv_index = haystack->index_->items_[pos];
if (kv_index < 0) {
  return;
}
```

so we have 2 early returns and use `kv_index` consistently

mycpp/gc_str.cc
```diff
@@ -525,6 +525,15 @@ List<Str*>* Str::split(Str* sep) {
   return this->split(sep, len(this));
 }
 
+int Str::hash(HashFunc h) {
+  // Use the MSB as a flag to indicate that we've already computed the hash.
+  // This keeps the header compact, since we don't have to add a bool.
+  if ((hash_value_ & 0x08000000) == 0) {
```
andychu, line 531: `0x0800_0000` isn't the MSB for 32 bits, it's `0x8000_0000`

That's what I meant by the previous comment :)

while I guess this works, let's use a bitfield to prevent confusion/mistakes

I haven't used bitfields before this project, but I think it will work fine here -- they seem to be very convenient

So we should have something like

```
unsigned hash : 31
unsigned is_hashed : 1
```

mycpp/gc_mylib.h
```diff
@@ -41,14 +41,33 @@ Tuple2<Str*, Str*> split_once(Str* s, Str* delim);
...
+  int last_offset = haystack->len_ - 1;
+  DCHECK(offset <= last_offset);
+
+  // To keep things compact, swap the target entry with the most recently
```
andychu, line 55: Does this break any ordering guarantees?

- the last kv_index of `haystack_` is supposed to be the most recently inserted one?
- but then swapping it breaks this property?


Or maybe this just needs some comments about what the invariant is

mycpp/gc_dict.h
```diff
@@ -307,49 +278,115 @@ void Dict<K, V>::clear() {
...
+
+  // Hash the key onto a slot in the index. If the first slot is occupied, probe
+  // until an empty one is found.
+  unsigned h = hash_key(key);
+  int init_bucket = h % capacity_;
+
+  // If there's a vacancy because of deletion on the probe path for this key, we
+  // should fill it before consuming a slot that has never been used. This
+  // should help keep the index somewhat compact.
+  int tombstone = kNotFound;
```
andychu, line 301: hm why is this var called "tombstone" ?  It's the return value?

I think of kDeletedEntry as the tombstone, so this reads funny

the tombstone meaning "this position in index_ used to point to some value in keys_ and values_, but it was deleted"

melvinw:
Nice catch on the spec tests! `OverallocatedStr()` was still initializing `hash_value_` the old way. Pushed a fix

mycpp/gc_dict.h
```diff
@@ -231,25 +202,25 @@ void Dict<K, V>::reserve(int n) {
     new_k = NewSlab<K>(capacity_);
```
andychu, line 202: line 196 can have its comment removed

I think we want to make `index_` bigger than `keys_` and `values_`, maybe multiply by 4/3 or 3/2 ?  

So the load factor is  3/4 or 2/3 (at most 3 out of 4 entries will not be kEmptyEntry or kDeletedEntry)


In theory this should affect performance -- the probing should be shorter

andychu (commented):
I think if we change the hash to a bitfield, it will also clarify the OverAllocatedStr issue as well

Right now we have an empty Str() constructor, which doesn't set fields, because:

1. we're using the "hanging array" pattern, which has to be dynamic
2. GLOBAL_STR() has to match Str


But maybe we can use the constructor


Actually I'm confused about global init too -- this is the thing that might change when we add string interning

So yeah string interning will affect this code



----

I don't quite understand the deletion/tombstone /probing invariants -- I think this could use some more comments, including in the unit tests

mycpp/gc_dict.h
andychu: Actually it's not sized differently now!  So we still have the TODO

I mentioned that in another comment

In theory I think if it's sized the same as keys_ and values_, then the performance will be worse ...

mycpp/gc_str.cc
melvinw: OH lol! Not sure how those got swapped... Nice eye. I guess this works because any bit is fine as long as we're consistently use the same one. Will fix.

Bitfields are somewhat annoying, because you have to guard them with a check on `__BYTE_ORDER` and define the layout twice, like below. Using the value instead of the layout of the int is a way to avoid this, but I get why it might seem more cumbersome. Are you willing to relax the "minimize ifdefs" goal for this?
```
#if __BYTE_ORDER == __BIG_ENDIAN
unsigned hash : 31
unsigned is_hashed : 1
#elif __BYTE_ORDER == __LITTLE_ENDIAN
unsigned is_hashed : 1
unsigned hash : 31
#else
// error
#endif
```

mycpp/gc_mylib.h
melvinw: Yes, this does. Just like in CPython 3. I thought I documented this somewhere... I was under the impression we were OK breaking ordering slightly in the presence of deletions. Do you want to make this try harder to preserve ordering?

mycpp/gc_dict.h
melvinw: Ah nice catch. Leftover from an old experiment

mycpp/gc_mylib.h
melvinw: It should be impossible for `offset` to be less than zero at this point, though. We check for that above and return early. Maybe I'm misunderstanding something...

I'm fine with changing the name, though

mycpp/gc_dict.h
melvinw: I dunno. It reads fine to me... It's not the return value all the time, like in this snippet from a few lines down:
```c++
      if (tombstone != kNotFound) {
        return tombstone;
      }
      return i;
```

Not really sure what name would be better without being verbose

mycpp/gc_mylib.h
andychu: I just meant to factor out the long expression, and give it a name

mycpp/gc_str.cc
andychu: I think you only need the `#ifdef` if you're sending it over the network, which we aren't doing

The other case you might need it is if you overlap integers/pointers, assuming they will line up, which we're not doing here

I think this is a "pure" usage of bitfields -- we have a 31 bit integer and a 1-bit int, and no other constraints

Only constraint is size

mycpp/gc_mylib.h
andychu: Hmm that wasn't my understanding of Python 3

Where did you see that?

I thought it always preserved insertion order

melvinw: I thought you pointed that out somewhere on Zulip. There was some mailing list entry with some code snippets that did this. This compaction thing isn't strictly necessary, though. We can just remove it from "fast path" and handle compaction whenever we rehash

mycpp/gc_str.cc
melvinw: Cool. Will tweak this

mycpp/gc_mylib.h
andychu: Yeah let's handle it when we rehash, and also add a unit test for our insertion order guarantee

The tradeoff appears to be 

1. saving space on deletion, vs.
2. dict ordering

And I think matching CPython with (2) is more important

My memory is that we saw at least 2 different ways to maintain the guarantee in the face of deletions

But yeah I agree the simplest thing is to handle it on rehashing

mycpp/gc_mylib.h
melvinw: Oh gotcha

mycpp/gc_dict.h
```diff
@@ -110,42 +101,43 @@ class Dict {
...
   void clear();
 
-  // Returns the position in the array.  Used by dict_contains(), index(),
-  // get(), and set().
+  // Returns an offset into the index for given key. If the key is not already
+  // in the table and there is room, the offset to an empty slot will be
+  // returned. The caller is responsible for checking if the index slot is empty
+  // before using it.
+  //
+  // Returns -1 if the dictionary is full. The caller can use this as a cue to
```
melvinw, line 131: note to self: s/-1/kNotFound/

mycpp/gc_dict.h
andychu: OK sounds good, yeah I didn't understand what this code was doing

mycpp/gc_dict.h
melvinw: Went with 3/2 as a multiplier for now. This was basically a random choice, though. We should profile at some point and tweak accordingly

mycpp/gc_heap_test.cc
```diff
@@ -16,7 +16,9 @@
   static_assert(offsetof(Str, field) == offsetof(GlobalStr<1>, field), \
                 "Str and GlobalStr should be consistent");
 ASSERT_GLOBAL_STR(len_);
-ASSERT_GLOBAL_STR(hash_value_);
+// NOTE: offsetof doesn't work with bitfields...
+// ASSERT_GLOBAL_STR(hash_);
+// ASSERT_GLOBAL_STR(is_hashed_);
```
melvinw, line 21: Meh :/

```
FAILED: _build/obj/clang-coverage/mycpp/gc_heap_test.o 
build/ninja-rules-cpp.sh compile_one clang coverage '' mycpp/gc_heap_test.cc _build/obj/clang-coverage/mycpp/gc_heap_test.o _build/obj/clang-coverage/mycpp/gc_heap_test.o.d
mycpp/gc_heap_test.cc:19:19: error: cannot compute offset of bit-field 'hash_'
ASSERT_GLOBAL_STR(hash_);
~~~~~~~~~~~~~~~~~~^~~~~~
mycpp/gc_heap_test.cc:16:31: note: expanded from macro 'ASSERT_GLOBAL_STR'
  static_assert(offsetof(Str, field) == offsetof(GlobalStr<1>, field), \
                ~~~~~~~~~~~~~~^~~~~~
/home/uke/oil_DEPS/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04/lib/clang/14.0.0/include/stddef.h:104:46: note: expanded from macro 'offsetof'
#define offsetof(t, d) __builtin_offsetof(t, d)
                       ~~~~~~~~~~~~~~~~~~~~~~^~
/home/uke/oil/mycpp/gc_str.h:83:12: note: bit-field is declared here
  unsigned hash_ : 31;
           ^
mycpp/gc_heap_test.cc:19:19: error: cannot compute offset of bit-field 'hash_'
ASSERT_GLOBAL_STR(hash_);
~~~~~~~~~~~~~~~~~~^~~~~~
mycpp/gc_heap_test.cc:16:64: note: expanded from macro 'ASSERT_GLOBAL_STR'
  static_assert(offsetof(Str, field) == offsetof(GlobalStr<1>, field), \
                                        ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
/home/uke/oil_DEPS/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04/lib/clang/14.0.0/include/stddef.h:104:46: note: expanded from macro 'offsetof'
#define offsetof(t, d) __builtin_offsetof(t, d)
                       ~~~~~~~~~~~~~~~~~~~~~~^~
/home/uke/oil/mycpp/gc_str.h:150:12: note: bit-field is declared here
  unsigned hash_ : 31;
           ^
mycpp/gc_heap_test.cc:20:19: error: cannot compute offset of bit-field 'is_hashed_'
ASSERT_GLOBAL_STR(is_hashed_);
~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
mycpp/gc_heap_test.cc:16:31: note: expanded from macro 'ASSERT_GLOBAL_STR'
  static_assert(offsetof(Str, field) == offsetof(GlobalStr<1>, field), \
                ~~~~~~~~~~~~~~^~~~~~
/home/uke/oil_DEPS/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04/lib/clang/14.0.0/include/stddef.h:104:46: note: expanded from macro 'offsetof'
#define offsetof(t, d) __builtin_offsetof(t, d)
                       ~~~~~~~~~~~~~~~~~~~~~~^~
/home/uke/oil/mycpp/gc_str.h:84:12: note: bit-field is declared here
  unsigned is_hashed_ : 1;
           ^
mycpp/gc_heap_test.cc:20:19: error: cannot compute offset of bit-field 'is_hashed_'
ASSERT_GLOBAL_STR(is_hashed_);
~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
mycpp/gc_heap_test.cc:16:64: note: expanded from macro 'ASSERT_GLOBAL_STR'
  static_assert(offsetof(Str, field) == offsetof(GlobalStr<1>, field), \
                                        ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
/home/uke/oil_DEPS/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04/lib/clang/14.0.0/include/stddef.h:104:46: note: expanded from macro 'offsetof'
#define offsetof(t, d) __builtin_offsetof(t, d)
                       ~~~~~~~~~~~~~~~~~~~~~~^~
/home/uke/oil/mycpp/gc_str.h:151:12: note: bit-field is declared here
  unsigned is_hashed_ : 1;
           ^
4 errors generated.
```

mycpp/gc_heap_test.cc
```diff
@@ -113,9 +115,9 @@ TEST offsets_test() {
     unsigned list_mask = List<int>::field_mask();
     ASSERT_EQ_FMT(0x0002, list_mask, "0x%x");
 
-    // in binary: 0b 0000 0000 0000 1110
+    // in binary: 0b 0000 0000 0001 1100
     unsigned dict_mask = Dict<int COMMA int>::field_mask();
-    ASSERT_EQ_FMT(0x0000e, dict_mask, "0x%x");
+    ASSERT_EQ_FMT(0x0001c, dict_mask, "0x%x");
```
melvinw, line 120: tweaked this because of the new `index_len_` member

melvinw:
OK. This should be ready for another round. The hash bitsets break the `GlobalStr`/`Str` layout compatibility checks because clang doesn't support `offsetof` on bitfields (g++ does, but it's one of these implementation-defined corners...). We can either:
1. Leave things as-is and informally enforce layout compatibility for strings
2. Go back to the mask approach
3. Keep track of hashes in `Dict`

What do you think?

mycpp/gc_dict.h
```diff
@@ -307,49 +283,92 @@ void Dict<K, V>::clear() {
...
+  }
+
+  // Hash the key onto a slot in the index. If the first slot is occupied, probe
+  // until an empty one is found.
+  unsigned h = hash_key(key);
+  int init_bucket = h % index_len_;
+
+  for (int i = 0; i < index_len_; ++i) {
+    // Start at init_bucket and wrap araound
+    int slot = (i + init_bucket) % index_len_;
```
andychu, line 305: I read this modulus can be slow ... and we also kinda do it twice on lines 301 and 305

It depends if `index_len_` is a power of 2, and right now we're growing it by 3/2

I think this is OK now but it might come up later, when we tune the growth factors

mycpp/gc_dict_test.cc
```diff
@@ -479,6 +479,112 @@ TEST test_tuple_key() {
...
+  // Reset and fill the table with keys that all has onto the same index slot
+  n = d->capacity_;
+  int target = n / 2;  // pick a slot in the middle to test wrap around
+  d->clear();
+  for (int i = 0; i < n; i++) {
+    d->set(target * i, i);
+  }
+  // Remove each entry one-by-one, stopping after each removal to check that
+  // the other keys can be set and retrieved without issue. This implicitly
+  // checks that special index entires like tombstones are working correctly.
```
andychu, line 576: s/entires/entries

mycpp/gc_dict.h
```diff
@@ -179,123 +175,103 @@ class Dict {
...
-    self->set(key, *v);
-    ++v;
-  }
-
-  return self;
-}
-#endif
-
 template <typename K, typename V>
 void Dict<K, V>::reserve(int n) {
```
andychu, line 184: yeah I think this should be `int desired_entries`, and the contract is that you can fit that number of entries

So the caller should not have to call RoundCapacity()

andychu (commented):
Looking good!

mycpp/gc_dict.h
```diff
@@ -307,49 +283,92 @@ void Dict<K, V>::clear() {
...
 template <typename K, typename V>
 void Dict<K, V>::set(K key, V val) {
-  int pos = position_of_key(key);
-  if (pos == -1) {  // new pair
-    reserve(len_ + 1);
+  DCHECK(obj_header().heap_tag != HeapTag::Global);
+  int pos = hash_and_probe(key);
+  if (pos == kNotFound) {
+    // No room. Resize and see if we can find a slot.
+    reserve(RoundCapacity(capacity_ * 2));
```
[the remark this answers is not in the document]
melvinw: Ah yeah. I was just using it here to handle the `capacity_ == 0`  case from [this thread](https://github.com/oilshell/oil/pull/1729#discussion_r1333843296). I'm fine dropping it though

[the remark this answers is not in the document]
andychu: Oh yeah the intention behind reserve() is that you call it with the exact number of things you need, and then it rounds up if necessary, based on an internal policy.

So we used to call it with reserve(n+1) when there was no key.  But it is valid for it to use  2*n behind the scenes -- that's hidden from the caller.

So 90% of the time, it should be a no-op, because it's already big enough

I think vector::reserve() works the same way


I think that logic is sound because it puts all the policy in one place.

The "attempts" logic was weird because reserve() is supposed to GUARANTEE you have enough space for that number of items.

It may be worth adding some comments/tests on that  ... I think we  are going to end up tuning how it works

mycpp/gc_dict.h
melvinw: Yeah, it can be expensive, but it doesn't show up in the profile right now (I think we have a few other bottlenecks elsewhere). If it becomes a problem we can split this loop back out into two loops over `[init_bucket, index_len)` and `[0, init_bucket)` like before. I like how compact the single loop is as long as we can get away with it

mycpp/gc_dict.h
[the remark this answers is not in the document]
andychu: yeah and probably calling the param `num_desired` is better than `n`, to make that clear

It ensures you have enough space for that number of items, but it can also allocate more, in order to save resizings

It is supposed to completely encapsulate the growth policy, like std::vector::reserve()

mycpp/gc_dict.h
melvinw: Sounds good

melvinw:
Pushed some fixes! Before we merge, we should figure what we're gonna do about the `Str`/`GlobalStr` bitfield thing https://github.com/oilshell/oil/pull/1729#issuecomment-1741621540

andychu:
Oh I think we can just leave that ... we're missing checks for 1 field, but the other fields are still checked, so that seems fine

mycpp/gc_dict.h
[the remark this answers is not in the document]
melvinw: Ah missed these! I'll replace the thing i pushed with `capacity_ + 1`

melvinw:
Ok great. Pushed a commit to simplify the resize branch in `set()` too

mycpp/gc_dict.h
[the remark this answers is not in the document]
melvinw: Whoops. Meant `len_ + 1`. `capacity_ + 1` would guarantee we allocate on every call to `set()`

mycpp/gc_dict.h
```diff
@@ -307,49 +284,117 @@ void Dict<K, V>::clear() {
...
+        return slot;
+      }
+    }
+
+    if (kv_index == kEmptyEntry) {
+      if (open_slot != -1) {
+        slot = open_slot;
+      }
+      // If there isn't room in the entry arrays, tell the caller to resize.
+      return len_ < capacity_ ? slot : kNotFound;
```
andychu, line 332: hm can we `DCHECK(len_ < capacity_)` because we got kEmptyEntry, and return `slot` ?

I think these conditions are equivalent:

- there's no room in the hash table for the new entry
- every entry i `index_` is either valid, or kDeletedEntry

In that case the iteration will proceed to the end of the loop, and then `kNotFound` will be returned ?

Maybe

```
if (kv_index == kEmptyEntry) {  // we found a place to put it
  DCHECK(len_ < capacity_);  // which means this must be true
  if (open_slot == -1) {
    return slot;  // put it in this empty slot
  } else {
    return open_slot;  // put it in the FIRST place  we found kDeletedEntry ??
  }
}
```

I guess I'm also a little confused if it matters if we return the position of the FIRST kDeletedEntry or the LAST ?  are they equivalent?

mycpp/gc_dict.h
```diff
@@ -307,49 +284,117 @@ void Dict<K, V>::clear() {
...
+      // looking for a key that was writen before this tombstone was written to
+      // the index we should continue probing until we get to that key. If we
+      // get to an empty index slot or the end of the index then we know we are
+      // dealing with a new key and can safely replace the tombstone without
+      // disrupting any existing keys.
+      open_slot = slot;
+    }
+  }
+
+  if (open_slot != -1) {
```
andychu, line 348: likewise it feels like when open_slot is set to an index that contains `kDeletedEntry`, then `len_ < capacity_` ?

So this could be

```
// We iterated through all the buckets, and did not find kEmptEntry

if (open_slot == -1) {  // We also did not find kDeletedEntry, so there's no place to put it
  return kNotFound;
} else {
  DCHECK(len_ < capacity_);
  return open_slot;  // Put it in the spot of a kDeletedEntry
}
```

?

andychu (commented):
Hm I proposed some changes ... if they are not correct then I'd like to know why :)

I think it's helpful to document the different conditions, since there are a bunch of them

andychu:
Yeah actually I think `tombstone_pos` is a clear name (while `tombstone` is confusing, because that's kDeletedEntry)

could be better than `open_slot`

melvinw:
Oh lol yes. You'll see in some of the intermediate commits that I tried adding the `DCHECK`s you mentioned, but I had to back them out because now that `index_` is bigger than `keys_`/`values_` it in't the case the encountering `kEmptyEntry` or even `kDeletedEntry` in the index implies `len_ < capacity_` anymore. Consider the following.

For `kEmptyEntry`. Suppose `index_len_` is twice `capacity_`. We could have `len_ == capacity_`, but still hash a new key onto an unoccupied slot. The assertion would cause a needless crash in this case, when the right response is to have the caller grow the table.

For `kDeletedEntry`. Again, suppose `index_len_` is twice `capacity_`. We could insert `capacity_` keys and get "lucky" that they all hash onto **[0, capacity_)** in the index. Then if we delete them all, that whole range will be `kDeletedEntry`. Then we could insert `capacity_` more keys and get "lucky" again and they all hash onto **[capacity_, index_len_)** in the index. If we then go to insert another key and it hashes somewhere in **[0, capacity_)** we'll see a `kDeletedEntry` and assert-failing in this case would also be wrong. We should have the caller grow the table.

mycpp/gc_dict.h
melvinw: > I guess I'm also a little confused if it matters if we return the position of the FIRST kDeletedEntry on the probe path, or the LAST ? are they equivalent?

It doesn't really matter if we pick the first, last, or anything in between. As long as we wait until we finish the probe sequence to be sure we are dealing with a new key before writing to whichever one we chose we're OK and don't have to worry about disrupting the paths to existing keys. Just did a quick test locally to confirm this.

The comment at the bottom of this loop captures this pretty well, I think.

melvinw:
> Yeah actually I think `tombstone_pos` is a clear name (while `tombstone` is confusing, because that's kDeletedEntry)
> 
> could be better than `open_slot`
> 
> Or I would even be more specific and say `first_tombstone_pos`

See my response to the inline comment about this. There's no special importance to this being the _first_ tombstone we saw. I feel pretty strongly that `open_slot` is fine given that.

andychu:
Ahh right, duh ... Thanks for going through all the iterations on this!  I like how it turned out

melvinw:
No problem!

PR #1730, opened 2023-09-16
[ysh] Expression Literals

PossiblyAShrub:
Adds support for expression literals:

```
ysh$ var l = ^[1 + 2]
ysh$ = type(l)
(Str)   'Expr'
ysh$ = evalExpr(l)
(Int)   3
```

### DIFF
--- a/library/func_misc.py
+++ b/library/func_misc.py
@@ -513,3 +513,19 @@ class Assert(vm._Callable):
             raise error.AssertionErr(msg, args.LeftParenToken())
 
         return value.Null
+
+
+class EvalExpr(vm._Callable):
+
+    def __init__(self, expr_ev):
+        # type: (expr_eval.ExprEvaluator) -> None
+        self.expr_ev = expr_ev
+
+    def Call(self, args):
+        # type: (typed_args.Reader) -> value_t
+        lazy = args.PosExpr()
+        args.Done()
+
+        result = self.expr_ev.EvalExpr(lazy, loc.Missing)
+
+        return result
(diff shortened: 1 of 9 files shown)


library/func_misc.py
```diff
@@ -513,3 +513,19 @@ def Call(self, args):
...
+    def __init__(self, expr_ev):
+        # type: (expr_eval.ExprEvaluator) -> None
+        self.expr_ev = expr_ev
+
+    def Call(self, args):
+        # type: (typed_args.Reader) -> value_t
+        lazy = args.PosExpr()
+        args.Done()
+
+        result = self.expr_ev.EvalExpr(lazy, loc.Missing)
```
andychu, line 529: Melvin just added args.LeftParenToken() or similar, so we can use that instead of `loc.Missing`

It will point to the `(` in `evalExpr()`

andychu (commented):
Perfect!  thanks for adding this

Just one comment about location info

---

Also related is that we probably want `$[x]` and `@[x]` in  expression mode too, for consistency with command mode.  (on the other hand, `^[x]` doesn't make sense in command mode)

Just like we have `$(echo hi)` and `@(slurp-lines)`

- `$[x]` is "stringify expression"
- `@[x]` is "stringify each element of List[T]` -> `List[str]`
  - also related is the join() builtin that Melvin implemented.  It now does essentially `@[x]` and then `join()`, although it's not implemented exactly that way

This can be done later, when we hit it, but just mentioning it for consistency

andychu:
OK this seems done, so I merged!  Thanks

andychu:
BTW the general direction I'm pushing in for YSH is still the flag parser and test framework

The `value.Expr` will be useful for the test framework, probably with `assert`

I got sidetracked by completion and the release, but hope to get back to it

PR #1733, opened 2023-09-19
[ysh] List->join

PossiblyAShrub:
I am making some progress on the argument parser. While doing that I found that I needed `List->join`.

### DIFF
--- a/core/shell.py
+++ b/core/shell.py
@@ -311,7 +311,7 @@ def AddMethods(methods):
         'append': func_misc.Append(),
         'extend': func_misc.Extend(),
         'pop': func_misc.Pop(),
-        'join': func_misc.Join(),
+        'join': func_misc.Join(),  # NOTE: This is both a method and a func
     }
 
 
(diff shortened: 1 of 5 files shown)


andychu:
OK great, this makes sense -- to expose join() as both function and method

So you can use it like

```
join(mylist)   # short
join(mylist, ':')
mylist->join(':')->upper()  # chained style
```

I mentioned this issue of "universal function call" in the latest blog post:

https://www.oilshell.org/blog/2023/09/release-0.18.0.html#design-issues-in-flight

---

Can you just add one line to `doc/ref/toc-ysh.md` next to `List append()` to  show that this exists ?

We will make a pass to fill in everything ... I still need to make some kind of checker/metric for that


---

Oh also definitely keep us updated on the flag parser on Zulip ... BTW I changed some stuff recently in `spec/ysh-builtin-argparse.test.sh ` -- it should be renamed

Main issue I can see there is that `var arg, i = parseArgs()` doesn't work from the API side ... it has to be `setvar arg, i = `, which is annoying

PR #1735, opened 2023-09-20
[mycpp] Unroll StackRoots

melvinw:


mycpp/gc_alloc.h
```diff
@@ -26,6 +26,11 @@ class StackRoots {
  public:
   // Note: void** seems logical, because these are pointers to pointers, but
   // the C++ compiler doesn't like it.
+  explicit StackRoots(void* root) {
+    n_ = 1;
```
andychu, line 30: does removing `n_` altogether do anything?

Then the destructor can have no loop

Maybe it will be optimized better because the class  no longer contains any members at all?

melvinw:
Ah, yes. A separate class is a good idea. I'll push that

mycpp/gc_alloc.h
```diff
@@ -22,6 +22,40 @@ extern MarkSweepHeap gHeap;
 #define VALIDATE_ROOTS 0
 
 // mycpp generates code that keeps track of the root set
+class StackRoot {
+ public:
+  StackRoot(void* root) {
+#if VALIDATE_ROOTS
```
andychu, line 28: Hm seems like we can extract a common ValidateRoot() free function, since it doesn't use `this` ?

It could take either `void*` or `RawObject*`, I think `RawObject*`  does give some type safety in some cases

melvinw: Ah sure. `RawObject*` seems logical

mycpp/gc_alloc.h
```diff
@@ -22,6 +22,40 @@ extern MarkSweepHeap gHeap;
...
+    RawObject* obj = *(reinterpret_cast<RawObject**>(root));
+    if (obj) {
+      RawObject* header = ObjHeader::FromObject(obj);
```
andychu: Also, by inspection, I wonder if we had a compile error here

I think this should be `ObjHeader* header`, not `RawObject*`

So it's masked by the `#ifdef` ?   Another reason to extract it, and put only the call site in the `#ifdef`, not the body

mycpp/gc_alloc.h
melvinw: Pushed

mycpp/gc_alloc.h
melvinw: Yup. There was :) fixed in the new commit

andychu:
Great thanks!  looks good

PR #1736, opened 2023-09-24
WIP: Implement args.ysh

PossiblyAShrub:
This is very much a work in progress, but we have the basic features working!

The snippet below now works!

```
source --builtin args.ysh

Args :spec {
  flag -v --verbose ('bool', false)
  flag -c --count ('int', 120)
  arg query
  rest files
}

= parseArgs(spec, :| -v hello example.sh other.py |)

# STDOUT: (List)   [<'verbose': True, 'count': 120, 'query': 'hello', 'files': ['example.sh', 'other.py']>, 4]
```

Here's a list of the changes I've had to make:
- Add support for `BlockArg` being passed to the `eval` builtin
- Partial `proc` argument binding
- Upgraded `dict()` to support casting from `BashAssoc` to `Dict` (`json read` returns `BashAssoc`, maybe it should be returning `Dict`?)
- Implemented `parseArgs`, `Args`, `flag`, `arg`, `rest`

There are a few hacks which need to be cleanup up:
- proc argument binding should be finalized, it's a mess right now
- eval is broken as now it only supports `BlockArg`s
- there are definitely a few more, and maybe some more tests to find some bugs


andychu:
Thanks for the update!  

Oh about default argument evaluation  -- I wanted to move that to the `core/code.py`, right now for funcs it's in `osh/cmd_eval.py` I think

AND I wanted to get rid of the "mutable default arg pitfall" in Python

```
def f(mylist=[]):
  mylist.append(42)  
  return mylist  # this is confusing because you're returning the SAME INSTANCE every time, leading to an ever-growing single list
```

Well I guess we can do that in 2 passes as well ... I guess I'm just giving a heads up that this will change

I think we will allow `def f(i=3)`, but not mutable containers `[]` and `{}` as defaults.  I think I mentioned this on the "#language-design > Things Oils Shipped Without" thread

----

I will look at the PR a bit more today

Thanks for the update!

osh/builtin_meta.py
```diff
@@ -39,18 +39,30 @@
...
+        self.expr_ev = expr_ev
         self.tracer = tracer
         self.errfmt = errfmt
 
     def Run(self, cmd_val):
         # type: (cmd_value.Argv) -> int
 
+        if cmd_val.typed_args:
+            t = typed_args.ReaderFromArgv(cmd_val.argv[1:], cmd_val.typed_args,
+                                          self.expr_ev)
```
andychu, line 58: Side note, not strictly related to this PR - I kinda think we should change `cmd_val` so it has evaluated arguments

I was thinking about the lazy evaluation semantics for awhile, and then I realized we want something at the call site

I think it will be

```
json write (x)  # eager
json write [x]  # lazy, synonym for (^[x])

```

there's a Zulip thread on it

osh/cmd_eval.py
```diff
@@ -2047,14 +2049,14 @@ def EvalBlock(self, block):
...
-        return namespace_
+        return namespace_, status
 
     def RunFuncForCompletion(self, proc, argv):
         # type: (Proc, List[str]) -> int
 
         # TODO: Change this to run YSH procs and funcs too
         try:
-            status = self.RunProc(proc, argv, loc.Missing)
+            status = self.RunProc(proc, argv, loc.Missing, None)
```
andychu, line 2059: Hm the I guess RunProc should have `Optional[ArgList]` then

core/code.py
```diff
@@ -98,77 +99,128 @@ def Call(self, args):
...
+                     for x in t.RestWords()]  # type: List[value_t]
+            val = value.List(items)
+
+            mem.SetValue(lvalue.Named(rw.name, rw.blame_tok), val,
+                         scope_e.LocalOnly)
+
+        npos = t.NumPos()
+        for i, p in enumerate(sig.pos_params):
+            if i >= npos and p.default_val:
+                # TODO: can we cache this? would require dependency checking...
```
andychu, line 178: Ah the defaults for word_params are computed in `osh/cmd_eval.py` when the `proc` is **defined**

I think we should keep that?  That's why I mentioned that issue about mutable default args ...

I can help with this ...

osh/cmd_eval.py
```diff
@@ -437,11 +437,12 @@ def _EvalRedirect(self, r):
                     # with globs
 
                     # mycpp needs this explicit declaration
-                    b = braces.BraceDetect(arg_word)  # type: Optional[word.BracedTree]
+                    b = braces.BraceDetect(
+                        arg_word)  # type: Optional[word.BracedTree]
```
andychu, line 441: Hm I'm getting different formatting

Can you try running

```
devtools/format.sh install-yapf
```

Then

```
$ devtools/format.sh yapf-version 
yapf 0.40.1
```
devtools/format.sh yapf-files
```

It should print version 0.40.1

And then `devtools/format.sh yapf-files osh/cmd_eval.py`

**Review by andychu — COMMENTED** [review]
Thanks for getting this going!

I made another pass over this, a few comments

I think there is a formatting issue -- I created some scripts to synchronize on the same version of yapf, so please try that, and maybe the diff will be slightly smaller

We can also merge some of this in pieces, I can help with that

File: osh/builtin_meta.py [hunk 5]
```diff
@@ -39,18 +39,30 @@
...
+
+            block = t.PosCommand()
+            t.Done()
+
+            unused, status = self.cmd_ev.EvalBlock(block)
```
andychu, line 63: is status unused here?    

I guess it's worth considering how to propgate status

```
$ eval 'false'; echo $?
1
```

osh/cmd_eval.py
PossiblyAShrub: Hmmm. My `yapf` version is also `0.40.1`. And, even after reinstalling, I cannot get the same result as you.

FWIW I ran:

```sh
$ devtools/format.sh install-yapf
$ devtools/format.sh yapf-version
$ devtools/format.sh yapf-files osh/cmd_eval.py
```

I am honestly not sure what the issue is then. Do you have any ideas?

In case it helps, here's what I get when manually checking the yapf version:

```sh
$ source _tmp/yapf-venv/bin/activate
(yapf-venv) $ pip list
Package            Version
------------------ -------
importlib-metadata 6.8.0
pip                22.0.4
platformdirs       3.10.0
setuptools         58.1.0
tomli              2.0.1
yapf               0.40.1
zipp               3.17.0
WARNING: You are using pip version 22.0.4; however, version 23.2.1 is available.
You should consider upgrading via the '/home/aolse/oils/oil/_tmp/yapf-venv/bin/python3 -m pip install --upgrade pip' command.
```

And even manually invoking this version of yapf:

```sh
(yapf-venv) $ python3 -m yapf -i osh/cmd_eval.py
(yapf-venv) $ git status
On branch argparse
Your branch is up to date with 'origin/argparse'.

nothing to commit, working tree clean
```

andychu: Oh I should have said that I do get some diffs in `osh/cmd_eval.py`, but not the same ones as in this PR

Maybe we can debug it on Zulip a little ...

It would be nice to get the same diffs, then possibly merge the formatting change separate from the  other changes, just so it's easier to see what happened

And smalller diffs tend to make merges easier

But it's not essential, we can play with it a little

andychu: I'll send a PR to show what I mean, maybe we can get to the bottom of this

spec/ysh-builtin-error.test.sh
```diff
@@ -114,7 +114,7 @@ error (100, status=42)
 
 #### Errors cannot take command args
 error uh-oh ('error', status=1)
-## status: 3
+## status: 2
```
andychu, line 117: Hmm why did this one change from 3 to 2?

spec/ysh-builtin-eval.test.sh
```diff
@@ -0,0 +1,41 @@
+# YSH specific features of eval
+
+## our_shell: ysh
+## oils_failures_allowed: 0
+
+#### Eval a block within a proc
+proc run (;;; block) {
+  eval (block)
```
andychu, line 8: it's cool that this works !

spec/ysh-user-feedback.test.sh
```diff
@@ -16,7 +16,7 @@ get_opt a 'lol'
 
 echo 'should not get here'
 
-## status: 1
+## status: 2
```
andychu, line 19: Hm I wonder why these changed too

andychu (commented):
OK I merged the formatting change into this one -- actually wasn't too bad

So you can pull now, and the diff is cleaner

And you should be able to run yapf more reliably now

---

I will look at this more closely, maybe we should chat about it on Zulip

The general idea looks right, but I didn't look at all the details

spec/ysh-builtin-error.test.sh
PossiblyAShrub: The extra `uh-oh` is registered as a usage error. We raise `error.Usage` on extra word arguments being passed to the `typed_args.Reader`.

This behavior now better matches other builtins (or system tools):

```sh
$ # for reference, on `master`
$ bin/ysh -c 'try { error --not-a-flag ("My error") }; echo $_status'
3
$ bin/ysh -c 'try { source --not-a-flag }; echo $_status'
  try { source --not-a-flag }; echo $_status
               ^~~~~~~~~~~~
[ -c flag ]:1: 'source' got invalid flag '--not-a-flag'
2
```

spec/ysh-user-feedback.test.sh
PossiblyAShrub: Because `'lol'` doesn't start with the `:` sigil, it fails with a usage error. We actually get a pretty nice error message here too!

```sh
$ bin/ysh -c 'proc myproc (out Ref) { setref out = 42 }; myproc '\''lol'\'   
  proc myproc (out Ref) { setref out = 42 }; myproc 'lol'
                                             ^~~~~~
[ -c flag ]:1: fatal: Ref param 'out' expected arg starting with colon : but got 'lol'
```

But I guess this actually goes against the spirit of this test. Should we change that line to `get_opt 'lol' :a`?

andychu: Oh yeah probably!  not sure what happened there, maybe a typo

andychu (commented):
Since you mentioned some issues with proc argument binding, I hope to have time to take a look tonight

Either make some changes to this PR, or maybe on master and then merge into this PR

andychu:
I'm going to merge a few of the smaller parts of this, to help split it up

I will try to do stuff tonight, so let me know if I collide with you.  I just merged master to re-run CI

Then we can focus on the core proc param binding parts

andychu:
OK I merged a few of the files ... yeah I agree that we may want to move some code around with respect to argument binding

- `ArgList` are expressions, like `List[expr_t]`
- cmd_value.Argv holds and `ArgList`

Arguably `cmd_value.Argv` should not hold `expr_t` ... it should hold `value_t` ???

Let me think about this

PossiblyAShrub:
> Arguably `cmd_value.Argv` should not hold `expr_t` ... it should hold `value_t` ???

Yeah, I think that this would make more sense. Then we can evaluate arguments (and handle any errors there) before running the builtin. We now have `value.Expr`, so the argument about lazy evaluation doesn't really hold anymore.

Other languages do this too (in their internal/native APIs). For example:
- Crafting Interpreters: https://craftinginterpreters.com/functions.html#native-functions
- Wren: https://wren.io/embedding/calling-c-from-wren.html#implementing-a-foreign-method
- Lua: https://www.lua.org/manual/5.4/manual.html#lua_CFunction

This also matches how arguments are handled within each of those languages. That is, arguments are evaluated before the function body is.

andychu:
OK I'm going to change that ... actually the only 2 builtins we have now that use expressions are

- json write
- error

So it should be easy to change

Yeah I think I had figured out we could do the  `cmd [lazyexpr]` syntax a few weeks ago vs `cmd (expr)`, and then never got back to it

andychu:
Let me know if you want to brainstorm about this more ... I think there are several parts and it makes sense to split up a bit

I like to have the "goal" of args.ysh and then kinda backtrack, and work our way back up to it, once we see what's missing

(today/tomorrow I might be a little busy but later in the week I'm free)

PossiblyAShrub:
Yeah, sorry I haven't had time to work on this for the past week.

I think I'm good to go ahead with the proc argument binding updates. Beyond that, working out a plan over a call would make sense. I am more free later in the week as well. I'll send you a message on zulip to coordinate that.

andychu:
BTW I merged master, and CI passed

But I just fixed the EvalBlock API, and we need to resolve those conflicts ... it was a small change to the return value

andychu:
Actually I think the `eval (cmd)` is the logical first thing to merge --  I might look at it tomorrow

andychu:
I merged the `eval` change BTW

https://github.com/oilshell/oil/commit/23a58f80405b064aeed391f892beffb8ba4e3b82

Now I'm thinking about how to make these all work

```
cd /tmp { echo hi }  # works
eval { echo hi }  # doesn't work yet

var b = ^(echo hi)

eval (b)  # works
cd /tmp (b)  # doesn't work

```

As mentioned we're a bit bottlenecked on the design -- some parts aren't clear, and still need to design

I put a threads on Zulip about it

---

I'm thinking this PR itself will mainly end up as args.ysh and tests

stdlib/prelude.ysh
```diff
@@ -0,0 +1,15 @@
+# These were helpful while implementing args.ysh
+# Maybe we will want to export them in a prelude so that others can use them too?
+#
+# Prior art: Rust has `todo!()` which is quite nice. Other languages allow
+# users to `raise NotImplmentedError()`.
```
andychu, line 5: Let's try these out for awhile

I added some comments in the code about this

```
# - 'pass' can be : or true in shell.  It's a little obscure / confusing, but
#   there is an argument for minimalism.  Although I prefer words like 'true',
#   and that already means something.
# - todo could be more static?  Rust presumably does it at compile time

```

Oh actually I JUST remembered that we used to have a `pass` **keyword**.  It was for left-to-right arg passing.

But then people complained that it broke their scripts -- `pass` already means something, it asks for a password or something

So yeah I'm thinking this will be problematic by default

Damn the flat shell namespace!

stdlib/args.ysh
```diff
@@ -24,140 +24,159 @@
...
+  var value = null
+  while (i < argc) {
+    var arg = argv[i]
+    if (arg->startswith('-')) {
+      for flag in (spec.flags) {
+        if ( (flag.short and flag.short === arg) or
+             (flag.long and flag.long === arg) ) {
+          case (flag.type) {
+            ('bool') | (null) { setvar value = true }
+            int {
```
andychu, line 147: This could be `('int')` to be more consistent with the line above

I guess it is a bit subtle to have an unquoted word -- it can also be `'int'`

stdlib/args.ysh
```diff
@@ -24,140 +24,159 @@
...
+  ##
+  ## See `Args` for examples of use.
 
-  # _this is part of this frame -- it's a duplicate that callers can look up I
-  # guess
-  # Well then all you need is getvar() ?  You don't even need bind-vars?
-  bind-vars (_this=this) {
-    # So now 'arg' and 'flag' can see _this?
+  var i = 0
+  var positionalPos = 0
```
andychu, line 132: On Zulip we mentioned `snake_case` for locals, and then `camelCase()` for functions

It is a little funny that it feels like `func` wants `namesLikeJS`, while proc wants `names_like_shell`, but that is probably too fine a distinction

stdlib/args.ysh
```diff
@@ -24,140 +24,159 @@
...
-#
-# - shvar does LocalOnly
-# - but 'proc arg' will push a new stack frame?
-#   - so we need something like setref?
-#
-
-proc bind-vars (; ; ...bindings; block) {
-  echo
+proc rest (name) {
+  ## Take the remaining positional arguments within an `Args`.
```
andychu, line 103: I wonder if we can just use python style docstrings, like 

`''' foo bar '''`

We didn't have those when I added the doc comments, which are `###`

andychu (commented):
OK I merged master into this branch, and added a few comments.

Good news is it passed the same tests without any changes!  So we're on the right track with proc/func params and args

I will just merge stdlib/*.ysh and the tests as is, and then we can iterate on this

Thanks for doing this!  

I will also write something on Zulip about what happened with procs and funcs -- there were some big refactorings, and major  Zephyr ASDL schema changes

PR #1737, opened 2023-09-26
[fanos] implement basic completion for PARSE keyword

Melkor333:
implement PARSE, return completions with newline separation. This PR is very minimal and I'd like to have a discussion on how to properly return the completions (since newline-separated is IMO bad).
It would also be great if we can add additional information like the symbol type (function, file, builtin, etc.).


Melkor333:
Closing this in favor of using `EVAL compexport -c *command*`, as proposed in Zulip.

PR #1739, opened 2023-09-29
test out yapf

andychu:


andychu:
@PossiblyAShrub Maybe you can send a PR from a totally separate branch, just like this

```
devtools/format.sh yapf-files osh/builtin_misc.py osh/cmd_eval.py core/state.py core/shell.py
```


I don't want to get too hung up on formatting, but maybe can remove this friction

I haven't been running yapf totally regularly, since a few decisions it makes are bad -- even though it's generally good

(also docformatter has messed up a few docstrings, though I think that tool was just a one-off)

core/state.py
```diff
@@ -2123,7 +2132,8 @@ def GetValue(self, name, which_scopes=scope_e.Shopt):
                         strs.append(_LineNumber(frame.call_tok))
 
                     elif case(debug_frame_e.Main):
-                        strs.append('0')  # Bash does this to line up with 'main'
+                        strs.append(
+                            '0')  # Bash does this to line up with 'main'
```
andychu, line 2136: Bad!  I dislike this

core/state.py
```diff
@@ -2022,7 +2029,8 @@ def GetValue(self, name, which_scopes=scope_e.Shopt):
         # if name not in COMPUTED_VARS: ...
 
         if name == 'ARGV':
-            items = [value.Str(s) for s in self.GetArgv()]  # type: List[value_t]
+            items = [value.Str(s)
+                     for s in self.GetArgv()]  # type: List[value_t]
```
andychu, line 2033: yuck!

core/state.py
```diff
@@ -693,8 +695,8 @@ def set_xtrace(self, b):
 
     def _SetArrayByNum(self, opt_num, b):
         # type: (int, bool) -> None
-        if (opt_num in consts.PARSE_OPTION_NUMS and
-                not self.mem.ParsingChangesAllowed()):
+        if (opt_num in consts.PARSE_OPTION_NUMS
+                and not self.mem.ParsingChangesAllowed()):
```
andychu, line 699: yuck!  dislike the `and`

osh/cmd_eval.py
```diff
@@ -437,11 +437,12 @@ def _EvalRedirect(self, r):
                     # with globs
 
                     # mycpp needs this explicit declaration
-                    b = braces.BraceDetect(arg_word)  # type: Optional[word.BracedTree]
+                    b = braces.BraceDetect(
+                        arg_word)  # type: Optional[word.BracedTree]
```
andychu, line 441: yuck!

andychu (commented):
OK I wonder if we should change our line length to 90 or 100, maybe that fixes the worst cases here

andychu:
They have column limit

https://github.com/google/yapf#column_limit

PossiblyAShrub:
I created https://github.com/oilshell/oil/pull/1740. Running `git diff` shows no difference?

andychu:
OK great, so then I will

1. try to land some of the formatting changes, BUT fix yapf config so it doesn't make the things I pointed out worse
1. merge them into your branch
1. then you should be able to pull, and run the formatter, and we both get the same results

The problem here is really that I haven't been running `yapf` consistently, because it makes things worse in 10% of cases

Time to fix that

andychu:
OK I think this is sorted out by the new `.style.yapf` and running `devtools/format.sh` with the virtualenv

So now we should be able to repeatably format the code

Only issue is sometimes we should manually put comments on the line above the code, to avoid wrapping, since yapf will wrap weirdly

PR #1740, opened 2023-09-29
test out yapf

PossiblyAShrub:
This is the diff after running:

```sh
devtools/format.sh yapf-files osh/builtin_misc.py osh/cmd_eval.py core/state.py core/shell.py
```


andychu:
OK I think this is sorted out by the new `.style.yapf` and running `devtools/format.sh` with the virtualenv

So now we should be able to repeatably format the code

Only issue is sometimes we should manually put comments on the line above the code, to avoid wrapping, since yapf will wrap weirdly

PR #1743, opened 2023-10-04
[mycpp] warn about excessive stack roots

melvinw:
Output on master.

```
WARNING: RootCompleter::Matches() has 29 stack roots. Consider refactoring this function.
WARNING: ::_PrintVariables() has 25 stack roots. Consider refactoring this function.
WARNING: SpecBuilder::Build() has 18 stack roots. Consider refactoring this function.
WARNING: Printf::_Format() has 23 stack roots. Consider refactoring this function.
WARNING: Printf::Run() has 17 stack roots. Consider refactoring this function.
WARNING: CommandEvaluator::_Dispatch() has 39 stack roots. Consider refactoring this function.
WARNING: ::_MakeAssignPair() has 19 stack roots. Consider refactoring this function.
WARNING: CommandParser::_MaybeExpandAliases() has 17 stack roots. Consider refactoring this function.
WARNING: CommandParser::ParseSimpleCommand() has 18 stack roots. Consider refactoring this function.
WARNING: Evaluator::Eval() has 19 stack roots. Consider refactoring this function.
WARNING: OilPrinter::DoCommand() has 18 stack roots. Consider refactoring this function.
WARNING: ::_PushOilTokens() has 25 stack roots. Consider refactoring this function.
WARNING: ::Main() has 100 stack roots. Consider refactoring this function.
```

### DIFF
--- a/soil/worker.sh
+++ b/soil/worker.sh
@@ -10,6 +10,7 @@ set -o pipefail
 set -o errexit
 
 REPO_ROOT=$(cd "$(dirname $0)/.."; pwd)  # tsv-lib.sh uses this
+EXTRA_MYCPP_ARGS="--stack-roots-warn 16"
 readonly REPO_ROOT
 
 source soil/common.sh
--- a/build/ninja-rules-py.sh
+++ b/build/ninja-rules-py.sh
@@ -10,6 +10,7 @@ set -o pipefail
 set -o errexit
 
 REPO_ROOT=$(cd "$(dirname $0)/.."; pwd)
+: ${EXTRA_MYCPP_FLAGS:=""}
 
 source build/dev-shell.sh  # python2 in $PATH
 source mycpp/common-vars.sh  # MYPY_REPO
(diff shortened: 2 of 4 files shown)


andychu:
I think this is a good report, but I'm not sure it should be on by default, since it will cause some non-actionable warnings -- not all of them should be fixed

I think we could pass an extra flag in the CI only or something

For now I think it should be a flag, and then we can turn it on when we need it ?

melvinw:
Sounds good. Added the flag. Where's the best spot to add this for CI? Perhaps [here](https://github.com/oilshell/oil/blob/master/soil/worker.sh#L18-L23)?

andychu:
One way to do it is to pass a flag to mycpp here, conditional on env variable set in CI:

https://github.com/oilshell/oil/blob/master/build/ninja-rules-py.sh#L77

Normally I would say that is a hack, but since it doesn't change the code generated, I think it's OK

Another option is to write the warnings to a file if we want to see them better, but I think stderr is fine to start

soil/worker.sh
```diff
@@ -10,6 +10,7 @@ set -o pipefail
 set -o errexit
 
 REPO_ROOT=$(cd "$(dirname $0)/.."; pwd)  # tsv-lib.sh uses this
+EXTRA_MYCPP_ARGS="--stack-roots-warn 16"
```
andychu, line 13: hm I would think this needs export, but let's just see what happens

andychu (commented):
Oops sorry I dropped this somehow, let's merge and see what happens

PR #1746, opened 2023-10-08
xxx dict optimizations

melvinw:


andychu:
I fixed the List<T> growth and I think I have a good idea for the Dict growth - https://oilshell.zulipchat.com/#narrow/stream/384167-performance/topic/Tuned.20List.3CT.3E.20growth.20policy

The Slab<K> and Slab<V> should follow the List logic -- they will be tuned for the pool

And then the `index_` will be a power of 2

The first size class for Slab<T*> is 5, so if you Roundup(), you get 8 for the index.  And then that's a 8 + (4*8) = 40 byte index, which also fits in the second pool

So it might be something like `RoundUp( len_ * 5/4)` for the index size

PR #1752, opened 2023-10-26
[test/spec] Spec tests for type -a

ellen364:
Partially addresses issue #259.

Possibly not good practice to open this before I've addressed the comments on the `shopt -s nocasematch` PR, but writing these spec tests was more relaxing than most things I could do. Hopefully they're useful.

### DIFF
--- a/spec/builtin-bash.test.sh
+++ b/spec/builtin-bash.test.sh
@@ -1,4 +1,4 @@
-## oils_failures_allowed: 4
+## oils_failures_allowed: 16
 ## compare_shells: bash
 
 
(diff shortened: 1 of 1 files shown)


spec/builtin-bash.test.sh
```diff
@@ -190,6 +190,68 @@ tar is /tmp/tar
 grep is /tmp/grep
 ## END
 
+#### type -a -> function
+f () { :; }
+type -a f
+## stdout-json: "f is a function\nf () \n{ \n    :\n}\n"
```
andychu, line 196: Oh there is a new style 

```
## STDOUT:
can have 
newlines here
## END
```

andychu (commented):
This is great thanks!  

Do you mind changing it to the new style?  This file was written before that existed, so it was a bit misleading

Changing the older test cases is welcome as well

(And it's perfectly welcome to have more than 1 PR in flight :) )

ellen364:
New and existing tests converted to use the new assertion style.

The multi-line style proved a little fiddly for test `type -a -> function` due to trailing spaces that the IDE was keen to remove. I think that will be more obvious now the initial version is committed because git will show any changes in trailing whitespace.

andychu:
Thanks!  For the trailing spaces, there are a variety of options:

- it's OK to use stdout-json if we care about the spaces
- we can `sed` out the spaces, if we don't care

In this case, there is probably an even bigger problem, in that we need to print source code.  There's a Zulip thread about this, and possibly a Github issue

https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Shells.20can.20print.2Freformat.20source.20code

In any case this is great progress, we can definitely make some of them pass pretty easily.

I didn't realize `type -a` was in #259  and it was also on `#oil-help` recently!

Issue #1732, opened 2023-09-19
Crash When Comparing Functions (and Other Values)


PossiblyAShrub:
This also happens with other values, ie `^[1] === ^[1]`.

Not sure if we want to define equality on these values, but we shouldn't crash.

### Traceback

```
$ bin/ysh -c '= str === str'
Traceback (most recent call last):
  File "/home/aolse/oils/oil/bin/oils_for_unix.py", line 208, in <module>
    sys.exit(main(sys.argv))
  File "/home/aolse/oils/oil/bin/oils_for_unix.py", line 183, in main
    return AppBundleMain(argv)
  File "/home/aolse/oils/oil/bin/oils_for_unix.py", line 138, in AppBundleMain
    return shell.Main('ysh', arg_r, environ, login_shell, loader, readline)
  File "/home/aolse/oils/oil/core/shell.py", line 1009, in Main
    cmd_flags=cmd_eval.IsMainProgram)
  File "/home/aolse/oils/oil/core/main_loop.py", line 353, in Batch
    is_return, is_fatal = cmd_ev.ExecuteAndCatch(node, cmd_flags=cmd_flags)
  File "/home/aolse/oils/oil/osh/cmd_eval.py", line 1859, in ExecuteAndCatch
    status = self._Execute(node)
  File "/home/aolse/oils/oil/osh/cmd_eval.py", line 1681, in _Execute
    status = self._Dispatch(node, cmd_st)
  File "/home/aolse/oils/oil/osh/cmd_eval.py", line 1084, in _Dispatch
    val = self.expr_ev.EvalExpr(node.e, loc.Missing)
  File "/home/aolse/oils/oil/ysh/expr_eval.py", line 303, in EvalExpr
    val = self._EvalExpr(node)
  File "/home/aolse/oils/oil/ysh/expr_eval.py", line 1024, in _EvalExpr
    return self._EvalCompare(node)
  File "/home/aolse/oils/oil/ysh/expr_eval.py", line 573, in _EvalCompare
    result = val_ops.ExactlyEqual(left, right)
  File "/home/aolse/oils/oil/ysh/val_ops.py", line 410, in ExactlyEqual
    raise NotImplementedError(left)
NotImplementedError: (value.Func callable:<library.func_misc.Str_ object at 0x7fd2f31dfd90>)
```

PossiblyAShrub:
[...]

PR #1754, opened 2023-10-31
[ysh] Fix Crash When Comparing Non-Comparable Types

PossiblyAShrub:
Issue: #1732

Now we raise a `error.TypeErr` when comparing non-comparable types.

```sh
ysh$ = {}->keys === {}->keys
  = {}->keys === {}->keys
  ^
[ interactive ]:1: fatal: Cannot compare this type, got BuiltinMethod
```

### DIFF
--- a/spec/ysh-expr-compare.test.sh
+++ b/spec/ysh-expr-compare.test.sh
@@ -370,3 +370,13 @@ for val in (unimpl) {
 }
 ## STDOUT:
 ## END
+
+#### Non-comparable types in case arms
+var myexpr = ^[123]
+
+case (myexpr) {
+  (myexpr) { echo 123; }
+}
+## status: 3
+## STDOUT:
+## END
--- a/spec/ysh-expr-compare.test.sh
+++ b/spec/ysh-expr-compare.test.sh
@@ -343,3 +343,30 @@ y=42
 x=42
 ## END
 
+#### Undefined comparisons
+shopt -s ysh:all
+
+func f() { true }
+var mydict = {}
+var myexpr = ^[123]
+
+var unimpl = [
+    / [a-z]+ /,  # Eggex
+    myexpr,  # Expr
+    ^(echo hello),  # Block
+    f,  # Func
+    mydict->keys,  # BoundFunc
+    # These cannot be constructed
+    # - Proc
+    # - Slice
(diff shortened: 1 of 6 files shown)


ysh/val_ops.py
```diff
@@ -406,7 +406,7 @@ def ExactlyEqual(left, right):
 
             return True
 
-    raise NotImplementedError(left)
+    raise error.TypeErr(left, "Cannot compare this type", loc.Missing)
```
andychu, line 409: Can we add an optional `loc_t` param  to ExactlyEqual(), and then pass `blame_loc=op` from `ysh/expr_eval.py` where it's called?

I think this will blame the `===` which is nicer

spec/ysh-expr-compare.test.sh
```diff
@@ -343,3 +343,30 @@ y=42
...
+    f,  # Func
+    mydict->keys,  # BoundFunc
+    # These cannot be constructed
+    # - Proc
+    # - Slice
+    # - Range
+]
+
+for val in (unimpl) {
+  try { _ val === val }
```
andychu, line 366: I changed `_` to `::`, but I didn't remove `_` yet

andychu (commented):
Thanks!

andychu:
OK cool it blames the `===` now

```
$ bin/ysh -c '= ^[1] === ^[1]'
  = ^[1] === ^[1]
         ^~~
[ -c flag ]:1: fatal: Cannot compare this type, got Expr
```

Though I think it should print both the type of left and right, I'm going to change it to do that, following some examples in ysh/expr_eval.py

I added `TypeErr` for the case where you expected one type, and got another

But there's also `TypeErrVerbose` which lets you construct a more detailed / arbitrary message

andychu:
Hm actually I tried this, and it made me realize that we CAN compare `^[42] === "hi"` as false

But we can't compare `^[42]` and `^[43]`


That seems like it could be a language design bug, let me think about that

andychu:
Brainstorming about that here - https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Equality.20on.20different.20types

But I think this PR is good now, we can consider that separately

andychu:
Merged, thanks!

PR #1755, opened 2023-11-02
[ysh] Improve error message when using setvar without var

PossiblyAShrub:
Previously:

```
$ cat example.ysh
proc foo() {
  read --line (&x)
  setvar x = x + "xyz"
  echo $x
}

foo
$ bin/ysh example.ysh
    setvar x = x + "xyz"
           ^
example.ysh:3: 'x' hasn't been declared
```

Now
```
$ cat example.ysh
proc foo() {
  read --line (&x)
  setvar x = x + "xyz"
  echo $x
}

foo
$ bin/ysh example.ysh
    setvar x = x + "xyz"
           ^
example.ysh:3: Declare 'var x' before using 'setvar'
```

### DIFF
--- a/osh/cmd_parse.py
+++ b/osh/cmd_parse.py
@@ -445,7 +445,7 @@ class VarChecker(object):
 
         if keyword_id == Id.KW_SetVar:
             if name not in top:
-                p_die("%r hasn't been declared" % name, name_tok)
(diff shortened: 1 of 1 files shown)


andychu:
Thanks!

Issue #609, opened 2020-02-23
Implement GLOBIGNORE


andychu:
Similar to #552 

Many usages shown here:

https://grep.app/search?q=GLOBIGNORE&case=true

https://news.ycombinator.com/item?id=22396824

andychu:
https://www.gnu.org/software/bash/manual/html_node/Filename-Expansion.html

```
The GLOBIGNORE shell variable may be used to restrict the set of file names matching a pattern. If GLOBIGNORE is set, each matching file name that also matches one of the patterns in GLOBIGNORE is removed from the list of matches. If the nocaseglob option is set, the matching against the patterns in GLOBIGNORE is performed without regard to case. The filenames . and .. are always ignored when GLOBIGNORE is set and not null. 
```

andychu:
We have many failing spec tests for this feature, and real shell scripts depend on it

PR #1757, opened 2023-11-18
[test/spec] Spec tests for GLOBIGNORE

ellen364:
Partially addresses #609

Cases generally use `echo *.*` because spec test setup create a subdirectory _tmp. That gets included when using `echo *` and would prove brittle if the test setup changes. The single character range case is a little awkward because it first removes the auto-generated _tmp dir.

### DIFF
--- a/spec/globignore.test.sh
+++ b/spec/globignore.test.sh
@@ -1,0 +1,159 @@
+
+## oils_failures_allowed: 14
+## compare_shells: bash
+
+#### Don't glob flags on file system with GLOBIGNORE
+# This is a bash-specific extension.
+expr $0 : '.*/osh$' >/dev/null && exit 99  # disabled until cd implemented
+touch _tmp/-n _tmp/zzzzz
+cd _tmp  # this fail in osh
+GLOBIGNORE=-*:zzzzz  # colon-separated pattern list
+echo -* hello zzzz?
+## stdout-json: "-* hello zzzz?\n"
+## N-I dash/mksh/ash stdout-json: "hello zzzzz"
+## status: 0
+
+#### basic star case -> ignore files with txt extension
+touch {basic.md,basic.txt}
+GLOBIGNORE=*.txt
+echo *.*
+## STDOUT:
+basic.md
+## END
+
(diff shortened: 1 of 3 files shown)


spec/globignore.test.sh
```diff
@@ -0,0 +1,159 @@
...
+touch {basic.md,basic.txt}
+GLOBIGNORE=*.txt
+echo *.*
+## STDOUT:
+basic.md
+## END
+
+#### basic question mark case -> ignore txt files with one char filename
+touch {1.txt,10.txt}
+GLOBIGNORE=?.txt
```
andychu, line 26: I think the `?.txt` and `*.txt` will not be separate in the code -- it all boils down to a glob() call?  But it's OK to have some redundant tests -- it can serve as documentation and help us understand the feature

andychu (commented):
Thank you!  I didn't know about globskipdots ... so many bash features :-P

Yesterday I put the `type -a` stuff on my TODO list on Zulip, so thanks for doing those tests as well!

test/spec.sh
```diff
@@ -290,10 +290,14 @@ sh-func() {
 
 glob() {
   # Note: can't pass because it assumes 'bin' exists, etc.
-  sh-spec spec/glob.test.sh --oils-failures-allowed 4 \
+  sh-spec spec/glob.test.sh --oils-failures-allowed 3 \
```
andychu, line 293: Oh also if you want something "boring" to do I would not mind if all these go inside the spec files

I added the `## oils_failures_allowed` lines a few months ago, and not all spec tests have been converted

Ideally I want to get rid of all these tiny functions -- that is a "refactoring in flight"

But it's not urgent, something to do in the background

andychu:
Oh also the `*.*` is fine, but `_tmp` is also sort of "legacy", because we used to NOT create a separate dir for each spec test case!

In theory we should be able to remove `_tmp`, but that's probably low priority.  It doesn't seem to be causing a huge problem

Maybe we can turn it off for SOME files, like have `## make_legacy_tmp_dir` at the top of each file ... then migrate away from it

PR #1761, opened 2023-12-02
Fix inverted proc/func naming convention

JNRowe:
Super trivial fix spotted while reading the documentation.

Let me take this moment to once again thank you for the excellent docs, I love reading them!  I *never* fail to learn something new about tooling I thought I knew well.

Thanks,

James


andychu:
Oh perfect, thank you !!

PR #1769, opened 2023-12-18
XXX [mycpp] add LinkedList

melvinw:
This commit adds LinkedList, which implements a simple doubly-linked list. This will be useful for building caches.

### DIFF
--- a/mycpp/gc_linked_list.h
+++ b/mycpp/gc_linked_list.h
@@ -1,0 +1,170 @@
+#ifndef MYCPP_GC_LINKED_LIST_H
+#define MYCPP_GC_LINKED_LIST_H
+
+#include "mycpp/common.h"  // DCHECK
+#include "mycpp/gc_obj.h"  // ObjHeader, maskbit
+
+template <typename T>
+class LinkedListNode {
+ public:
+  LinkedListNode() = delete;  // A list node with no object makes no sense.
+  LinkedListNode(T obj) : obj_(obj), prev_(nullptr), next_(nullptr) {
+  }
+
+  T obj() const {
+    return obj_;
+  }
+
+  LinkedListNode* prev() const {
+    return prev_;
+  }
+  LinkedListNode* next() const {
+    return next_;
+  }
+
+  // Replace the nodes predecessor with the given node.
+  void set_prev(LinkedListNode* node) {
+    prev_ = node;
+  }
+
+  // Replace the nodes sucessor with the given node.
+  void set_next(LinkedListNode* node) {
+    next_ = node;
+  }
+
+  static constexpr ObjHeader obj_header() {
+    return ObjHeader::ClassFixed(field_mask(), sizeof(LinkedListNode<T>));
+  }
+
+  static constexpr uint32_t field_mask() {
+    return maskbit(offsetof(LinkedListNode, obj_)) |
+           maskbit(offsetof(LinkedListNode, prev_)) |
+           maskbit(offsetof(LinkedListNode, next_));
+  }
+
+ private:
+  T obj_;
+  LinkedListNode* prev_;
+  LinkedListNode* next_;
+};
+
+template <typename T>
+class LinkedList {
+ public:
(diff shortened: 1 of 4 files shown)


andychu:
Hmm is there any reason to make it GC-managed?  I think the ownership is clear for a cache -- it's global and can be torn down at process exit

i.e. We could just have our own simple manually managed linked list.  

Oh OK well I guess a reason is if the entries themselves are GC objects.  

In this case the value is a `regex_t` which is NOT a GC object.  Actually that will complicate tracing.  And the key is a `Str`, but we could also have the cache own copies of the Str.

---

Tracing is still expensive on some workloads

http://travis-ci.oilshell.org/github-jobs/5644/raw-vm.wwz/_tmp/perf/osh-parse.report.txt

I would tend to go for the minimal thing that will solve regcomp() first ... i.e. whatever makes it fast

Not sure if it has to be integrated with the GC

andychu:
Oh yeah and remember we actually have no code hooks for deallocating ... So we have to figure out where to put the `regfree()`.   

I guess that's a main goal -- to prevent that potentially infinite memory leak (while not redoing regcomp() and regfree() every time unnecessarily)

The dealloc hook / finalizer is a pretty hard problem and I don't think we need it in general, since this is I think the first time its come up

mycpp/gc_linked_list.h
```diff
@@ -0,0 +1,170 @@
...
+  void set_next(LinkedListNode* node) {
+    next_ = node;
+  }
+
+  static constexpr ObjHeader obj_header() {
+    return ObjHeader::ClassFixed(field_mask(), sizeof(LinkedListNode<T>));
+  }
+
+  static constexpr uint32_t field_mask() {
+    return maskbit(offsetof(LinkedListNode, obj_)) |
```
andychu, line 40: This is tricky because whether this bit is set depends on whether T is a pointer or not

For Slab<T> and Tuple{2,3,4} I think we're using `std::is_pointer<T>` at compile time to conditionally set the mask!  That took a bit of figuring out ...

melvinw: Nice catch

melvinw:
> Hmm is there any reason to make it GC-managed? I think the ownership is clear for a cache -- it's global and can be torn down at process exit
> 
> i.e. We could just have our own simple manually managed linked list.
> 
> Oh OK well I guess a reason is if the entries themselves are GC objects.
> 
> In this case the value is a `regex_t` which is NOT a GC object. Actually that will complicate tracing. And the key is a `Str`, which in general IS a GC object

Oh, yeah. I guess this doesn't need to be managed by the GC (marking all the nodes could be wasteful...). I went with this since we needed `Str*` as keys. More commentary on this directly below.

> But we could also have the cache own copies of the Str as a `char*` or whatever

This probably isn't necessary. The cache type that uses this list will still have to have a `Dict<Str*, LinkedListNode<...>*>` anyway, right? It should be safe to just have the linked list use `Tuple2<Str*, regex_t>` since the lifetimes should be coupled. Does this make sense?

> Oh yeah and remember we actually have no code hooks for deallocating ... So we have to figure out where to put the regfree().

Oh, yeah, this could be messy. The clearest spot to add this is probably at evict-time. Two ideas below.

We could have `LRUCache::Set()` return the handle to the value of the evicted value if it decided to evict something. Then the caller could unpack and do the appropriate cleanup. I don't love this.

Maybe we could have the user provide a cleanup method in the constructor? We could have `LRUCache<K, V>::LRUCache(int capacity, void(V)* cleanup)`. Then whenever the cache decides to evict something, it calls the provided cleanup func on the victim

andychu:
Hm yeah not sure, I'd be tempted just to hard-code regfree() and see what it looks like

It doesn't have to be generic at first

I think the other  instances of LRU caches  are in Python code, and it seems like we're not sure how or if to unify them ...

melvinw:
Hmm ok. Let me ditch this pr and just send some specialized thing then

PR #1771, opened 2023-12-21
[docs] add -v expression to known differences

### LINKED ISSUE #1532: [[ -v ]] accepts an expression
from e-mail

```
$ declare -A foo
$ foo['bar']=2
$ [[ -v foo['bar'] ]] && echo exists
exists
$

Behaviour in OSH:
$ declare -A foo
$ foo['bar']=2
$ [[ -v foo['bar'] ]] && echo exists
$ 
```

I would like to know if this blocks a real script from running

Melkor333:
fixes #1532

edit: I was in a hurry, this obviously doesn't fix but document it :)


andychu:
Hm I added some failing spec  tests for this `[[ -v expr ]]` behavior ...

I wonder if we can support it, not sure how hard it is ...

andychu:
The annoying thing is I can't find this behavior documented

e.g. here

https://www.gnu.org/software/bash/manual/bash.html#index-_005b_005b

or here

https://www.gnu.org/software/bash/manual/bash.html#index-test

or here

```
-v varname

    True if the shell variable varname is set (has been assigned a value).
```

and also apparently there was a change in bash 5.2 around `[[ -v a[@] ]]`

andychu:
I tweaked the docs, added failing spec tests, and merged.


Thanks for pushing on this!

PR #1777, opened 2023-12-30
[builtins/dict] add => values() method

Melkor333:
C++ probably fails because I don't wrap the elements in a `value.X` but we'll see :)

### DIFF
--- a/builtin/method_dict.py
+++ b/builtin/method_dict.py
@@ -41,6 +41,5 @@ class Values(vm._Callable):
         dictionary = rd.PosDict()
         rd.Done()
 
-        keys = [k for k in dictionary.values()]  # type: List[value_t]
-        return value.List(keys)
-
+        values = dictionary.values()  # type: List[value_t]
+        return value.List(values)
--- a/builtin/method_dict.py
+++ b/builtin/method_dict.py
@@ -27,3 +27,20 @@ class Keys(vm._Callable):
 
         keys = [value.Str(k) for k in dictionary.keys()]  # type: List[value_t]
         return value.List(keys)
+
+
+class Values(vm._Callable):
+
+    def __init__(self):
+        # type: () -> None
+        pass
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+
+        dictionary = rd.PosDict()
+        rd.Done()
+
+        keys = [k for k in dictionary.values()]  # type: List[value_t]
+        return value.List(keys)
+
--- a/doc/ref/chap-type-method.md
+++ b/doc/ref/chap-type-method.md
@@ -141,17 +141,17 @@ rather than its value.
 Add an element to a list.
 
     var fruits = :|apple banana pear|
-    call fruits -> append("orange") # returns null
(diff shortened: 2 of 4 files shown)


Melkor333:
btw the `X` is already missing in the `doc/ref/toc-ysh.md`

builtin/method_dict.py
```diff
@@ -27,3 +27,20 @@ def Call(self, rd):
...
+        # type: () -> None
+        pass
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+
+        dictionary = rd.PosDict()
+        rd.Done()
+
+        keys = [k for k in dictionary.values()]  # type: List[value_t]
```
andychu, line 44: should be called `values` or `v` or `vals` :-)

andychu (commented):
Great, thank you!

How about filling in a one line description for keys() and values() in `doc/ref/chap-type-method.md` ?

And perhaps 2 examples like the ones in `start() end()` etc.


I've been doing that lately, we are slowly  creeping up :-)

Melkor333:
I hated the emptiness of that page so I added more examples…

Used `=` for the `value()` to show that it's not strictly strings (as it is with `keys()`)

doc/ref/chap-type-method.md
```diff
@@ -138,27 +138,77 @@ rather than its value.
 
 ### append()
 
+Add an element to a list.
+
+    var fruits = :|apple banana pear|
+    call fruits -> append("orange") # returns null
```
andychu, line 144: I think we want to encourage a style of no space for ->`, but spaces for `=>` to make it look more different?

```
call fruits->append('orange')
```

vs.

```
var v = d => values() => reversed()
```

Because `=>` can be chained and `->` can't.  `->` generally has no return value

What do you think?

builtin/method_dict.py
andychu: reminder :)

Oh also I think it should just be

```
vals = dictionary.values()
```

No need for list comprehension

In the keys() case we need the list comprehension, because we have to add the `value.Str()` wrapper

andychu (commented):
Thanks for adding the documentation!    much appreciated

andychu:
Oh just to clarify the types a bit, there are extra "wrappers" for YSH level vs. "mycpp" level:

Here's the type of a mycpp str:

    s = 'foo'  # type: str

and mycpp dict:

    d = {'foo': 42}   # type:  Dict[str, int]  

From `core/value.asdl`, here's are YSH wrappers

    ysh_str = value.Str('foo')  # type: value.Str

And then a mycpp Dict with a value.Str:

    var mycpp_dict = {'key': value.Str('foo')}  # type: Dict[str, value_t]

And then a YSH dict!

    var ysh_dict = value.Dict(mycpp_dict)  # type: value.Dict


So yeah these extra wrappers take a bit of getting used to!

Melkor333:
Thanks for your patience with me!
I hope the styling is now a bit better :)

andychu:
Thank you!  I'm going to put this on #help-wanted as another great example

PR #1778, opened 2024-01-02
Str => replace()

PossiblyAShrub:
## Changes

- Implemented `Str => replace()`
- Added documentation in `doc/ref`
- Fixed a bug where the mycpp `replace` would loop indefinitely if `old_len == 0`


doc/ref/toc-ysh.md
```diff
@@ -237,7 +237,7 @@ X [Builtin Sub]   _buffer
 
 ```chapter-links-type-method
   [Primitive] Bool   Int   Float   Str   Slice   Range   
-  [Str]       X find(eggex)   X replace(eggex, template)
+  [Str]       X find(eggex)   replace()
```
PossiblyAShrub, line 240: The `find(eggex)` breaks the link generation on `replace`. This is an issue in the `doctools/help_gen.py` `TOPIC_RE`. Fixing that is probably for another PR.

PossiblyAShrub:
This is now ready for a review. I also ended up fixing an issue where the mycpp implementation of `replace` would enter an infinite loop if `old_len == 0`. I am not sure if this was possible to trigger before `Str => replace ()`.

PossiblyAShrub:
> This is now ready for a review.

@andychu bump. Just in case you missed my message :)

andychu:
Oh sorry!!   I think I got a whole bunch of push notifications in my e-mail and yeah I missed it

I try to review within 24 hours, so definitely ping me if it's been more than that

andychu:
I cloned this and played with it -- looks very good!  I'll have some more comments tomorrow

builtin/method_str.py
```diff
@@ -130,3 +135,148 @@ def Call(self, rd):
...
+        self.mem = mem
+        self.expr_ev = expr_ev
+
+    def EvalSubstExpr(self, expr, blame_loc):
+        # type: (value.Expr, loc_t) -> str
+        res = self.expr_ev.EvalExpr(expr.e, blame_loc)
+        if res.tag() == value_e.Str:
+            return cast(value.Str, res).s
+
+        raise error.TypeErr(res, "expected expr to eval to a string",
```
andychu, line 153: How about writing this "expected Expr to evaluate to Str" to be consistent with how we spell the types in other cases?

builtin/method_str.py
```diff
@@ -130,3 +135,148 @@ def Call(self, rd):
...
+        eggex_val = None  # type: value.Eggex
+        subst_str = None  # type: value.Str
+        subst_expr = None  # type: value.Expr
+
+        pattern = rd.PosValue()
+        with tagswitch(pattern) as case:
+            if case(value_e.Eggex):
+                # HACK: mycpp will otherwise generate:
+                #  value::Eggex* eggex_val ...
+                eggex_val_ = cast(value.Eggex, pattern)
```
andychu, line 176: I think this is the reason for the awkward UP pattern, e.g. I think you can use `UP_pattern` and `UP_subst` to be consistent with the rest of the codebase


There is a Zulip thread about that, let me find this

builtin/method_str.py
```diff
@@ -130,3 +135,148 @@ def Call(self, rd):
...
+
+            else:
+                raise error.TypeErr(subst,
+                                    'expected substitution to be Str or Expr',
+                                    rd.LeftParenToken())
+
+        count = rd.NamedInt("count", -1)
+        rd.Done()
+
+        if count == 0:
```
andychu, line 206: I think we should check that count is not -2 or less ??   -1 means unset, and 0 to N is valid?

Weirdly Python does not seem to check this either

```
>>> p.sub('zz', '123 x 123',count=1)
'zz x 123'
>>> p.sub('zz', '123 x 123',count=-1)
'123 x 123'
>>> p.sub('zz', '123 x 123',count=-2)
'123 x 123'

```

It's sometimes possible to avoid these checks if the code "implicitly" handles it correctly, but I'm worried about some "accidental" behavior with negative numbers

Adding a test would help

builtin/method_str.py
```diff
@@ -130,3 +135,148 @@ def Call(self, rd):
...
+            return value.Str(string)
+
+        if string_val:
+            if subst_str:
+                s = subst_str.s
+            if subst_expr:
+                s = self.EvalSubstExpr(subst_expr, rd.LeftParenToken())
```
andychu, line 213: one of these is supposed to be true, so that means `assert s is not None`?

I guess there is a (2x2) matrix here that would be nice to simplify, though not sure it's possible with our idioms

builtin/method_str.py
```diff
@@ -130,3 +135,148 @@ def Call(self, rd):
...
+            return value.Str(result)
+
+        if eggex_val:
+            ere = regex_translate.AsPosixEre(eggex_val)
+            cflags = regex_translate.LibcFlags(eggex_val.canonical_flags)
+
+            pos = 0
+            parts = []  # type: List[str]
+            replace_count = 0
+            while pos < len(string):
```
andychu, line 226: Some comments about the algorithm here would help

I'm wondering if we can consolidate it with anything in `func_eggex`

Though we don't need to do that to start, since the tests look good here

doc/ref/chap-type-method.md
```diff
@@ -27,6 +27,25 @@ This chapter in the [Oils Reference](index.html) describes YSH types and methods
 
 ### replace()
 
+Replace substrings or eggex matches within a string.
+
+    var mystr = "code is 1234"
+    = mystr=>replace("code", "Code")  # => "Code is 1234"
```
andychu, line 33: FWIW I started writing this `mystr => replace(...)` with spaces

vs. `call mylist->pop()`

That is up for debate, but I think it makes them look more different and is a nice style

doc/ref/chap-type-method.md
```diff
@@ -27,6 +27,25 @@ This chapter in the [Oils Reference](index.html) describes YSH types and methods
...
+    var mystr = "year is 1988"
+    = mystr=>replace(/ <capture digit{4} as year : int> /, ^"$[year + 1]")
+    # => "year is 1989"
+
+To limit the number of replacements, pass in a named count argument. By default
+the count is `-1` which means "replace all".
+
+    var mystr = "bob has a friend named bob"
+    = mystr=>replace("bob", "Bob", count=1)   # => "Bob has a friend named bob"
+    = mystr=>replace("bob", "Bob", count=-1)  # => "Bob has a friend named Bob"
```
andychu, line 47: Weirdly Python says that the count must be non-negative, and 0 means replace ALL

not replace zero!

I'm wondering if they have any reason for that, or it's just that people think replacing zero is useless.  But I think it could be consistent for some algorithms, not sure.

---

https://docs.python.org/3/library/re.html#re.sub

> The optional argument count is the maximum number of pattern occurrences to be replaced; count must be a non-negative integer

doc/ref/toc-ysh.md
andychu: Yeah I can take a look at this

I noted on Zulip that I want to fix a few other things about how these are formatted

https://oilshell.zulipchat.com/#narrow/stream/202008-oil-documentation/topic/Re-format.20doc.2Fref.20TOC.20.20a.20bit

mycpp/gc_str.cc
```diff
@@ -286,6 +286,11 @@ BigStr* BigStr::rjust(int width, BigStr* fillchar) {
 }
 
 BigStr* BigStr::replace(BigStr* old, BigStr* new_str) {
+  // Use -1 as in python2: "aaaa".replace(-1) -> "AAAA"
```
andychu, line 289: Oh interesting, now I see that Python 2 replace() is inconsistent with Python re.sub !!   funny

0 means 0 occurences, so yeah that's a good justification

It also silently accepts any negative number, so I think that's fine for us as well.  I think we just make sure that -2 and -99 behave exactly the same way as -1.

mycpp/gc_str.cc
```diff
@@ -298,12 +303,20 @@ BigStr* BigStr::replace(BigStr* old, BigStr* new_str) {
 
   // First pass: Calculate number of replacements, and hence new length
   int replace_count = 0;
-  while (p_this <= last_possible) {
-    if (memcmp(p_this, old_data, old_len) == 0) {  // equal
-      replace_count++;
-      p_this += old_len;
-    } else {
-      p_this++;
+  if (old_len == 0) {
```
andychu, line 306: why do we need to special case old_len == 0 ?

OK this is a little weird and I honestly wonder if the code was correct

```
>>> 'xxx'.replace('', 'y', 3)
'yxyxyx'

```

I think we should check if there are unit tests in `gc_str_test.cc` that cover this

Whenever we add more if statements in the C++, I think we should have more C++ unit tests

because it's not obvious if the spec tests cover them

there could have been a bug here perhaps

andychu (commented):
This looks very good, thank you!

I added some cases in `test/ysh-runtime-errors` just to help me review, so you should pull that

And then I made a bunch of comments

I'm curious what happened with `old_len`

core/state.py
```diff
@@ -939,6 +939,23 @@ def __exit__(self, type, value, traceback):
         self.mem.PopCall(True)
 
 
+class ctx_Argv(object):
+    """For $1, $2, $3, etc."""
```
andychu, line 943: I just pushed a change noting that $0 isn't set, which is conventionally the whole match rather than a submatch


And Python does support the equivalent of it

It might be that we need to turn state.dollar0 into a stack ... Let me think if there's another way to implement it.  I guess we could also change GetArgNum() or something



```
>>> p=re.compile(r'\d+')
>>> 

>>> p.sub('yy', 'foo 123')
'foo yy'
>>> 
>>> p.sub(r'-\g<0>-', 'foo 123')
'foo -123-'
```

doc/ref/chap-type-method.md
```diff
@@ -27,6 +27,25 @@ This chapter in the [Oils Reference](index.html) describes YSH types and methods
...
+Replace substrings or eggex matches within a string.
+
+    var mystr = "code is 1234"
+    = mystr=>replace("code", "Code")  # => "Code is 1234"
+    = mystr=>replace(/ d+ /, "<redacted>")  # => "code is <redacted>"
+
+Replacement expressions can use captures from the given eggex.
+
+    var mystr = "year is 1988"
+    = mystr=>replace(/ <capture digit{4} as year : int> /, ^"$[year + 1]")
```
andychu, line 39: This should also document positional captures like $1 $2

and $0 when we implement it

test/ysh-runtime-errors.sh
```diff
@@ -854,6 +854,45 @@ test-method-type-errors() {
...
+     _should-run '= "foo" => replace("o", "x", count=-2)'
+   fi
+   # Replace empty string?  Weird Python behavior
+   _should-run '= "foo" => replace("", "-")'
+   _should-run '= "foo" => replace("", "-", count=2)'
+
+   # Use Expr with string
+   _should-run '= "foo" => replace("o", ^"-")'
+   # $0 is regular $0 here
+   _should-run '= "foo" => replace("o", ^"-$0")'
```
andychu, line 880: I just pushed this change noting the difference between

```
replace("o", Expr)  # string to expr
replace(/[o]/, Expr)  # eggex to expr
```

I think there is an argument for disallowing the former, since it probably won't support  `$0`  ??

That is, we could have

```
replace(Str, Str)
replace(Eggex, Str)
replace(Eggex, Expr)
```

but not

```
replace(Str, Expr)
```

Either that or support $0 in the last case.  It doesn't seem to add much power ?

andychu (commented):
I noticed a few things and added some more comments

I think it's nice that we are making replacement consistent, whereas Python re.sub and ''.replace() aren't consistent



Also I want to test this  "empty string replacement" semantic -- honestly what Python does doesn't quite make sense to me, I wonder what JS and other languages do

mycpp/gc_str.cc
PossiblyAShrub: I tried altering the previous string replacement algorithm to support `old_len == 0`, however it ended up so full of special cases that I decided to handle `old_len` separately.

Previously the code would not terminate on an empty string. I opted to follow the python2 behaviour here. However, the `Str => replace()` in YSH could behave differently. Did you have any thoughts there?

Re unit tests: sure, I can write some. (By the way, I did sneak in some spec tests which cover this behaviour. See the ones with empty strings and/or counts.)

andychu: Oh I didn't realize you fixed a bug!  That's important, we should note it in the "change description"

Since I only look at the  whole diff, it's usually OK to squash and then write a PR description

That will help review the changes

---

OK I didn't see the spec tests, but I don't think it's a bad idea to simply copy them in to `gc_str_test.cc`, since it will help us debug ...

I'm still kinda thinking about empty string, even though it is consistent between JS and Python

andychu: BTW I use `devtools/git.sh rebase-staging`, although the command it runs is actually shorter :-P

https://github.com/oilshell/oil/blob/master/devtools/git.sh#L30

It's not essential to do that every time, because when I merge PRs I squash

But it can be useful to consolidate things under a single big change description.  This PR can use that, if there is both a bug fix and a new feature

core/state.py
andychu: I was thinking about this state.dollar0 issue -- I was hesistant to **couple** the unrelated regex and the `argv[0]` semantic

It's kind of mixing two different things up

I think we might not need a stack -- we can just create a separate variable, and then GetArgNum(0) can just test if there's an "active" regex match?

Just a thought

(BTW `state.GetArg0()` does not need to be changed, that's only for word eval slicing)

PossiblyAShrub:
I've addressed your feedback, this is ready for another look.

I decided to implement $0 for both `Str => replace(Str, Expr)` and `Str => replace(Eggex, Expr)`. Even though the former isn't very powerful, being able to write the shorter `mystr => replace("some long string", ^"'$0'")` is nice.

builtin/method_str.py
```diff
@@ -130,3 +136,196 @@ def Call(self, rd):
...
+            cflags = regex_translate.LibcFlags(eggex_val.canonical_flags)
+
+            # Walk through the string finding all matches of the compiled ere.
+            # Then, collect unmatched substrings and substitutions into the
+            # `parts` list.
+            pos = 0
+            parts = []  # type: List[str]
+            replace_count = 0
+            while pos < len(string):
+                indices = libc.regex_search(ere, cflags, string, 0, pos)
```
andychu, line 268: I'm not sure eflags (execution flags) should always be 0 ...  In `builtin/method_str.py` there is some logic around `REG_NOTBOL`, where it's only 0 when `pos == 0` when you're searching multiple times

This affects whether `^` can be found in the middle of the string, which would be a bug

It's a bit of a weird API but it kinda makes sense and I think other regex engines have it

Let me see if I can come up with a test case

core/state.py
```diff
@@ -1366,6 +1366,10 @@ def GetArg0(self):
     def GetArgNum(self, arg_num):
         # type: (int) -> value_t
         if arg_num == 0:
+            # $0 may be overriden, eg. by Str => replace()
+            if "0" in self.var_stack[-1] and (
```
andychu, line 1370: I think you could have a variable `regex_match = self.var_stack[-1].get('0')` and then this subexpression would only appear once,

I think the code would be slightly shorter

doc/ref/chap-type-method.md
```diff
@@ -27,6 +27,56 @@ This chapter in the [Oils Reference](index.html) describes YSH types and methods
...
+    # => "Year: 1989, Month: 06, Day: 08"
+
+    var mystr2 = "year is 1988"
+    = mystr2 => replace(/ <capture digit{4} as year : int> /, ^"$[year + 1]")
+    # => "year is 1989"
+
+In some cases it might be useful to refer to the entire capture itself in a
+substitution string. This can be achieved with `$0`.
+
+    var mystr = "replace with mystr=>replace()"
```
andychu, line 52: add spaces around => ?

doc/ref/chap-type-method.md
```diff
@@ -27,6 +27,56 @@ This chapter in the [Oils Reference](index.html) describes YSH types and methods
...
 ### replace()
 
+Replace substrings or eggex matches within a string.
+
+    var mystr = "code is 1234"
+    = mystr => replace("code", "Code")  # => "Code is 1234"
+    = mystr => replace(/ d+ /, "<redacted>")  # => "code is <redacted>"
+
+Replacement expressions can use captures from the given eggex. Captured values
+can be read with `$1`, `$2`, etc. or, for named captures, with the names given
```
andychu, line 37: I think it's more readable if you structure it like

... Refer to captured values by number ...

<example>

... Or by name ...

<example>

That makes the blocks of code shorter and easier to digest

---

Actually I would even do that with

Replace substring

<example>

or Eggex

<example>

i.e. so the examples are just 1 line

andychu (commented):
Looks great!   Thank you


I think there might be an issue with `^` -- that requires a special case eflags

Let me change the survey to see what Python and JS do, I don't think I tested that case

andychu:
OK I added some Python and JS demos here with `^` for beginning of line/string - https://github.com/oilshell/oil/commit/a821fc257dc46a60a5056815ea5ea99b629bcd78

Although actually now I realize sometimes `^` means beginning of line, and sometimes it means beginning of string, and I should probably test that too !!

(Python has a mode for this, not sure about JS)

So we should have another spec test that is exactly like those

My suspicion is that it will fail unless REG_NOTBOL is set whenever pos is not  zero

builtin/method_str.py
andychu: Hm I guess I was wrong about eflags=0 causing this bug?  I saw you added the test, but it already passes without changing the code?

That surprises me but OK, glad it works :)  I can look into it separately ... curious what exactly would fail

andychu: Hm OK it looks like it is fine and taken care of by `cflags`

`eflags` seems to be only for `leftMatch()` vs. `search()`, I got confused

But at least we have that test case now!

Looks like this is ready to merge then?

PossiblyAShrub:
> Looks like this is ready to merge then?

Yep, this is ready to be merged.

andychu:
Merged, thank you!  That was probably the trickiest method implemented yet :-)

I'm glad to have this in the Oils 0.20.0 release, which means we have a very complex and powerful Eggex API!  

It would be nice to write a blog post about how it's better on many fronts ...

Although I still wonder about exporting vars directly in `if (s ~ pat)` rather than going through `_group()`

...

Also I have an idea for the flag parsing -- a low level `ctx` builtin, and then everything else in YSH.  It will have to wait until after 0.20.0, but we can discuss it if interested.  It has some similarity to the context managers we're using in many places, including in this PR

PR #1779, opened 2024-01-03
[ysh] Make ^"..." syntax sugar for ^["..."]

PossiblyAShrub:
I decided to implement this in a PR separate from https://github.com/oilshell/oil/pull/1778.

### DIFF
--- a/ysh/expr_to_ast.py
+++ b/ysh/expr_to_ast.py
@@ -691,7 +691,11 @@ class Transformer(object):
                 return cast(BracedVarSub, pnode.GetChild(1).tok)
 
             elif typ == grammar_nt.dq_string:
-                return cast(DoubleQuoted, pnode.GetChild(1).tok)
+                s = cast(DoubleQuoted, pnode.GetChild(1).tok)
+                # sugar: ^"..." is short for ^["..."]
+                if pnode.GetChild(0).tok.id == Id.Left_CaretDoubleQuote:
+                    return expr.Literal(s)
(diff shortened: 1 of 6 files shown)


ysh/expr_to_ast.py
```diff
@@ -691,7 +691,11 @@ def Expr(self, pnode):
                 return cast(BracedVarSub, pnode.GetChild(1).tok)
 
             elif typ == grammar_nt.dq_string:
-                return cast(DoubleQuoted, pnode.GetChild(1).tok)
+                s = cast(DoubleQuoted, pnode.GetChild(1).tok)
+                # sugar: ^"..." is short for ^["..."]
+                if pnode.GetChild(0).tok.id == Id.Left_CaretDoubleQuote:
+                    return expr.Literal(s)
```
andychu, line 697: It might also be possible to change the evaluator, i.e. `ysh/expr_eval.py`

If a `DoubleQuoted` node has  an `Id.Left_CaretDoubleQuote`, then it could "stop" evaluating it

It may depend if there are some "LST invariants" we would to preserve, but we can think about that later I guess

This has a good test, which helps any future refactoring!

andychu (commented):
Looks great, thanks!

andychu:
Hm the CI has been flaky lately, so I restarted it in the Github Actions UI

I believe this is due to us using SSH to publish the HTML.  For some reason it seems to fail quite often

Probably should change to HTTP

PR #1798, opened 2024-01-17
fix README.md

AtomsForPeace:
I got this working on my machine with this small change.


andychu:
Oh excellent thanks!  Yes I changed the default target to ASAN a couple months ago (which is very useful)

Glad you got it working!

Issue #1767, opened 2023-12-12
echo builtin should disallow typed args


andychu:
From #1762 

```
andy@hoover ~/git/oilshell/oil (dev-hoover-9)$ bin/ysh -c 'ls (x)'
  ls (x)
      ^
[ -c flag ]:1: fatal: Undefined variable 'x'
andy@hoover ~/git/oilshell/oil (dev-hoover-9)$ bin/ysh -c 'echo (x)'
  echo (x)
        ^
[ -c flag ]:1: fatal: Undefined variable 'x'
```

bar-g:
Hm, I then noticed some amount of special parsing may be present in bash's echo as well:

```
$ echo ()
>
```
(PS2 Prompt)

and `echo (x)` returns unexpected symbol »x«.

(Whereas both `ls ()` and `ls (x)` give syntax errors in bash.)

No idea if that's a bash feature or a bug with no need to mimic.

andychu:
I believe this is because `echo ()` is starting a shell function named `echo`!

I was confused as why `ls ()` gives a syntax error, but I believe it's because I have an alias `ls` in my bashrc

After alias expansion it's a syntax error

`find ()` behaves like `echo ()`

bar-g:
> I believe this is because echo () is starting a shell function named echo!

Ah, ok, indeed:
```
$ echo ()
> { true; }
$ declare -f echo
echo () 
{ 
    true
}
$
```

Looks like it's the same in osh and ysh and safe from variable expansion:
```
$ x="() { true; }"
$ echo $x
() { true; }
$ declare -f echo
$
```
(Unlike with x='-n' or '-e'.)

----
So it seems this is yet another general echo and write pitfall. But one that can be worked around using quotes to actually successfully echo "( anything in parenthesis )".

Maybe osh and ysh can work around it in non-interactive or not is-main modes?

----


> I was confused as why ls () gives a syntax error, but I believe it's because I have an alias ls in my bashrc

You seem correct. I can confirm, I had, too, and after `unalias ls` I also see the same interactive PS2  prompt entering behavior here for echo, ls, and find (in bash, osh and ysh).

Whereas I wonder why your initial ls and echo example show no difference, while you were saying it's special to echo.

> https://github.com/oilshell/oil/issues/1762#issuecomment-1852831193
> (The reason [for this issue] is that echo uses a special flag parser due to lack of --, and this accidentally made it accept (x))

[...]

PR #1799, opened 2024-01-18
add breaking test for #1767

AtomsForPeace:
How does this look? 
I added an expected failure but the osh case doesn't fail.
I've just tried the check the status codes of the commands. 

I assume this is because the current failure also returns a status code of 2. 
Is there another way I could test this then? Maybe the error text that's returned?
Is there a way to see a diff in the expected vs actual result of one of these tests?

### DIFF
--- a/spec/builtin-io.test.sh
+++ b/spec/builtin-io.test.sh
@@ -48,9 +48,8 @@ echo -e "\\"
 
 #### echo builtin should disallow typed args
 echo (42)
-## STDOUT:
-status=2
-## END
+## status: 2
+## OK mksh/zsh status: 1
 
 #### echo -en
 echo -en 'abc\ndef\n'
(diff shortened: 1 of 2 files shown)


andychu:
Oh I see this is a difference between OSH and YSH!  I didn't realize that about the bug


```
$ bin/osh -c 'echo (42)'
  echo (42)
       ^
[ -c flag ]:1: 'echo' got unexpected typed args
andy@hoover:~/git/oilshell/oil$ bin/ysh -c 'echo (42)'

```

In that case, I suggest  keeping this test, and then adding another one in one of the  `spec/ysh-*` files

I guess it could go in `spec/ysh-builtins`

andychu:
Oh also I just sent you an invite to work on `oilshell/oil` directly

Sorry there is an issue where we can't run our CI from `AtomsForPeace/oil`, even if I give you permissions.  It's a github actions secrets issue

I'll make a note of this on Zulip, we might be able to change the auth on our CI server (?)

AtomsForPeace:
So I'm a bit confused with what the tests are showing.
echo (x) gives me an undefined variable 'x' error as expected so that one's fine
echo (42) returns a blank line in the interactive shell but the test says it returns a status of 2 which I don't understand.

AtomsForPeace:
Closing this in favour of #1804

Issue #576, opened 2020-01-15
implement $SECONDS


andychu:
This should be pretty straightforward as it's just exposing a timestamp I think

In #572 the `sbp` prompt uses it

AtomsForPeace:
Closed with #1801

PR #1801, opened 2024-01-19
add failing spec test.  #576

AtomsForPeace:
Is this added in the correct file?

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -1969,6 +1969,10 @@ class Mem(object):
         if name == '_':
             return value.Str(self.last_arg)
 
+        if name == 'SECONDS':
+            seconds = self.GetSeconds()
+            return value.Int(seconds)
+
         # In the case 'declare -n ref='a[42]', the result won't be a cell.  Idea to
         # fix this:
         # 1. Call self.unsafe_arith.ParseVarRef() -> BracedVarSub
--- a/core/state.py
+++ b/core/state.py
@@ -396,19 +396,6 @@ def _SetOptionNum(opt_name):
     return opt_num
 
 
-class SecondsState(object):
-
-    def __init__(self):
-        # type: () -> None
-        self.start = int(time_.time())
-        self.latest = self.start
-
-    def GetSeconds(self):
-        # type: () -> int
-        self.latest = int(time_.time())
-        return self.latest - self.start
-
-
 class MutableOpts(object):
 
     def __init__(self, mem, opt0_array, opt_stacks, opt_hook):
@@ -1069,7 +1056,7 @@ class Mem(object):
         self.debug_stack = debug_stack
 
         self.pwd = None  # type: Optional[str]
-        self.seconds_state = None  # type: Optional[SecondsState]
+        self.seconds_start = int(time_.time())  # int
 
         self.token_for_line = None  # type: Optional[Token]
         self.loc_for_expr = loc.Missing  # type: loc_t
@@ -1117,10 +1104,7 @@ class Mem(object):
 
     def GetSeconds(self):
         # type: () -> int
-        if self.seconds_state:
-            return self.seconds_state.GetSeconds()
-        self.seconds_state = SecondsState()
-        return self.seconds_state.GetSeconds()
+        return int(time_.time()) - self.seconds_start
 
     def ParsingChangesAllowed(self):
         # type: () -> bool
--- a/core/state.py
+++ b/core/state.py
@@ -8,7 +8,7 @@
 state.py - Interpreter state
 """
 from __future__ import print_function
-from time import time
+import time as time_  # avoid name conflict
 
 from _devbuild.gen.id_kind_asdl import Id
 from _devbuild.gen.option_asdl import option_i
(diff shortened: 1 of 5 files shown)


andychu:
The exact file is kind of a judgement call

Although it looks like this file is named for the `times` builtin specifically

I think `spec/vars-special` is a bit better, that includes $PWD and $HOME, so $SECONDS makes sense

Thanks for adding this!

andychu:
This is looking good

You may have already seen it, but the  HTML spec test report is here - http://travis-ci.oilshell.org/github-jobs/5915/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

The way I find that is to go to 

And then search for the PR number, and click "All Jobs and Tasks", and then `ovm-tarball`, and then  "Results"

Or you can click directly on the failing `ovm-tarball` link on the front page

So it looks like everything is working as intended

andychu:
Also we sometimes use this idiom to disable say dash, because it doesn't support `$SECONDS`

```
$ fgrep 'case $SH' spec/* |head

spec/array-compat.test.sh:case $SH in mksh) exit ;; esac
spec/array-compat.test.sh:case $SH in mksh) exit ;; esac
spec/assign-extended.test.sh:case $SH in (mksh) exit 0; esac
spec/assign.test.sh:case $SH in *dash) exit 99 ;; esac  # stderr unpredictable

```

So we are just testing if `$SH` is a certain shell, and then aborting the spec test

It's also OK to just let dash run, and then write a different assertion for it

There are `OK` and `N-I` tags to mark that different output is OK

AtomsForPeace:
Ok, thanks. Now I've excluded dash.
I'm trying to figure out where to put the logic now. 
At first look a SecondsState class in core.state.py seems to make sense.
I can't import time it seems. Do I need to access the time module some other way or expose it somehow?

andychu:
Hm we are importing time in a few places

```
$ grep 'import time' */*.py
benchmarks/time_.py:import time
builtin/printf_osh.py:import time as time_  # avoid name conflict
client/headless_demo.py:    import time
core/completion.py:import time as time_
core/main_loop.py:        import time
core/pyos.py:import time
core/shell.py:import time
```

What's the error when you import it?


BTW it's typical to make it pass in Python first ... the C++ can be done as a separate step, though sometimes it just works

AtomsForPeace:
Aha. So what do I run to run the tests?
I tried test/unit.sh all and they all pass.

What I was doing earlier was running: 
test/stateful.sh soil-run

The error I got was something like this:
error: ‘time’ was not declared in this scope; did you mean ‘time_::time’?

I guess that was just the wrong tests though.

andychu:
Oh I would just run the spec tests, like

```
test/spec.sh vars-special
```

and

```
test/spec.sh vars-special -r 5 -v
```

This page has useful commands:

AtomsForPeace:
Yeah, I was doing that when writing the test but I wanted to add a unit test for the python code.

andychu:
Oh in that case `test/unit.sh all` is right, and I add the tests next in `foo_test.py` next to `foo.py`

There are also commands to run one test -- `test/unit.sh unit` which basically just sets PYTHONPATH

If the spec tests have good coverage, unit tests aren't 100% necessary.  But if I have something tricky I definitely use unit tests

--

Also I see the compile errors in the CI now -- that should be easy to fix by changing it to `time_.time` like in the other places

The way we translate to C++ is very particular -- it can only handle the idioms we use in the code :)  But it works well because the C++ compiler is strict

AtomsForPeace:
Are the CI failures related to my changes?

AtomsForPeace:
Are the CI failures related to my changes?

andychu:
Yes I think they're related due to the `time_.time` thing I mentioned

But if you get the tests passing for Python alone, that's enough

If you look at the github actions logs, they will have a link to this site where we publish:

So I generally go directly there and search to see the logs, or you can follow the links

---

But yeah getting it work in Python is the first step

Although in this case if you follow the import style of `time_.time` elsewhere in the codebase, I think it should work

AtomsForPeace:
Thanks for the hint!

core/state.py
```diff
@@ -1100,6 +1115,13 @@ def SetPwd(self, pwd):
         """Used by builtins."""
         self.pwd = pwd
 
+    def GetSeconds(self):
+        # type: () -> int
+        if self.seconds_state:
+            return self.seconds_state.GetSeconds()
```
andychu, line 1121: BTW it feels like this will compute the number of seconds since the FIRST time you invoked `$SECONDS`?

But really it should be since the startup time of bash?

So I think it's OK to eagerly initialize seconds_state in the constructor, which is called in main() I think


```
$ bash -c 'sleep 1; echo $SECONDS'
1
$ bash -c 'sleep 2; echo $SECONDS'
2
```

BTW we could put an "ad hoc" test in `demo/bash-seconds.sh` or something

These are NOT automated, but it's better than nothing.  I don't like to have tests that "sleep" in our automated tests, but in this case it seems hard to test in any other way

The spec test are still useful for the `0` case where you don't sleep, but then we can have another sanity check

AtomsForPeace:
Hey, think it might be ready for another look now

andychu:
Thank you!  great work :-)

andychu:
BTW I tweaked the tests a little bit

https://github.com/oilshell/oil/commit/064800ac5c0de923a42c79d46d222e575c960507

Asserting stdout is helpful since it's more "direct"

The shell conditions are still non-obvious to read for me :)   I didn't realize mksh and zsh implement $SECONDS, but apparently so!

PR #1804, opened 2024-01-20
 echo builtin should disallow typed args #1767

AtomsForPeace:
Switched from my fork to this repo.

I'm a bit confused with what the tests are showing. I've added two to the ysh-builtins.
echo (x) gives me an undefined variable 'x' error as expected so that one's fine
echo (42) returns a blank line in the interactive shell but the test says it returns a status of 2 which I don't understand.

Whereas osh seems to work for echo (42) but gives the incorrect error for echo (x).


andychu:
Oh I think the issue is that (perhaps confusingly) the default shell in the `ysh-builtins` file is still `osh`!  That is kind of historical

It has to be overridden with `## our_shell: ysh`, but that would break tests in that file probably

So if I grep, then `spec/ysh-bugs` already has `our_shell: ysh`, so maybe move it there?

I'll think about if we need to change anything here


```
$ grep our_shell spec/*
grep: spec/bin: Is a directory
spec/README.md:    ## our_shell: osh               # shell we run with
grep: spec/stateful: Is a directory
grep: spec/testdata: Is a directory
spec/ysh-assign.test.sh:## our_shell: ysh
spec/ysh-bin.test.sh:## our_shell: ysh
spec/ysh-bugs.test.sh:## our_shell: ysh
spec/ysh-builtin-error.test.sh:## our_shell: ysh
```







```
$ test/spec.sh ysh-builtins -r 18-19 -v -d
        oils_failures_allowed: 6

ysh-builtins.test.sh
case    line    osh
 18     246     pass    echo builtin should disallow typed args - literal
case: 18
osh stdout:

osh stderr:
  echo (42)
       ^
[ stdin ]:1: 'echo' got unexpected typed args


 19     250     FAIL    echo builtin should disallow typed args - variable
case: 19
[osh status] Expected 2, got 1
osh stdout:

osh stderr:
  echo (x)
        ^
[ stdin ]:1: fatal: Undefined variable 'x'



                osh
        pass    1
        FAIL    1
        total   2

ysh-builtins: FATAL: Got 1 failures (1 oils failures), but 6 are allowed

```

AtomsForPeace:
Note: 
osh error for echo (x):
```
fatal: Undefined variable **'x'**
```
Desired error:
```
'echo' got unexpected typed args
```

The error for echo (x) is thrown during the parsing in ysh/expr_eval.py. echo (42) throws an error later on down the line. The error for echo (x) is thrown here because it looks up the variable before the error that echo (42) throws is reached. 

If you define x and then run echo (x) you get the same error as echo (42). i.e.
```
x=1
echo (x)
```

AtomsForPeace:
@andychu So it seems that in ysh simple_echo is set to true where as it's false in osh. Any idea where that's set or why that might be the case?

andychu:
Oh I see!  So we have this `shopt --set ysh:upgrade` or `ysh:all` mechanism to turn OSH into YSH

https://www.oilshell.org/release/0.19.0/doc/options.html

And `simple_echo` is part of that.  


So basically `simple_echo` is off in OSH, and on in YSH.  And it doesn't check for typed args!  That makes sense


---

OK so  this `_SimpleFlag` function is a little hacky, but I don't think we touch it

I think all we need is a check for `cmd_val.typed_args` in the `simple_echo` case

The schema is in `core/runtime.asdl`

If ASDL is confusing let's chat on Zulip about it, but basically it just turns into a normal Python class with a `typed_args` attribute, that could be `None` or something else

If you print it out in the `simple_echo` case that should give a clue

And then we usually `raise error.Usage` to show an error, you can see that in the other builtins

---

Thanks for looking into this!

builtin/io_osh.py
```diff
@@ -10,6 +10,7 @@
 from frontend import flag_spec
 from frontend import lexer
 from frontend import match
+from frontend.typed_args import DoesNotAccept
```
andychu, line 13: Our C++ translation (mycpp) usually requires `from frontend import typed_args`

there are a couple exceptions but that's the general rule

andychu (commented):
This makes sense! 

Looks like github says there is a conflict.  I think if you fix that then the CI will run

If it passes locally and passes in the CI that's good

---

The duplicate errors is a separate issue  .. let me try to find a Zulip thread about that

It's mostly cosmetic

andychu:
Thanks, I noticed a merge issue that caused the test failures, and pushed a fix

The issue with 

```
echo (42)  # status 2
echo (x)  # status 1, not status 2
```

is that arguments get evaluated **before** the builtin `echo` is called.

So `x` is an  undefined var, giving status 1, which means we never get to echo.

If it is defined as 42, then 42 is a typed arg passed to  echo, and it fails now.

---

Thanks for getting to the bottom of this!  The tests look good and that's often the way I figure things out.  I didn't realize this bug was YSH only at first

andychu:
Oh I had to adjust the ysh-builtins test cases to say `shopt -s ysh:all`, because they run with osh not ysh

That is what I was saying above about `## our_shell: ysh`, or moving to `spec/ysh-bugs` which already has `## our_shell: ysh`

https://github.com/oilshell/oil/commit/3000eac44d7acf565cd607c648cb5a22f6e2772f

Either way the code was right, but the tests weren't quite tickling it.

And I added `var x = 43` so the error we're testing isn't masked by the undefined variable

Thanks!

PR #1814, opened 2024-01-29
feat: check if pwent exists

### LINKED ISSUE #1774: Detect {get,set}pwent() functions at configure time to build on Android Bionic
Weird and quite irrelevant:

For fun I tried to build the official oils 0.19.0 tar file on [termux](https://termux.dev/en/) with the instructions from the `README-native.txt`.

It fails with the following error:
```
~/.../oil/oils-for-unix-0.19.0 $ _build/oils.sh
...
CXX _gen/osh/arith_parse.cc
CXX _gen/ysh/grammar_tables.cc
CXX cpp/core.cc
cpp/core.cc:150:3: error: use of undeclared identifier 'setpwent'
  setpwent();
  ^
cpp/core.cc:153:13: error: use of undeclared identifier 'getpwent'
    entry = getpwent();
            ^
2 errors generated.
```

This is due to the android libc alternative bionic [not having the pwent functions](https://en.m.wikipedia.org/wiki/Bionic_(software)#Differences_from_POSIX).

The fix is very easy: comment out all 3 `...pwent` function calls. 🎉
Given I only got the python build working inside a [ubuntu proot](https://github.com/termux/proot-distro) I consider this a great success story!!

AFAIU `complete -A user ...` is the only thing requiring this code block and it just gives an empty result with the functions commented.

Now I'm not sure if/how that could be made to work with something like `#ifdef` or if an eventual termux build should just use a `patchfile` anyway (given the requirement of `libc` isn't met).

Melkor333:
Fixes #1774 by checking if `*pwent` functions exist.


Melkor333:
Tested compilation on my Laptop and my phone, it works 🥳 
```
$ # Android / manually removed the generated `#define` on my laptop:
$ result/bin/ysh -c 'compgen -A user sam'
  compgen -A user sam
  ^~~~~~~
[ -c flag ]:1: errexit PID 201633: command.Simple failed with status 1

$ # Normal compilation on my laptop still works
$ ysh -c 'compgen -A user sam'
samuelh 
```

Is this OK or should there be a helpful error like `compiled without gepwent support. user completion not possible`?

build/detect-pwent.c
```diff
@@ -0,0 +1,14 @@
+#include <stdio.h>  /* readline needs this, issue #21 */
+#include <pwd.h>   // passwd
+
+static int test_event_hook(void) {
```
andychu, line 4: This looks great, can remove all the readline stuff

cpp/core.cc
```diff
@@ -164,6 +165,9 @@ List<PasswdEntry*>* GetAllUsers() {
   endpwent();
 
   return ret;
+  #else
+  throw Alloc<RuntimeError>(StrFromC("Compiled without gepwent support. user completion not possible"));
```
andychu, line 169: I think this might cause a crash?

Although it's a little cheesy, I think it could be OK to return the empty list and print a warning to stderr about HAVE_PWENT?

So it won't crash but there will be some spam

I'm open to other ideas

andychu (commented):
Looks great, thank you!

Melkor333:
@andychu I don't know how to fix the checks. I assume they don't run `configure` beforehand and use some predefined headers but I don't know where to put the `#define` statement...

andychu:
OK it was just a unit test failing!  I disabled the test when \*pwent() are not available.

----

The commits didn't cause any mails, so feel free to push as much as you want

The contributor build is all automated on Debian and Ubuntu now

http://travis-ci.oilshell.org/github-jobs/6088/#job-dev-setup-debian

(One last issue, I want to remove stuff from `/wedge` and put it all in `~/wedge` so you don't need root.  However that's not essential now, and will take awhile)

Let's talk about it on Zulip ... I saw the Nix thread

andychu:
This just landed on `master`!  Definitely give the tarball a whirl and let us know on Zulip if it works 

http://travis-ci.oilshell.org/github-jobs/6089/

I'm excited for this :-) :-)

---

BTW for the 0.0001% of people who care about the obsolete `expr` command on Android, that's my code :)  It's  probably my most deployed code by orders of magnitude

I think most people replace it with `busybox` though!

That is, I rewrote `expr` for  Rob Landley's toybox project.  Android is the major user of toybox, somewhat for licensing reasons.  That's what I worked on right before Oils!

PR #1817, opened 2024-02-01
Document Eggex Flags

PossiblyAShrub:
The [current docs](https://www.oilshell.org/release/0.19.0/doc/eggex.html#flags-and-translation-preferences) do not cover any of the [currently implemented eggex flags](https://github.com/oilshell/oil/blob/139146035a4f76a9fbaf88eb9ee61e3759c84be5/ysh/regex_translate.py#L371). I have added documentation for those flags:

- `i`/`reg_icase` [REG_ICASE](https://www.gnu.org/software/libc/manual/html_node/Flags-for-POSIX-Regexps.html)
- `reg_newline` [REG_NEWLINE](https://www.gnu.org/software/libc/manual/html_node/Flags-for-POSIX-Regexps.html)

There were no tests for `reg_newline`, so I added some in `spec/ysh-reggex-api`.

### DIFF
--- a/doc/eggex.md
+++ b/doc/eggex.md
@@ -311,6 +311,21 @@ Flags and translation preferences together:
 
     / digit+ ; ignorecase ; python /  # could translate to (?i)\d+
 
+The following flags are currently supported:
+
+#### `i` / `reg_icase` (Ignore Case)
+
+**With** this flag, the match _is not_ case sensitive.
+
+**Without** this flag, the match _is_ case sensitive.
+
+#### `reg_newline` (Multiline)
+
+**With** this flag, `$` will match before a newline and `^` will match after a
+newline. Newlines are also ignored in `dot` and `!...` patterns.
+
+**Without** this flag, the newline `\n` is treated as an ordinary character.
+
 ### Multiline Syntax
 
 You can spread regexes over multiple lines and add comments:
--- a/doc/eggex.md
+++ b/doc/eggex.md
@@ -320,17 +320,17 @@ Use this flag to ignore case when matching. For example, `/'foo'; i/` matches
 
 #### `reg_newline` (Multiline)
 
-With this flag, `%end`/`$` will match before a newline and `%start`/`^` will
-match after a newline.
+With this flag, `%end` will match before a newline and `%start` will match
+after a newline.
 
     u'abc123\n' ~ / digit %end ; reg_newline /    # true
     u'abc\n123' ~ / %start digit ; reg_newline /  # true
 
-Without the flag, `%start`/`^` and `%end`/`$` only match from the start or end
-of the string, respectively.
+Without the flag, `%start` and `%end` only match from the start or end of the
+string, respectively.
 
-    u'abc123\n' ~ / digit $ /                     # false
-    u'abc\n123' ~ / ^ digit /                     # false
+    u'abc123\n' ~ / digit %end /                  # false
+    u'abc\n123' ~ / %start digit /                # false
 
 Newlines are also ignored in `dot` and `![abc]` patterns.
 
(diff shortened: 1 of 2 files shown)


spec/ysh-regex-api.test.sh
```diff
@@ -84,6 +84,44 @@ yes
...
+  echo 'BUG'
+}
+if ($'abc\n123' ~ / ^ digit /) {
+  echo 'BUG'
+}
+
+if ($'abc123\n' ~ / digit $ ; reg_newline /) {
+  echo 'yes'
+}
+if ($'abc\n123' ~ / ^ digit ; reg_newline /) {
```
andychu, line 100: BTW as of a few weeks ago, you can use `u'abc\n123'` instead of the `$''` bash syntax.     `u` ensures it's a unicode string, and `b` is any byte string (but they are both of type `Str`, like `$''` is of type `Str`)




Also the eggex spelling of `^` and `$` is `%begin` and `%end` which is a sign it, though we accept the other spellings to be familiar

doc/eggex.md
```diff
@@ -311,6 +311,21 @@ Flags and translation preferences together:
 
     / digit+ ; ignorecase ; python /  # could translate to (?i)\d+
 
+The following flags are currently supported:
```
andychu, line 314: I would say "In Oils, the following flags ..."

There is this idea that people can implement Eggex on top of other regex engines ... not sure how appealing it is, but it is supposed to be more abstract

doc/eggex.md
```diff
@@ -311,6 +311,21 @@ Flags and translation preferences together:
...
+
+#### `i` / `reg_icase` (Ignore Case)
+
+**With** this flag, the match _is not_ case sensitive.
```
andychu, line 318: I think it can be more readable to give an example, like

> Use this flag to ignore case when matching.  For example `/'foo'; i/` matches 'FOO', but `/'foo'/` doesn't.

It's also shorter -- I think the two sentences here are a bit redundant

doc/eggex.md
```diff
@@ -311,6 +311,21 @@ Flags and translation preferences together:
...
+
+**Without** this flag, the match _is_ case sensitive.
+
+#### `reg_newline` (Multiline)
+
+**With** this flag, `$` will match before a newline and `^` will match after a
```
andychu, line 324: We should use the eggex names `%end` and `%start` here.

And I think it should be `dot` and `![abc]` to be clearer, since `!...` could mean something else

andychu (commented):
Thanks for adding this!

BTW the very last link in this doc:


http://travis-ci.oilshell.org/github-jobs/6097/ovm-tarball.wwz/_release/VERSION/doc/ysh-regex-api.html

Links to this help topic

http://travis-ci.oilshell.org/github-jobs/6097/ovm-tarball.wwz/_release/VERSION/doc/ref/chap-expr-lang.html#re-flags

But it's kind of sparse with no examples, and "duplicate" docs are fine (and inevitable )

PossiblyAShrub:
I've updated the documentation. The `reg_newline` examples are a few lines long, but I think that's reasonable because the behaviour is a little more complex. This is ready for another review.

doc/eggex.md
```diff
@@ -311,6 +311,37 @@ Flags and translation preferences together:
...
+
+Use this flag to ignore case when matching. For example, `/'foo'; i/` matches
+'FOO', but `/'foo'/` doesn't.
+
+#### `reg_newline` (Multiline)
+
+With this flag, `%end` will match before a newline and `%start` will match
+after a newline.
+
+    u'abc123\n' ~ / digit %end ; reg_newline /    # true
```
andychu, line 326: Looks great, but I think we should insert `=` before each example line, to show it's an expression, not a command

I have tried to be consistent about examples always being valid commands.   And `= expr` is a valid command

(Also at some point we really need to make the examples runnable, but that's a bit further down the road)

PossiblyAShrub: Done.

PR #1819, opened 2024-02-03
Prototype the ctx builtin

PossiblyAShrub:
This is a rough draft of the `ctx` builtin. It has implemented these verbs:

- `push`
- `set`
- `emit`

Here's an example which will now run:

```
var p = {}
ctx push (p) {
  ctx set (help="This is a help message")
  ctx emit flags ({short_name: '-v'})
  ctx emit flags ({short_name: '-h'})
}
# p => {'help': 'This is a help message', 'flag': [{short_name: '-v'}, {short_name: '-h'}]}
```

### DIFF
--- a/builtin/pure_ysh.py
+++ b/builtin/pure_ysh.py
@@ -102,6 +102,23 @@ class Shvar(vm._Builtin):
         return 0
 
 
+class ctx_Context(object):
+    """For ctx push (context) { ... }"""
+
+    def __init__(self, mem, context):
+        # type: (state.Mem, Dict[str, value_t]) -> None
+        self.mem = mem
+        self.mem.PushContextStack(context)
+
+    def __enter__(self):
+        # type: () -> None
+        pass
+
+    def __exit__(self, type, value, traceback):
+        # type: (Any, Any, Any) -> None
+        self.mem.PopContextStack()
+
+
 class Ctx(vm._Builtin):
 
     def __init__(self, mem, cmd_ev):
@@ -111,13 +128,8 @@ class Ctx(vm._Builtin):
 
     def _Push(self, context, block):
         # type: (Dict[str, value_t], command_t) -> int
-
-        # TODO: ctx_Ctx
-        self.mem.PushContextStack(context)
-        self.cmd_ev.EvalCommand(block)
-        self.mem.PopContextStack()
-
-        return 0
+        with ctx_Context(self.mem, context):
+            return self.cmd_ev.EvalCommand(block)
 
     def _Set(self, updates):
         # type: (Dict[str, value_t]) -> int
--- a/builtin/pure_ysh.py
+++ b/builtin/pure_ysh.py
@@ -4,7 +4,7 @@ builtin/pure_ysh.py - YSH builtins that don't do I/O.
 from __future__ import print_function
 
 from _devbuild.gen.runtime_asdl import (cmd_value, scope_e)
-from _devbuild.gen.syntax_asdl import loc
+from _devbuild.gen.syntax_asdl import loc, loc_t
 from _devbuild.gen.value_asdl import (value, value_e, value_t, LeftName)
 from core import error
 from core import state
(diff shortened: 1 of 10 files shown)


builtin/pure_ysh.py
```diff
@@ -102,6 +102,106 @@ def Run(self, cmd_val):
...
+
+        return 0
+
+    def Run(self, cmd_val):
+        # type: (cmd_value.Argv) -> int
+        _, arg_r = flag_spec.ParseCmdVal('ctx',
+                                         cmd_val,
+                                         accept_typed_args=True)
+
+        verb = arg_r.Peek()
```
andychu, line 162: BTW there is a Peek2() that gives a location that could be more specific?  I didn't test it, but if you do `ctx zzz` it could be more precise

builtin/pure_ysh.py
```diff
@@ -102,6 +102,106 @@ def Run(self, cmd_val):
...
+        field = None  # type: Optional[str]
+        field_loc = None  # type: Optional[loc_t]
+        if verb is None:
+            raise error.Usage('expected a verb (push, set, emit)',
+                              arg_r.Location())
+        if verb not in ("push", "set", "emit"):
+            raise error.Usage("unknown verb '%s'" % verb, arg_r.Location())
+        arg_r.Next()
+
+        if verb == "emit":
```
andychu, line 172: It's kind of weird that this is here, not under Emit(), but I can see why you did that

I guess one way of using the "natural" order  is to call AtEnd() in every branch?

Likewise I think the usage error on `not in ("push", "set", "emit")`  goes naturally at the end

spec/ysh-builtin-ctx.test.sh
```diff
@@ -0,0 +1,112 @@
...
+
+#### mini-argparse
+proc parser (; place ; ; block_def) {
+  var p = {}
+  ctx push (p, block_def)
+  call place->setValue(p)
+}
+
+var Bool = "Bool"
+var Int = "Int"
```
andychu, line 68: Yeah I think we need to define `Bool Int` etc. as either a Dict of methods or a `value.Type`

andychu (commented):
OK this is awesome!  I'm glad the idea actually worked

I like that this is making use of the rich proc signatures, rich call sites, and `value.Place`

So I think the theory from September was right -- flag parsing will actually test and validate the language design !  And it's passing the tests :-) :-) :-)

(flag parsing being a form of serialization, and serialization is really hard in a lot of languages because you're bridging interior and exterior, going outside the type system into "strings" !  e.g. see protobuf and serde)

---

I made a couple minor comments ... and I think probably the `ctx_Context` or `ctx_Ctx` is a good idea

Originally I thought we would do

```
ctx push
ctx emit
ctx pop
```

However I think it's not safe if you throw `error` in the middle and catch it -- then you will miss the pop and start other `ctx emit` in the wrong place

This is basically the same rationale as `with ctx:` in the Python language itself -- it correctly handles exceptions

And `error` is an exception.  (TBH I've been wondering if we should just call it raise or throw for that reason ...)

----

It could be useful to have a test for nested `ctx push`

I've been wondering if the single ctx stack is enough, and I think so, but we'll see.  You will only be constructing one flag def or VM definition or config file at a time

PossiblyAShrub:
I've cleaned this up and added some docs over the builtin. The PR is ready for another review.

builtin/pure_ysh.py
```diff
@@ -102,6 +102,106 @@ def Run(self, cmd_val):
...
+    def __init__(self, mem, cmd_ev):
+        # type: (state.Mem, CommandEvaluator) -> None
+        self.mem = mem
+        self.cmd_ev = cmd_ev  # To run blocks
+
+    def _GetContext(self):
+        # type: () -> Dict[str, value_t]
+        ctx = self.mem.GetContext()
+        if ctx is None:
+            raise error.FatalRuntime(
```
andychu, line 133: I think maybe error.Expr is more conventional and will also use status 3

I think that is the "default" for errors in new YSH features and the ctx is an expression, so it seems to make sense

andychu (commented):
OK excellent!  One comment but I'll just merge, since it can maybe be done along with argparse stuff

PR #1821, opened 2024-02-04
ysh-faq: echo_flags facts and consistent write -n -- $prefix example

bar-g:


doc/ysh-faq.md
```diff
@@ -72,22 +72,24 @@ The rules are subtle, so we avoid inventing new ones.
...
+To echo escape sequences, use a statically-parsed string literal, rather than
 `echo -e`:
 
-    echo $'tab \t newline \n'   # YES
-    echo j"tab \t newline \n"   # TODO: J8 notation
+    echo $'tab: \t newline: \n' # YES
+    echo u'tab: \t A: \u{41}'   # J8 unicode notation
 
-    echo -e tab \t newline \n'  # NO
+    echo -e "tab: \t newline: \n"    # NO (backslash)  => Invalid char escape
```
andychu, line 84: I think these examples should use single quotes, not double quotes, because there is an extra meaning of `\\` in double quotes

doc/ysh-faq.md
```diff
@@ -72,22 +72,24 @@ The rules are subtle, so we avoid inventing new ones.
 
 ## How do I write the equivalent of `echo -e` or `echo -n`?
 
-YSH removed these flags to `echo` because it has good alternatives
+YSH removed these flags to `echo` to solve their interference problem with
+variable substitutions (see next) and double-quoted backslash.
 
-To escape variables, use the statically-parsed string language, rather than
+To echo escape sequences, use a statically-parsed string literal, rather than
```
andychu, line 78: How about "To echo special characters denoted by escape sequences"

doc/ysh-faq.md
```diff
@@ -72,22 +72,24 @@ The rules are subtle, so we avoid inventing new ones.
...
 `echo -e`:
 
-    echo $'tab \t newline \n'   # YES
-    echo j"tab \t newline \n"   # TODO: J8 notation
+    echo $'tab: \t newline: \n' # YES
+    echo u'tab: \t A: \u{41}'   # J8 unicode notation
```
andychu, line 82: I think this should be exactly the same string with `newline: \n`, and we could possibly link to `doc/ref` to show the full syntax

andychu (commented):
Thanks, sorry for the delayed response, I was traveling a bit

doc/ysh-faq.md
bar-g: very good

doc/ysh-faq.md
bar-g: Maybe newline \n as unicode? I think it's good to include some general idea of differences.

doc/ysh-faq.md
bar-g: Hm, but it's the NO side showing problems, and there actually are two things involved in why echo does not work in the same way it does in shell. So, I find it only helpful to show both issues on separate example lines.

Also the string type supporting substitutions is probably more important for echo in general.

doc/ysh-faq.md
bar-g: `echo u'tab: \t newline: \u{A}'`

bar-g: include idea of usages rather

doc/ysh-faq.md
bar-g: In another way also "The example should use double quotes, because they support substitutions".

So actually, what's the ysh way to insert substitutions? `echo $'tab:\t'$name$'continued'`?

Because these don't work:
```
echo u'tab:\t'"$name"u'continued'
echo u'tab:\t'${name}u'continued'
```

bar-g:
PS: I didn't know how to insert internal links to the doc.

doc/ysh-faq.md
bar-g: I think there may still be no proper equivalent to `echo -e "\t $var"`, or do you know one?

```
ysh ysh-0.20.0$ echo u'\t'$indented
  echo u'\t'$indented
            ^~~~~~~~~
[ interactive ]:1: Unexpected token after YSH single-quoted string
```

As of this, I could only use echo's default space separation of arguments as current workaround in the FAQ:

```
echo $'tab: \t' $indented  # YES
echo u'tab: \t' $indented  # J8 with \u{unicode} notation
echo b'tab: \t' $indented  # J8 with \u{unicode} and \y{byte} notation

echo -e "tab: \t $indented"   # NO (backslash)  => Error: Invalid char escape
echo -e "tab: \\t $indented"  # NO (echo_flags) => Prints: "-e tab..."
```

bar-g: With something before the u'-string, it' s not an error, but instead it's not recognized as such:
```
ysh ysh-0.20.0$ echo ${first}u'\t'$second
EXTENDEDu\tTEXT
ysh ysh-0.20.0$ 
```

bar-g: Added this one:
`echo $[ u'tab: \t' ++ indented ] # YES (generic YSH, not "IFS separated)`
Is that it?

andychu: Yes, using `++` is a good way to combine different quote types!

That's useful in this section

doc/ysh-faq.md
```diff
@@ -72,47 +72,56 @@ The rules are subtle, so we avoid inventing new ones.
...
-`echo -e`:
+To echo special characters denoted by escape sequences, choose a
+statically-parsed string literal, rather than `echo -e`:
 
-    echo $'tab \t newline \n'   # YES
-    echo j"tab \t newline \n"   # TODO: J8 notation
+    echo $'tab: \t' $indented       # YES ($IFS separated, shell compatible)
+    echo u'tab: \t' $indented       # J8 with \u{unicode} notation
+    echo b'tab: \t' $indented       # J8 with \u{unicode} and \y{byte} notation
+    echo $[u'tab: \t ' ++ indented] # string addition (not $IFS separated)
```
andychu, line 84: I feel this section has gotten a bit complex

I do like the addition of the `++`, and double quotes are worth a mention

 I will merge this into another branch and then minimize / polish it a bit

doc/ysh-faq.md
```diff
@@ -72,47 +72,56 @@ The rules are subtle, so we avoid inventing new ones.
...
-    echo $x             # an even shorter way
+    echo $x             # an even shorter "unmolested" way
 
-So `echo` is technically superfluous in YSH, but it's also short, familiar, and
-correct.
+So `echo` is technically superfluous in YSH, but it's also a short, familiar,
+and correct way to compose words to print. 
 
-YSH isn't intended to be compatible with POSIX shell; only OSH is.
+Whereas `write`, by default, prints one line per argument given.
```
andychu, line 122: Note that you can also use `write --sep '' $x $y $z` to print them with no newlines

andychu:
OK I tightened it up a bit and landed it here:

https://github.com/oilshell/oil/commit/7b2b5036d8e728bd6fd7198c6ae1bc4d43c9a821

I generally try to use as few words as possible while still being accurate / focusing on the main point.

Thanks for the contribution!

bar-g:
Alright, I see you like the minimal style.

I think it's good as long as it's not breaking the reading comprehension flow. I guess the FAQ can work with some other string literal details left out, no problem. (Even less as the reference docs are becoming better.)

Issue #1784, opened 2024-01-13
'const c' without null on RHS can be re-declared


bar-g:
[EDIT after comments]

need `=> is_empty?` and `=> is_set?` (methods returning bool)

For example, as global variables or constants may need to be shared (potentially later also local variables by "entailing" selected "entailed" vars to dynamic scope maybe) between different code parts (func/proc/module), people need a way to check for them, e.g. to init constants (first and only write).
https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

Creating a (shared) global from within a module's proc/func with `setglobal` would overwrite it unconditionally, otherwise.

---

[Original quest where this came up was: Robust declaration of a globally shared variable. [It also yielded the more specific #1786 and implement try ( ) in addition to try { } #1787 :]

It can only be re-declared as long as it is of NoneType, but not with empty types.
```
ysh ysh-0.19.0$ const c
ysh ysh-0.19.0$ const c
ysh ysh-0.19.0$ const c = {}
ysh ysh-0.19.0$ const c = {}
  const c = {}
        ^
[ interactive ]:2: fatal: Can't assign to readonly value 'c'
```

It seems problematic that the error is fatal, either way, when empty or filled.

Trying it out, strangely `try` does not even find 'const', though?
```
ysh ysh-0.19.0$ try const c = {}

  try const c = {}
      ^~~~~
[ interactive ]:25: 'const' not found
```

Do you think `try const c = {}`  would be a simple and elegant way to ensure existence of a global, without ever resetting it? (After making it fail gracefully.)

----

In any case, how to manually determine if a global var (or "const" dict) has already been declared, i.e. to avoid resetting the required shared resource after it was already initialized and possibly updated by another script/module?

In the git's chap-type-method.md I found `=> keys()`, but it also fails fatal, 
```
ysh ysh-0.19.0$ try echo @[g => keys()]

  try echo @[g => keys()]
                  ^~~~
[ interactive ]:63: fatal: Undefined variable 'keys'
```

and triggers other strange errors after `try`, `=` or `call`:

```
ysh ysh-0.19.0$ echo @[c => keys()]


ysh ysh-0.19.0$  #  so, dict c is empty
ysh ysh-0.19.0$  try call c => keys()

   try call c => keys()
                     ^
[ interactive ]:52: Space required before (
ysh ysh-0.19.0$ try = c => keys()

  try = c => keys()
      ^
[...]

PR #1823, opened 2024-02-06
#1784: const variable with nothing on RHS should not be allowed to be redeclared

AtomsForPeace:


test/ysh-runtime-errors.sh
```diff
@@ -491,6 +491,11 @@ test-var-decl() {
   _expr-error-case 'setvar x, y = 1, 2, 3'
 }
 
+test-const-decl() {
+  _error-case 'const x = {}; const x = {};'
+  _error-case 'const x; const x;'
+}
+
```
andychu, line 498: Thanks, this is great!  I was just thinking about this bug, definitely needs to be fixed

Minor issue is that we don't have "allowed failures" for `test/ysh-runtime-errors.sh`, only for spec tests

So we could either 

- try to fix this bug and merge with this test
- Or add to  `spec/ysh-assign` so we can set `oils_allowed_failures:` .  Tests first are welcome as usual
- 

It's no problem to have  a bit of duplicate coverage in both test files, since this is just 2 lines

---

I haven't thought about / look at what the actual bug is, but we can chat about it

AtomsForPeace: Yeah, just having a poke about to see where it could be for now. I was thinking about option 1 but option 2 works well as well. Could be nice to have allowed failures in all test files.

AtomsForPeace:
Seems like it could be on line 698 in osh/cmd_eval.py.
There's a comment that makes me think this is intentional:
# Note: not respecting const since they should be initialized

AtomsForPeace:
> Seems like it could be on line 698 in osh/cmd_eval.py. There's a comment that makes me think this is intentional: # Note: not respecting const since they should be initialized

@andychu any thoughts on this?

andychu:
Oh sorry for not responding!  I was traveling a a few days ago, I didn't realize you had replied

Let me see

andychu:
Hm actually I am not sure what that comment means, it could have been a brain fart

https://github.com/oilshell/oil/blob/master/osh/cmd_eval.py#L698

I wonder if we change flags to include `SetReadOnly` if the keyword is `KW_Const` (like below), does it fix the bug?

If so then that's probably the right thing to do

Or it could be something else

AtomsForPeace:
Bump

andychu:
Oh sorry I didn't realize this was ready to review!  

I didn't see a message?  Sometimes people push "in between" commits and they don't want a review yet

We had a similar mix-up a few weeks ago, so let me write something in `Contributing` and `README.md` about this

i.e. basically send a message saying "this is ready for review", and similar

andychu:
This looks great, thank you!

I auto-formatted the code with the `yapf` formatter and factored out `flags`.  (There are some tips on yapf at https://github.com/oilshell/oil/wiki/Oil-Dev-Cheat-Sheet -- I just updated them)

----

I'm puzzled by that comment

> # Note: not respecting const since they should be initialized

I think what that meant is that we used to have a **static** const check that might catch this, so maybe I thought we didn't need to do the **dynamic** check by passing flags.

But that was only within `proc` and `func`, so the global `const` was a bit broken.

---

I learned the hard way that variable declaration and mutation is tricky at the global level.  

Related links - https://www.oilshell.org/release/0.20.0/doc/variables.html#appendix-b-problems-with-top-level-scope-in-other-languages (JavaScript and Racket have some problems with this.)

`const` is also a bit weird in dynamic languages, I think it makes sense that Python doesn't have it.  Right now we're using it because it helps with `source`-ing multiple files in the same namespace.

PR #1824, opened 2024-02-08
word_parse.py: parse_backslash (errors + hinting questions)

bar-g:
This is to provide consistent short error messages, and adds minimal hinting questions for context and **immediate help-on-error**.

Note how the messages even improve more once `parse_backslash` gets renamed into `strict_backslash_escape`.

Example error messaage:
```
Invalid char escape in "-string literal (parse_backslash, no verbatims)
       Allowed: \", \$, or \\ to denote ", $, or \
       * use \\ to denote a \ ?
       * 'chain'"different"$'literals' ?
       * (ysh) concat "- ++ u'- or b'-literals ?
```

### DIFF
--- a/osh/word_parse.py
+++ b/osh/word_parse.py
@@ -667,7 +667,7 @@ class WordParser(WordEmitter):
 
                 # x = $'\z' is disallowed; ditto for echo $'\z' if shopt -u parse_backslash
                 if is_ysh_expr or not self.parse_opts.parse_backslash():
-                    p_die("Invalid char escape in C-style string literal", tok)
+                    p_die("Invalid char escape in C-style string literal. Literal: '\\\\' (parse_backslash)", tok)
 
                 tokens.append(tok)
 
@@ -944,7 +944,7 @@ class WordParser(WordEmitter):
                         if (is_ysh_expr or
                                 not self.parse_opts.parse_backslash()):
                             p_die(
-                                "Invalid char escape in double quoted string",
+                                "Invalid char escape in double quoted string. Literal: '\\\\' (parse_backslash)",
                                 self.cur_token)
                     elif self.token_type == Id.Lit_Dollar:
                         if is_ysh_expr or not self.parse_opts.parse_dollar():
@@ -1611,7 +1611,7 @@ class WordParser(WordEmitter):
             ch = lexer.TokenSliceLeft(tok, 1)
             if not self.parse_opts.parse_backslash():
                 if not pyutil.IsValidCharEscape(ch):
-                    p_die('Invalid char escape (parse_backslash)',
+                    p_die("Invalid char escape. Literal: '\\\\' (parse_backslash)',
                           self.cur_token)
 
             part = word_part.EscapedLiteral(self.cur_token,
--- a/osh/word_parse.py
+++ b/osh/word_parse.py
@@ -667,7 +667,7 @@ class WordParser(WordEmitter):
 
                 # x = $'\z' is disallowed; ditto for echo $'\z' if shopt -u parse_backslash
                 if is_ysh_expr or not self.parse_opts.parse_backslash():
-                    p_die("Invalid char escape in C-style string literal. Use '\\\\' to get \\ (parse_backslash).", tok)
+                    p_die('Invalid char escape in C-style string literal. Use \\ to get \ (parse_backslash).', tok)
 
                 tokens.append(tok)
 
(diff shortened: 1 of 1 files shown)


osh/word_parse.py
```diff
@@ -667,7 +667,7 @@ def ReadSingleQuoted(self, lex_mode, left_token, tokens, is_ysh_expr):
 
                 # x = $'\z' is disallowed; ditto for echo $'\z' if shopt -u parse_backslash
                 if is_ysh_expr or not self.parse_opts.parse_backslash():
-                    p_die("Invalid char escape in C-style string literal", tok)
+                    p_die('Invalid char escape in C-style string literal. Use \\\\ to get \ (parse_backslash).', tok)
```
andychu, line 670: Hm this seems overly specific, because the example given above is `\z`

I think what we could do is have some kind of error code, and the error code could explain the hints like using `\\\\` 


Though I don't even think that's a good workround, because mostly you will use regular single quotes like `'\'`

bar-g: Providing context and hints right with the error is immensely helpful. Needs to be provided directly, I think.

What do you think of adding them separately underneath? For example:
```
'Invalid char escape in C-style string literal (parse_backslash).
    Example options:          * \\\\ denotes \
                       posix: * '-string
                       ysh:   * r'-string
```

bar-g:
```
'Invalid char escape in C-style string literal (parse_backslash).
\t* \\\\ to denote \ ?
\t* '-string ? (posix)
\t* r'-string ? (ysh)'
```

bar-g: > I don't even think that's a good workround, because mostly you will use regular single quotes like '\'

Workaround for what?
And to the contrary, if a $'...' string is used (explicit shell "raw" string supporting escape sequences), single quotes won't allow that.

bar-g: Please have a look at the changed strings.

Seems much better now. Adding "bullet questions" below the error could become a good general theme.

bar-g:
The error + question style works out quite good:
```
Invalid char escape in C-style $'-string literal (parse_backslash)
	* use \\ to denote a \ ?
	* 'chain'"different"$'literals' ?
	* (ysh) use u'- or b'-literal instead ?
	* (ysh) concat r'- ++ u'- or b'-literals ?
```

andychu:
I think this belongs in `doc/error-catalog.md` and then we should print URLs in the error messages

We want something like this - https://doc.rust-lang.org/error_codes/error-index.html



This kind of description doesn't scale -- there will always be more context and hints you can add

 (and it's very hard to read and edit in the source code)

bar-g:
> (and it's very hard to read and edit in the source code)

Yes it is, I noticed that. I think referencing makes sense, to keep it maintainable.

However, for a helpful interpreter I think at least short errors, info and hints must be compiled into the interpreter, and be printed by default. Mentioning some ID, or better even, also printing more detailed info if some doc package is installed could be a separate good thing.

So what's there?
* short-error-message (what's wrong), better maintained is source code, or not? ever re-used?
* short-error-info (e.g. what's allowed, reasoning, comparisions)
* short-bullet-question hints

andychu:
OK I did something about this:


http://travis-ci.oilshell.org/github-jobs/6335/ovm-tarball.wwz/_release/VERSION/doc/error-catalog.html#oils-err-10

https://github.com/oilshell/oil/commit/36fcf3dfb66bcd15b069a2e14b8c36df990e24aa

You can download a new tarball from the CI if you want to check it out (task `cpp-tarball` or `ovm-tarball`, let me know if you haven't done that before)


Feedback welcome!

Instructions on how to update it are at the top of the doc.  That is welcome too 

Although really we need to link more to the Oils reference, which is `ref/chap-*.html` etc.   I might be missing a couple links there

And then we need to write missing help topics too

bar-g:
Ok, let's see how that goes.

But don't you think adding at least one zero to each error category would make sense?

100->1000
200->2000

bar-g:
I looked at the  examples that you provided, and saw you only added some error explaining info after the bullet-question hints. (Wereas I had been adding such short error info (context/reason/what's good) right behind the error (what's wrong).

May we add the error info lines before the question-hints, as ordered list maybe?

And quite important, would you actually accept the lines that I prepared in the pull requests so far?

bar-g:
Closing in favor of  error-catalog.md: add parse_backslash info and hints #2317

PR #1832, opened 2024-02-11
Use the new ctx builtin in parseArgs()

PossiblyAShrub:
## Changes:
- In the `ctx` builtin change `error.FatalRuntime(3, ...)` to `error.Expr(...)`
- Use the `ctx` builtin for the arg-spec construction API

### DIFF
--- a/spec/ysh-stdlib-args.test.sh
+++ b/spec/ysh-stdlib-args.test.sh
@@ -9,7 +9,7 @@
 
 source --builtin args.ysh
 
-Args (&spec) {
+arg-parse (&spec) {
   flag -v --verbose ('bool')
   arg src
   arg dst
@@ -37,11 +37,12 @@ json write (i)
 
 source --builtin args.ysh
 
-Args (&spec) {
-  description = '''
+arg-parse (&spec) {
+  # TODO: implement description, prog and help message
+  description '''
      Reference Implementation
   '''
-  prog = "program-name"
+  prog "program-name"
 
   arg -v --verbose (Bool, help = "Verbose")
   arg src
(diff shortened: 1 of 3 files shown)


spec/ysh-stdlib-args.test.sh
```diff
@@ -9,7 +9,7 @@
 
 source --builtin args.ysh
 
-Args (&spec) {
+arg-parse (&spec) {
```
andychu, line 12: We don't have to resolve this now, but it's hard to find a good name for this

I would think of it as noun-vs-verb

I think the "spec" should be a noun.  Though I've come to dislike "spec" because it seems like a wishy-washy and overused word

'arg-parse' seems like a verb.  `arg-parser` would be closer

But we can think about it

stdlib/args.ysh
```diff
@@ -34,82 +34,76 @@
 #     - that's what it's doing, it's mutating "self"
 #     - _this is the thing we're currently creating
 
-proc Args ( ; out; ; block) {
+proc arg-parse (; place ; ; block_def) {
   ## Create an args spec which can be passed to parseArgs.
```
andychu, line 38: Now that I see we have these `##` comments, I wonder if we do something like Elixir

https://hexdocs.pm/elixir/1.12/writing-documentation.html

And have a no-op `doc` verb, so it would be 

```
proc arg-parse {
  doc '''
  Multi-line 
  doc
  string.  And leading spaces are stripped according to the rule of multi-line strings
  '''
}
```

Right now we have

```
proc p {
  ### single line doc string
}
```

```
proc p {
  ### single line doc string
  ### this seems
  ### ugly
  ### and noisy

}
```

andychu (commented):
OK this looks like good progress!

PR #1836, opened 2024-02-13
[ysh] Implement Str => trim*, create UTF-8 decoder

PossiblyAShrub:
## Changes
- Implement unicode-aware `Str => trim()`, `Str => trimLeft()` and `Str => trimRight()`
- Added UTF-8 decoding to `osh/string_ops.py` (this is not entirely correct as we ignore surrogate pairs, but the plan is to re-implement decoding in `fastfunc` anyways)


andychu:
I pointed on Zulip some unfinished test code I had for UTF-8 decoding (in C, not Python)

The new JSON/J8 encoder and decoder both use the "Bjoern DFA" in C.  I think we can reuse that, although there is a slight modification we need to make (surrogate pair roudn trip in `spec/ysh-json`)

And testing is one way I justified that ... e.g. usually we want to write the "executable spec" in Python, and I actually did that with `data_lang/qsn_.py` which is now deleted.    But that implementation was always quite slow with 1-byte heap-allocated strings

But for UTF-8, if we do it in C, then we can do exhaustive testing very fast, in seconds

---

So yeah I think the Bjoern DFA was "proven" in the last month, and it makes sense to centralize on that

After Eggex and JSON, Unicode is one of the last things  to fix the problem where  "Shell is a language of strings that's bad at strings" - https://oilshell.zulipchat.com/#narrow/stream/266575-blog-ideas/topic/Shell.20Is.20a.20Language.20of.20Strings.20That.27s.20Bad.20At.20Strings

So if you are interested in going kinda deep on this, I can use the help :)

andychu:
Hm maybe to narrow the conversation a bit, I think it's possible to land this without changing all of UTF-8 (although I think that's a fun separate conversation)

I think we could possibly write two functions:

```
fastfunc.DecodeUtf8(s: str) -> List[int]  # wrapper for Bjoern DFA, we have to decide what to do about errors

fastfunc.EncodeUtf8(runes: List[int]) -> str    # encoding is easier, there's a copy in in data_lang/utf8_test.cc
```

And then the trim() function could use these two?

1. decode
2. process code points in Python
3. encode

Although I guess this is not very efficient

I guess the problem is if you have a 1MB string, and it has a single space at the beginning and at the end, and then it has `xyz`

    ' xyz....some other stuff...  xyz '

Then it's easy to tell that you don't need to decode and encode everything

Hm

andychu:
Oh also I followed the link in `core/shell.py`, that doesn't quite agree with the one you added

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#white_space

I find things like this highly annoying

> Note: Changes to the Unicode standard used by the JavaScript engine may affect programs' behavior. For example, ES2016 upgraded the reference Unicode standard from 5.1 to 8.0.0, which caused U+180E MONGOLIAN VOWEL SEPARATOR to be moved from the "Space_Separator" category to the "Format (Cf)" category, and made it a non-whitespace. Subsequently, the result of ["\u180E".trim().length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) changed from 0 to 1.


basically if the Unicode database changes, then this function changes ...

which means Oils needs to reference a Unicode version technically, ugh :-(

for a very simple function.

I wonder if we an option for ASCII spaces?  Or we could even have `unicode_version` as a param to trim()!

PossiblyAShrub:
Ah, the DFA is pretty cool! I agree that we should use that instead, especially if it's been proven. Instead of allocating a list of codepoints, why not lazily decode? We'd have an API similar to the decoder I've already written:

```python
# decode one codepoint, or return -1 for error. cursor is an in/out param (return a Tuple in python?)
fastfunc.DecodeUtf8Codepoint(s: str, cursor: *int) -> int

# No string re-encoding needed
```

In the future, we can look into a backwards decoder which starts at the end of the string and moves to the beginning. Then we only have to scan from the start to first non-whitespace codepoint and end to last non-whitespace codepoint. (Although, there could an issue though if a user passes in a string like `" xyz <some non-utf8 bytes> xyz "`.)

I can help out with the Unicode work. I think once we have enough Unicode utilities a good chunk of the string API can be completed. That will be helpful for `argparse` (and maybe unit testing?) which needs to manipulate strings (eg. formatting a `--help` message.)

> Oh also I followed the link in `core/shell.py`, that doesn't quite agree with the one you added
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#white_space
> 
> I find things like this highly annoying
> 
> > Note: Changes to the Unicode standard used by the JavaScript engine may affect programs' behavior. For example, ES2016 upgraded the reference Unicode standard from 5.1 to 8.0.0, which caused U+180E MONGOLIAN VOWEL SEPARATOR to be moved from the "Space_Separator" category to the "Format (Cf)" category, and made it a non-whitespace. Subsequently, the result of ["\u180E".trim().length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) changed from 0 to 1.
> 
> basically if the Unicode database changes, then this function changes ...
> 
> which means Oils needs to reference a Unicode version technically, ugh :-(
> 
> for a very simple function.
> 
> I wonder if we an option for ASCII spaces? Or we could even have `unicode_version` as a param to trim()!

Good catch! I like the `unicode_version` param. I can imagine that most users will ignore it, but this gives YSH a good look for those who care about that level of stability.

I see that I am also missing `\t`, `\v` and `\f` in the `SPACES` list, I'll make sure to add those.

andychu:
OK cool, one semantic that could make sense is by default strip the 6 chars in the firs 6 rows of the table

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#white_space


```
U+0009 	Character tabulation 	<TAB> 	Horizontal tabulation 	\t
U+000B 	Line tabulation 	<VT> 	Vertical tabulation 	\v
U+000C 	Form feed 	<FF> 	Page breaking control character ([Wikipedia](https://en.wikipedia.org/wiki/Page_break#Form_feed)). 	\f
U+0020 	Space 	<SP> 	Normal space 	
U+00A0 	No-break space 	<NBSP> 	Normal space, but no point at which a line may break 	
U+FEFF
```

And then only if you pass `unicode_version=something`, "other unicode space characters" are included


Basically I think the default of those 6 spaces is going to work "almost all the time".  Most people will NOT need to pass `unicode_version` ever, probably like 99.999% of programs

But then we have both stability, and a path to evolve that matches JavaScript and future Unicode standards

andychu:
And yeah I think we can decode lazily

I think we will need this for

```
for rune in (runes(s)) {
}
```


We should test the case with invalid bytes in the middle too, and specify what it does

andychu:
Hm I had a thought about this, which is similar to the `unicode_version=` idea, but less work

1. By default, we use the 6 chars in the JavaScript table.  Doesn't depend on a Unicode DB or libc.
2. If the user opts in, then libc `isspace()` is called.  We return true if **either** this returns true, or it's in the fixed list of 6 chars.

---

I made a demo of isspace() here and it seems reasonable.  Although I guess I didn't test the exotic spaces

https://github.com/oilshell/oil/commit/239d4c0dda02cbcc6481aca2586edad6a2f0e736

Weirdly it disagrees (on my machine) that U+00A0 which is &nbsp; is  a space - https://www.compart.com/en/unicode/U+00A0

But that's why we can have the stable fallback.

I guess the param could be called `trim(more_spaces=true)` or something -- it's an unspecified list from the system libc, which we don't control

andychu:
And I guess the reason to not to **always** use `isspace()` is so that scripts are more stable.  Unless you pass the keyword param, the behavior won't depend on the exact system you're running on, only the Oils version

PossiblyAShrub:
I'm not sure if enabling libc `isspace()` should be a separate parameter. How about something like `classdb="unicode8.0.0"` or `classdb="libc"`? Otherwise, by default, we only strip ASCII white-space. (We can probably find a better name than `classdb`.)

Given that `isspace()` is locale dependent \[0\], I agree that we shouldn't rely on `libc` behaviour _by default_.

\[0\]: https://sourceware.org/glibc/manual/2.39/html_mono/libc.html#Case-Conversion

andychu:
Oh well I was saying the libc should be the ONLY param?

I like "unicode8.0.0" less because that means we have to update it in perpetuity :-P  I like delegating to libc, which already has to update it

It keeps Oils smaller

andychu:
Oh and  I updated the Unicode doc again, and I think `trim()` is weirdly one of the ONLY functions that could possible change

https://github.com/oilshell/oil/blob/master/doc/unicode.md

The other being upper() and lower(), so I would also like to delegate those to libc toupper() and tolower()  (which we're doing in C++ already -- Python is slightly different which is immaterial)

PossiblyAShrub:
Not bundling in a unicode database sounds reasonable. If user's need it, we can always add a way to reference an external unicode db.

I still don't like the name `more_chars`; it's too vague. What do you think of a more explicit name such as `libc_fallback`?

andychu:
We can  do something besides more_spaces/more_chars

But `libc_fallback` I think is too legalistic (even though this will be rare).  If you ask most shell users what that means, you'll probably get a blank stare


It could be `use_locale` or something, though I'm not sure if that's technically accurate

andychu:
maybe `more_unicode`?

Since it should be uncommon, a slightly longer name is OK

andychu:
I found an example of the unicode DB updates.  I assume that for 90% of  releases, the list of space characters doesn't change (or maybe they will never change again)

So this is a good reason to avoid burdening ourselves with the database (as JavaScript and Julia do).  The libc strategy seems like it will make Oils more stable


https://juliastrings.github.io/utf8proc/releases/


```
[utf8proc 2.9.0](https://github.com/JuliaStrings/utf8proc/releases/tag/v2.9.0) (2023-10-32):

    Unicode 15.1 support (https://github.com/JuliaStrings/utf8proc/issues/253).

[utf8proc 2.8.0](https://github.com/JuliaStrings/utf8proc/releases/tag/v2.8.0) (2022-10-30):

    Unicode 15 support (https://github.com/JuliaStrings/utf8proc/issues/247).

[utf8proc 2.7.0](https://github.com/JuliaStrings/utf8proc/releases/tag/v2.7.0) (2021-12-16):

    Unicode 14 support ([#233](https://github.com/JuliaStrings/utf8proc/issues/233)).
```

andychu:
Let me know if I should review this now, I took a quick skim and it's looking close to done?

(remember github sends me a notification on each push, which is fine, but then a review should be explicitly requested ... I felt bad I missed the last one by several days!)

PossiblyAShrub:
> Let me know when I should review this, I took a quick skim and it's looking close to done?

Yep, this was almost done! I wanted to give it go over my code once more before I asked for a review.

This is now ready for a review.

osh/string_ops.py
```diff
@@ -53,7 +55,95 @@ def _Utf8CharLen(starting_byte):
...
+
+    if b & 0b11000000 != 0b10000000:
+        raise error.Expr(INVALID_CONT, loc.Missing)
+
+    return b & 0b00111111, cursor
+
+
+def DecodeUtf8Char(s, start):
+    # type: (str, int) -> int
+    """Given a string and exclusive bounds [start, end), decode the Unicode
```
andychu, line 75: Looks like this no longer takes end, only start

osh/string_ops.py
```diff
@@ -53,7 +55,95 @@ def _Utf8CharLen(starting_byte):
         e_strict(INVALID_START, loc.Missing)
 
 
-def _NextUtf8Char(s, i):
+def _ReadOneUnit(s, cursor):
+    # type: (str, int) -> Tuple[int, int]
+    """Helper for DecodeUtf8Char"""
+    if cursor >= len(s):
+        raise error.Expr(INCOMPLETE_CHAR, loc.Missing)
```
andychu, line 62: Hm yeah I remember this location issue, the problem is basically parsing strings at runtime, and that has to be connected back to the source location somehow

I imagine the error location is not good now

Or I guess it would just be ideal if it points to the `(` in  `trim()`?

(I just did a huge refactoring of `test/ysh-runtime-errors.sh`, the file we use to test and eyeball the errors)

osh/string_ops.py
```diff
@@ -53,7 +55,95 @@ def _Utf8CharLen(starting_byte):
...
+    # encodings in table 3.7 from [0].
+    #
+    # | Scalar Value               | 1st Byte | 2nd Byte | 3rd Byte | 4th Byte |
+    # +----------------------------+----------+----------+----------+----------+
+    # | 00000000 0xxxxxxx          | 0xxxxxxx |          |          |          |
+    # | 00000yyy yyxxxxxx          | 110yyyyy | 10xxxxxx |          |          |
+    # | zzzzyyyy yyxxxxxx          | 1110zzzz | 10yyyyyy | 10xxxxxx |          |
+    # | 000uuuuu zzzzyyyy yyxxxxxx | 11110uuu | 10uuzzzz | 10yyyyyy | 10xxxxxx |
+    #
+    # [0] https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf
```
andychu, line 101: Hm I just scrolled through this doc, I had never seen it

That's a doozy

Someone should write a much shorter description :)   I believe I have "all" of the UTF-8 encoding/decodnig  in my head now (not all of Unicode, but I believe the only other thing Oils needs is uppercase/lowercase)


The point of the skeleton in `data_lang/utf8_test.cc`  was to test that claim :)  By exhaustive testing

I still want to do that, although there is a lot of other stuff to do, and this PR looks like it will get us very close to where we want to be

I'm a little worried about the speed, because we're creating 1 byte strings, but correctness is more important than speed.  Decoding only the beginning and end definitely helps

andychu (commented):
This looks excellent, thank you!  Great tests

- Can you update the trim() trimLeft() trimRight() topics in `doc/ref/chap/type-method`?  And in `toc-ysh` they can be marked implemented
- Also do the decoding errors give at least some reasonable position?


I guess I could clone and run the spec tests, BTW there is a flag `-d` that shows more errors, like this

```
test/spec.sh ysh-methods -r 0 -v -d  # show more detail
```

andychu:
OK I just tried the `-d` thing, but didn't see the error locations

So I merged master into this branch, since I did that huge refactor to `test/ysh-runtime-errors`

andychu:
OK actually the error location is pretty good!  

```
==== CASE: -c   var badUtf = b'\yF9'

  # error is missed
  call " a$[badUtf]b " => trim()
  echo status=$_status

  # error is found
  call "$[badUtf]b " => trim() =====


status=0
    call "$[badUtf]b " => trim()
          ^~
[ -c flag ]:8: fatal: Invalid start of UTF-8 character
```

osh/string_ops.py
PossiblyAShrub: It may also be nice to emit the offending byte offset. For example, in python 3.10:

```python
>>> bytes([0x20, 0xF9]).decode('utf8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf9 in position 1: invalid start byte
```

But, for now, I've decided to keep the error messages consistent with the other UTF-8 string operations.

osh/string_ops.py
PossiblyAShrub: > I'm a little worried about the speed, because we're creating 1 byte strings, but correctness is more important than speed. Decoding only the beginning and end definitely helps

Yikes. I just took a look at the generated cpp, and yes, we do this. I did a simple grep for `ord(some_string->at(some_index))` and it appears we do this elsewhere too:

```
$ grep -E 'ord\(.*->at' _gen/bin/oils_for_unix.mycpp.cc
                      d = len(s) >= 2 ? ord(s->at(1)) : 0;
        delim_byte = ord(arg->d->at(0));
          tok = location::LeftTokenForWord(trail->words->at(0));
          tok = location::LeftTokenForWord(trail->words->at(-1));
    return ord(this->src_line->content->at(pos));
      start_num = ord(part->start->at(0));
      end_num = ord(part->end->at(0));
    left_tok = location::LeftTokenForWord(words->at(i));
    right_tok = location::RightTokenForWord(words->at(-1));
    left_tok = location::LeftTokenForCompoundWord(words->at(0));
  b = ord(s->at(cursor));
  b = ord(s->at(start));
  byte_as_int = ord(s->at(i));
    byte_as_int = ord(s->at(i));
      return Alloc<CharCode>(ord(tok->tval->at(0)), false, tok);
                      cval = num::ToBig(ord(consts::LookupCharC(tok_str->at(1))));
```

I guess this would be fixed by small strings? It might also be worth optimizing this into something like `int BigStr::ord_at`, which would avoid the allocation entirely.

In this case, this should be fine for now? The plan is to replace all of the python utf-8 decoding with something like the Bjorn DFA.

PossiblyAShrub:
I've fixed the docstring on `DecodeUtf8Char`. Since the error message is alright, I don't think it needs to be updated in this PR. This is ready for another review.

osh/string_ops.py
andychu: Yes this is OK for now, we have that issue in other places too

Small string optimization can fix that ... I think it's feasible but not trivial, I did a little experiment in the code awhile ago

https://github.com/oilshell/oil/blob/master/mycpp/small_str_test.cc

andychu:
Looks great, thank you!

PR #1847, opened 2024-02-23
[toc-ysh] relatable 'echo' and 'write' short descriptions

### LINKED ISSUE #1843: [osh] echo -E "a\nb"
Bash also parses `echo -E`.

man bash:  "The -E option disables the  interpretation  of  these escape  characters,  even  on  systems  where they are interpreted by default."
```
$ echo -E "a\nb"
a\nb
$ echo -E -eenEEnneeeE -nnnee 'ALL-PERMUTATION-OPTS-TRIMMED\n'
ALL-PERMUTATION-OPTS-TRIMMED

```

```
osh-0.20.0$ echo -E 'a\nb'
-E a\nb
```

bar-g:
Have short descriptions in overview that are useful for comparing.

ysh-echo: The  -E flag is also part of the flag permutations dropped when printing, but at the end not listing flags here, instead mentioning their default print operation mode (to contrast write and echo) and saying echo to have no flags, and later no echo_flags bugs.

For osh-echo the -E flag might rather be a not too significant *missing feature*, so also not worth mentioning in the docs.

### DIFF
--- a/doc/ref/toc-ysh.md
+++ b/doc/ref/toc-ysh.md
@@ -127,8 +127,8 @@ Siblings: [OSH Topics](toc-osh.html), [Data Topics](toc-data.html)
                   is-main                false when sourcing a file
                   use                    change first word lookup
   [I/O]           ysh-read               flags --all, -0
-                  ysh-echo               print args on line, flawless simple_echo <!-- after #1772: print args on line, no echo_flags flaws -->
-                  write                  Like echo, with --, --sep, --end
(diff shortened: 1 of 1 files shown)


bar-g:
Sorry, I seem to have pressed the wrong button in responding. It started separate "reviews".

andychu:
I'm going to close this since I'm getting lost in all the threads

- `-E` isn't implemented, so this is premature
- the other changes don't seem like improvements, which I mentioned

PR #1848, opened 2024-02-23
[toc-ysh] complete "write" short description

bar-g:
This adjusts the "write"-side, to show the relation between ysh-echo and write in the toc overview.

(echo-side is 1847)


bar-g:
merged change into #1847

PR #1855, opened 2024-03-03
Add OILS-ERR-200 - Unexpected typed args passed to an external command

PossiblyAShrub:
I came across this while working on `args.ysh` and got this confusing error message when I forgot to `source --builtin args.ysh`:

```
  arg-parse (&spec) {
            ^
[ stdin ]:1: fatal: Unexpected typed args passed to external command 'arg-parse'
```

### DIFF
--- a/core/executor.py
+++ b/core/executor.py
@@ -311,8 +311,9 @@ class ShellExecutor(vm._Executor):
         environ = self.mem.GetExported()  # Include temporary variables
 
         if cmd_val.typed_args:
-            e_die('%r appears to be external. External commands don\'t accept typed args (OILS-ERR-200)' % arg0,
-                  cmd_val.typed_args.left)
+            e_die(
+                '%r appears to be external. External commands don\'t accept typed args (OILS-ERR-200)'
+                % arg0, cmd_val.typed_args.left)
 
         # Resolve argv[0] BEFORE forking.
         if run_flags & USE_DEFAULT_PATH:
--- a/core/executor.py
+++ b/core/executor.py
@@ -311,7 +311,7 @@ class ShellExecutor(vm._Executor):
         environ = self.mem.GetExported()  # Include temporary variables
 
         if cmd_val.typed_args:
-            e_die('Unexpected typed args passed to external command %r' % arg0,
+            e_die('Unexpected typed args passed to external command %r (OILS-ERR-200)' % arg0,
                   cmd_val.typed_args.left)
 
         # Resolve argv[0] BEFORE forking.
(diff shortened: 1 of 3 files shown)


doc/error-catalog.md
```diff
@@ -169,7 +169,21 @@ They're numbered starting from `200`.
...
+```
+  cat ("myfile")
+      ^
+[ -c flag ]:1: fatal: Unexpected typed args passed to external command 'cat' (OILS-ERR-200)
+```
+
+- External programs cannot accept [typed
+  arguments](ref/chap-cmd-lang.html#typed-arg). Try passing untyped arguments
+  instead. (eg. `cat myfile`)
+- Did misspell a [YSH proc](ref/chap-cmd-lang.html#proc-def)?
```
andychu, line 186: You can also add the exact problem you had!  

> Did you forget to source a file that contains the proc / shell function you wanted to run?  If a name is not found, YSH assumes it's an external command.

andychu (commented):
Looks good!

I wonder if we should consider changing the error, like

```
'arg-parse' appears to be an external command, and it got unexpected typed args
```

I think this error could be fairly common

andychu:
Or maybe:

> 'arg-parse' appears to be an external command, which you can't pass typed args to

or

> 'arg-parse' appears to be external.  External commands don't accept typed args

doc/error-catalog.md
```diff
@@ -169,7 +169,21 @@ They're numbered starting from `200`.
...
+test/runtime-errors.sh test-external_cmd_typed_args
+-->
+
+```
+  cat ("myfile")
+      ^
+[ -c flag ]:1: fatal: Unexpected typed args passed to external command 'cat' (OILS-ERR-200)
+```
+
+- External programs cannot accept [typed
```
andychu, line 183: I might also rephrase this to compare / contrast

- Builtin commands and user-defined procs may accept [typed args], but external commands never do.

core/executor.py
```diff
@@ -311,8 +311,9 @@ def RunSimpleCommand(self, cmd_val, cmd_st, run_flags):
         environ = self.mem.GetExported()  # Include temporary variables
 
         if cmd_val.typed_args:
-            e_die('Unexpected typed args passed to external command %r' % arg0,
-                  cmd_val.typed_args.left)
+            e_die(
+                '%r appears to be external. External commands don\'t accept typed args (OILS-ERR-200)'
```
andychu, line 315: Generally speaking I use single quotes unless there's a single quote, then I switch to double quotes, to avoid \

But this is a small thing

andychu:
Looks great, thank you!  

I agree this one is very common -- I've run into it

The other one I ran into the other day was `if ($x)` vs `if (x)`

From the Zulip feedback, the command/expression distinction is not obvious to people, so we have to do some work there.  I also would like coarse syntax highlighting for that reason

Issue #1837, opened 2024-02-13
ysh-simple to compose simpler


bar-g:
I was wondering about what you meant when you wrote about having "Lazy Arg Lists", and found in the doc it's actually named it lazy-expr-arg (https://www.oilshell.org/release/0.20.0/doc/ref/chap-cmd-lang.html#YSH-Simple)

After reading it I felt there may even be some more potential to simplify ysh-simple, also for improving on the ability to compose.

If command mode could also accept multiple typed args in separate parens and brackets, there would simply only be typed-args and expr-args.

```
myproc word (x, y; n=3)                         # one word and three typed arguments
myproc word (x) (y) (;) (n=3)                   # synonym
myproc word (x) [count > threshold] (;) (n=3)   # this could also support multiple expression args

```

Could this be handy and simple to understand? The ysh-simple doc would only have to present non-exclusive typed-arg, expr-arg and block-arg.

bar-g:
Maybe the semicolon `(;)` could even be made unnecessary, by parsing/filtering out all named args first **no matter where they are positioned**?

The remaining args are then words, typed, expressions or blocks, all in their own order.
This could allow for more common command flexibility, for example:
`myproc [count > threshold] word1 (->x) word2 (->y)`

bar-g:
Spreads could maybe be passed as `(positional;)` and `(;named)`.

bar-g:
What? Maybe think a bit about `myproc "$@"`.

The `(;)` was dealt with in the first comments topic.

bar-g:
Just noticed the general case for `myproc --name1=(expr)` would need to be `myproc --name1="( expr )"` (double quoted), to allow spaces of the expression in the word.

bar-g:
Updated the example, showing `@ARGV` vs. `= ARGV` behavior.

bar-g:
> `p @words (...typed; ...named; block)`

But these only iterate over argument types.

Whereas in the command interface language "$@" iterates over _**all**_ flags, options, and arguments, i.e. words, and may also be called argument vector.

I notice, that you didn't use `@ARGV` in your the example above, and I think that is correct and important.

I'd say `"$@"` **_and_** `@ARGV` are on the shell/command language side of things.

And two important properties of the shell command language are:
* The flexibility it has to construct combinations of commands + flags + options + arguments, for example to meet domain specific requirements or common practice.
*  Ability to use xargs, e.g. for mapping pipes into such command vectors.

[...]

PR #1856, opened 2024-03-03
Misc. Argparse Updates

PossiblyAShrub:
I started adding support for `default` in `parseArgs`, but then got carried away :P

## Changes

- Set `default` values for flags and args in `parseArgs`
- Add `doc/ref` entries for `Args Parser`, `parser`, `flag`, `arg`, `rest` and `parseArgs()`
- Enforce required positional arguments
- Move default and help into named arguments for `flag` and `arg`
- Raise an error on duplicate flag/arg names

### DIFF
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1187,7 +1187,7 @@ Flags:
 Bash has this, but OSH won't implement it.
 
 
-## Args
+## Args Parser
 
 YSH includes a command-line argument parsing utility called `parseArgs`. This
 is intended to be used for command-line interfaces to YSH programs.
@@ -1300,7 +1300,6 @@ Default values for an argument can be set with the `default` named argument.
     # => args.count = 2
     # => args.optimize = true
 
-
 Each name passed to `flag` must be unique to that specific `parser`. Calling
 `flag` with the same name twice will raise an error inside of `parser`.
 
@@ -1351,7 +1350,7 @@ strings (usually `ARGV`.)
 
 The returned `args` is a dictionary mapping the names of each `arg`, `flag` and
 `rest` to their captured values. (See the example at the [start of this
-topic](#Args).)
+topic](#Args-Parser).)
 
 `parseArgs` will raise an error if the `ARGV` is invalid per the parser
 specification. For example, if it's missing a required positional argument:
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1214,7 +1214,6 @@ Then, create an argument "spec":
 
       arg src (help='Source')
       arg dest (help='Dest')
-      arg times (help='Foo')
 
       rest files
     }
@@ -1224,17 +1223,16 @@ Finally, parse `ARGV` (or any other array of strings) with:
     var opt, i = parseArgs(spec, ARGV)
 
 The returned `opt` is a `Dict` containing key-value pairs with the parsed
-values (or defaults) for each flag and argument. For example, given `ARGV = :|
--v -P 12 |`, `opt` would be:
+values (or defaults) for each flag and argument. For example, given
+`ARGV = :| mysrc -P 12 mydest a b c |`, `opt` would be:
 
     {
-        "verbose": true,
+        "verbose": false,
         "max-procs": 12,
         "invert": true,
-        "src": null,
-        "dest": null,
-        "times": null,
-        "files": [],
+        "src": "mysrc",
+        "dest": "mydest",
+        "files": ["a", "b", "c"]
     }
 
 `i` is the last parsed index into `ARGV` which is useful if you wanted to do
@@ -1242,6 +1240,6 @@ further parsing of `ARGV` after `parseArgs`.
 
 Inside of `arg-parse`, there are three ways you can add to the spec:
 
-- `flag (short, long ; type='bool' ; default=null, help=null)` -- For "flags" like `--verbose` or `-N`
-- `arg (name ; ; default=null, help=null)` -- For positional arguments
+- `flag (short, long ; type='bool' ; default=null, help=null, required=false)` -- For "flags" like `--verbose` or `-N`
+- `arg (name ; ; default=null, help=null, required=true)` -- For positional arguments
 - `rest (name)` -- Take the remaining positional arguments and store them in `opts[name]`
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1277,7 +1277,7 @@ The above example declares a flag "--verbose" and a short alias "-v".
 - `true` if the flag was passed at least once
 - `false` otherwise
 
-Flags can also accept values, for example, if you wanted to accept an integer count:
+Flags can also accept values. For example, if you wanted to accept an integer count:
 
     parser (&spec) {
       flag -N --count ('int')
@@ -1337,6 +1337,9 @@ Capture zero or more positional arguments not already captured by `arg`. So,
 for `ARGV = :| hello file.txt message.txt README.md |`, we would have
 `args.query = "file.txt"` and `args.files = ["file.txt", "message.txt",
 "README.md"]`.
+
+Without rest, passing extraneous arguments will raise an error in
+`parseArgs()`.
 
 `rest` can only be called _once_ within a `parser`. Calling it multiple times
 will raise an error in `parser`.
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1186,3 +1186,62 @@ Flags:
 
 Bash has this, but OSH won't implement it.
 
+
+## Flags
+
+### parseArgs()
+
+YSH includes a `argparse`-like utility called `parseArgs`. This is intended to
+be used for command-line interfaces to programs.
+
+To use it, first import `args.ysh`:
+
+    source --builtin args.ysh
+
+Then, create an argument "spec":
+
+    arg-parse (&spec) {
+      flag -v --verbose (help="Verbosely")  # default is Bool, false
+
+      flag -P --max-procs ('int', default=-1, doc='''
+        Run at most P processes at a time
+        ''')
+
+      flag -i --invert ('bool', default=true, doc='''
+        Long multiline
+        Description
+        ''')
+
+      arg src (help='Source')
+      arg dest (help='Dest')
+      arg times (help='Foo')
+
+      rest files
+    }
+
+Finally, parse `ARGV` (or any other array of strings) with:
+
+    var opt, i = parseArgs(spec, ARGV)
+
+The returned `opt` is a `Dict` containing key-value pairs with the parsed
+values (or defaults) for each flag and argument. For example, given `ARGV = :|
+-v -P 12 |`, `opt` would be:
+
+    {
+        "verbose": true,
+        "max-procs": 12,
+        "invert": true,
+        "src": null,
+        "dest": null,
+        "times": null,
+        "files": [],
+    }
+
+`i` is the last parsed index into `ARGV` which is useful if you wanted to do
+further parsing of `ARGV` after `parseArgs`.
+
+Inside of `arg-parse`, there are three ways you can add to the spec:
+
+- `flag (short, long ; type='bool' ; default=null, help=null)` -- For "flags" like `--verbose` or `-N`
+- `arg (name ; ; default=null, help=null)` -- For positional arguments
+- `rest (name)` -- Take the remaining positional arguments and store them in `opts[name]`
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1203,11 +1203,11 @@ Then, create an argument "spec":
     arg-parse (&spec) {
       flag -v --verbose (help="Verbosely")  # default is Bool, false
 
-      flag -P --max-procs ('int', default=-1, doc='''
+      flag -P --max-procs ('int', default=-1, help='''
         Run at most P processes at a time
         ''')
 
-      flag -i --invert ('bool', default=true, doc='''
+      flag -i --invert ('bool', default=true, help='''
         Long multiline
         Description
         ''')
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1191,16 +1191,16 @@ Bash has this, but OSH won't implement it.
 
 ### parseArgs()
 
-YSH includes a `argparse`-like utility called `parseArgs`. This is intended to
-be used for command-line interfaces to programs.
+YSH includes a command-line argument parsing utility called `parseArgs`. This
+is intended to be used for command-line interfaces to YSH programs.
 
 To use it, first import `args.ysh`:
 
     source --builtin args.ysh
 
-Then, create an argument "spec":
+Then, create an argument parser:
 
-    arg-parse (&spec) {
+    arg-parse (&parser) {
       flag -v --verbose (help="Verbosely")  # default is Bool, false
 
       flag -P --max-procs ('int', default=-1, help='''
@@ -1220,9 +1220,9 @@ Then, create an argument "spec":
 
 Finally, parse `ARGV` (or any other array of strings) with:
 
-    var opt, i = parseArgs(spec, ARGV)
+    var args, last_idx = parseArgs(parser, ARGV)
 
-The returned `opt` is a `Dict` containing key-value pairs with the parsed
+The returned `args` is a `Dict` containing key-value pairs with the parsed
 values (or defaults) for each flag and argument. For example, given
 `ARGV = :| mysrc -P 12 mydest a b c |`, `opt` would be:
 
@@ -1235,11 +1235,11 @@ values (or defaults) for each flag and argument. For example, given
         "files": ["a", "b", "c"]
     }
 
-`i` is the last parsed index into `ARGV` which is useful if you wanted to do
-further parsing of `ARGV` after `parseArgs`.
+`last_idx` is the last parsed index into `ARGV` which is useful if you wanted
+to do further parsing of `ARGV` after `parseArgs`.
 
-Inside of `arg-parse`, there are three ways you can add to the spec:
+Inside of `arg-parse`, there are three ways you can add to the parser:
 
-- `flag (short, long ; type='bool' ; default=null, help=null, required=false)` -- For "flags" like `--verbose` or `-N`
-- `arg (name ; ; default=null, help=null, required=true)` -- For positional arguments
+- `flag (short, long ; type='bool' ; default=null, help=null)` -- For "flags" like `--verbose` or `-N`
+- `arg (name ; ; help=null)` -- For positional arguments
 - `rest (name)` -- Take the remaining positional arguments and store them in `opts[name]`
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1187,9 +1187,7 @@ Flags:
 Bash has this, but OSH won't implement it.
 
 
-## Flags
-
-### parseArgs()
+## Args
 
 YSH includes a command-line argument parsing utility called `parseArgs`. This
 is intended to be used for command-line interfaces to YSH programs.
@@ -1198,9 +1196,9 @@ To use it, first import `args.ysh`:
 
     source --builtin args.ysh
 
-Then, create an argument parser:
+Then, create an argument parser **spec**ification:
 
-    arg-parse (&parser) {
+    parser (&spec) {
       flag -v --verbose (help="Verbosely")  # default is Bool, false
 
       flag -P --max-procs ('int', default=-1, help='''
@@ -1220,11 +1218,11 @@ Then, create an argument parser:
 
 Finally, parse `ARGV` (or any other array of strings) with:
 
-    var args, last_idx = parseArgs(parser, ARGV)
+    var args = parseArgs(spec, ARGV)
 
 The returned `args` is a `Dict` containing key-value pairs with the parsed
 values (or defaults) for each flag and argument. For example, given
-`ARGV = :| mysrc -P 12 mydest a b c |`, `opt` would be:
+`ARGV = :| mysrc -P 12 mydest a b c |`, `args` would be:
 
     {
         "verbose": false,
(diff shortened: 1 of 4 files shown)


doc/ref/chap-builtin-cmd.md
```diff
@@ -1186,3 +1186,60 @@ Flags:
...
+
+      arg src (help='Source')
+      arg dest (help='Dest')
+
+      rest files
+    }
+
+Finally, parse `ARGV` (or any other array of strings) with:
+
+    var opt, i = parseArgs(spec, ARGV)
```
andychu, line 1223: 'opt' is an OK name, but can also be confused with optimization or something

I kind of like the idea that you "parseArgs" and you get "flags"

So then  flag.verbose rather than opt.verbose, because it's concrete.  It's not strictly accurate because you also have flag.src, which is not a flag, but somehow I think that is not horrible

---

Again a separate conversation, although I guess we are documenting it now, so maybe now is the time to flesh out the names

stdlib/args.ysh
```diff
@@ -167,5 +190,26 @@ func parseArgs(spec, argv) {
     setvar args[spec.rest] = rest
   }
 
+  # Set defaults for flags and args
+  for flag in (spec.flags) {
+    if (flag.name not in args) {
+      if (flag.required) {
```
andychu, line 196: I'm not sure we should have the idea of "required flags"

I believe most command line programs follow the convention that

- "required" things are ALWAYS args
- but some args might not be required
- flags are always optional

If you look at `grep`, that's true

```
grep mypat  # stdin

grep mypat MYFILE

grep -C 1 mypat  # optional flag

```

i.e. there are no required flags

I think it's OK for the user to add their own validation with a BETTER error in case they want required flags, like

```
if (flag.message === null) {
  error 'Require --message' (status=2)
}
```

stdlib/args.ysh
```diff
@@ -74,11 +85,18 @@ proc flag (short, long ; type='bool', default=null, help=null) {
...
+  }
+
+  # TODO: Should use "trimPrefix"
+  var name = long[2:]
+
+  ctx emit flags ({short, long, name, type, default, help, required})
 }
 
-proc arg (name ; ; default=null, help=null) {
+proc arg (name ; ; default=null, help=null, required=true) {
```
andychu, line 99: Hm what happens if you do this?

```
arg pos1 (required=false)
arg pos2 (required=true)
```

Then you have something not required before something required

I think it could make sense for all args to be required

And then the user can use `rest args` if they want their own validation



```
```

stdlib/args.ysh
```diff
@@ -60,10 +60,21 @@ proc arg-parse (; place ; ; block_def) {
...
+
+      setvar names[x.name] = null
+    }
+  }
+
   call place->setValue(p)
 }
 
-proc flag (short, long ; type='bool', default=null, help=null) {
+proc flag (short, long ; type='bool' ; default=null, help=null, required=false) {
```
andychu, line 77: hm we didn't talk about this but I also think we should allow all of

```
flag --long-only
flag -v
flag -v --verbose
```

I believe this is common, if you look at `grep --help`, etc.  There are some flags that are only long

This requires some more subtle validation, but we can do it with `...word_args`

andychu (commented):
Thanks for the update!


- I like default and help being keyword args -- good
- I want to argue that flags should always be optional, and args should always be required -- by our framework   
  -  The user still has flexibility to make optional args (`rest`) and required flags (default and explicit check)
- I noted that I think `flag --long` only is OK, but that doesn't have to be part of this PR
- naming is still an annoyingly hard problem :)

andychu:
I also think the  `args <=> required`, `flags <=> optional` rule kind of mirrors the two kinds of typed arguments

`positional <=> required`,  `named <=> optional`

It's sort of an easy rule to remember I think?

andychu:
Just to throw something concrete out there, I think

```
parser (&spec) {
  ...
}

var flag, i = parseArgs(spec, ARGV)
```

seems fairly consistent.  You make a parser, and then you parser with it.

---

Sometimes I like doing `parser (&p) { ... }`, but some people may object to single letter vars, especially for globals

bar-g:
> I also think the args <=> required, flags <=> optional rule kind of mirrors the two kinds of typed arguments

Yes, exactly. And flags and options map to named args.
More parallels and nice defaults kept coming up after my first idea to composable args #1837.

I think there may be three different general ways to organize positional arg lists, and all have their merits and valid use cases:

```
# 1.) For iterating through the CLI call references (words),
#    with a Forth-like Quality (not against grain) (https://www.oilshell.org/blog/2017/01/13.html)
@ARGV    # list of words (command as dispatched, with flags, opts, args)
         # NOTE: containing complete typed-CLI that supports typed args, e.g.:
         # myproc --flag -f some.file --option=(var) -- (arg1) word (arg2) word [arg3] {...}

# possibly also make list results from parsing available for 'for'-loop usage?:
@FLAGS   # list of words (flags)
@OPTIONS # dict of words (opt='value', also opt=(val), all expanded to long opts if spec avail.?)
@ARGS    # list of words, also (arg), [arg], {...}

shift    # removes elements from the beginning of ARGV and corresponding others above 


# 2.) "typed-CLI" For iterating over the passed, and already parsed arguments (per type).
# (e.g. for each dict arg found in ARGV take two word args)

WORDV   # list of untyped (string) words found in ARGV
NAMEDV  # dict of named args (as found within ARGV)
        # includes  --flags as NAMEDV['flag'] = true (Bool),
        # and --opt="value" as NAMEDV['opt']  = "value"
BLOCKV  # list of block type 

TYPEDV  # list of all other typed args (as found within ARGV)
        # also/or? have lists of each type?


# 3.) "typed-func interface" For accessing a group of predefined function call params
#    (i.e. ';'-separated), optionally iterating over additional splats
myproc (a, b ; x, y ; name='value') {...}
named vars from positional-args, named-args and named splats
```

And I think these could ideally work automatically by default, and leaving "specs" mostly for overriding or refining defaults, e.g. having short and long options etc.

This should allow for composable things like `proc repeat` and `proc timeout` or for a simple `runproc myproc @ARGV` or `runproc @ARGV` at the end of a script to be enough to provide CLI access.

andychu:
OK let me reply on that design proposal, but bottom line is that I don't think it's a good idea to **automatically** map between typed args and string flags

It's too much magic

Raku has that, and I don't really see that it's a feature people like

You always need a little bit of code to bridge the gap.  Our philosophy is not to make everything declarative, but to allow them mix imperative and declarative.  This makes the whole system smaller

doc/ref/chap-builtin-cmd.md
bar-g: For my two cents, I find 'opt' ok, not much to confuse in relation to parsing args, or is it?

I find `flag.something` that isn't a flag is much worse (it's concretely wrong).

If you really think opt could be confused with optimization here, maybe the full word option could also work.

doc/ref/chap-builtin-cmd.md
```diff
@@ -1186,3 +1186,60 @@ Flags:
...
+YSH includes a `argparse`-like utility called `parseArgs`. This is intended to
+be used for command-line interfaces to programs.
+
+To use it, first import `args.ysh`:
+
+    source --builtin args.ysh
+
+Then, create an argument "spec":
+
+    arg-parse (&spec) {
```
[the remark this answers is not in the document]
bar-g: Ok, my first pass on this whole naming problem...
Building on these conventions:

    args   are the ones passed  to  funcs/procs (and cli?)
    params are the ones defined for funcs/procs (and cli?)


I came to this idea:
```
cli-define (&cli_params) {
   flag
   arg
   rest
}

var params, parsed_idx = parseArgs(ARGV, cli_params)
```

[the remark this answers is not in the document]
bar-g: If `var params` is a too broad global variable name, it could be `var cli_params`,
and `cli-define (&cli_definition)` (maybe even more compelling).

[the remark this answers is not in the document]
bar-g: Singular names for dicts? `param.key` or `params.key`?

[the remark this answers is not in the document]
bar-g: I really think the cli definition should have the `option` definition separatly from `flag` definitions.

It's adheres to common meaning, brings structure into the cli_definition, and defining a flag becomes a nice shortcut, a flag is an option of type bool with implicit default=true.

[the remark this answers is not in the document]
bar-g: If the convention is to capitalize types would it make sense to require corresponding strings in the definitions, e.g 'Bool' or 'Int'?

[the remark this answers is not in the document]
bar-g: probably better: `parse_args_idx`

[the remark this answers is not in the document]
bar-g: When starting from the example given at https://www.oilshell.org/blog/2023/06/ysh-sketches.html#declare-data-hay-aint-yaml-argparse I came to this idea to make code better readable:

```
Args (&mycp_cli) {  # mycp_cli is data created with code
  flag -v --verbose "Show verbose output"
  flag -R --recursive "Copy recursively"
  opt -f --filter default=null "Possibly only copy camelCase files?"
  arg src
  arg dest      
}

var flag, opt, arg, pos = parseArgs(mycp_cli, ARGV)
if (opt.filter) { # e.g. modify arg.src }
cp $[arg.src] $[arg.dest]
```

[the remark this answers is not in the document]
bar-g: s/opt/option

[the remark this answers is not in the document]
andychu: Aidan posted a survey of Python argparse, nodejs yargs, and Rust clap on Zulip:

https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/parseArgs.20Design.20Survey

We should probably move the discussion there

PossiblyAShrub:
I have settled on naming:

```
parser (&spec) {  # spec for "specification"
  flag -v --verbose
  arg query
  rest files
}

# last_index has been removed in favour of `rest rest` for `parseArgs` chaining
var args = parseArgs(spec, ARGV)
```

The `required` named arguments have also been removed.

I also filled in the `doc/ref` entry for the `Args` topic. This covers `parser`, `flag`, `arg`, `rest` and `parseArgs`. It's just a first draft, so I need to edit it, but any first opinions are welcome :)

While writing in `doc/ref`, I uncovered the some limitations, issues and design decisions. I've also got some TODOs littered throughout `args.ysh`. Here they are for reference. Most of them can be left for future PRs.

- How can we explicitly pass false to a boolean flag? `--verbose=false`?
  - Maybe `--no-XXXX` variants of flags? (eg. --no-verbose)
- Chaining parsers / sub-commands (more about this on the "parseArgs survey" zulip thread)
  - Either will allow parser nesting
  - Or can use `rest rest` and `parseArgs` again on `rest`
- Need to document the help named argument. Punting while we do not generate help messages
- Missing some types from `flag (type=...)`. We also need to validate this somehow... (really, this depends on type objects like `Str` and `Bool`)
- Need to raise an error for `flag --name` and then `arg name`? (I should fix that in this PR...)

bar-g:
How would you `shift` without pos/index?

Couldn't `proc parser` rather be `func defineArgs`? IMHO plays better with `parseArgs`.

As code is read more often then written, this should also make for more readable code:

```
defineArgs (&myproc_cli) {
   flag ...
   option ...
   arg ...
   unparsed
}

var flag, option, arg, unparsed = parseArgs(myproc_cli, ARGV)

# Now the code talks about concrete things like `flag.verbose` (a bool),
# `option.max-procs`, `arg.src`, `arg.dest` etc.
```

bar-g:
Oh, `unparsed` wouldn't even need to be called within `defineArgs`.

andychu:
We did a survey on Zulip, and almost all libraries call it "parser" or "argparser"

https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/parseArgs.20Design.20Survey

Having separate `flag option arg` isn't terrible, but the problem is that you will often pass this thing to more procs and functions.  So it's more annoying to pass 3 things around than 1.

And there are no name conflicts, so it's best to use 1 thing.

bar-g:
ok, yep, and thanks (still not used to passing contexts)

PossiblyAShrub:
To address @bar-g's comments a bit more:

> How would you `shift` without pos/index?
> ```
> defineArgs (&myproc_cli_spec) {
>    [...]
>    unparsed
> }
> 
> var flag, option, arg, unparsed = parseArgs(myproc_cli_spec, ARGV)
> ```
>
> [...]
>
> Oh, `unparsed` wouldn't even need to be called within `defineArgs`.

We don't really need a `last_index` or `unparsed` as these are both covered by `rest`. For example, to get a "git-like" interface:

```
parser (&spec) {
  flag -v --verbose
  arg action
  rest rest
}

parser (&commit_spec) {
  flag -m --message ('str')  # Would be Str once we have type objects
}

var args = parseArgs(spec, ARGV)

case (args.action) {
  commit {
    var commitArgs = parseArgs(commit_spec, args.rest)
  }
  ...
}
```

Shifting can then be done by mutating `args.rest` (a list.)

I also had this idea to nest `parser`s (example below) but that's for another PR.

```
parser (&gitSpec) {
  program "git"

  flag -v --verbose

  parser (&diffSpec) {
    program "diff"

    flag --staged
    rest refs
  }
  subcommand (diffSpec)  # Invoked with `git [-v|--verbose] diff [...args]`
}
```

> Couldn't `proc parser` rather be `func defineArgs`? IMHO plays better with `parseArgs`.
>
> As code is read more often then written, this should also make for more readable code:
> 
> ```
> defineArgs (&myproc_cli_spec) {
>    flag ...
>    option ...
>    arg ...
>    unparsed
> }

Well, first we should really make sure that we call this parser. I've found that most argument parsing libraries use this term -- a few actually do not name this at all!

So then, why not `proc defineArgsParser`? It's a good name, but it's long. Argument parsing is supposed to be a really common part of YSH scripts. As a form of "huffman-coding", we want to make these common parts shorter to write.

Also, with some ideas around modules, it may be possible that users refer to `parser` with `args parser (&spec) { ... }`. But that's probably too theoretical to be a good argument.

Finally, I should note that, even without "arg" in the name, the reader should only need to scan ahead a few lines to fully disambiguate that this is an "argument parser." For example:

```
parser (&spec) {
  flag -v --verbose  # Ah! this is concerned with arguments/flags
  ...
```

> ```
> var flag, option, arg, unparsed = parseArgs(myproc_cli_spec, ARGV)
>
> # Now the code talks about concrete things like `flag.verbose` (a bool),
> # `option.max-procs`, `arg.src`, `arg.dest` etc.
> ```

My main gripe with this is when you don't have any one of `option`s, `flag`s or `arg`s. You have to remember the order and may get a bunch of `_`s everywhere. For example:

```
var _flags, _options, args, _unparsed = parseArgs(spec, ARGV)
var _, _, args, _ = parseArgs(spec, ARGV)  # I don't think we can even reuse `_`...
```

You could then argue that we pack this in a `Dict` like so:

```
var parsed = parseArgs(spec, ARGV)

# Super explicit
= parsed.args.src
= parsed.args.options.maxProcs
= parsed.args.flags.verbose
```

My argument above no longer holds. But now there is another issue:
- This is long and perhaps overly explicit
- I am not sure why we really want to separate `args` from `options` (ie. is the explicitness beneficial?)
  - Keep in mind that there shouldn't be name conflicts between args and flags/options

---

I also notice that you are advocating for making `flag` separate from `option`. It would be nice to reduce the number of "atoms" you need to build an argument parser spec. Do you think that `flag (type=...)` is somehow worse than a separate `option (type=<type>)` where `type != Bool`?

bar-g:
Thanks, for the elaborate reply. I was mainly thinking of making the resulting code "speak", yet the argument needing to pass only one `args` dict instead of several tipped it.

Only two things I still have.

* In the doc maybe have:

```
    ## parseArgs
    ### parser spec   (oh, now that I write this, what about parserSpec() ???)
    ### parseArgs     (not parser)
    ...
```

* One thing that could be a bit more readable and understandable: use `rest unparsed` instead of `rest rest` example?

bar-g:
Sorry, got it, command mode, not func. So maybe actually `### parser (&spec)` as a subsection.

PossiblyAShrub:
About the docs, we already do have that, I think? (I've written `parser` instead of `parser (&spec)` in the [TOC](http://travis-ci.oilshell.org/github-jobs/6420/ovm-tarball.wwz/_release/VERSION/doc/ref/toc-ysh.html#builtin-cmd) for consistency.) You can view the docs for my latest commit [here](http://travis-ci.oilshell.org/github-jobs/6420/ovm-tarball.wwz/_release/VERSION/doc/ref/chap-builtin-cmd.html#Args).

> One thing that could be a bit more readable and understandable: use `rest unparsed` instead of `rest rest` example.

So you mean:

```
parser (&spec) {
  flag -v --verbose
  arg action
  rest unparsed  # CHANGED
}

parser (&commit_spec) {
  flag -m --message ('str')  # Would be Str once we have type objects
}

var args = parseArgs(spec, ARGV)

case (args.action) {
  commit {
    var commitArgs = parseArgs(commit_spec, args.unparsed)  ## CHANGED
  }
  ...
}
```

?

If we do go this way, I think `rest unparsed` is better than `rest rest`. However, I'm still not sure if this is the exact style that we will endorse.

bar-g:
> So you mean: [...] parseArgs(commit_spec, args.unparsed)  ## CHANGED

Yeah, that's what I meant in this direction.

Docs: Sorry, what I meant was:
* Maybe change the title `## Args` into `## parseArgs`? But now, comparing with the others, maybe yet something better... "CLI", "UI",  "Calls", "Command Parsing",... "CLI Parsing"?

Currently `### parse` comes only after explaining `parse (&spec)` and  before explaining use of `parseArgs`.
* change current  `### parse`, to `parseArgs`?
* re-insert  `### parse`, to actually come before `parse`?

> My main gripe with this is when you don't have any one of options, flags or args. You have to remember the order and may get a bunch of _s everywhere.

Hm, my thought here was `parseArgs` could return a list created by appending the dicts `option`, `flag`, `arg` and the list `unparsed`, in whatever order they where used in the `parser` (or `parser-spec`?) proc, with `unparsed` required to be last.

So, just leaving out non-used dicts in the returned list, and thus allow to whip up:

```
parser (&spec) { flag -v verbose }    # maybe even?: parser (&p) flag -v verbose   
var flag = parseArgs(spec, ARGV)

# or maybe even all-in-one together?:
var flag = parseArgs('flag -v verbose', ARGV)
```

> I am not sure why we really want to separate args from options (ie. is the explicitness beneficial?)

I guess, because they usually need to be handled differently, i.e. iterating over them, requiring, verification, errors, etc.

> I also notice that you are advocating for making flag separate from option. It would be nice to reduce the number of "atoms" you need to build an argument parser spec.

I think that reducing the widely known flag/option/arg distinction would make the following script much worse to understand.

Hm, how worse would passing multiple dicts be? `sub-proc (flag, option, arg)`, ok at least I see plural form is better  `sub-proc (flags, options, args)`.

bar-g:
> Shifting can then be done by mutating args.rest (a list.)

Hm, good point.

So, I see, for a complex CLI a first pass of parseArgs() could just collect and remove only a couple optional global flags and `--` and create the list args.unparsed. And this list could then be searched for keywords or symbols to segment the entire command into parts. Which can be sent through parseArg() individually, and all of them together finally compose the complete operation.

PossiblyAShrub:
> * Maybe change the title `## Args` into `## parseArgs`? But now, comparing with the others, maybe yet something better... "CLI", "UI",  "Calls", "Command Parsing",... "CLI Parsing"?

Good point. "Args" could be more descriptive. I'm thinking "Arg Parser" or, if I have room, "Arg Parsing". (It's unfortunate that our current TOC layout enforces short topic names...)

> Hm, my thought here was `parseArgs` could return a list created by appending the dicts `option`, `flag`, `arg` and the list `unparsed`, in whatever order they where used in the `parser` (or `parser-spec`?) proc, with `unparsed` required to be last.

Ah, that feels dangerous. I don't know of any other API which reorders results like that. It's a subtle issue that I would encounter only after using the API for a _significant_ amount of time. For example, I may (incorrectly) assume that the order is fixed -- until I reorder my `parser` block and everything breaks. Providing a variable length list may be even more dangerous. It's also harder to see, just from reading the code, if that order is correct. You'd need to check back and forth between the `parser (&spec)` block and `var ... = parseArgs(spec, ARGV)` to verify this.

> I guess, because they usually need to be handled differently, i.e. iterating over them, requiring, verification, errors, etc.

I'm still not understanding the *motivation* here. I don't think the separation alone is _useful_. It only serves to create more friction.

I can see how "flattening" everything into a single `Dict` is problematic:

```
for name, value in (args) {
  # We loose information about if `name` is a flag, option or positional
  # So now we need an oracle ... bad!
  if (oracle_isAnOption(name) and oracle_valueIsBadForOption(name, value)) {
    error "Bad value for flag --$name" (status=2)
  }
}

# However, if we knew every name was a flag, we can skip out on `oracle_isAnOption`!
for name, value in (flags) {  # flags, not args
  # But we still need to know something about what makes `value` good or bad.
  # The oracle remains!
  if (oracle_valueIsBadForOption(name, value)) {
    error "Bad value for flag --$name" (status=2)
  }
}
```

To summarize, just knowing if a param is a flag, option or positional is _not enough_ to do anything useful with it. In most cases you need _context_ around that argument. This includes tasks like:
 - Verification (What makes a value good or bad?)
 - Requiring (Not all options/args are always required. They often depend on other arguments, eg. `git commit -m`, `-m` is not always required.)
 - Errors (Good error messages need context on _why_ that state is an error.)

That context is best found by either embedding it into your program (the programmer knows what their options mean) or, for dynamic utilities over top of `arg.ysh`, more information needs to be provided about each option.

> I think that reducing the widely known flag/option/arg distinction would make the following script much worse to understand.

I don't think the distinction between flags (as options for boolean values) and options (for non-boolean values) is so extreme. Or, at least, not as prevalent as you argue. For example, here's a snippet of man page for grep:

```
     The following options are available:

     -A num, --after-context=num
             Print num lines of trailing context after each match.  See also the
             -B and -C options.

     -a, --text
             Treat all files as ASCII text.  Normally grep will simply print
             “Binary file ... matches” if files contain binary characters.  Use
             of this option forces grep to output lines matching the specified
             pattern.

     -B num, --before-context=num
             Print num lines of leading context before each match.  See also the
             -A and -C options.

     -b, --byte-offset
             The offset in bytes of a matched pattern is displayed in front of
             the respective matched line.
```

`--text`, a flag, and `--before-context <num>`, an option, are listed together and given the same name.

> So, I see, for a complex CLI a first pass of parseArgs() could just collect and remove only a couple optional global flags and `--` and create the list args.unparsed. And this list could then be searched for keywords or symbols to segment the entire command into parts. Which can be sent through parseArg() individually, and all of them together finally compose the complete operation.

Yes, exactly.

bar-g:
> \# However, if we knew every name was a flag, we can skip out on `oracle_isAnOption`!
> for name, value in (flags) {  # flags, not args
>   \# But we still need to know something about what makes `value` good or bad.

I thought with "true" (boolean) flags the case is clear. The value is good if it evaluates to true or false. So that separates them, (boolean) flags can be iterated over separtately from generic options and args. So they won't get in the way and create any friction in the implemenation of option validation.


> --text, a flag, and --before-context <num>, an option, are listed together and given the same name.

Well, ok, but that's a user facing doc. Don't think it necessarily applies to implementing (boolean) flags vs. generic options in useful ways. 

Even if I assume option validation can be made really easy, and it can fall back to default boolean validation,  I somehow still think it could beneficial to have (boolean) "feature" flags separated a bit mentally, and especially visually in the code referring to them. 

flag.debug (not required but always defined)
arg.foo (required)
option.bar (maybe there, or not)

Somehow, flags are like the list of options that don't need any further handling, these are are just always there when defined and work.

Generally, I think all flags are also options but not all options are flags.

andychu:
Hmm I didn't read the whole discussion here, but I would say

1.  We don't have to resolve everything in this PR.  As long as it's progress, that's good.  (I started the naming discussion because it was documented more)
2. We should avoid doing anything "creative".  People have different opinions about flags, but all flag libraries work.  Nobody says "I can't use Go because its flag library is too weak".   

Basically our flag library is something basic that works, and if people don't like it, they can write their own

That was the point of basing everything on the `ctx` builtin, which is a minimal mechanism that lets you write nice declarative APIs, and compose them

It's probably better to test patterns **on top** of this library rather than building everything in.

e.g. generating completion spec from the data, or generating custom help messages, or generating HTML help, etc.

or re-creating clang/npm/docker command lines, etc.  Those are usually better tests than trying to imagine what people might want

andychu:
Also, I don't see any need to distinguish between "flags" and "options" here.  I think we just have flags and arguments

---

Shell has a notion of options like

-  `sh +o errexit` and `sh +e` are equivalent
- `compgen -A action` is also weird, and `complete -A action`

However I don't think that's in scope for this flags library. That's exactly the case where you write it yourself.

e.g. I wrote a custom flag parser in Python for Oils, in part because Python's flag parsing libraries don't support that pattern.   Likewise if someone wanted to emulate `sh` CLI syntax in YSH, then the YSH language is powerful enough to let you do that.

PossiblyAShrub:

bar-g:
Concerning the "burden" of having to pass multiple dicts: I think that is a only a problem as long as typed args only compose in too limited ways.

So this shouldn't be hindering having better readable code because it refers to more contextual variables like: flag.debug, arg.foo and option.bar (which just may exist, or not).

I already have an idea to solve the composing problem, I'm working it into the consolidated issue version that I'm preparing to replace:

* #1837

We can check this later.

---

> Our flag library

> I wrote a custom flag parser in Python for Oils, in part because Python's flag parsing libraries

Hm, I'm quite wondering about all this "flags" talk, as the more general, and commonly used terms clearly seems argument parsing and options.

I just did a search for "python flag parsing" and then searching the results it hardly contained "flag". It's all argparse and options.

I'm glad that the prominent words used in the implementation are parse and Arg.

Then using _only_ the flag keyword is strange, as this feature isn't limited to flags, it supports generic options.

It may not seem as worse if it were only using "option" (i.e. for flags as well)...

However, @PossiblyAShrub already explicitly showed how flattening everything into "flags" isn't a good idea. And flattening everything into "options" would be the same. Why would you want to flatten everything into either "flag" or "option", although the difference is exactly that flags don't need any "oracle"? They are trivial and always defined. (Unlike generic "options", which are thus good to keep separate, to iterate and think about.)

bar-g:
Hm, not as long alternative for explicit flag and option packing:

```
ysh ysh-0.20.0$ var args = {}

ysh ysh-0.20.0$ setvar args['flag'] = {}

ysh ysh-0.20.0$ setvar args['flag']['debug'] = false

ysh ysh-0.20.0$ = args.flag.debug

(Bool)   false

```

bar-g:
If mulitple vars become simple to compose into proc calls, it seems parseArgs could return empty dicts or lists without problem:

    ysh ysh-0.20.0$ var a, b, c = [ {}, {key: 'value'}, {} ]
    ysh ysh-0.20.0$

bar-g:
If really wanting only selected dicts?: `parseFlags()`, `parseOptions`, `parseRest()`

doc/ref/chap-builtin-cmd.md
```diff
@@ -1186,3 +1186,188 @@ Flags:
...
+store the integer `5` under `args.count`. If the user passes in a non-integer
+value like `ARGV = :| --count abc |`, `parseArgs` will raise an error.
+
+Default values for an argument can be set with the `default` named argument.
+
+    parser (&spec) {
+      flag -N --count ('int', default=2)
+
+      # Boolean flags can be given default values too
+      flag -O --optimize ('bool', default=true)
```
andychu, line 1296: How do you set it to false?  (may have missed it on the thread)

Historically there are two solutions I think:

- GNU is `--no-optimize`
- Go is `--optimize=0` , but NOT `--optimize 0` as that's 2 words

That's the first thing that pops into my mind when it says `--verbose` is true if it's passed

----

OK now I see the TODO below!  I think we should follow some existing example like GNU, see if they have a special API for `--no-optimize` or `--optimize=0`

stdlib/args.ysh
```diff
@@ -3,50 +3,50 @@
...
 # }
 #
-# var opt, i = parseArgs(spec, ARGV)
+# var args = parseArgs(spec, ARGV)
 #
-# echo "Verbose $[opt.verbose]"
+# echo "Verbose $[args.verbose]"
 
 # TODO: See list
 # - It would be nice to keep `flag` and `arg` private, injecting them into the
```
andychu, line 30: Not related to this PR, but just reminding myself that we discussed the flag/arg namespace issue here

https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Flag.20parsing.20words.20.2F.20.22dialect.22.20scope.20-.20namespace.2C.20eval

stdlib/args.ysh
```diff
@@ -3,50 +3,50 @@
...
-#   arg times 'Foo' (Int)
+#   arg src (help='Source')
+#   arg dest (help='Dest')
+#   arg times (help='Foo')
+#
+#   rest files
 # }
 #
-# var opt, i = parseArgs(spec, ARGV)
+# var args = parseArgs(spec, ARGV)
```
andychu, line 25: Having this code example here makes me wonder if we need docstrings in YSH, probably with `'''` like Python.  Since we already have that syntax.

It also reminds me that at some point we will move the tests from spec tests to an actual YSH test framework!    In pure YSH code, so people can read the tests for examples.

That will be exciting :)

stdlib/args.ysh
```diff
@@ -60,10 +60,23 @@ proc arg-parse (; place ; ; block_def) {
     setvar p.rest = null
   }
 
+  var names = {}
+  for items in ([p.flags, p.args]) {
+    for x in (items) {
+      if (x.name in names) {
+        error "Duplicate flag/arg name $[x.name] in spec" (status=3)
```
andychu, line 67: Aside - I wonder if `(status=3)` is hard to remember (for us, not for users, where it doesn't really matter)

I think in the Python code it's easy enough to to do `raise error.Expr` or `raise error.TypeErr`

I think the main reason I created the status 3 is because bash/dash doesn't use that code, so it makes spec  tests a little more distinctive

stdlib/args.ysh
```diff
@@ -74,22 +87,30 @@ proc flag (short, long ; type='bool', default=null, help=null) {
...
-  ctx emit flags ({short, long, type, default, help})
+  # bool has a default of false, not null
+  if (type === 'bool' and default === null) {
+    setvar default = false
+  }
+
+  # TODO: validate `type`
+
+  # TODO: Should use "trimPrefix"
+  var name = long[2:]
```
andychu, line 98: We now have `name => trimLeft('--')` -- @passionsocks just landed it yesterday!

stdlib/args.ysh
```diff
@@ -129,27 +140,42 @@ func parseArgs(spec, argv) {
...
       for flag in (spec.flags) {
         if ( (flag.short and flag.short === arg) or
              (flag.long and flag.long === arg) ) {
           case (flag.type) {
             ('bool') | (null) { setvar value = true }
             int {
               setvar i += 1
-              setvar value = int(argv[i])
+              if (i >= len(argv)) {
+                error "Expected integer after '$arg'" (status=2)
```
andychu, line 157: I think `2` is probably right, but again it makes me wonder if this API is ergonomic for users

We also have that issue when you do `case (_status) { (2) { echo } }` to some degree

andychu (commented):
Thank you, great progress!  I made a bunch of notes mostly about things I thought while reading the code

We may follow up on these laters

I think you've written more YSH code than I have now :)  I think it looks pretty nice and is readable

I'm thinking about docstrings and a unit test framework

andychu:
Merged.   Thanks for writing the `doc/ref` entries and spec tests, that helps a lot in reviewing the code!

PR #1857, opened 2024-03-03
add Test for "${v[@]+foo}"

Melkor333:
As discussed [in zulip](https://oilshell.zulipchat.com/#narrow/stream/307442-nix/topic/Replacing.20bash.20with.20osh.20in.20Nixpkgs.20stdenv/near/424272584).

dash is behaving differently than bash an mksh, but I'd say we follow bash, as it makes a bit more sense? :)


andychu:
Hopefully what I said will be clear from examples in the files nearby

As long as it exits 0 with  `test/spec.sh var-op-test`, it should be right

Melkor333:
> We could also add the test case with ! for Nix, in spec/var-ref ? (that file got migrated to the new style)

AFAIU the `!` is completely irrelevant in the nix case. The substitution `${!var:+${!var}} -> ${hook[@]:+${hook[@]}}` should work without issues. But I can add it aswell to make sure the expansion after the + isn't critical.

One thing though: I figured that the `:` is irrelevant to bash. But it's only documented with `:+`. I guess it makes sense to add the `:` back to the test case? (The `oils` code doesn't seem to support this undocumented behaviour :) )

andychu:
I think the `:+` and `+`  is for empty vs. undefined (which is different in shell)

There's no harm in testing it all ways -- `+` and `:+`, and `!` and no  `!`

It's really better because often assumptions are proven to be wrong :)

andychu:
Not sure if you're still in the middle of this, but there is a failure in the CI - let me know if you need help

http://travis-ci.oilshell.org/github-jobs/6487/

Could be related to `oils_failures_allowed` ?

Melkor333:
> Not sure if you're still in the middle of this, but there is a failure in the CI - let me know if you need help
> 
> http://travis-ci.oilshell.org/github-jobs/6487/
> 
> Could be related to `oils_failures_allowed` ?

It's very strange. According to travis it has 1 failure and 1 allowed failure. I honestly don't get why it fails with exit 1!?

andychu:
Ah OK, there is a pattern of zsh failures:

http://travis-ci.oilshell.org/github-jobs/6487/ovm-tarball.wwz/_tmp/spec/osh-py/var-op-test.html

compare no zsh:

http://travis-ci.oilshell.org/github-jobs/6488/ovm-tarball.wwz/_tmp/spec/osh-py/var-op-test.html

So simply removing `zsh` from compare_shells should work!

andychu:
Arguably it's interesting to test `zsh`, but I think right now we don't really care ... 

I usually add zsh if I need another "vote" on what the behavior should be.

Melkor333:
> Arguably it's interesting to test `zsh`, but I think right now we don't really care ...
> 
> I usually add zsh if I need another "vote" on what the behavior should be.

Oh got it, I thought it only counts `osh` tests :)
Anyway i added the zsh custom behaviours. I can also revert this change and just remove zsh (e.g. to not unnecessarily increase CI load). But I thought it might be good to have it if more tests come with an alpha release :)

btw. my added test is wildly different on shells if it is `argv.p "${x:+something}"` versus just `echo "v=${x:+something}"` (bash just *deletes* the argument). I made the test so it works on all shells, but I could change it aswell.

andychu:
OK I'm back from traveling, let me look at this ... 

Sorry if you had to go through CI to test ... I think `test/spec.sh var-op-test` should pass or fail locally, so you don't have go through CI, but I noticed some of the tests aren't very isolated

andychu:
OK there was still one zsh failure, which I fixed

This is definitely a good bug, and I will look at it after I write release notes

Though I'm not 100% confident it's the same as the Nix issue.  It's a good IMO to test EXACTLY what's in the code, not just something similar.

PR #1891, opened 2024-03-29
[docs/error-catalogue] Add OILS-ERR-201, Str with arithmetic operators

PossiblyAShrub:
From zulip: [#language-design > Awk has annoying string <-> int conversion](https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Awk.20has.20annoying.20string.20.3C-.3E.20int.20conversion).

> ```
> ysh ysh-0.21.0$ = '100' + '10a'
>   = '100' + '10a'
>           ^
> [ interactive ]:6: fatal: Binary operator expected numbers, got Str and Str
> ```
> 
> Maybe worth an entry in the error catalogue? `+` technically works on number formatted strings, we should note that.

### DIFF
--- a/doc/error-catalog.md
+++ b/doc/error-catalog.md
@@ -200,13 +200,13 @@ test/runtime-errors.sh test-arith_ops_str
 [ -c flag ]:1: fatal: Binary operator expected numbers, got Str and Str (OILS-ERR-201)
 ```
 
-- The `+`, `-`, `*` and `/` operators are intended to represent their
-  _arithmetic_ counterparts.
-- The arithmetic operators can be used on strings, provided that they are
-  formatted as numbers. You can explicitly parse a string into a number with the
+- Use `++` if you intended to _concatenate_ strings or lists.
+- The arithmetic operators (`+`, `-`, `*`, `/`) may be used on strings,
+  provided that they are formatted as numbers. For example, `= '10' + '1'` will
+  result in `(Int) 11`.
+- You can explicitly parse a string into a number with the
   [`int()`](ref/chap-builtin-func.html#int) and
   [`float()`](ref/chap-builtin-func.html#float) functions.
-- Use `++` if you intended to _concatentate_ strings or lists.
 
 ## Appendix
 
--- a/doc/error-catalog.md
+++ b/doc/error-catalog.md
@@ -195,18 +195,22 @@ test/runtime-errors.sh test-arith_ops_str
 -->
 
 ```
-  = "100" + "10a"
-          ^
-[ -c flag ]:1: fatal: Binary operator expected numbers, got Str and Str (OILS-ERR-201)
+  = "age: " + 100
+            ^
+[ -c flag ]:1: fatal: Binary operator expected numbers, got Str and Int (OILS-ERR-201)
+
+  = 100 + myvar
+        ^
+[ -c flag ]:2: fatal: Binary operator expected numbers, got Int and Str (OILS-ERR-201)
 ```
 
-- Use `++` if you intended to _concatenate_ strings or lists.
-- The arithmetic operators (`+`, `-`, `*`, `/`) may be used on strings,
-  provided that they are formatted as numbers. For example, `= '10' + '1'` will
-  result in `(Int) 11`.
-- You can explicitly parse a string into a number with the
-  [`int()`](ref/chap-builtin-func.html#int) and
-  [`float()`](ref/chap-builtin-func.html#float) functions.
+- Did you mean to use `++` to concatenate strings/lists?
+- The arithmetic operators (`+`, `-`, `*`, `/`) may be used with strings,
+  provided that they are formatted as numbers. For example, `= 10 + '1'  # =>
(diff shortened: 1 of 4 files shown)


doc/error-catalog.md
```diff
@@ -187,6 +187,27 @@ test/runtime-errors.sh test-external_cmd_typed_args
...
+test/runtime-errors.sh test-arith_ops_str
+-->
+
+```
+  = "100" + "10a"
+          ^
+[ -c flag ]:1: fatal: Binary operator expected numbers, got Str and Str (OILS-ERR-201)
+```
+
+- The `+`, `-`, `*` and `/` operators are intended to represent their
```
andychu, line 203: I would start this with more "actionable" / direct advice  questions, particularly:

> Did you mean to use `++` to concatenate strings? 

> `+` is for addition on numbers, while `++` is for concenation.

I think that is going to be the 99% case

Also might be reasonable to try to use something "realistic" like `= "spam" + myvar` or something

doc/error-catalog.md
```diff
@@ -187,6 +187,27 @@ test/runtime-errors.sh test-external_cmd_typed_args
...
+  _arithmetic_ counterparts.
+- The arithmetic operators can be used on strings, provided that they are
+  formatted as numbers. You can explicitly parse a string into a number with the
+  [`int()`](ref/chap-builtin-func.html#int) and
+  [`float()`](ref/chap-builtin-func.html#float) functions.
+- Use `++` if you intended to _concatentate_ strings or lists.
```
andychu, line 209: typo "concatenate"

andychu (commented):
This is great, thank you!  I agree people will run into this.

Sometimes it helps to have a "user report" so we know where the "mismatch" in thinking occurs, but here I think we can predict it


As mentioned, my comments are mostly around style -- I would make it more "pointed" about switching to `++`

There could also be other contexts that people get this error in, that require a different solution other than `++`

Maybe the number is formatted incorrectly, like they have `100K` ?  That  won't work with YSH .  (I guess the solution there is a conversion function using a regex)

---

So yeah I would start with the "short hints/tips" first, and then the links to reference material  should be lower down

doc/error-catalog.md
PossiblyAShrub: > Also might be reasonable to try to use something "realistic" like `= "spam" + myvar` or something

Hmm, I want to make sure that the example makes the error clear. I worry that `= "spam" + myvar` doesn't make the issue obvious enough.

PossiblyAShrub:
I've reordered the help items and made them more actionable. I've also fleshed out the `ysh-arith` topic and added details about string coercion. This is ready for another review.

doc/error-catalog.md
PossiblyAShrub: I decided to include two examples of the error, the Str + Str case (where concatenation should be used) and the Str + Int case (where the Str cannot be coerced).

doc/ref/chap-expr-lang.md
```diff
@@ -209,7 +209,32 @@ Note that these are distinct from `!  &&  ||`.
 
 ### ysh-arith
 
-    +  -  *  /   //   %   **
+YSH supports most of the arithmetic operators from Python. Notably, `%` differs
+from python as it represents remainder in YSH.
```
andychu, line 213: Thanks for documenting this!

I would phrase this as - `/ and %` differ from Python since they round toward zero, not negative infinity.


https://www.oilshell.org/blog/2024/03/release-0.21.0.html#integers-dont-do-whatever-python-or-c-does

So BOTH `/` and `%` in YSH are different from Python, for negative numbers.  They follow C and JavaScript more.

I would indeed call our operation "remainder", although Python's could be called "remainder" too -- because they both go with division.  (Python chooses the divmod() name for some reason)

doc/ref/chap-expr-lang.md
```diff
@@ -209,7 +209,32 @@ Note that these are distinct from `!  &&  ||`.
...
+Use `%` to compute the _remainder_ of integer division. The left operand must
+be an `Int` and the right a _positive_ `Int`.
+
+    = 1 % 2  # -> (Int) 1
+    = -4 % 2  # -> (Int) 0
+
+Use `**` for exponentiation. The left operand must be an `Int` and the right a
+_positive_ `Int`.
+
+All arithmetic operators may coerce either of their operands from strings to a
```
andychu, line 234: If we want to be more precise here, we could mention the operators that call _ConvertToInt like `//` and bit shift:


https://github.com/oilshell/oil/blob/master/ysh/expr_eval.py#L88

And then the ones that call `_ConvertToNumber` like `+ - * /`

The latter work on floats too

andychu (commented):
Thanks, this is  definitely missing documentation!

(slight clarifications)

PossiblyAShrub:
I've noted that `/` also differs from Python. I also linked to those release notes as they provide some context to _why_ we are different.

I also mentioned the difference between _CoerceToInt `//` and _CoerceToFloat `*` operators.

PR #1894, opened 2024-04-04
FIX: Correctly handle the case in Proc args binding when both  typed positional varargs and block arg are present

glyh:
Continuation of https://github.com/oilshell/oil/pull/1893. Fixes https://github.com/oilshell/oil/issues/1892, maybe fixes https://github.com/oilshell/oil/issues/1849 .

### DIFF
--- a/ysh/func_proc.py
+++ b/ysh/func_proc.py
@@ -355,7 +355,7 @@ def _BindTyped(
             num_params += len(rest_val)
             mem.SetLocalName(lval, value.List(rest_val))
         else:
-            if num_args > num_params:
+            if num_args > num_params + (1 if block_param else 0):
                 # Too many arguments.
                 raise error.Expr(
                     "%r takes %d typed args, but got %d" %
--- a/ysh/func_proc.py
+++ b/ysh/func_proc.py
@@ -330,37 +330,19 @@ def _BindTyped(
     num_args = len(pos_args)
     num_params = 0
 
-    i = 0
-
     if group:
         for p in group.params:
-            if i < num_args:
-                val = pos_args[i]
+            if num_params < num_args:
+                val = pos_args[num_params]
             else:
-                val = defaults[i]
+                val = defaults[num_params]
                 if val is None:
                     raise error.Expr(
                         "%r wasn't passed typed param %r" %
                         (code_name, p.name), blame_loc)
 
             mem.SetLocalName(LeftName(p.name, p.blame_tok), val)
-            i += 1
-        num_params += len(group.params)
-
-    # Special case: treat block param like the next positional arg
-    if block_param:
-        if i < num_args:
-            val = pos_args[i]
-        else:
-            val = defaults[i]
-            if val is None:
-                raise error.Expr(
-                    "%r wasn't passed block param %r" %
-                    (code_name, block_param.name), blame_loc)
-
-        mem.SetLocalName(LeftName(block_param.name, block_param.blame_tok),
-                         val)
-        num_params += 1
+            num_params += 1
 
     # ...rest
 
@@ -369,14 +351,30 @@ def _BindTyped(
         if rest:
             lval = LeftName(rest.name, rest.blame_tok)
 
-            rest_val = value.List(pos_args[num_params:])
-            mem.SetLocalName(lval, rest_val)
+            rest_val = pos_args[num_params:-1] if block_param else pos_args[num_params:]
+            num_params += len(rest_val)
+            mem.SetLocalName(lval, value.List(rest_val))
         else:
             if num_args > num_params:
                 # Too many arguments.
                 raise error.Expr(
                     "%r takes %d typed args, but got %d" %
                     (code_name, num_params, num_args), blame_loc)
+
+    # Special case: treat block param like the next positional arg
+    if block_param:
+        if num_params < num_args:
+            val = pos_args[num_params]
+        else:
+            val = defaults[num_params]
+            if val is None:
+                raise error.Expr(
+                    "%r wasn't passed block param %r" %
+                    (code_name, block_param.name), blame_loc)
+
+        mem.SetLocalName(LeftName(block_param.name, block_param.blame_tok),
+                         val)
+        num_params += 1
 
 
 def _BindNamed(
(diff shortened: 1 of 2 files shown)


andychu:
OK cool, the tests ran!

http://travis-ci.oilshell.org/github-jobs/6593/

There are some failures, let me look at it a bit later

I can help interpret

glyh:
I added a new spec test in spec/ysh-proc.test.sh but I'm now sure how should I inspect the information from a block.

ysh/func_proc.py
```diff
@@ -369,15 +351,31 @@ def _BindTyped(
         if rest:
             lval = LeftName(rest.name, rest.blame_tok)
 
-            rest_val = value.List(pos_args[num_params:])
-            mem.SetLocalName(lval, rest_val)
+            rest_val = pos_args[num_params:-1] if block_param else pos_args[num_params:]
```
andychu, line 354: Hmm why move `if block_param` after the `rest` section?  

Wouldn't keeping it before that simplify this condition?

ysh/func_proc.py
```diff
@@ -369,15 +351,31 @@ def _BindTyped(
...
+            num_params += len(rest_val)
+            mem.SetLocalName(lval, value.List(rest_val))
         else:
-            if num_args > num_params:
+            if num_args > num_params + (1 if block_param else 0):
```
andychu, line 358: Likewise can we remove `if block_param` by handling it first?

I might have missed something

andychu (commented):
Thanks for looking into this !  This looks very promising

I need to clone the branch and look at the before/after a little, but here are a couple comments

andychu:
OK I played with it a bit more ... I do see the problem, and I see the logic behind your fix - thank you

I wonder if there are any other problems here, let me see

andychu:
Also I'm thinking about our docs:

https://www.oilshell.org/release/0.21.0/doc/proc-func.html

>  (The block param is passed as a typed, positional arg, although this detail usually doesn't matter.)

Clearly I did not think this through!

I think the design is a little weird

andychu:
I think this came about in unifying

```
cd /tmp { echo hi }
```

```
cd /tmp (myblock)
```

But yeah now I wonder if it should be


```
cd /tmp ( ; ; myblock)
```

for consistency.  I don't think this occurs much in real code.  It's only for testing and metaprogramming and so forth.

There are also builtins like `fopen` and `shvar` ... hm


-----


Another option is to change the signature ???  We could try to make it so there are 3 sections, not 4

```
proc p (w1, w2; t1, t2; named1, named2) {
}
```

Hm

glyh:
It does still make sense and code like this work:
```
# NOTE: Usage is like below
# withMutex lock1 lock2 lock3 {
#   echo hello
# }
proc withMutex(;...mutexesWithBlk) {
  if (len(mutexesWithBlk) == 1) {
    eval (mutexesWithBlk[0])
  } else {
    var mutexesRest = mutexesWithBlk[1:]
    var mutexCurrent = mutexesWithBlk[0]
    {
      if flock $fd {
        withMutex (...mutexesRest) 
      } else {
        echo "Failed to grab the lock `$mutexCurrent`"
        exit 1
      }
    } {fd}<$mutexCurrent
  }
}

```

glyh:
I don't really understand the `cd /tmp ( ; ; myblock)` syntax. I thought `;` only occurs in definition sites.

andychu:
So `;` does have to be used  at call sites in one case now:

```
call f(...args ; ...kwargs)
```

which is like Python's

```
f(*args, **kwargs)
```

We only have a single operator `...`, so we have to use `;` to disambiguate.

----

I think the argument for `cd (; ; myblock)` is to make a **fully general wrapper**

### How do you wrap any proc?

Suppose you have:

```
proc p(w1, w2; t1, t2; n1, n2; block) {
}
```

How do you wrap it, like Python's `f(*args, **kwargs)`

I think you need

```
proc wrapper(...words; ...typed; ...named; block) {
  p @words (...typed ; ...named ; block)
}
```

This is more **consistent** than:


```
proc wrapper(...words; ...typed; ...named; block) {
  p @words (...typed ++ [ block ]; ...named)
}
```

@bar-g ran into this bug, and may have opinions

He wanted `p @ALL_ARGS`, but I think it's not desirable to unify them


I think `p @words (...typed ; ...named ; block)`  makes sense

---

I think it's nice because the signature matches the call site

Thoughts?

andychu:
I am not sure of the exact fallout of changing it, but I would like a PR that does `cd /tmp ( ; ; block)` to compare the two

Thanks for doing this -- it's very helpful in clarifying our design

glyh:
I think it makes sense to have different treatment for different cases

```
# The following are equivalent
p (...a)
p (...a;;)
# The following are equivalent
p(...a; ...b; c)
p(...(a ++ [c]); ...b)
```

andychu:
I am going to write something about this.  I think we want to make **call** site and **def** site consistent

and we have **two** options in that case

```
proc p (words; typed; named; block) { ... }

cd ( ; ; block)
```

OR

```

proc p (words; typed; named) {  ...  }  # No Block section !!

cd (block)  # FIRST typed arg
```

Either of these consistent.  But neither the OLD behavior or this NEW behavior is consistent.  We still have a problem, even with the PR

---

I believe  that an argument for the first style is `...` should work APART from blocks

```
cd (...typed) {
  echo hi
}

cd (...typed ; ; ^(echo hi) )
```

should work consistently


It shouldn't be

```
cd (... [ typed ++ [^(echo hi)] ])
```

in other words

I think that works, but it's kinda crazy !!

andychu:
Yeah I think that is the core argument

```
cd (... [ typed ++ [^(echo hi)] ])
```

is way too complex.  We are appending one element to a list, and then splatting it

---


```
cd (...typed ; ; ^(echo hi) )
```

is more understandable

bar-g:
Sorry, different from said, I will only be able to look at this next week.

andychu:
Hm I played around with the other option -- `p @words (...typed; ...named; block)` and it appears to work fine

https://github.com/oilshell/oil/commit/610e5e62407eb603ee9ab2146f57a315a75a8e5a

We still have `eval (b)` rather than `eval (; ;b)` though, because there's a difference between a `value.Command` that's positional (first section), and a block argument (fourth section)

So right now I think this works -- the main reason it would change is because of one of our 16-17 use cases from the blog/Zulip.  We are trying to encode many idioms in this reflective interface

andychu:
The alternate solution seems to have resolved this issue -- thank you for digging deep into the code!

Issue #861, opened 2020-11-13
context manager style for cleanup / trap handlers


andychu:
https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/Seeking.20Feedback.20on.20OSH.20and.20Oil/near/216440385

That is I kind of like my failure / cleanup in scope.  In Python

```
with mycleanup():
  main()
```

rather than

```
register_signal_handler()
main()
```

andychu:
#850 is related

I think Danilo Spinella also mentioned this

andychu:
- Hm test frameworks have this issue too: sharness, bats, etc.  #929

PR #1899, opened 2024-04-05
Implement semaphores in stdlib/synch, backed by named pipes

glyh:
This PR implements several synchronization primitives for ysh: Semaphores, Mutexes, RWLocks and CondtionalVariables. For now only tests for semaphores are written. 

### Some possible improvement
- [x] Design a better interface. I think this would be done when meta objects landed. 
- [ ] <del> Find a way to hide redundant job control informations. </del> 
https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Job.20control.20stderr
- [ ] No integration with exeception control fllow.
- [ ] <del> No observability is implemented. </del> Impossible without shared memory.

### Other information
- Not very useful. No shared-memory parallelization is possible in shell as jobs are implemented with IPC.  We still have go-channel style IPC though.
- Some syntax sugar that provides guard-like behavior should be preferred.

### Use case
- For emulating `make -j`, semaphores can be used to synchronize between jobs.

### Status
Ready to merge.


glyh:
We can build some other synchronization primitives atop.

glyh:
Also it would be nice to supress the job control information when it's not done interactively or directly by user.

glyh:
I'm still now sure about using funcs all the way, as my code is spammed with `call` all around now.

glyh:
Looks like there's some cryptic bugs occuring on my end, hold on let me create a demo in spec.

glyh:
The failing test is here, I can't find a way to make it work correctly. Basically I create a semaphore, upped it 4 times and down it 4 times. But the job seems to be hanging.

glyh:
http://travis-ci.oilshell.org/github-jobs/6614/cpp-spec.wwz/_tmp/spec/ysh-py/ysh-stdlib-synch.test.html#L31

This specific test is failing on my machine but passing on CI, I'm more than confused now.

glyh:
Yeah I can confirm this is race condition. If I modified the implementation to

image

With that test I got

glyh:
Tried to fix with https://stackoverflow.com/questions/8297415/in-bash-how-to-find-the-lowest-numbered-unused-file-descriptor but ysh doesn't support this syntax.

glyh:
Same bug reproduced with bash
```bash
#!/usr/bin/env bash

sema=`mktemp -u`
mkfifo $sema

semaDown() {
  head -c 1 < $sema > /dev/null
  echo down
}

semaUp() {
  {
    echo -n -e '1' > $sema
    echo up
  } &
}

{
  semaUp
  semaUp
  semaUp
  semaUp
} &

semaDown
semaDown
semaDown
semaDown

echo 1
```

This is probably a hard-no to use pipes for semaphore.

glyh:
Strangely this code actually works
https://unix.stackexchange.com/questions/618042/fifo-based-semaphore-explanation

glyh:
Okay hopefully this is fixed now.

glyh:
For now this is not very useful, however, because no variables are shared between jobs in ysh. We need to add some features to share variables across process.

glyh:
Blocked by https://github.com/oilshell/oil/issues/1903

spec/ysh-stdlib-synch.test.sh
```diff
@@ -0,0 +1,56 @@
...
+
+#### semaphore
+source --builtin synch.ysh
+
+var s = semaNew(1)
+{ 
+  sleep 0.5
+  call semaDown(s)
+  echo 1
+} &
```
andychu, line 13: Style note: `fork { myblock }` is preferred in YSH  (we could omit the `&` operator)


we also have `forkwait { myblock }`

stdlib/synch.ysh
```diff
@@ -0,0 +1,179 @@
...
+
+############################
+### FIFO File Desriptors ###
+############################
+
+func fifoFdNew() {
+  # WARN: this section should be critical but for now it's not
+  # A solution may be retry on fail.
+  #====================
+  var fifo = $(mktemp -u)
```
andychu, line 13: Functions will likely become **pure**, disallowing `$()`

https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Why.20are.20functions.20pure.2C.20with.20IO.20type.3F

We may want to use a proc instead


```
var fifo
fifo-fd-new (&fifo)
```

stdlib/synch.ysh
```diff
@@ -0,0 +1,179 @@
...
+func semaDown(sema) {
+  read <$sema
+}
+
+func semaUp(sema; delta = 1) {
+  {
+    for _ in (0 .. delta) {
+      echo >$sema
+    }
+  } &
```
andychu, line 46: can use `fork { }` instead

stdlib/synch.ysh
```diff
@@ -0,0 +1,179 @@
...
+  exec {fd}>-
+  exec {fd}<-
+  rm $fifoFile
+}
+
+#################
+### Semaphore ###
+#################
+
+func semaNew(value) {
```
andychu, line 31: Likewise, these should be procs, with kebab-case e.g.

https://www.oilshell.org/release/0.21.0/doc/style-guide.html

```
sema-new
sema-up
sema-down
```

andychu (commented):
This is great, I really like the experimentation

And thanks for finding the json error

I am not sure we need all of these primitives -- I think we should implement **apps** like

- xargs -p
- make -j

and then do the **minimal** necessary to implement them

I think some of the primitives are a bit too low-level, for YSH

But we want more than shell, for sure

---

I think for now we should rename this to `stdlib/draft-synch.ysh`, to note that it's a draft

It's great to show people what you can do

Later, if we use it as part of xargs or make, then we can upgrade it to `stdlib/synch.ysh` -- and also add documentation

Thanks!

glyh:
Great, I'll fix them later. For now I'm still struggling to make mkfifo actually spit out the data I fed them.

andychu:
Yes another issue is that `sema-destroy` may not be guaranteed to be called -- because YSH has "exceptions"

We have a lot of block-like constructs, so we may need another

- #861 

is kind of related

stdlib/synch.ysh
glyh: How do you return a non int value in a proc then? For now these have to be function as some functions returns non-int. And since purity is transitive the called function is also forced to be function.

glyh:
CI is failing as I rename functions to kebab case but didn't change them to proc. I guess I also can't change it to proc as some functions returns dict.

glyh:
@andychu I think this is ready for a merge.

andychu:
- OK as mentioned, one issue is that sema-destroy won't be called in case of an error ... we might need some scoping.  But we can handle it later
- Please rename to `stdlib/draft-synch.ysh` or `stdlib/draft-semaphore.ysh` for now
  - since we don't have documentation, and there are probably some other TODOs


We still have to test it against the make/xargs use cases, etc.  I may be able to help with that -- I like to try out the APIs

spec/ysh-stdlib-synch.test.sh
```diff
@@ -0,0 +1,56 @@
...
+  sema-down (s)
+  echo 1
+}
+fork { 
+  sleep 1
+  sema-down (s)
+  echo 2
+}
+fork { 
+  sleep 1.5
```
andychu, line 20: We should reduce these sleep values when we add more test cases

Because I don't want to make the suite slow

spec/ysh-stdlib-synch.test.sh
```diff
@@ -0,0 +1,56 @@
+# spec/ysh-stdlib
+
+## our_shell: ysh
+
+#### semaphore
```
andychu, line 5: this should be more descriptive

spec/ysh-stdlib-synch.test.sh
```diff
@@ -0,0 +1,56 @@
...
+sema-destroy (s)
+## STDOUT:
+1
+4
+2
+5
+3
+## END
+
+#### semaphore init and multiple down
```
andychu, line 39: some comments would help here

stdlib/draft-synch.ysh
```diff
@@ -0,0 +1,51 @@
...
+
+proc fifo-fd-destroy(; fd) {
+  var fifoFile = $(readlink /proc/$$/fd/$fd)
+  exec {fd}>&-
+  exec {fd}<&-
+  rm $fifoFile
+}
+
+#################
+### Semaphore ###
```
andychu, line 28: I prefer to use

```
#
# Semaphore
# 
```

which is easier to maintain

andychu (commented):
OK I made some minor comments, but let's merge for the sake of progress

Thanks for this nice experiment

Issue #1872, opened 2024-03-20
Proposal: implement a macro system for YSH


glyh:
Macros are powerful tools for writing concise yet powerful codes. For now YSH doesn't seem to have one, however.

### Pros
- Having a macro system elminates the need of some other language constructs. For example, I would like ysh to have some constructs that mimics makefile, but with macros I can rock my own `makefile`
- Easy implementation of noncrucial language extension, e.g. [if-not](https://clojuredocs.org/clojure.core/if-not), [destructuring](https://clojure.org/guides/destructuring), [data transformation](https://github.com/redplanetlabs/specter). I would like to highlight having something like specter is extremely useful in shell where transforming data is a normal task. 
- Slime down the language itself so many constructs doesn't need to be hard coded to the lanugage spec itself. Furthermore, this will allow users to decide whether to use some constructs or not, which may be flexible.

### Cons
- Maintainence overhead
- Abusing macros yield cryptic codes. But that's rarely the case in real life
- Debugging with macros is a problem. We may need to take a look at [racket's approach](https://docs.racket-lang.org/macro-debugger/index.html) to see if there's any space for improvement.

### Implementatoin Issues
If I remember correctly, ysh parses the code staically, so there might be some modifications on the parser's side to delegate AST construction to the macro system.  I personally prefer to have macro epansion as a single layer before execution, i.e. our macro system is more like racket's than common lisp's(main difference being whether you can expand a macro in runtime or not) 

### Inspirations
- [Elixir's metaprogramming system](https://elixirschool.com/en/lessons/advanced/metaprogramming) is a good demonstration of how macros should look like in a dynamic-typed ruby-like languge.
- [TCL](https://www.tcl.tk/about/language.html) seems to build its entire language on string replace, and thus it's a kind of macro system. It may play especially well with shells as string is the API for unix in many scenarios.

### Further information

I would like to help implementing this if the maintainer is interested. For now I think we need to collect more opinons from developers and end-users.

glyh:
I would like to help implement this feature if possible.

glyh:
I personally think elxir's macro style would be a good fit for ruby/python like languages:

https://elixirschool.com/en/lessons/advanced/metaprogramming

glyh:
Reopened as only macros can solve scoping issues. https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Semaphores.3F/near/431423521

PR #1912, opened 2024-04-06
Implement stdlib/termcodes

glyh:
This is a draft PR. 

### Issues
- There's no way to actually test this because termcode commands has effect not on stdout/stderr but on other environments. 
- There's some biolerplates in the code. Begs macro system.
 #1872
 
### TODOs
- [ ] Auto adapting colors with `$COLORTERM`
- [ ] Maybe delegate jobs to `tput` from ncurses if it's present. Since it's not really part of coreutils so this better stay optional. Not sure yet. 
- [ ] Nesting styles doesn't need to be fixed. 

### Resources 
- https://en.wikipedia.org/wiki/ANSI_escape_code
- https://invisible-island.net/xterm/ctlseqs/ctlseqs.html

### Preview

```
source ./draft-termcodes/xterm.ysh 
echo $["I'm doubly underscorred" => withStyle(STYLE_DOUBLY_UNDERLINED)]
echo $["I'm clickable!" => withLink("https://www.google.com") => withFg(sgr24(0xff00ff))]
echo $["I'm crossed out" => withStyle(STYLE_CROSSED_OUT)]
```

image

### DIFF
--- a/stdlib/draft-termcodes/ansi.ysh
+++ b/stdlib/draft-termcodes/ansi.ysh
@@ -64,13 +64,13 @@ func withFg(text, color) {
     } else {
       setvar sgr4_id += 90 - 8
     }
-    return ("$TERM_ESC[${sgr4_id}m${text}$TERM_ESC[m")
+    return ("${TERM_CSI}${sgr4_id}m${text}${TERM_CSI}m")
   }
 }
 
 func withBg(text, color) {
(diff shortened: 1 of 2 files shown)


glyh:
Closing, nested style is not really a thing in xterm. It's better to implement this as 3rd party libraries as manual control for each differently styled segment is required.

PR #1915, opened 2024-04-06
Implement stdlib/random

glyh:
This is a draft PR implementing some common random functions for std. 

Probably need something like bigint before implementing too much interfaces for random numbers. For now I think we're good with this one function.


glyh:
Wrong commits pushed, close and reopen this.

PR #1917, opened 2024-04-06
Stdlib random

glyh:
Continuation of #1915

This is a draft PR implementing some common random functions for std.

Probably need something like bigint before implementing too much interfaces for random numbers. For now I think we're good with this one function.


glyh:
Discussions are welcomed.

PR #1922, opened 2024-04-08
Fix nix-shell for `test/spec-py.sh ysh-all`

glyh:
NOTE: `bin/ysh` does work for me, but `test/spec-py.sh` is still broken
@andychu Help wanted. 

Related https://github.com/oilshell/oil/issues/1918

https://oilshell.zulipchat.com/#narrow/stream/417617-help-wanted/topic/Fix.20nix-shell

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -53,8 +53,17 @@ let
   #   Clang for coverage too
 
   # nixpkgs: busybox linux only; no smoosh
+
+  # busybox links xargs, we need that to run tests, though
+  busybox_no_symlinks = busybox.override {
+     extraConfig = ''
+       CONFIG_INSTALL_APPLET_DONT y
+       CONFIG_INSTALL_APPLET_SYMLINKS n
+     '';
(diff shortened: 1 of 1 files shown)


andychu:
I am not sure if anyone else is using the Nix build -- I don't use it

@Melkor333  ??

If this works for you, I guess we merge it?   Right now it is "dev-contributed", although if we can make `build/{dev,nix}-shell.sh correspond more, that would be nice

with "wedges" I think that's possible -- it just puts re2c in your $PATH, etc.

glyh:
It doesn't work, see zulip chat.

glyh:
Is there a recommended workflow if python2 is not working on my machine? It begs PYTHONHOME but I found no document on how to set this. The latest doc was like 10 years ago.

glyh:
It doesn't work on Arch. And I do believe it make sense to use a reproducible build environment for this project as it pulls too many dependencies.

PR #1928, opened 2024-04-08
direnv support

glyh:
### Status
Ready for merge.

### Description
Direnv is a tool for user to separate their shell config into different folder specific profile. When `cd`ing into the `oil` folder, if direnv is allowed to run the script, it will automatically run the `build/dev-shell.sh` script. and dump environment variables into current shell.

### Usage
- [Install direnv and its shell hook](https://direnv.net/#getting-started)
- cd into git repo root
- Run `direnv allow`
- Now whenever you cd into oil repo or one of its subdirs, direnv will load environment variables directly for you.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+. build/dev-shell.sh 
--- a/.envrc
+++ b/.envrc
@@ -1,3 +1,12 @@
 #!/usr/bin/env bash
 
+#### Description
+# Direnv is a tool for user to separate their shell config into different folder specific profile. When `cd`ing into the `oil` folder, if direnv is allowed to run the script, it will automatically run the `build/dev-shell.sh` script. and dump environment variables into current shell.
+
+#### Usage
+# - Install direnv and its shell hook on https://direnv.net/#getting-started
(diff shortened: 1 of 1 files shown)


.envrc
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
```
andychu, line 2: Please add some comments on what this is, what the purpose is (run bin/ysh etc. without changing $PATH)

how to use it, how to install, etc.


I've actually never used direnv before :)

andychu (commented):
OK thanks, I think this is a good idea

Just needs some comments

glyh:
This is somewhat broken, I need to fix it.

.envrc
glyh: Okay this is done

andychu: OK great, can you add the comments to the top of the script?  So people know how to use it


Right now it looks like it's only in the PR, and will get lost

glyh: done

andychu:
Great, thank you!  I will "advertise" this on Zulip

PR #1939, opened 2024-04-10
variables.md: setglobal to create global

bar-g:


andychu:
Great thanks!

I think this was probably left over from an earlier iteration of Oil / YSH

andychu:
Actually yeah I think back then I wanted a 1:1 correspondence between shell and "Oil", but I quickly realized that this made everything way too coplex

"Oil" was being compromised by POSIX shell / bash

PR #1945, opened 2024-04-23
Pretty Printing for Oils values

justinpombrio:
Ready for review! Code has been formatted by yapf. There are still C++ errors. (What's a good way to view them? I get lost in the Travis links.)


data_lang/pretty.py
```diff
@@ -1,36 +1,145 @@
...
+                            node,
+                            frag.indent,
+                            frag.flat
+                        ))
+
+def _ValueToDoc(val):
+    # type: (value_t) -> doc_t
+    """Convert an Oils value into a `doc`, which can then be pretty printed."""
+
+    with tagswitch(val) as case:
```
andychu, line 124: I think we can fix the C++ compile errors using this pattern, basically add `UP_val = val` above this line, and then the downcasts are `val = cast(value.Bool, UP_val)` instead


https://www.oilshell.org/release/0.21.0/doc/oils-repo/mycpp/README.html#val-up_val-val-downcasting-pattern

andychu (commented):
I took a quick skim, looking very good!

I noticed the C++ failures here

http://travis-ci.oilshell.org/github-jobs/6780/

Generally it is OK to do Python first and worry about C++ later ... I view them as separate steps.  Sometimes it "just works"

Here I think there is a simple fix that will clear up 90% of the errors

It's not obvious from the compiler output, but I documented it in mycpp/README

data_lang/pretty.py
justinpombrio: Oh, it's just about variable reuse! The all caps UP_ made me think there was something a lot more magic going on, like the compiler treating casts of things containing "UP_" differently.

I'll just give the casted values more specific names, as is appropriate since they have different types and hold different data.

andychu: Actually mycpp DOES have a rule for `UP_` now :-P

```
$ grep 'UP_' mycpp/*.py
mycpp/cppgen_pass.py:            #    src = cast(source__SourcedFile, UP_src)
mycpp/cppgen_pass.py:            # -> source__SourcedFile* src = static_cast<source__SourcedFile>(UP_src)
mycpp/cppgen_pass.py:                    if to_cast.name.startswith('UP_'):
```

This is not really ideal -- there are threads on Zulip about "improving the cast pattern"

I'm open to work on that (remember I want to rewrite mycpp completely) -- but I guess SEPARATELY from this PR, since a lot of code would change

I think for this PR it's better to keep consistency with the `UP_val` style

It's possible some other schemes might work, I haven't thought about it

There is an assymmetry between vars "going in" to the switch and vars "coming out", and I think we mostly solved that, and it's documented

andychu:
Hm I guess there are some more C++ errors, but I would just ignore them until the code behaves how we want.  

sometimes it "just works", but not today :-/  I want people to be able to contribute as if it's "just python", and then C++ is a separate thing

justinpombrio:
J8 printing is still present in this PR, but is there a reason to have it in addition to pretty printing? If the aim is that J8 printing is for the subset of data values that can be re-parsed after being printed, ought some of its settings change now?

data_lang/pretty.py
```diff
@@ -1,40 +1,546 @@
...
+    def _BashAssoc(self, vassoc):
+        # type: (value.BashAssoc) -> MeasuredDoc
+        if len(vassoc.d) == 0:
+            return _Text("(BashAssoc)")
+        mdocs = []
+        for k2, v2 in iteritems(vassoc.d):
+            mdocs.append(
+                _Concat([
+                    _Text("["),
+                    self._StringLiteral(k2),
```
justinpombrio, line 461: Note that I'm printing this like a plain string. That effects both the style (not colored the same as a Ysh Dictionary key) and the quotes (quotes are always present, unlike Ysh Dictionary keys which omit quotes if they're identifier-like). It could instead be printed as a _DictKey, which would change both of those properties.

data_lang/pretty.py
```diff
@@ -1,40 +1,546 @@
...
+from mycpp import mops
+from mycpp.mylib import log, tagswitch, BufWriter, iteritems
 
 _ = log
 
+KEY_STYLE = ansi.GREEN
+NUMBER_STYLE = ansi.YELLOW
+NULL_STYLE = ansi.BOLD + ansi.RED
+BOOL_STYLE = ansi.BOLD + ansi.BLUE
+CYCLE_STYLE = ansi.BOLD + ansi.YELLOW
```
justinpombrio, line 118: These styles (and the choice of which things are styled) can of course be whatever you like. I picked these out of a hat. Not sure if there's anything more principled to do. Could copy nodejs styling, I suppose.

data_lang/pretty.py
```diff
@@ -1,40 +1,546 @@
...
+                return self._BashArray(varray)
+
+            elif case(value_e.BashAssoc):
+                vassoc = cast(value.BashAssoc, val)
+                return self._BashAssoc(vassoc)
+
+            else:
+                ysh_type = value_str(val.tag(), dot=False)
+                id_str = ValueIdString(val)
+                return _Text("<" + ysh_type + id_str + ">")
```
justinpombrio, line 543: Are there any other values cases we should handle? E.g. I added a case for `range`.

andychu:
This looks excellent, thank you! 

I read through the  comments, tests, and code -- looks very nice and clean

---

I only have very superficial comments

- string type should have a style
- types should have a style, both in (BashAssoc ...) and <Eggex 0x123>

https://www.oilshell.org/release/0.21.0/doc/pretty-printing.html#screenshots

I realized that the type is redundant in those cases, because it looks like

```
(BashAssoc)   (BashAssoc ["a"]=v)
```

But let's think about this LATER -- I think we will "harmonize" it with the ASDL type names and so forth

---

Oh actually I just realized that BashAssoc and BashArray should use fastfunc.ShellEncodeString() instead of J8EncodeString(), because otherwise we're mixing syntaxes.    JSON double quoted strings are not shell strings.

---

I'm a little worried about the number of GC objects, but

- we are tuning the allocators a lot
- It will be fun to test performance on the ASDL pretty printing, which generates 109 MB of text in a realistic case.
  - there is a benchmark in `data_lang/pretty-benchmark.sh` and I think it's algorithmically slow (quadratic or worse)
- so maybe this algorithm will beat that naive one, AND look better and be more correct

But yeah performance is a separate thing from this PR, and is not an issue for `=`

---

Another tangent:

- I think the `pretty.asdl` lines could be indented, like `frontend/syntax.asdl` 
- But it's not worth it to write an ASDL pretty printer since we have <1000 lines of code in it
- I have been thinking of that "polyglot" formatter that only understands `{} ()` and `#` comments, which ASDL happens to respect
  - basically it makes some kind of "coarse tree", and indents only??

This relates to my `doctools/micro-syntax` experiment ... not important right now, but I like to identify these little problems / use cases for text processing :)

andychu:
And BTW it looks like ASDL pretty printing will plug very cleanly into this (as we expected) ...

The main difference I think is that it's generated code that will be constructing the `doc_t`.  We need some metaprogramming to turn the statically typed structures into homogeneous ones, but this is an orthogonal issue



I kinda want to do it, to test my knowledge of the algorithm/usage

Though I am backed up right now with benchmarks, blog posts, and releases!

andychu:
Here is a test case 

```
osh-0.21.0$ declare -A assoc=(['k']=$'foo \x01\u03bc')
osh-0.21.0$ = assoc

(BashAssoc)   (BashAssoc ["k"]="foo \u0001μ")  # should use the original syntax

```


-----

Shell double quotes different than JSON double quotes:

```
osh-0.21.0$ declare a=("\u03bc")
osh-0.21.0$ = a
(BashArray)   (BashArray "\\u03bc")  # not respected
```

data_lang/pretty.py
andychu: We might want to, but we can add them later

andychu:
> J8 printing is still present in this PR, but is there a reason to have it in addition to pretty printing? If the aim is that J8 printing is for the subset of data values that can be re-parsed after being printed, ought some of its settings change now?

Oh do you mean that J8 now has extra configurability that we're not using?  because pretty printing does it

we no longer need the option for cycles in J8, etc.

That's true, but that can be done separately ...  

---

I am also thinking of the converse `json write (val, pretty=true)` which would do wrapping, which could be useful for users

But let's also leave that for later

justinpombrio:
> string type should have a style; types should have a style, both in (BashAssoc ...) and <Eggex 0x123>

I've revamped the styles. It's now:

- Copying nodejs style: strings are green, numbers are yellow, dict keys are uncolored.
- Null is red, cycles are magenta, and types (BashArray, BashAssoc, <Eggex 0x123>) are cyan.
- Additionally, keywords (true, false, null, ... in cycles) are bold.

> I realized that the type is redundant in those cases

I think it would be very weird to omit the "(Type)   " prefix, but only sometimes. If you'd like to not print the "(Type)   " prefix, that's easy and we should just make that change now.

> Oh actually I just realized that BashAssoc and BashArray should use fastfunc.ShellEncodeString()

Done. Though I'm having trouble testing it. Here's my test case (from you):
```
#### Bash Assoc: string formatting
declare -A assoc=(['k']=$'foo \x01\u03bc')
= assoc
## stdout: (BashAssoc)   (BashAssoc ['k']=$'foo \u0001μ')
```

Is that what's expected? That's with passing `0` as the "ysh fallback" arg to ShellEncodeString. Passing 1 instead gives:
```
(BashAssoc ['k']=b'foo \u{1}μ')
```

> I'm a little worried about the number of GC objects

Yeah, it's definitely a lot of small objects. It is guaranteed linear time, at least. (I'm interpreting what you said as there being another algorithm that's quadratic? If this one is quadratic there's a bug.) If speed is a problem, there's a few options:
- smart constructors
- arena allocation, if that's something we can do
- if really needed, there are more complex options

> I think the pretty.asdl lines could be indented, like frontend/syntax.asdl

Oh that was so bothering me. It was only de-dented to copy the j8 one. Indented now.

justinpombrio:
So I'm still not seeing how to tell why CI failed. Github says "Some checks were not successful". I click on "details" next to a step with an "X" on it. It doesn't matter which one, because they all show a console with similar info. On the left there's a white navigation bar, with the same list of jobs. On the console, it says "View CI results here", which brings me to a Travis page with the same list of jobs. Clicking on one brings me to a list of "tasks". Some of the tasks say "status: -1". If I click on those, there are some C++ errors, printed without syntax highlighting which makes them kind of hard to read. Is that where I should be looking?

andychu:
Ah OK I usually start here

then I click "interactive" which is the first failure

then I click on the first task that has status: 1 in red

http://travis-ci.oilshell.org/github-jobs/6874/interactive.wwz/_tmp/soil/logs/process-table.txt

first error I see is

```
_gen/bin/oils_for_unix.mycpp.cc:28775:29: error: expected unqualified-id before 'typename'
   pretty_asdl::MeasuredDoc* typename = nullptr;
                             ^~~~~~~~
_gen/bin/oils_for_unix.mycpp.cc:28778:29: error: expected nested-name-specifier before ')' token
   StackRoot _root1(&typename);
                             ^
```

I believe this must be because `typename` is a keyword in C++, so we can't use it as a Python var name!

andychu:
> > I realized that the type is redundant in those cases
> 
> I think it would be very weird to omit the "(Type) " prefix, but only sometimes. If you'd like to not print the "(Type) " prefix, that's easy and we should just make that change now.

Hm well I want  to maintain a separation between pretty printing `= x ` and `json write (x)` (and `json8 write (x)`) in users minds

So the Dict/List output not to be parse-able

There are other ways to accomplish that, but I think we can consider it separately.   I will start a thread about it on Zulip


(also removing the type right now, will break all those unrelated tests)




> > Oh actually I just realized that BashAssoc and BashArray should use fastfunc.ShellEncodeString()
> 
> Done. Though I'm having trouble testing it. Here's my test case (from you):
> 
> ```
> #### Bash Assoc: string formatting
> declare -A assoc=(['k']=$'foo \x01\u03bc')
> = assoc
> ## stdout: (BashAssoc)   (BashAssoc ['k']=$'foo \u0001μ')

OK that's good!  It's an equivalent spelling.  That is expected


> 
> Yeah, it's definitely a lot of small objects. It is guaranteed linear time, at least. (I'm interpreting what you said as there being another algorithm that's quadratic? If this one is quadratic there's a bug.) If speed is a problem, there's a few options:
> 
> * smart constructors
> * arena allocation, if that's something we can do
> * if really needed, there are more complex options

Yeah we have an ASDL algorithm that is naive/slow.  I am pretty sure this one will beat it -- it will be fun to test

andychu:
Merged, thank you!   This seemed to come together very smoothly

I'll start some Zulip threads about

- the type name issue, the exact format (brainstorming)
- ASDL printing --- I thought about this last night, and I think it actually doesn't involve code gen.  Because should probably just use code gen to flatten heterogeneous typed -> homogeneous, and then pretty printing can be another step on top of the homogeneous tree
  - it's already 2 step, I think it might 3 step

PR #1948, opened 2024-04-28
always clean up child process on exit

### LINKED ISSUE #1003: limit size of process and jobs dictionary

melvinw:
I only half-way fixed #1003 in 12697216532eea6b04c98a5f59015372b8efd27a. We should remove all processes from the child process map on exit, not just the ones that have been backgrounded and assigned job IDs.

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -1154,7 +1154,8 @@ class Process(Job):
                     print_stderr('[%d] Done PID %d' % (self.job_id, self.pid))
 
                 self.job_list.RemoveJob(self.job_id)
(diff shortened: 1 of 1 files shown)


andychu:
Awesome, thanks for finding this !!!

It's always these subtle things that cause the hardest to find bugs

We should have some test like #1949 , but this is visible in the benchmarks

PR #1951, opened 2024-04-29
xxx clear roots after forking

melvinw:
this might be a bad idea? let's find out...


andychu:
This is interesting, and I hadn't thought of it as an optimization

I want to get our measurement situation under control a bit and then we can see if it's a bottleneck

---

I think there could be some unsafety introduced when throwing exceptions though, it's possible we may end up lower on the call stack and use an object in an exception handler.

I am not sure if that happens now, but it also seems hard to guarantee it wono't ever happen

PR #1952, opened 2024-04-30
Add failing test case for correctness errors in osh/string_ops utf8 decoders

PossiblyAShrub:
I had this in my work tree before starting a fix for the J8 UTF-8 bug. It can be merged on it's own if you want.


andychu:
Thanks!

PR #1953, opened 2024-05-01
performance experiments

melvinw:


andychu:
Thanks, this makes sense and definitely reduces allocations

https://oilshell.zulipchat.com/#narrow/stream/384167-performance/topic/New.20bin-true.20workload.20exposes.20some.20problems

Though it doesn't seem to move that /bin/true benchmark at all, which I'm a little surprised by

I think the ability to control sizes makes sense, though I wonder if we can do it for Dict and List uniformely, like:

```
d = NewDict()
mylib.ReserveDict(d, 16)

L = []
mylib.ReserveList(L, 32)
```

Lists are VERY common and this might give us more control

Lists are much more common than dicts, probably by a factor of 100x, due to word eval and ASDL syntax tree and all that

...

Maybe we can keep this in our back pocket?

---

The other idea I have is to add a 3rd pool.   That is easy to try

melvinw:
going to break this up into separate PRs

PR #1954, opened 2024-05-01
Add compgen -k and -e spec tests

KingMob:
## About 
As discussed on [Zulip](https://oilshell.zulipchat.com/#narrow/stream/146045-shell-autocompletion/topic/Bash.20completion.20compatibility.20status.3F), this adds spec tests for `compgen -k` and `compgen -e`. bash passes all tests, and osh does not.

## Notes
In particular, `compgen -e` might be easy to implement, since it's very similar to `compgen -v`, which is already done. The difference is `-e` completes exported vars, and `-v` completes _all_ vars.

`compgen -k` returns/completes shell reserved keywords.

### DIFF
--- a/spec/builtin-completion.test.sh
+++ b/spec/builtin-completion.test.sh
@@ -1,5 +1,5 @@
 
-## oils_failures_allowed: 3
+## oils_failures_allowed: 9
 ## compare_shells: bash
 
 #### complete with no args and complete -p both print completion spec
@@ -162,6 +162,38 @@ PATH
 PWD
 ## END
 
+#### compgen -e with global/local exported vars
+export v1_global=0
+f() {
+  local v2_local=0
+  export v2_local
+  compgen -e v
+}
+f
+## STDOUT:
+v1_global
+v2_local
+## END
+
+#### compgen -e on known, but unexported, var
+unexported=0
+compgen -e unexported
+## status: 1
+## stdout-json: ""
+
+#### compgen -e on unknown var
+compgen -e __nonexistent__
+## status: 1
+## stdout-json: ""
+
+#### compgen -e P
+cd > /dev/null  # for some reason in bash, this makes PIPESTATUS appear!
+compgen -e P | grep -E '^PATH|PWD' | sort
+## STDOUT:
+PATH
+PWD
+## END
+
 #### compgen with actions: function / variable / file 
 mkdir -p $TMP/compgen2
 touch $TMP/compgen2/{PA,Q}_FILE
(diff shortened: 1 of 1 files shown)


spec/builtin-completion.test.sh
```diff
@@ -162,6 +162,37 @@ PATH
 PWD
 ## END
 
+#### compgen -e with global/local exported vars
+export v1_global=0
+f() {
+  export local v2_local=0
```
andychu, line 168: nitpick : should be

```
local v2_local=0
export local
```

export local actually exports a variable named "local"!

andychu (commented):
Thank you, this is great!  

At the top of the file, oils_failures_allowed should be adjusted

It's not trivial to find but I go to

and then

http://travis-ci.oilshell.org/github-jobs/6838/

and you can see the red row here:

http://travis-ci.oilshell.org/github-jobs/6838/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

So basically we allow the failures to ensure there are no regressions once fixed

spec/builtin-completion.test.sh
KingMob: Sorry about that. I'm not much of a shell scripter. 

I'm more the sort who liked your blog posts, thought I'd give osh a try, and have just been trying to get this one thing working so I can test it as a daily driver :P 

Wait, should that not be `export v2_local` instead of `export local`?

andychu: Oh yes, typo!   the var name comes after export

No problem, it is easy to make these mistakes, hence Oils :)

KingMob:
OK, I bumped up the number of failing tests for the completion spec. I was a little confused that one test I didn't touch, `compgen with action and suffix: helptopic` also started failing, but iiuc, it fails because I was testing with the minimal build, which lacks help text. The test is correct in Travis.

---

I tried a full build to see if it would make the help spec test succeed, and got this error when fetching deps:

```
root@3bddf505c6b0:/oil# build/deps.sh fetch
# ... lots of stuff

File '_build/deps-source/busybox/busybox-1.35.0.tar.bz2' already there; not retrieving.

tar (child): bzip2: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
```

This is in my Ubuntu 22.04.3 LTS Docker image. 

If I run `apt-get install bzip2`, then it will finish just fine after that. I see that bzip2 is in the wedge deps for Alpine and Fedora, but not Ubuntu/Debian. Maybe it should be? I can submit a PR for that if you like.

andychu:
Great thank you!

And yes good call about bzip2.  That does appear to be missing.   A PR is appreciated!

PR #1955, opened 2024-05-02
[data_lang/utf8] Create UTF8 decoder to round-trip surrogates. Fix J8 lone surrogate spec test

PossiblyAShrub:
I ended up writing a new UTF-8 decoder from scratch. It follows the Crockford style but 1) is stateless and 2) can round-trip surrogates. I've also written some fairly comprehensive tests for the decoder.

Now the `J8EncodeOne` function is able round-trip surrogates!

### DIFF
--- a/data_lang/j8.h
+++ b/data_lang/j8.h
@@ -2,9 +2,10 @@
 #define DATA_LANG_J8_H
 
 #include <stdio.h>   // sprintf
-#include <string.h>  // memcmp
+#include <string.h>  // memcmp, memcpy, strlen
 
 #include "data_lang/utf8_impls/bjoern_dfa.h"
+#include "data_lang/utf8.h"
 
 #define J8_OUT(ch) \
   **p_out = (ch);  \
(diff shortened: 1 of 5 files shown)


andychu (commented):
OK wow, looks great !!   I am not sure why I didn't think of doing it this way, which is clearer

I thought you would need the state param

I think the problem is that I thought there would be a buffer overflow.

And there was a comment in the decode() loop about a subtle invariant -- the buffer must be NUL terminated, because if it were ONLY length-prefixed, then you could have

    (start byte, cont, cont, GARBAGE)

But instead we know we have:

   (start byte, cont, cont, NUL)  # this will always be a UTF-8 decoding error

But yeah our usage Bjorn DFA already relies on that property, but I didn't know that when I started it.  I think I fixed an ASAN bug

The whole "p_in p_out" style without lengths to determine the end of buffer is a bit scary, but I think it is a logical way to do this, and we have very good tests

---

I will look at this a bit more tonight!  I want to test surrogate halves in J8 vs. JSON -- JSON is legacy with surrogate halves

And I want to poke at the error messages

It also reminds me I think we have some max code point bugs in source code, e.g. when you write `'\u{fffff}` in code, not data ...  i.e. going the opposite direction, when encoding rather than decoding

data_lang/utf8.h
```diff
@@ -0,0 +1,178 @@
+#ifndef DATA_LANG_UTF8_H
+#define DATA_LANG_UTF8_H
+
+#include <stdio.h>
+
+#include <stdint.h> // uint32_t
+#include <stddef.h> // size_t
+
+#ifdef __cplusplus
```
andychu, line 9: hm what breaks if we don't have this?  I don't recall using `__cplusplus` in our codebase

`j8.h` is shared between C++ and CPython C, and it seems to work

data_lang/utf8.h
```diff
@@ -0,0 +1,178 @@
...
+    result->error = UTF8_ERR_TRUNCATED_BYTES;
+    return;
+  }
+  result->bytes_read += 1;
+
+  // Continuation bytes follow the bit pattern 10xx_xxxx. We need to a)
+  // validate the pattern and b) remove the leading '10'.
+  if ((byte & 0xC0) == 0x80) {
+    result->codepoint <<= 6;
+    result->codepoint |= byte & 0x3F;
```
andychu, line 88: ( Too bad we can't use binary literals in C!  it's a lot easier to see that way )

data_lang/utf8.h
```diff
@@ -0,0 +1,178 @@
...
+
+    _cont(input, &result);
+    _cont(input, &result);
+    if (result.error) return result;
+
+    if (result.codepoint < 0x800) {
+      result.error = UTF8_ERR_OVERLONG;
+    }
+
+    if (result.codepoint >= 0xD800 && result.codepoint <= 0xDFFF) {
```
andychu, line 143: I like to write these sorta like Python style `3 <= x < =5` like this

```
(3 <= x && x <= 5)
```

andychu (commented):
OK this is awesome!  only some superficial comments

I pushed a slight change to the spec test

data_lang/j8.h
```diff
@@ -2,9 +2,10 @@
 #define DATA_LANG_J8_H
 
 #include <stdio.h>   // sprintf
-#include <string.h>  // memcmp
+#include <string.h>  // memcmp, memcpy, strlen
 
 #include "data_lang/utf8_impls/bjoern_dfa.h"
```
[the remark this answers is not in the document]
PossiblyAShrub: Not yet... I still need to update the other `*EncodeOne` functions. I'll leave that for another PR.

data_lang/j8.h
```diff
@@ -90,46 +91,40 @@ static inline int J8EncodeOne(unsigned char** p_in, unsigned char** p_out,
   //
   // UTF-8 encoded runes and invalid bytes
   //
-  unsigned char* start = *p_in;  // save start position
-  uint32_t codepoint = 0;
-  uint32_t state = UTF8_ACCEPT;
+  Utf8Result_t result = utf8_decode(*p_in);
```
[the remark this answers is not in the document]
PossiblyAShrub: Playing around with goldbolt (C: https://godbolt.org/z/1jb3xo9qT; C++: https://godbolt.org/z/zbnqh3eW8,) there is practically no difference. In each case, we essentially end up writing the struct directly into the caller's stack frame.

But, with these functions being inline, I don't think the passing style matters at all. The compiler will specialize to the same code either way.

I see that out params are the norm, so I'll use that for consistency.

data_lang/utf8.h
PossiblyAShrub: Yeah, we don't need that here. All the code in this file fits within the shared subset of C and C++.

data_lang/utf8.h
PossiblyAShrub: Just came out in C23^[ref](https://en.wikipedia.org/wiki/C23_(C_standard_revision)#Constants)! Should be able to start using them in a few decades :P

PossiblyAShrub:
I've made changes per your feedback. This is ready for another look.

andychu:
Great thanks!

Yeah we should remove Bjorn DFA altogether, just so we have more minimal code.  (We don't need the surrogate property in the other case, but that's OK)

Also there is a placeholder thread about the Python/C++ wrappers for this function.  I guess I want to try to avoid heap allocations on every iteration of `for rune in (runes(s)) { in the C++ case`  ???  Not quite sure

I think mycpp does have a special optimization where you can reeturn `Tuple<3>()` by value, not as a GC object, and maybe we can use that

To a first approximation we count allocations (and those are much more expensive than say branches, which is why I think the speed of the Bjorn DFA style is optional now)

andychu:
And it's great you know UTF-8 inside and out now :)

PR #1960, opened 2024-05-07
core: skip tracer on ext/heredoc thunks

melvinw:
The tracer overhead is guaranteed to be a waste in both of these cases. Skipping it decreases `benchmarks/osh-runtime/bin_true` wall time by 5-7%.


andychu:
Hmmm, surprised the effect is so big here

I was hoping to add more shell-level tracing, e.g. with TSV/JSON output and so forth

There was a Zulip thread about "zero overhead tracing as a motivation for performance"

I guess our language is kinda slow for that, and ebpf does a lot better ... but I still think there are some use cases for shell-level tracing (e.g. enhance set -x) and OILS_TRACE_DIR ...

melvinw:
Sounds good. We could tweak `SkipTrace` to accept a tracer to make the decision a little smarter, too.

If I remember correctly the main thing this skips is an allocation in `tracer.OnNewProcess()`. I had a more targeted patch that deferred that to later that might help in lieu of this too https://github.com/oilshell/oil/commit/d9cb6a2c21072476fbee00a9941c27b5a6afa6a7

Let me know what you decided and I'll adjust.

melvinw:
One more clarification just in case it's not clear: we only skip the tracer in the child after forking (and only for ExternalThunk and HereDocWriterThunk). We still call `OnProcessStart()` from the parent unconditionally

PR #1961, opened 2024-05-07
[data_lang] Remove remaining uses of Bjorn DFA and delete unused UTF-8 decoders

PossiblyAShrub:
The goal is to get rid of the other decoders so we can focus only on `data_lang/utf8.h`.


cpp/data_lang.cc
```diff
@@ -165,20 +165,18 @@ BigStr* ShellEncodeString(BigStr* s, int ysh_fallback) {
...
-  for (int i = start; i < end; ++i) {
-    // This var or a static_cast<> is necessary.  Should really change BigStr*
-    // to use unsigned type
-    uint8_t c = s->data_[i];
-    decode(&state, &codepoint, c);
-    if (state == UTF8_REJECT) {
+  Utf8Result result;
+
+  for (int i = start; i < end;) {
+    utf8_decode(reinterpret_cast<unsigned char*>(&s->data_[i]), &result);
```
andychu, line 171: All the reinterpret_cast<> in our code bugs me, though I checked CPython and it the same issue.

They do casting whenever they pick a byte out of a string

The char vs. unsigned char thing is such a wart in C

----

I think this would be slightly cleaner as `s->data_ + i`

It's basically staying in "pointer land" rather than getting a value and taking its address

Also the precedence of `&` and `->` and `[]` confuses me so I like that style

cpp/NINJA_subgraph.py
```diff
@@ -96,9 +96,7 @@ def NinjaGraph(ru):
         ],
         # This includes in the release tarball.  (It doesn't change any
```
andychu, line 97: I would delete this command and the headers=[] line

andychu (commented):
Great, thanks for tracking down all the usages!

(It is a bit annoying how we have the same bindings in pyext/fastfunc and cpp/data_lang, hence the goal to reduce C code as much as possible)

PossiblyAShrub:
Thanks for the quick review! I've made your suggested changes. This is ready for another look.

PR #1962, opened 2024-05-08
Tabular alignment

justinpombrio:


andychu:
Hm is the intention to do dicts as well?  I think we talked about this on Zulip, but I'm not sure what the conclusion is

I think it looks better for an array of numbers

The nested test cases look a little weird, but maybe those are uncommon cases

data_lang/pretty_test.txt
```diff
@@ -205,184 +284,173 @@ Input
...
->       1
+>     "variety-pack": [
+>         null,
+>         [
+>             "Los",
+>             "pollitos",
+>             "dicen",    "pío",
+>             "pío",      "pío"
+>         ],
+>         [
```
andychu, line 430: This one seems to look weirder than it used to?  I wonder if having the rule only apply to "leaf" containers, i.e. containers that don't have any containers, would help

Although arguably you may want to format dicts like

```
{k: [1,2], k2: [3,4], ... }
```

etc

justinpombrio:
> Hm is the intention to do dicts as well? I think we talked about this on Zulip, but I'm not sure what the conclusion is

I'd suggest not doing it for dicts, because (i) it would need to align both the keys and the values, which might look stranger, and (ii) since you're aligning both, they'll tend to be longer, making dicts use the tabular layout less often, making it less useful and more surprising to users; and (iii) I've seen tabular alignment for lists elsewhere, but never for dicts.

> The nested test cases look a little weird, but maybe those are uncommon cases

You mean the formatting of `[1, [2, [3, [4, [5, [6]]]]]]`? That is not using the new tabular alignment (there aren't enough elements for it to get chosen by the PP algorithm), it's behaving the same as before. The reason it looks worse than before is because I changed the test cases from having an indentation level of 2 to 4, to match the real behavior.

I agree that `[1, [2, [3, [4, [5, [6]]]]]]` formats poorly. I added it because it's a case where Wadler-style printing doesn't work well. Though it's only this bad when you have a (hopefully rare) combination of elements smaller than the indentation level (4 chars) _and_ deep nesting _and_ the nesting is non-uniform.

andychu:
Sorry for the delayed response

Let me poke at this a bit more tonight,  I think it's probably fine but I thought it may be weird if we do tabular alignment of lists but not dicts

But I can see they are slightly different

andychu:
Oh one thing that could help us decide is to do the ASDL formatting, because we actually have real testdata for that, not synthetic testdata

e.g. we can do `osh -n configure` and then  eyeball it to make sure that the AST prints/fits nicely, based on real data

data_lang/pretty.py
```diff
@@ -429,6 +460,60 @@ def _Join(self, items, sep, space):
...
+        sixth,
+        seventh,
+        eighth
+        ```
+
+        The first "single line" style is used if the items fit on one line.  The
+        second "tabular' style is used if the flat width of all items is no
+        greater than `self.max_tabular_width`. The third "multi line" style is
+        used otherwise.
+        """
```
andychu, line 489: Maybe write something about why we're not using a more obvious rule?  It's expensive?

e.g. as long as 2 items fit on every line, it could be tabular, etc.

e.g. if I have a screen of 80, and I have 10 strings of length no more 30, then 2 can fit  on every line


I can sorta see why that requires a search, but it's probably good to document it

andychu (commented):
Thanks, I tested this out and it seems nice!

I bumped the width to 16 to accomodate floats, and added a little demo

data_lang/pretty.py
```diff
@@ -429,6 +460,60 @@ def _Join(self, items, sep, space):
...
+
+        if len(items) == 0:
+            return _Text("")
+
+        max_flat_len = items[0].measure.flat
+        seq = [items[0]]
+        for item in items[1:]:
```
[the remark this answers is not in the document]
justinpombrio: Done. 2/3 of the loops like this could use `i != 0`, the third one really wanted `i != len(items)` b.c. the padding depends on the item before it.

data_lang/pretty.py
justinpombrio: Done.

That's a good thing to document, really, because the reason for it isn't obvious and explains a bit about the sorts of things you can and can't do with the pretty printing language.

I documented it in a comment (`#`) instead of a doc string (`"""`) b.c. it's about the implementation rather than the interface.

andychu:
Merged, thanks!

PR #1965, opened 2024-05-10
[data_lang] Use new utf8 decoder in osh/string_ops

PossiblyAShrub:
Now all utf8 decoding goes through the new decoder!

Because our decoder has separate codes for each encoding error, we can give much more informative errors compared to, eg. Python:

```
ysh ysh-0.21.0$ b'\yC1\y80' => trimStart()
  b'\yC1\y80' => trimStart()
                          ^
[ interactive ]:7: Space required before (
ysh ysh-0.21.0$ = b'\yC1\y80' => trimStart()
  = b'\yC1\y80' => trimStart()
  ^
[ interactive ]:8: fatal: Utf8 Error: Decoded Overlong at 0

ysh ysh-0.21.0$ python3
Python 3.10.4 (main, Jul 18 2023, 22:16:42) [GCC 13.1.1 20230429] on linux
>>> bytes([0xC1, 0x80]).decode('utf8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc1 in position 0: invalid start byte
```


PossiblyAShrub:
This is ready for a review.

Some changes I am thinking about for future PRs:
- We should unify `DecodeUtf8Char` and `NextUtf8Char` as they are often called together leading to double the work in decoding.
- I ran into an issue with `utf8_decode` expecting a nul-terminated string. I used a work-around, but, if possible, it might be better to change `data_lang/{utf8,j8}.h` to avoid the nul-terminator.

spec/var-op-len.test.sh
```diff
@@ -113,21 +113,21 @@ true
...
-[ stdin ]:3: warning: Invalid UTF-8 continuation byte
-[ stdin ]:3: warning: Invalid UTF-8 continuation byte
-[ stdin ]:3: warning: Invalid start of UTF-8 character
-[ stdin ]:3: warning: Invalid start of UTF-8 character
-[ stdin ]:3: warning: Invalid UTF-8 continuation byte
-[ stdin ]:3: warning: Invalid UTF-8 continuation byte
-[ stdin ]:3: warning: Invalid UTF-8 continuation byte
-[ stdin ]:3: warning: Invalid start of UTF-8 character
-[ stdin ]:3: warning: Invalid start of UTF-8 character
+[ stdin ]:3: warning: Utf8 Error: Bad Encoding at 0
```
andychu, line 116: Some cosmetic issues

- Utf8 -> UTF-8   -- should be able to `sed` this out in most places
- "at 5" -> how about "at offset 5 in string" ?

Or I'm not sure if this helps, but "at offset 5 in string of length 10" might narrow down the possibilities for readers


----

I wanted to  integrate these errors with the `loc_t`  system, but that's separate from this change

andychu (commented):
This is great, thanks for consolidating all of this!  

I am kinda annoyed we have to "wrap things twice", but it looks like you didn't have a problem here

Also  it looks like returning `Tuple2<>` by value works, which is great (so we don't have GC objects per code point)

----

And maybe having to wrap things twice is good "discouragement" from writing (unsafe) C code ...

I think it's very justified for UTF-8, but in general I like as much typed Python as possible

---

Just some cosmetic comments

PossiblyAShrub:
I settled on `"UTF-8 Error: Bad Encoding at byte index %d in string of length %d"`. It's a little long, but much less ambiguous.

cpp/data_lang.cc
```diff
@@ -160,6 +160,23 @@ BigStr* ShellEncodeString(BigStr* s, int ysh_fallback) {
   return buf->getvalue();
 }
 
+Tuple2<int, int> Utf8DecodeOne(BigStr* s, int start) {
+  assert(0 <= start && start <= len(s));
```
andychu, line 164: I think this should be `start < len(s)`, because `s[len(s)]` is not a valid part of the string

I think of it like a "half open interval" in math `[0, n)` is written

`0 <= i && i < n`

in both Python and C++

pyext/fastfunc.c
```diff
@@ -80,6 +80,33 @@ func_PartIsUtf8(PyObject *self, PyObject *args) {
...
+func_Utf8DecodeOne(PyObject *self, PyObject *args) {
+  char *string;
+  size_t length;
+  int start;
+
+  if (!PyArg_ParseTuple(args, "s#i", &string, &length, &start)) {
+    return NULL;
+  }
+  // Bounds check for safety
+  assert(0 <= start && start <= length);
```
andychu, line 93: ditto, I think this should be start < length

andychu (commented):
Sorry one small thing I missed the first time around

cpp/data_lang.cc
PossiblyAShrub: Ah, logically this is true, but remember that we have a nul-terminator. So, `s[len(s)]` -- and in turn `fastfunc.Utf8DecodeOne(s, len(s))` -- is actually well defined.

I also have tests over this behaviour:

```python
s = 'h\xE2\xA1\x80\xC5\x81'

# UTF8_ERR_END_OF_STREAM = 6
self.assertEqual((-6, 0), fastfunc.Utf8DecodeOne(s, 6))
```
(https://github.com/oilshell/oil/pull/1965/files#diff-6d07d51003e574f4cbdd6d2a6ff6a13c1e60dd0c854b23bf9da960497ebef712R52) (+ an identical one for cpp)

I don't believe we rely on this from the python/mycpp side, but we do rely on it in `data_lang/j8.h`. While we could restrict `fastfunc.Utf8DecodeOne` so that $0 \le i \lt \text{len}(s)$, I don't want to have the interface deviate from `utf8_decode` for consistency.

andychu: Hm I think it's the difference between a Python string/mycpp string (has a length), and a C string (has no length)

But if you think it's OK, I would add a comment by both asserts that says this case is well defined, just because it "looks weird"

I agree the way we use it in j8.h is a little weird, but it is correct, and is probably the best solution.

pyext/fastfunc.c
```diff
@@ -89,7 +89,10 @@ func_Utf8DecodeOne(PyObject *self, PyObject *args) {
   if (!PyArg_ParseTuple(args, "s#i", &string, &length, &start)) {
     return NULL;
   }
-  // Bounds check for safety
+
+  // Bounds check for safety. start can equal length because the string has a
+  // nul-terminator. So utf8_decode(s + len(s)) is valid and would decode that
```
andychu, line 94: Hm actually the fact that you would get a zero code point from decoding the terminator seems bad !!!  That NUL terminator is not part of the string, just like it's not in C.

Or maybe it's OK because the caller should always check the error and not use the code point zero ?

---

It still seems a bit confusing but I don't want to have us blocked on this, since it's probably not causing a problem now


Maybe I need to look at the tests more closely

I think it's possible the unit tests may be asserting some properties that we don't need in real code, i.e. the `s[len(s)]` case

PossiblyAShrub: Yeah, there's some subtlety around zero-codepoints. I *cannot* assume inside `utf8_decode` that those denote the end of a string. Instead, we return both that the decoded codepoint is zero and that this *may* be the end of the string. But, like the surrogate errors, the caller can decide if that's fatal or not. We need this because, unlike `data_lang/j8.h`, `osh/string_ops.py` is perfectly fine with zero-codepoints in the middle of the string.

Although, I did add the `mylib.ByteAt(s, i) == 0` check in `osh/string_ops.py`... so `utf8_decode` doesn't actually see any nul-terminators in the middle of the code. I should probably do this instead:

```python
codepoint_or_error, _bytes_read = fastfunc.Utf8DecodeOne(s, i)
if codepoint_or_error == UTF8_ERR_END_OF_STREAM:
  # utf8_decode found a nul-terminator, but we allow those in the middle of our strings
  return 0
elif codepoint_or_error < 0:
  raise error
return codepoint_or_error
```

andychu: Oh hm I didn't look carefully at the special case for NUL.

I think if we can get rid of that we should ... The definition of an Oils/YSH string is the same as a Python 2 string basically -- it's an array  of bytes, which can include NULs, which may or may not be UTF-8 encoded

That is for PURE data operations like string trim() etc.

BUT when you make a syscall like `chdir()`, then Unix does NUL terminated strings, so it's OK to truncate in that case. 

But for regular string processing we shouldn't treat it as special.

----

j8.h seems to deal with NUL fine, without any special cases

```
ysh andy@andy-VirtualBox:~/git/oilshell/oil$ var x = b'a\y00\y00b'

ysh andy@andy-VirtualBox:~/git/oilshell/oil$ = x
(Str)   "a\u0000\u0000b"

ysh andy@andy-VirtualBox:~/git/oilshell/oil$ json write (x)
"a\u0000\u0000b"
```

The subtlety is that

- we rely on the trailing NUL only for INVALID sequences.  If you have a 4 byte sequence that's cut off at 3 bytes, we don't know until we hit the NUL.  We need teh NUL.
- we do NOT rely on it for VALID sequences.  There we can rely on the length check

So yeah I wrote that big comment in `j8.h` because I found this confusing. 

---

But I think we should try to design the API to get rid of that special case in Python -- it will affect the forward string iterator too

---

I also think if we never actually access `s[len(s)]`, and can rely on the length check, we should not handle that case and make the assertion fail

That is I think we should distinguish our Python-level APIs from the C one

- utf8decode() is  a low level C helper.
- J8 encoding, utf-8 validity, code point iteration are Python-level APIs

I think it's more important to be strict about the Python APIs, which use GC strings, with internal NULs

And utf8decode() is basically an implementation detail, which uses C strings which are buffers.

andychu:
Also this issue is tricky, so we have `test/spec.sh nul-bytes` for some test cases

We might want to add some for string strip / string iteration there

(Also CPython sometimes asserts for internal NUL, whereas I believe the right behavior for Oils is to only "truncate at the edges" for `chdir()` etc. , but be able to deal with arbitrary binary data  with every in-memory operation)

PossiblyAShrub:
> That is I think we should distinguish our Python-level APIs from the C one
> 
> - utf8decode() is a low level C helper.
> - J8 encoding, utf-8 validity, code point iteration are Python-level APIs
> 
> I think it's more important to be strict about the Python APIs, which use GC strings, with internal NULs
> 
> And utf8decode() is basically an implementation detail, which uses C strings which are buffers.

Okay, I get your point now. Originally I didn't want this distinction, ie. `fastfunc.Utf8DecodeOne` should be a *very thin* wrapper around `utf8_decode`. But I see how that's undesirable.

I'll throw up another PR with a few changes:
- Make the `fastfunc.Utf8DecodeOne` "safe" and work with the python2/oils string model (over the C one)
- Remove the NUL special casing in our `osh/string_ops.py` UTF-8 functions (let `fastfunc.Utf8DecodeOne` handle that)
- Add some tests to validate that string APIs like `Str => trim*()` are resilient to zero-codepoints

PR #1967, opened 2024-05-12
[fastfunc] Use the python2/oils string model in Utf8DecodeOne

PossiblyAShrub:
Per the discussion in #1965, I've made the following updates:

- Made the `fastfunc.Utf8DecodeOne` "safe" by exposing the python2/oils string model (over the C one)
- Removed the NUL special casing in our `osh/string_ops.py` UTF-8 functions (`fastfunc.Utf8DecodeOne` now handles that)
- Added some tests to validate that string APIs like `Str => trim*()` are resilient to zero-codepoints

### DIFF
--- a/cpp/data_lang.cc
+++ b/cpp/data_lang.cc
@@ -175,13 +175,6 @@ Tuple2<int, int> Utf8DecodeOne(BigStr* s, int start) {
     codepoint_or_error = decode_result.codepoint;
   }
 
-  // Follow the python2/oils string model. See func_Utf8DecodeOne in
-  // pyext/fastfunc.c for details.
-  if (decode_result.error == UTF8_ERR_END_OF_STREAM) {
-    codepoint_or_error = 0;
-    decode_result.bytes_read = 1;
-  }
-
   return Tuple2<int, int>(codepoint_or_error, decode_result.bytes_read);
 }
 
--- a/cpp/data_lang.cc
+++ b/cpp/data_lang.cc
@@ -161,10 +161,8 @@ BigStr* ShellEncodeString(BigStr* s, int ysh_fallback) {
 }
 
 Tuple2<int, int> Utf8DecodeOne(BigStr* s, int start) {
-  // Bounds check for safety. start can equal length because the string has a
-  // nul-terminator. So utf8_decode(s + len(s)) is valid and would decode that
-  // terminator (also setting UTF8_ERR_END_OF_STREAM).
-  assert(0 <= start && start <= len(s));
+  // Bounds check for safety
+  assert(0 <= start && start < len(s));
 
   const unsigned char* string = reinterpret_cast<unsigned char*>(s->data());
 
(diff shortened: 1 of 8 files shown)


cpp/data_lang.cc
```diff
@@ -161,10 +161,8 @@ BigStr* ShellEncodeString(BigStr* s, int ysh_fallback) {
 }
 
 Tuple2<int, int> Utf8DecodeOne(BigStr* s, int start) {
-  // Bounds check for safety. start can equal length because the string has a
-  // nul-terminator. So utf8_decode(s + len(s)) is valid and would decode that
-  // terminator (also setting UTF8_ERR_END_OF_STREAM).
-  assert(0 <= start && start <= len(s));
+  // Bounds check for safety
+  assert(0 <= start && start < len(s));
```
andychu, line 165: This can be DCHECK(), which is on everywhere except the release build

(it doesn't exist in CPython, only in our C++)

pyext/fastfunc.c
```diff
@@ -104,6 +102,14 @@ func_Utf8DecodeOne(PyObject *self, PyObject *args) {
     codepoint_or_error = decode_result.codepoint;
   }
 
+  // utf8_decode treats zero-bytes as C-style NUL-terminators. But python2/oils
+  // strings treat these as zero-codepoints. Translate END_OF_STREAM errors
+  // (resulting from zero-bytes) to valid zero-codepoints.
+  if (decode_result.error == UTF8_ERR_END_OF_STREAM) {
```
andychu, line 108: Hm it seems like this works but isn't it cleaner to not return UTF8_ERR_END_OF_STREAM in the first place?

Basically I imagine



- `0xce 0x00` -- truncated encoding relies NUL bytes, returns  UTF8_ERR_TRUNCATED_BYTES
- 0x00` after some other valid encoding -- this is UTF8_OK 

I think it's possible to distinguish these cases in utf8decode()?  Or would it also need a length?

andychu (commented):
Thanks for following up on this!  (delayed review due to Mother's day :) )

andychu:
Thinking about it a little more, I think the caller is responsible for not calling utf8_decode() on the trailing NUL, in the valid case

Because you don't want to get UTF8_OK in that case

But in the invalid case, the caller doesn't know there's a NUL there, but it must supply it

---

I think we actually don't need UTF8_END_OF_STREAM at all?  I think we can just get rid of it, and make the caller is responsible 

1. NUL terminating every string
2. not over-running the buffer -- i.e. don't call it when on the  NUL **past** the end of the string, only ones **before** the end of the string (which it knows)

It is a bit weird and subtle, but I think it makes sense

(and this issue is why I was initially confused about the whole state machine / "inverting" the Crockford code)

PossiblyAShrub:
Yeah, you were right about removing the `END_OF_STREAM` error state; it simplified the code while preserving correctness. The "nul-terminator required but you must keep track of the buffer end" rule is certainly subtle, so I made sure to note it in the doc-comment.

andychu:
Looks very nice now, thanks!

PR #1977, opened 2024-05-26
[mycpp] add some CFG tests for examples

melvinw:
This commit instruments cppgen_pass.py to emit a control flow graph and
adds some tests for a selection of the trickier examples. It also
refactors ControlFlowGraph block handling some more and adds better
break/continue helpers for loops.


mycpp/cppgen_pass.py
```diff
@@ -2402,6 +2462,8 @@ def visit_func_def(self, o: 'mypy.nodes.FuncDef') -> T:
             return
 
         func_name = o.name
+        if self.decl:
+            self.cfgs[_StripMycpp(o.fullname)] = pass_state.ControlFlowGraph()
```
melvinw, line 2466: TODO: use the new `SymbolPath` stuff here once it's on master

mycpp/pass_state.py
```diff
@@ -282,3 +299,31 @@ def __exit__(self, *args) -> None:
...
+
+def DumpControlFlowGraphs(cfgs: dict[str, ControlFlowGraph], facts_dir='_tmp/mycpp-facts') -> None:
+    """
+    Dump the given control flow graphs and associated facts into the given
+    directory as text files that can be consumed by datalog.
+    """
+    edge_facts = '{}/cf_edge.facts'.format(facts_dir)
+    os.makedirs(facts_dir, exist_ok=True)
+    with open(edge_facts, 'w') as cfg_f:
+        for func, cfg in cfgs.items():
```
melvinw, line 327: TODO: sort by func

mycpp/cppgen_pass.py
```diff
@@ -2142,55 +2177,63 @@ def _write_str_switch(self, expr, o):
...
-        self.def_write_ind('}\n')
+            self.indent -= 1
+            self.def_write_ind('}\n')
 
     def visit_with_stmt(self, o: 'mypy.nodes.WithStmt') -> T:
         """
         Translate only blocks of this form:
 
-        with switch(x) as case:
+        WIth switch(x) as case:
```
andychu, line 2236: this got mangled

andychu (commented):
Hm my first thought is - can this be its own pass?    like mycpp/control_flow_pass.py which is like const_pass.py

In principle it seems like it only needs to hook into a few things


- loops - for while
- conditionals - if, switch str_switch, tagswitch
- exceptions

And it has

- a loop stack
- a dict per function node



The 3 switch constructs are the ones that seem most tightly coupled

Aside from that do you see any obstacles?

---

I want to understand the dependencies between passes better.  Right now they are all mixed up in that big constructor to the pass


It doesn't seem like this needs to do anything with expressions?   which suggests it might be a standalone pass -- 

---

Although ternary expressions are technically control flow

Do we do anything about those, and is correctness affected if you don't handle `x() if b else y()`  ?

mycpp/TEST.sh
```diff
@@ -246,6 +246,22 @@ test-invalid-examples() {
   done
 }
 
+test-control-flow-graph() {
+  local mycpp=_bin/shwrap/mycpp_main
+  ninja $mycpp
+  for ex in mycpp/examples/*.py; do
+
+		local data_dir=testdata/control-flow-graph/$(basename -s .py $ex)
```
andychu, line 254: looks like there are some tabs here

andychu:
Also it would be nice to have some technical design info about the control flow graph, either in some docstrings or a separate markdown file

Is there one "obvious" kind of control flow graph?  Or are there are different ones that are appropriate for different analyses?

What properties does the graph obey?

I have not worked with them very much -- I only refactored the Python bytecode generator in `opy/`.   There I think it is mostly used to desugar into jumps ??   The bytecode only has jumps, but the source code has `for while if ternary-if`, etc.



I am not sure there is any analysis or optimization there based on the graph, though I Haven't looked at that code in a long time

melvinw:
> Also it would be nice to have some technical design info about the control flow graph, either in some docstrings or a separate markdown file

Ah sure. I'll flesh out the doc string. If it starts to sprawl too much I'll move it to a separate file.

> Is there one "obvious" kind of control flow graph? Or are there are different ones that are appropriate for different analyses?

What we're building here is a pretty fine-grained or "full flow graph" (according to Wikipedia). Every statement has a node. There are other flavors that group statements into blocks. Otherwise there isn't much variation I think? Anything interesting comes from either the semantics of the underlying language or the facts that you annotate the graph with when doing dataflow analysis (this is where you see lots of variation I imagine). I will include notes about this in the docs. 

> What properties does the graph obey?

Nothing special. It's just a directed graph with no islands. If you remove the back-edges generated by loops the remaining edges should form a DAG.

andychu:
OK great, if we can have more pure functions in mycpp/util.py and fewer methods like _write_cases() then I think that will improve things

The visitor tends to "couple" everything together, when really there can be separate functions

I was thinking of `mycpp.asdl`, and we probably still want that long term, but separate functions sounds like a good idea

to avoid duplication but also reuse the same structure

melvinw:
OK @andychu I split this back out into a separate pass and pulled the switch case collection into a util. I also added a few details to the docstring for `ControlFlowGraph`. Have another look?

mycpp/cppgen_pass.py
```diff
@@ -556,6 +567,10 @@ def accept(self, node: Union[Statement, Expression]) -> Optional[T]:
                 return res
             else:
                 try:
+                    cfg = self.current_cfg()
+                    if cfg and not isinstance(node, Block) and not isinstance(node, ForStmt) and not isinstance(node, WhileStmt):
+                        cfg.AddStatement()
```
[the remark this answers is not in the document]
melvinw: fixed this in the new pass

mycpp/control_flow_pass.py
```diff
@@ -0,0 +1,418 @@
...
+            if (isinstance(stmt, ExpressionStmt) and
+                    isinstance(stmt.expr, StrExpr)):
+                continue
+
+            if isinstance(stmt, FuncDef) and stmt.name == '__repr__':
+                continue
+
+            self.accept(stmt)
+
+    def visit_global_decl(self, o: 'mypy.nodes.GlobalDecl') -> T:
```
andychu, line 312: I think we can delete all this stuff, i.e. all the empty methods

But yeah the other passes are a mess too, so we can do that all at once

I think I wasn't sure what "dialect" of Python we were using until the translation was done, so I kept them all "just in case"

But now it's done so we can go back and gradually clean things up

mycpp/control_flow_pass.py
```diff
@@ -0,0 +1,418 @@
...
+                except UnsupportedException:
+                    pass
+                return None
+
+    # Not in superclasses:
+
+    def visit_mypy_file(self, o: 'mypy.nodes.MypyFile') -> T:
+        # Skip some stdlib stuff.  A lot of it is brought in by 'import
+        # typing'.
+        if o.fullname in ('__future__', 'sys', 'types', 'typing', 'abc',
```
andychu, line 83: Can you factor this out into something like util.ShouldSkipPyFile()  ?

mycpp/control_flow_pass.py
```diff
@@ -0,0 +1,418 @@
...
+                # keep trying, so we can produce more error
+                # messages. Generate a temp of the right type to keep
+                # from causing more downstream trouble.
+                except UnsupportedException:
+                    res = self.alloc_temp(self.node_type(node))
+                return res
+            else:
+                try:
+                    cfg = self.current_cfg()
+                    if cfg and not isinstance(node, Block) and not isinstance(node, ForStmt) and not isinstance(node, WhileStmt):
```
andychu, line 70: OK I didn't notice this in the previous incarnation -- the accept() has some logic

That is interesting, maybe put a comment here to explain it

andychu (commented):
Thanks!  Sorry you had to shuffle it around a lot, but I think this is much better

There is some of the duplication we were worried about, but it's not too much

If I had done this initially, then we wouldn't have the `if self.decl` and `if self.forward_decl` stuff all over cppgen_pass

I think you can delete the empty methods, but it's also OK to wait until the whole thing works end-to-end, so we are sure we haven't missed any cases

(e.g. ternary if and such)

mycpp/control_flow_pass.py
melvinw: oh nice! i wasn't sure what the defaults for `ExpressionVisitor` and `StatementVisitor` were. let's drop these for now

melvinw:
> Thanks! Sorry you had to shuffle it around a lot, but I think this is much better

No sweat. I agree, it's much more obvious what this is doing now without all the other noise

andychu:
Looks great, thanks!

PR #1978, opened 2024-05-28
mycpp: avoid subclass name colissions

melvinw:
Subclass names can collide, e.g. `builtin.func_misc.Type` and `builtin.meta_osh.Type` which share the base `core.vm._Callable`. This isn't a big deal until you need to query the subclass->base relationship, which a future commit will do.

This commit also changes `pass_state.Virtual.virtuals` into a mapping from (class, method) -> (base, method). If an entry maps to `None` that means the given method is part of the base class. This will allow a future commit to find the base for a given subclass.

The commit also changes mycpp to use fully-qualified class names when interacting with `pass_state.Virtual` to avoid ambiguity.

### DIFF
--- a/mycpp/cppgen_pass.py
+++ b/mycpp/cppgen_pass.py
@@ -18,7 +18,7 @@ from mypy.nodes import (Expression, Statement, NameExpr, IndexExpr, MemberExpr,
 
 from mycpp import format_strings
 from mycpp.crash import catch_errors
-from mycpp.util import log, join_name_cpp, split_py_name
+from mycpp.util import log, join_name, split_py_name
 from mycpp import util
 
 from typing import Tuple, List
@@ -2414,7 +2414,7 @@ class Generate(ExpressionVisitor[T], StatementVisitor[None]):
         if not self.decl and self.current_class_name:
             # definition looks like
             # void Class::method(...);
-            func_name = join_name_cpp((self.current_class_name[-1], o.name))
+            func_name = join_name((self.current_class_name[-1], o.name))
         else:
             # declaration inside class { }
             func_name = o.name
@@ -2502,7 +2502,8 @@ class Generate(ExpressionVisitor[T], StatementVisitor[None]):
 
             # e.g. class TextOutput : public ColorOutput
             if base_class_name:
-                self.always_write(' : public %s', join_name_cpp(base_class_name, True))
+                self.always_write(' : public %s', join_name(base_class_name,
+                                                            strip_package=True))
 
             self.always_write(' {\n')
             self.always_write_ind(' public:\n')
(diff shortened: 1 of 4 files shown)


mycpp/pass_state.py
```diff
@@ -18,17 +18,17 @@ class Virtual(object):
...
 
     def __init__(self) -> None:
-        self.methods: dict[str, list[str]] = defaultdict(list)
-        self.subclasses: dict[str, list[str]] = defaultdict(list)
-        self.virtuals: list[tuple[str, str]] = []
-        self.has_vtable: dict[str, bool] = {}
-        self.can_reorder_fields: dict[str, bool] = {}
+        self.methods: dict[tuple[str], list[str]] = defaultdict(list)
+        self.subclasses: dict[tuple[str], list[tuple[str]]] = defaultdict(list)
+        self.virtuals: dict[tuple[tuple[str], str], Optional[tuple[str, str]]] = {}
```
andychu, line 23: It might be nice to define a type alias `SymbolName = tuple[str]` or `SymbolPath`

mycpp/cppgen_pass.py
```diff
@@ -2687,7 +2688,7 @@ def visit_class_def(self, o: 'mypy.nodes.ClassDef') -> T:
                                 callee.name == '__init__'):
                             base_constructor_args = expr.args
                             #log('ARGS %s', base_constructor_args)
-                            self.def_write(' : %s(', base_class_name)
+                            self.def_write(' : %s(', join_name_cpp(base_class_name, True))
```
andychu, line 2691: should call this with strip_package=True (generally we follow the rule that optional args are always named)

Also why does strip_package exist?  It could use a comment

andychu (commented):
Thanks, this seems reasonable

mycpp/pass_state.py
melvinw: Good idea. Will add this

mycpp/cppgen_pass.py
melvinw: Ah, yeah `strip_package` thing is awkward. Apparently we omit the top level directories (e.g. `core`, `builtin`, `ysh`) from the c++ namespaces. Will add a comment

mycpp/util.py
```diff
@@ -14,8 +14,33 @@
...
 
+SymbolPath = tuple[str]
+
 
 def log(msg: str, *args: Any) -> None:
     if args:
         msg = msg % args
     print(msg, file=sys.stderr)
+
+def join_name(parts: SymbolPath, strip_package: bool = False, delim: str = '::') -> str:
```
andychu, line 25: I think it's better not to have the param if we don't use it, if it's always ::

It can be a constant in the file, or just repeated

melvinw: Another PR I have sets this to `.` for use with the datalog files

PR #1982, opened 2024-06-02
[mycpp] add souffle binary build library

melvinw:
Splitting this out from #1981


andychu:
Thanks, this generally looks good

Can you write a self-contained shell smoke test ?

- to run the built souffle on some kind of "hello world"
- maybe compile and run the output (if it's easy)

We actually moved most things to the "wedge" system, which

1. runs on contributor machines, and is automated with CI tasks dev-setup-{alpine,debian}
2. gets baked into the Docker images (so we don't rebuild re2c/souffle every time)

Example here with uftrace:

https://github.com/oilshell/oil/blob/master/deps/source.medo/uftrace/WEDGE#L81



I should be able to make the wedge itself and the Docker image/layers

But the automated smoke test example will help

melvinw:
Pushed a skeleton for the wedge with a smoke test

andychu:
OK thanks, I will run this locally and add it to the `cpp-*` CI images (the ones that have mycpp)

Might take a few days, I'll let you know

melvinw:
Sounds good. Let me know if you run into any issues!

PR #1987, opened 2024-06-03
[build] check in souffle c++ 

melvinw:
Souffle takes forever to build. To make this stuff more accessible to
all devs,  this commit adds the souffle C++ headers to `vendor/souffle`
and checks in the generated C++ for `mycpp/datalog/call-graph.dl`. It
also tweaks the ninja libraries to accommodate the new process.


andychu:
Can you add something to `soil/worker.sh` to show that it builds and runs?    I guess we can use the `deps/source.medo/souffle/smoke-test.dl`

i.e. check that into `prebuilt/`, and then have a little soil task that builds and runs the C++, similar to the smoke test you already wrote


As something that's de-coupled from whatever we do with mycpp

andychu:
I think it could belong in the `cpp-small` task

prebuilt/NINJA_subgraph.py
```diff
@@ -74,4 +76,8 @@ def NinjaGraph(ru):
             '//cpp/frontend_flag_spec',
         ])
 
-    ru.souffle_cpp('mycpp/datalog/call-graph.dl')
+    # NOTE: To avoid failures on these in CI, these targets are only enabled if
+    # the user has the souffle executable in their PATH.
+    if subprocess.call('which souffle > /dev/null', shell=True) == 0:
+        ru.souffle_cpp('mycpp/datalog/call-graph.dl')
+        ru.souffle_cpp('deps/source.medo/souffle/smoke-test.dl')
```
melvinw, line 83: @andychu thoughts on this?

melvinw:
Good idea re: the test. Pushed a commit to add that

prebuilt/NINJA_subgraph.py
andychu: If we are always prebuilding the datalog, can we simply have NO Ninja support at all for Souffle?

Logically it could be outside the build system

We could just have a shell script you run every time you change the datalog ...

---

I am not sure what the right tradeoff is yet -- we'll have to work with it

But if it is reasonable to work with, there's no rule that everything has to be in Ninja

the `prebuilt/` stuff is generally a special case -- it is already handled with shell scripts, outside of Ninja

melvinw: Oh right. I kept these ninja targets around because they were shorter than typing out `build/ninja-rules-datalog.sh compile_souffle ... ...`, but we could definitely achieve the same effect with a script... I'll push a commit to replace this with a script

melvinw: OK pushed this. Should be good to go now

andychu:
Great thanks!

PR #1988, opened 2024-06-05
FIX: comments in j8 examples should use # instead of //

ah-edg:
Use `#` instead of `//` in the examples to match the description


andychu:
Great thank you!  

It may grow `//` comments eventually, but I think it's better to be minimal now... :)

PR #1989, opened 2024-06-05
[mycpp] annotate CFG with function calls

melvinw:
This commit adds the first concrete `Fact` and uses it to annotate statements that make function calls. You can feed this in to `_bin/datalog/call-graph` and see a list of statements that might lead to the GC being invoked. This output still doesn't impact the generated C++, though.

### DIFF
--- a/mycpp/control_flow_pass.py
+++ b/mycpp/control_flow_pass.py
@@ -1,17 +1,18 @@
 """
 control_flow_pass.py - AST pass that builds a control flow graph.
 """
+import collections
 from typing import overload, Union, Optional, Dict
 
 import mypy
 from mypy.visitor import ExpressionVisitor, StatementVisitor
 from mypy.nodes import (Block, Expression, Statement, ExpressionStmt, StrExpr,
-                        ForStmt, WhileStmt, CallExpr, FuncDef, IfStmt)
+                        CallExpr, FuncDef, IfStmt, NameExpr, MemberExpr)
 
-from mypy.types import Type
+from mypy.types import CallableType, Instance, Type, UnionType
 
 from mycpp.crash import catch_errors
-from mycpp.util import split_py_name
+from mycpp.util import join_name, split_py_name
 from mycpp import util
 from mycpp import pass_state
 
@@ -24,19 +25,98 @@ class UnsupportedException(Exception):
 
 class Build(ExpressionVisitor[T], StatementVisitor[None]):
 
-    def __init__(self, types: Dict[Expression, Type]):
+    def __init__(self, types: Dict[Expression, Type], virtual, local_vars):
 
         self.types = types
-        self.cfgs = {}
+        self.cfgs = collections.defaultdict(pass_state.ControlFlowGraph)
         self.current_statement_id = None
+        self.current_class_name = None
         self.current_func_node = None
         self.loop_stack = []
+        self.virtual = virtual
+        self.local_vars = local_vars
+        self.imported_names = set()  # MemberExpr -> module::Foo() or self->foo
+
+        # HACK for conditional import inside mylib.PYTHON
+        # in core/shell.py
+        self.imported_names.add('help_meta')
 
     def current_cfg(self):
         if not self.current_func_node:
             return None
 
-        return self.cfgs.get(split_py_name(self.current_func_node.fullname))
+        return self.cfgs[split_py_name(self.current_func_node.fullname)]
+
+    def resolve_callee(self, o: CallExpr) -> Optional[util.SymbolPath]:
+
+        if isinstance(o.callee, NameExpr):
+            return split_py_name(o.callee.fullname)
+
+        elif isinstance(o.callee, MemberExpr):
+            if isinstance(o.callee.expr, NameExpr):
+                is_module = (isinstance(o.callee.expr, NameExpr) and
+                             o.callee.expr.name in self.imported_names)
+                if is_module:
+                    return split_py_name(
+                        o.callee.expr.fullname) + (o.callee.name, )
+
+                elif o.callee.expr.name == 'self':
+                    assert self.current_class_name
+                    return self.current_class_name + (o.callee.name, )
+
+                else:
+                    local_type = None
+                    for name, t in self.local_vars.get(self.current_func_node,
+                                                       []):
+                        if name == o.callee.expr.name:
+                            local_type = t
+                            break
+
+                    if local_type:
+                        if isinstance(local_type, str):
+                            return split_py_name(local_type) + (
+                                o.callee.name, )
+
+                        elif isinstance(local_type, Instance):
+                            return split_py_name(
+                                local_type.type.fullname) + (o.callee.name, )
+
+                        elif isinstance(local_type, UnionType):
+                            assert len(local_type.items) == 2
+                            return split_py_name(
+                                local_type.items[0].type.fullname) + (
+                                    o.callee.expr.name, )
+
+                        else:
+                            assert not isinstance(local_type, CallableType)
+                            # primitive type or string. don't care.
+                            return None
+
+                    else:
+                        # context or exception handler. probably safe to ignore.
+                        return None
+
+            else:
+                t = self.types.get(o.callee.expr)
+                if isinstance(t, Instance):
+                    return split_py_name(t.type.fullname) + (o.callee.name, )
+
+                elif isinstance(t, UnionType):
+                    assert len(t.items) == 2
+                    return split_py_name(
+                        t.items[0].type.fullname) + (o.callee.name, )
+
+                elif o.callee.expr and getattr(o.callee.expr, 'fullname',
+                                               None):
+                    return split_py_name(
+                        o.callee.expr.fullname) + (o.callee.name, )
+
+                else:
+                    # constructors of things that we don't care about.
+                    return None
+
+        # Don't currently get here
+        raise AssertionError()
 
     #
     # COPIED from IRBuilder
@@ -69,10 +149,9 @@ class Build(ExpressionVisitor[T], StatementVisitor[None]):
                     cfg = self.current_cfg()
                     # Most statements have empty visitors because they don't
                     # require any special logic. Create statements for them
-                    # here. Blocks and loops are handled by their visitors.
-                    if (cfg and not isinstance(node, Block) and
-                            not isinstance(node, ForStmt) and
-                            not isinstance(node, WhileStmt)):
+                    # here. Don't create statements from blocks to avoid
+                    # stuttering.
+                    if cfg and not isinstance(node, Block):
                         self.current_statement_id = cfg.AddStatement()
 
                     node.accept(self)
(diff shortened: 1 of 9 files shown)


mycpp/cppgen_pass.py
```diff
@@ -2419,10 +2418,14 @@ def visit_func_def(self, o: 'mypy.nodes.FuncDef') -> T:
...
                 #log('arg_names %s', arg_names)
                 #log('local_vars %s', self.local_vars[o])
-                self.prepend_to_block = [
-                    (lval_name, c_type, lval_name in arg_names)
-                    for (lval_name, c_type) in self.local_vars[o]
-                ]
+                self.prepend_to_block = []
+                for (lval_name, lval_type) in self.local_vars[o]:
+                    c_type = lval_type
+                    if not isinstance(lval_type, str):
```
andychu, line 2424: hm can we remove this and always use a Type instance rather than a string?

mycpp/control_flow_pass.py
```diff
@@ -24,19 +25,98 @@ class UnsupportedException(Exception):
...
+        self.imported_names.add('help_meta')
 
     def current_cfg(self):
         if not self.current_func_node:
             return None
 
-        return self.cfgs.get(split_py_name(self.current_func_node.fullname))
+        return self.cfgs[split_py_name(self.current_func_node.fullname)]
+
+    def resolve_callee(self, o: CallExpr) -> Optional[util.SymbolPath]:
```
andychu, line 50: docstring for this method?

mycpp/control_flow_pass.py
```diff
@@ -24,19 +25,98 @@ class UnsupportedException(Exception):
...
+        self.current_class_name = None
         self.current_func_node = None
         self.loop_stack = []
+        self.virtual = virtual
+        self.local_vars = local_vars
+        self.imported_names = set()  # MemberExpr -> module::Foo() or self->foo
+
+        # HACK for conditional import inside mylib.PYTHON
+        # in core/shell.py
+        self.imported_names.add('help_meta')
```
andychu, line 42: We shouldn't duplicate this imported_names structure here, and also in cppgen_pass.py

Because it is heuristic, and the two heuristics can get out of sync

Generally we want to compute things in one pass, and then pass something more structured and rigid on to future passes

---

I just did some of that to fix rooting ... in the DECL pass we now compute `ctx_member_vars`, and then we pass that info onto the IMPL pass

(and eventually we are going to get rid of the self.decl and self.forward_decl booleans and factor out all the duplication)

mycpp/control_flow_pass.py
```diff
@@ -24,19 +25,98 @@ class UnsupportedException(Exception):
...
+                        else:
+                            assert not isinstance(local_type, CallableType)
+                            # primitive type or string. don't care.
+                            return None
+
+                    else:
+                        # context or exception handler. probably safe to ignore.
+                        return None
+
+            else:
```
andychu, line 99: this also seems like it may duplicate some logic in cppgen_pass

I think we may want to pass the result of resolve_callee onto a future stages, in addition to using it in datalog

mycpp/cppgen_pass.py
melvinw: I wanted to do this, but the casting case makes it difficult: https://github.com/oilshell/oil/blob/master/mycpp/cppgen_pass.py#L1482-L1499). We would need some way of knowing whether to call `GetCType()` or `_GetCTypeForCast()` when iterating over the list of locals. Any way of communicating that felt messier than falling back on a string

mycpp/control_flow_pass.py
melvinw: Ah sure. I thought this logic was simple enough and unlikely to change, but it's better to be safe. I'll pass this in.

mycpp/control_flow_pass.py
melvinw: I guess the if conditions overlap with `GetCType()` somewhat, but I think a shared function would be overkill. Or do you mean specifically the `UnionType` thing?

I can definitely stash these in a map so future passes can grab it easily, though.

mycpp/control_flow_pass.py
```diff
@@ -236,3 +351,73 @@ def visit_try_stmt(self, o: 'mypy.nodes.TryStmt') -> T:
...
+
+    # Expressions
+
+    def visit_member_expr(self, o: 'mypy.nodes.MemberExpr') -> T:
+        self.accept(o.expr)
+
+    def visit_yield_expr(self, o: 'mypy.nodes.YieldExpr') -> T:
+        self.accept(o.expr)
+
+    def visit_op_expr(self, o: 'mypy.nodes.OpExpr') -> T:
```
andychu, line 366: I think a lot of these methods can be left out, since the default in

~/wedge/oils-for-unix.org/pkg/mypy/0.780/mypy/traverser.py

does the same thing

I guess that is sorta nice about the visitor compared to switch

Although I think the visitor is worse in other cases

mycpp/cppgen_pass.py
andychu: So we have to distinguish the cases

    x = 42
    x = cast(SomeType, y)

How about  adding a boolean field to local_var_list that tells if it's casted or not?

mycpp/control_flow_pass.py
```diff
@@ -24,19 +25,111 @@ class UnsupportedException(Exception):
...
+        `moduleA` would is named `moduleA.foo`. Calls to free-functions defined
+        in imported modules are named the same way.
+        """
+
+        if isinstance(o.callee, NameExpr):
+            return split_py_name(o.callee.fullname)
+
+        elif isinstance(o.callee, MemberExpr):
+            if isinstance(o.callee.expr, NameExpr):
+                is_module = (isinstance(o.callee.expr, NameExpr) and
```
andychu, line 70: I think the use of `self.imported_names` should be eliminated -- that is the main source of duplication I see

We are trying to distinguish the cases

    a.b  # attribute of class/struct  a->b
    a.b()  # module and function a::b()

And the heuristic is to look at if a is an imported name

I think we should figure that out ONCE, and then change the representation for later  passes

e.g. we could introduce a new node type even ...  

Or maybe wrap it it in a new type

At the very least it could save this calculation in a dict for later


----

Logically I think the disambiguation belongs in an EARLY pass.   Though maybe the DECL pass can be the place to stick it.

We really should create something like `mycpp.asdl` that defines the language.

So we know that

    a.b() is different than a.b
    x = 42 is different than x = cast(T, 42)

in terms of what we need to compile

If the early stages do the disambiguation, then the later stages don't have to worry about it, and they have a nicer tree  to work with

---

Logically it does seem like we need this to make a correct control flow graph

Right now the disambiguation part is mixed up with the control flow part.  It was always mixed up but I think this is the right time to fix it -- when we have some significant new functionality that hinges on it

mycpp/control_flow_pass.py
melvinw: Oh, cool. That's convenient. I'll remove the ones that don't have any extra logic.

mycpp/cppgen_pass.py
melvinw: Yeah, I had something like this in my first draft of this. It felt kinda clunky. I guess I don't feel too strongly about this, though. I'll revert to that.

andychu:
Thinking about this a bit more, I think ideally we want a structure like

    MyPy AST for Typed Python -->
    mycpp AST resembling C++ ->
    forward decls -->   (this may turn into proper modules eventually, instead of 1 huge translation unit)
    DECL / prototypes ->
    control flow -->    # I suppose this goes here because it only affects IMPL?   Could be up for grabs
    IMPL

andychu:
And we would want a separate `mycpp.asdl` for ALL the passes

But doing that all at once is too much.

----

So  as an intermediate step, we could create a new first phase -- maybe the **heuristic mycpp disambiguation phase** that fills in what's missing in Python


- `obj->field` vs. `module::Func()` or `module.Class()` -- using the imported name heuristic
- member vars -- these are not listed explicitly in Python -- there is a heuristic
- virtual -- this does not appear in the source
- casting -- there are some special considerations here -- we are using the `UP_foo` convention
- I just added a `ctx_Foo` convention for rooting, since context managers only go on the stack

So instead of creating a whole new tree right now, this first phase could just output a bunch of dicts keyed by nodes

e.g. for every MemberExpr, we can create a dict in the first pass that says if it's `obj->field` or `module::Func`

And then we can use that in all passes.

Likewise for the `local_var_list` -- that is used for "hoisting" declarations to the top of every function, before we see them.

---

And then gradually we can move `virtual` there, and `ctx_member_vars`, etc.

We could even call this the "HeuristicPass" or something

So it would be something like


```
member_expr_lookup = ...
local_vars = ...
p1 = heuristic_pass.Build(member_expr_lookup, local_vars, ...)
```

So right now it would output a bunch of dicts

But later it could construct a real IR like `mycpp.asdl`

andychu:
So basically I think this is a good way to gradually move stuff out of `cppgen_pass`

The problem is we want to use `self.imported_names` to check `MemberExpr` in both control flow and in IMPL

But we don't want to duplicate that logic. Instead we can move it to the EARLY heuristic pass, and use that info in as many subsequent stages as we want

So I think it should be a relatively flexible way of organizing the code, and lead to many quick refactorings

andychu:
Hm I tried to delete some of the "pass through" methods from `const_pass.py` and it didn't work

I think the problem is that we are not using that LATER API `traverser.py` 

I probably copied from some earlier API, ugh

So yeah maybe we don't worry about it now, although this causes a bunch of boilerplate I would like to avoid

Maybe the heuristic_pass can be the same thing as the const_pass ... logically that doesn't seem too bad -- all we are doing is building a dict const_lookup

And then we can add a couple other dicts there too

melvinw:
closing this while we figure out the IR stuff

melvinw:
no reason we can't just tack the IR stuff onto this change. reopening.

mycpp/pass_state.py
```diff
@@ -6,13 +6,42 @@
...
 _ = log
 
 
+class ModuleMember(object):
+    def __init__(self, module_path: SymbolPath, member: str) -> None:
+        self.module_path = module_path
+        self.member = member
+
+
+class StaticObjectMember(object):
```
andychu, line 25: Could use a comment for this ... I forgot that we also have AsdlType::CreateNull() -- I think that is the only instance of this!

Usually `::` is a C++ namespace generated for a Python module


----

Also I wonder if we should turn on type checking of mycpp in the CI?  Since we have more type annotations now

andychu (commented):
This looks great, thank you!

Nitpick: I feel like "ast_pass" is a bit of generic name ... aren't they all AST passes sorta?


Otherwise, I think it makes sense to merge this as is pretty much?   Does it pass CI?

melvinw:
> Nitpick: I feel like "ast_pass" is a bit of generic name ... aren't they all AST passes sorta?

Ah lol good point. I'll rename this to `ir_pass`.

andychu:
OK great, thanks for doing this!

I realize it was a lot of yak shaving, but I think this is the right thing for the future, and also the right time to do it, because everything works now :)

melvinw:
No sweat!

PR #2001, opened 2024-06-21
add .html to header link toc-ysh.md

dezren39:
i'm guessing this should be .html? but also these links dont work from markdown :( i'm only guessing this works


andychu:
Oh yes, thanks for fixing this!

To test it you would have to install our tools including CommonMark:

And run

`build/doc.sh all-ref`

And then look in `_release/VERSION/doc/ref`

But this looks correct so I merged it!

PR #2008, opened 2024-06-25
feat: noclobber shell option

momiji:


core/process.py
```diff
@@ -183,6 +186,7 @@ def __init__(
             mem,  #type: state.Mem
             tracer,  # type: Optional[dev.Tracer]
             waiter,  # type: Optional[Waiter]
+            mutable_opts,  #type: MutableOpts
```
andychu, line 189: If we just reading this bit, please use `self.exec_opts.noclobber()`

There is some metaprogramming that allows that simpler interface

self.mutable_opts is for mutating options

core/process.py
```diff
@@ -399,10 +408,21 @@ def _ApplyRedirect(self, r):
...
+                        self.errfmt.PrintMessageEx(
+                            "I/O redirect error: can't overwrite existing file %r: %s"
+                            % (arg.filename, pyutil.strerror(e)),
+                            self.mutable_opts.Get(option_i.verbose_errexit),
+                            r.op_loc)
+                    else:
+                        self.errfmt.PrintMessageEx(
+                            "I/O redirect error: can't open file %r: %s" %
+                            (arg.filename, pyutil.strerror(e)),
+                            self.mutable_opts.Get(option_i.verbose_errexit),
```
andychu, line 421: Hm why not just use PrintMessage() ?  It already doesn't print the code?

I agree that there were some duplicate messages from redirects.  It's a little hard to sort out, because there are multiple cases.    Sometimes it prints too much, sometimes too little, etc.


   `verbose_errexit`  isn't really the right option here anyway -- that is for `set -e` failures

core/process.py
```diff
@@ -399,10 +408,21 @@ def _ApplyRedirect(self, r):
...
+                            r.op_loc)
+                    raise IOError(
```
andychu, line 423: Hm I don't think this is the right way to do it, we can maybe tackle this separately from noclobber

I agree the messages can be tweaked, but we should do it in a more systematic way


I use `test/runtime-errors.sh` to eyeball the errors

spec/sh-options.test.sh
```diff
@@ -318,16 +330,8 @@ echo $?
 set +x
 echo $SHELLOPTS | grep -q xtrace
 echo $?
-## STDOUT:
-1
-0
-1
```
andychu, line 324: Unfortunately it looks I made a change while you were moving repos, and this file got "clobbered"

Please revert this change -- hopefully there is an easy way to do it

andychu (commented):
This is great, thank you!   

I am glad noclobber works, and I agree the error messages can be tweaked

I'd suggest landing  just the noclobber change first

---

Or if you want to put them both together, write some cases in `test/runtime-errors.sh` -- 

e.g.


    $ test/runtime-errors.sh test-control_flow

Then we can look at the output,  to make sure we don't silence other errors

spec/sh-options.test.sh
momiji: ok done

core/process.py
momiji: That's right, this is completly wrong...

core/process.py
momiji: the thing is there is error code clobber issue, this is why printing message and raisin a false error seemed a good idea.
otherwise, we'll need something other than IOError/OSError, but I wasn't able to achieve this as there are impacts with c++ I believe.
and as this is the only place in code that did double print message, I found this dirty trick.
but we want to keep caller to catch oserror coming from posix.close, write and so on.

core/process.py
momiji: replaced with exec_opts. I tried to use bool instead and call with exec_opts.noclobber() but it return always False, so I guess these are lately set.

momiji:
BTW, I have absolutely no idea where this vigoroustelevision comes from...

momiji:
do you think this can be merged soon, either on master of the soil-staging?
BTW, what does "soil-staging" mean? is it just a copy of master or something else?

osh/cmd_eval.py
```diff
@@ -1481,12 +1481,14 @@ def _DoRedirect(self, node, cmd_st):
...
-                self.errfmt.PrintMessage(
-                    'I/O error applying redirect: %s' %
-                    pyutil.strerror(io_errors[0]),
-                    self.mem.GetFallbackLocation())
+                # if errno is 0, error has already printed
+                if io_errors[0].errno:
+                    # core/process.py prints cryptic errors, so we repeat them
+                    # here.  e.g. Bad File Descriptor
+                    self.errfmt.PrintMessage(
+                        'I/O redirect error: %s' %
```
andychu, line 1489: I think the old string was more descriptive: "I/O error applying redirect"

The new one doesn't seem grammatically correct to me

core/process.py
andychu: This IOError(0) trick is too subtle

As mentioned, we need to have some demos in `test/runtime-errors.sh` to tickle all the error cases that are affected

And I'd like to see examples of before and after

But let's do that in another change

andychu (commented):
The core logic looks good - glad the tests pass.  That part can be merged

I'd suggest reverting the error message changes, or moving them to another PR

This is another issue we can tackle separately

andychu:
Hi @momiji , wondering if you saw my comments?

What do you think of merging the minimal thing, and then tackling the error messages separately?  I agree something needs to be done there, but I think we haven't figured it out yet

andychu:
I am looking at this again, I ran the spec tests and the output looks like

```
osh stderr:
rm: cannot remove '/home/andy/git/oilshell/oil/_tmp/spec-tmp/sh-options.test.sh.3956901/27-osh/no-clobber': No such file or directory
[ stdin ]:5: I/O redirect error: can't overwrite existing file '/home/andy/git/oilshell/oil/_tmp/spec-tmp/sh-options.test.sh.3956901/27-osh/no-clobber': File exists
```

andychu:
Hm, this is the error without the specal `0` case

There is some redundancy, but I have seen this elsewhere

Also I think it is better to quote the code.  Somehow the code quotation was omitted in the original



```
osh stderr:
rm: cannot remove '/home/andy/git/oilshell/oil/_tmp/spec-tmp/sh-options.test.sh.3959868/27-osh/no-clobber': No such file or directory
  echo foo > $TMP/no-clobber
           ^
[ stdin ]:5: Can't open '/home/andy/git/oilshell/oil/_tmp/spec-tmp/sh-options.test.sh.3959868/27-osh/no-clobber': File exists
[ stdin ]:5: I/O error applying redirect: File exists

```

andychu:
OK I merged this, with minor changes

We can track the error message issue separately - it isn't specific to noclobber - https://github.com/oils-for-unix/oils/issues/2035

Issue #1654, opened 2023-06-16
ERR trap executed when errexit is ignored


heftig:
The bash manpage mentions:

> The `ERR` trap is not executed if the failed command is part of the command list immediately following a `while` or `until` keyword, part of the test in an `if` statement, part of a command executed in a `&&` or `||` list except the command following the final `&&` or `||`, any command in a pipeline but the last, or if the command's return value is being inverted using `!`. These are the same conditions obeyed by the `errexit` (`-e`) option.

Osh 0.16.0 apparently handles none of these conditions. In addition, some cases cause the trap to be run twice:

```sh
osh-0.16.0$ trap 'echo 1' ERR
osh-0.16.0$ echo 2 && false
2
1
1
osh-0.16.0$ echo >&2 2 | false
1
2
1
```

bar-g:
Hi, concerning bash neglecting errors, maybe see:
https://www.oilshell.org/release/0.16.0/doc/error-handling.html

There are more error exit options in osh than in bash, also to print more info.
Some are automatically enabled in ysh, you could run your simple examples there.

bar-g:
Ok, sorry, I see, from `shopt -p` even if verbose_errexit is enabled there is are no messages.

Yep, seems the ERR trap is not getting disabled together with `set +o errexit`.

heftig:
I mean that the `ERR` trap shouldn't trigger for e.g. `if false; then`. That's just unhelpful.

bar-g:
Yes, I see your point, now. Let's see what Andy responds.

Originally I had mistakenly assumed you had switched errorexit on in osh, and your problem were that ERR trap is executed twice, but that seems once for the single command, and once for the pipeline. Which is basicly fine, I think.

Another question is if it might make sense for oils-for-unix to internally prevent repeated execution of the same traps? And wether to always execute EXIT after other signals like (only?) bash. Maybe there is a good combination that obsoletes workarounds, and makes trap usage actually practical, by default.

andychu:
Thanks for the report and for reproducing/confirming the bug -- that's especially helpful

The ERR feature is based on bash, and was implemented very recently, so it's probably just an oversight

andychu:
This is now fixed, and will be out with the next release

It passes all these tests - http://op.oils.pub/github-jobs/7329/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap-err.html

Feel free to test it out

@momiji has a pending PR for `set -o errtrace`

[...]

PR #2010, opened 2024-06-26
feat; errtrace shell option

momiji:
Additional fixes for;:
- shopt -o: show all options, like bash
- refactoring: as it is not possible to have 2 optional args, I added missing args where needed
- *2 on spec-cpp counts as there are 2 tests so comparison was wrong

Also note dash/ash does not time builtin, some tests may be removed for them.

### DIFF
--- a/doc/ref/chap-option.md
+++ b/doc/ref/chap-option.md
@@ -101,6 +101,7 @@ These options are from bash.
 ## Other Option
 
     noclobber   # Redirects don't overwrite files
+    errtrace    # Inherits ERR trap in shell functions
 
 ## Compat
 
--- a/opy/_regtest/src/core/state.py
+++ b/opy/_regtest/src/core/state.py
@@ -78,6 +78,7 @@ SET_OPTIONS = [
     ('x', 'xtrace'),
     ('f', 'noglob'),
     ('C', 'noclobber'),
+    ('E', 'errtrace'),
     ('h', 'hashall'),
     (None, 'pipefail'),
 
@@ -112,6 +113,7 @@ class ExecOpts(object):
     self.noglob = False  # -f
     self.noexec = False  # -n
     self.noclobber = False  # -C
+    self.errtrace = False  # -E
     # We don't do anything with this yet.  But Aboriginal calls 'set +h'.
     self.hashall = True  # -h is true by default.
 
--- a/spec/builtin-trap-bash.test.sh
+++ b/spec/builtin-trap-bash.test.sh
@@ -555,3 +555,37 @@ g
 return-helper.sh
 profile [x y]
 ## END
+
+
+#### trap ERR with errtrace (-E)
+# ERR trap is not inherited by shell functions unless the -o errtrace option is set
+set -E
+trap "echo ERR" ERR
+echo 0
+false
+echo 1
+cat <( date X ; echo 2 )
+echo $( date X ; echo 3 )
+{ date X ; echo 4 ; }
+{ date X ; echo 5 ; } & wait
+( date X ; echo 6 ; )
+true
+## status: 0
+## stdout-json: "0\nERR\n1\nERR\n2\nERR 3\nERR\n4\nERR\n5\nERR\n6\n"
+
+
+#### trap ERR without errtrace (+E)
+# ERR trap is not inherited by shell functions unless the -o errtrace option is set
+set +E
+trap "echo ERR" ERR
+echo 0
+false
+echo 1
+cat <( date X ; echo 2 )
+echo $( date X ; echo 3 )
+{ date X ; echo 4 ; }
+{ date X ; echo 5 ; } & wait
+( date X ; echo 6 ; )
+true
+## status: 0
+## stdout-json: "0\nERR\n1\n2\n3\nERR\n4\n5\n6\n"
--- a/builtin/trap_osh.py
+++ b/builtin/trap_osh.py
@@ -47,13 +47,16 @@ class TrapState(object):
         self.hooks = {}  # type: Dict[str, command_t]
         self.traps = {}  # type: Dict[int, command_t]
 
-    def ClearForSubProgram(self):
-        # type: () -> None
+    def ClearForSubProgram(self, inherit_errtrace):
+        # type: (bool) -> None
         """SubProgramThunk uses this because traps aren't inherited."""
 
         # bash clears DEBUG hook in subshell, command sub, etc.  See
         # spec/builtin-trap-bash.
+        handler = self.hooks.get('ERR', None)
         self.hooks.clear()
+        if handler and inherit_errtrace: self.hooks['ERR'] = handler
+
         self.traps.clear()
 
     def GetHook(self, hook_name):
--- a/core/executor.py
+++ b/core/executor.py
@@ -158,8 +158,8 @@ class ShellExecutor(vm._Executor):
         # type: () -> None
         assert self.cmd_ev is not None
 
-    def _MakeProcess(self, node, inherit_errexit=True):
-        # type: (command_t, bool) -> process.Process
+    def _MakeProcess(self, node, inherit_errexit, inherit_errtrace):
+        # type: (command_t, bool, bool) -> process.Process
         """Assume we will run the node in another process.
 
         Return a process.
@@ -189,7 +189,8 @@ class ShellExecutor(vm._Executor):
                                         node,
                                         self.trap_state,
                                         self.multi_trace,
-                                        inherit_errexit=inherit_errexit)
+                                        inherit_errexit,
+                                        inherit_errtrace)
         p = process.Process(thunk, self.job_control, self.job_list,
                             self.tracer)
         return p
@@ -396,7 +397,7 @@ class ShellExecutor(vm._Executor):
             pi = process.Pipeline(self.exec_opts.sigpipe_status_ok(),
                                   self.job_control, self.job_list, self.tracer)
             for child in node.children:
-                p = self._MakeProcess(child)
+                p = self._MakeProcess(child, True, self.exec_opts.errtrace())
                 p.Init_ParentPipeline(pi)
                 pi.Add(p)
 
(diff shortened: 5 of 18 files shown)


spec/builtin-trap-bash.test.sh
```diff
@@ -555,3 +555,37 @@ g
...
+false
+echo 1
+cat <( date X ; echo 2 )
+echo $( date X ; echo 3 )
+{ date X ; echo 4 ; }
+{ date X ; echo 5 ; } & wait
+( date X ; echo 6 ; )
+true
+## status: 0
+## stdout-json: "0\nERR\n1\nERR\n2\nERR 3\nERR\n4\nERR\n5\nERR\n6\n"
```
andychu, line 574: We have a more modern style of

```
## STDOUT:
0
ERR
## END
```

for multi-line assertions now

spec/builtin-trap-bash.test.sh
```diff
@@ -555,3 +555,37 @@ g
 return-helper.sh
 profile [x y]
 ## END
+
+
+#### trap ERR with errtrace (-E)
+# ERR trap is not inherited by shell functions unless the -o errtrace option is set
```
andychu, line 561: It seems like it affects both shell functions and subshells ??

opy/_regtest/src/core/state.py
```diff
@@ -78,6 +78,7 @@ def Disable(self):
     ('x', 'xtrace'),
     ('f', 'noglob'),
     ('C', 'noclobber'),
+    ('E', 'errtrace'),
```
andychu, line 81: Oh anything in `opy/_regtest` does not need to be changed

This is a test file that's not active

Please revert this -- thanks for searching

doc/ref/chap-option.md
```diff
@@ -101,6 +101,7 @@ These options are from bash.
 ## Other Option
 
     noclobber   # Redirects don't overwrite files
+    errtrace    # Inherits ERR trap in shell functions
```
andychu, line 104: same comment - describe how it changes the behavior for subprocesses?


Also mention that this is `-E`

spec/builtin-trap-bash.test.sh
```diff
@@ -555,3 +555,37 @@ g
...
+false
+echo 1
+cat <( date X ; echo 2 )
+echo $( date X ; echo 3 )
+{ date X ; echo 4 ; }
+{ date X ; echo 5 ; } & wait
+( date X ; echo 6 ; )
+true
+## status: 0
+## stdout-json: "0\nERR\n1\n2\n3\nERR\n4\n5\n6\n"
```
andychu, line 591: I'm a bit confused about this option

I think it might be useful to have a simpler test ... one I came up with:

```
$ bash -c 'trap "echo err \$LINENO" ERR;  set +E; f() { false; }; f '
err 1
err
andy@hoover:~/git/oilshell/oil$ bash -c 'trap "echo err \$LINENO" ERR;  set -E; f() { false; }; f '
err 1
err 1
err

```

andychu (commented):
Thanks for working on this!   

I see diffs from noclobber change for some reason.  I guess it should be based off master

---

I wonder at a high level  what you use this for?


I'm a bit worried about the correctness ... there are several known bugs around the ERR trap

I will bump those threads on Zulip

andychu:
Also about the 2x, are you running with 2 OSH ?   Not sure where that comes from

What does the output of your spec tests look like?

andychu:
Oh also I think `shopt -o` should be a separate change

bash has both `shopt -o` and `shopt -p` which is very confusing ... they print a different set of options

andychu:
I think the issue is that `shopt -o` is printing the wrong set of options

This sums it up ... yeah let's put this in another change, because I don't think the change here is right

```
andy@hoover:~/git/oilshell/oil$ bin/osh -c 'shopt -p | wc -l; shopt -o | wc -l; shopt -p -o |wc -l; set -o | wc -l' 
53
53
12
12
andy@hoover:~/git/oilshell/oil$ bash -c 'shopt -p | wc -l; shopt -o | wc -l; shopt -p -o |wc -l; set -o | wc -l' 
57
27
27
27

```

momiji:
OK I'll split and correct what you've mentioned.

2x is related to the cpec-spp.sh test that compares osh python and osh native.
Because of that counts are doubled, because an expected failure will happen twice instead of once, so we need to compare numbers with x2.
And as written in the code, this use case onmy happend when the native bin option is used.

andychu:
I realized there a bunch of bugs in trap ERR, not strictly related to this change

I created a new file `spec/builtin-trap-err`

http://op.oils.pub/github-jobs/7313/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap-err.html


- also we can test against busybox ash, an dmksh

----

Main finding

- OSH already "inherits" traps in FUNCTIONS
- but NOT in subshells

So this change is good, but there are a bunch of edge cases to clean up

I guess the question is if we want to disable it in functions by default -- mksh does NOT do this, but busybox ash does.  I think we should agree with bash and busybox ash, so we do it

---

Let me work on fixing some of those bugs

andychu:
You can see my commits here, I don't think they will collide with your spec tests

https://github.com/oilshell/oil/commits/master/


But it may be good to move your tests to `builtin-trap-err`, so we can test against ash and mksh

I try not to "just implement  bash bugs", but really test if shells agree

andychu:
I think the issue is that `shopt -o` is printing the wrong set of options

This sums it up ... yeah let's put this in another change, because I don't think the change here is right

```
andy@hoover:~/git/oilshell/oil$ bin/osh -c 'shopt -p | wc -l; shopt -o | wc -l; shopt -p -o |wc -l; set -o | wc -l' 
53
53
12
12
andy@hoover:~/git/oilshell/oil$ bash -c 'shopt -p | wc -l; shopt -o | wc -l; shopt -p -o |wc -l; set -o | wc -l' 
57
27
27
27

```

momiji:
I've been working on this all day and it's getting more and more complex.
I added many more tests (not yet commited in a new branch to replace this one), and discover new use case that work and some that don't work, and I think I need some time now to better understand the code now.
For example, I've just understood that the _NoForkLast is removing fo_fork of the last command, and because of this there is no wa to retrieve the error status to call trap err.
However, I remember something about it in the documentation, let's continue on experimentations...

momiji:
BTW, the new branch will be called errtrace-shell-option-2

andychu:
Hm we might want to chat about it

I made 2 significant changes to `trap ERR` here ... one of them is NOT related strictly to your change (don't run trap ERR in errexit conditions) and the other one is (don't run it in shell functions by default)

This was an old bug #1654 

https://github.com/oilshell/oil/commit/10ef2c48fda5537f80cff2fc402d7f9e41ad7de7


Basically I did this because to refresh my memory / learn about trap ERR

TBH I didn't understand your PR and this was my way of trying to understand it !   

The key bit is shell functions and child processes are two **different** things that `set -E ` affects

momiji:
I must redo the other branch as the rebase just broke everything, let's keep in touch when I'm ready with all this stuff and have identified all use cases / tests.

andychu:
OK I'm down to 4 failures in OSH, and hopefully I can fix 3 of them today

As mentioned 1 was related to `errtrace`, the others were not -- I just didn't want to build on shaky foundations

https://op.oils.pub/github-jobs/7322/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap-err.html

----

The core of this change is exactly right -- the SubProgramThunk change, so thank you for finding that!

So I will ping you when I stop changing the code ... (although I did separate it into a separate spec file so at least that part wouldn't conflict!)

Thanks for looking at this!

andychu:
OK I fixed #1654, so I won't be touching these files anymore

---

Conveniently it passes all these tests, and the only 2 failures left are `set -o errtrace`!

 http://op.oils.pub/github-jobs/7329/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap-err.html

So I think your change can fix subprocess test case - thanks for finding this!

And then look for `InsideFunction()` in `MaybeRunErrTrap()` -- this also needs an `errtrace` conditional


(feel free to look at recent commits to see what I did, though I think the new test cases are pretty precise)

---

I may look at `shopt -o` separately , I think that is slightly different

andychu:
BTW I fixed `shopt -o` here

https://github.com/oilshell/oil/commit/7ed0ab60e7d91db2435f1782c6bbc2efcbb72165

In bash, `shopt -o` doesn't show all options, it shows the ones managed by `set -o`

---

We *should* have something else to print ALL options, like `shopt --tsv` !    We can use help on that

It could have 3 columns I suppose?   (set or shopt, option name, boolean)   Could discuss this design on Zulip


---

In any case I would very much appreciate  the `errtrace` change, now that everything else is in order

momiji:
Sorry I've been very busy these last 2 days, I only started reworking from master this morning and u fortunately the day gor busy again ... I'll try to create a -3 branch correctly with all master changes. Otherwise most probably tomorrow evening. It's near 8pm here.

andychu:
No problem, I appreciate the time and help!

Historically many people are interested in helping on Oils, but they only care about the new YSH and don't actually know shell

The fact that you know shell (and use it!) is very important  IMO :)

----

I want to write some blog posts about how it works, but some concrete tips:

(1) Oils is a big "tree walk interpreter" like all shells.  I call it "Crafting Interpreters" + 20 Unix syscalls.

i.e. shell is very elaborate, but conceptually we try to do things "by the book".


(2) you can print the syntax tree with `-n`

It's just a big tree of "structs" and arrays essentially, representing the source code

```
$ bin/osh -n -c 'echo hi | wc'
(command.Pipeline
  children: [
    (command.Simple
      blame_tok: <Id.Lit_Chars echo>
      more_env: []
      words: [{<Id.Lit_Chars echo>} {<Id.Lit_Chars hi>}]
      do_fork: T
    )
...
```

momiji:
The PrintMessageEx was a try to mimic the difference between error display.  

Sometime we only have the error message, and other time there is a more visual error display the command and the location of the error.

I believe I haven't yet fully understand the conditions between them, I just thought it was `verbose_errexit`, but I may be wrong.

I found the idea good, as modern languages all display errors this way, it's just that I don"t wan't bash to behave like this for compatibility reason (again).

I'll try to give it a look also, and see if there are any tests later.

andychu:
There is a new branch for this, so I will close this one

Feel free to reopen if there's anything else here

PR #2011, opened 2024-06-26
Fixes spec cpp counts

momiji:
Without the fix, the following fails with:

```
ysh-builtin-shopt: FATAL: Got 2 failures (1 oils failures), but 1 are allowed
```

With the fix, it is fine, as both osh python and osh native works the same.

```
$ test/spec-cpp.sh run-file ysh-builtin-shopt
ysh-builtin-shopt.test.sh
  0       3     pass    pass    shopt supports long flags
  1      17     pass    pass    shopt supports 'set' options
  2      30     pass    pass    shopt --unset errexit { }
  3      51     pass    pass    shopt -p works correctly inside block
  4      65     pass    pass    shopt --set GROUP { }
  5     111     pass    pass    shopt and block status
  6     124     pass    pass    shopt usage error
  7     137     pass    pass    shopt -p
  8     162     FAIL    FAIL    TODO: all options as a table

                osh     osh-cpp
        pass    8       8
        FAIL    1       1
        total   9       9
ysh-builtin-shopt: note: Got 2 allowed oils failures (exit with code 0)
```

### DIFF
--- a/test/sh_spec.py
+++ b/test/sh_spec.py
@@ -79,7 +79,8 @@ OTHER_OSH = ('osh_ALT',)
 YSH_CPYTHON = ('ysh', 'ysh-dbg')
 OTHER_YSH = ('oil_ALT',)
 
-# Forw now, only count the Oils CPython failures
+# For now, only count the Oils CPython failures.  TODO: the spec-cpp job should
+# assert the osh-cpp and ysh-cpp deltas.
 OTHER_OILS = OTHER_OSH + OTHER_YSH + ('osh-cpp', 'ysh-cpp')
 
 
--- a/test/sh_spec.py
+++ b/test/sh_spec.py
@@ -1395,9 +1395,10 @@ def main(argv):
   # spec/smoke.test.sh -> smoke
   test_name = os.path.basename(test_file).split('.')[0]
 
-  allowed = opts.oils_failures_allowed
+  multiplier = 2 if opts.oils_cpp_bin_dir else 1
+  allowed = opts.oils_failures_allowed * multiplier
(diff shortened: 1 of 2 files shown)


test/sh_spec.py
```diff
@@ -1395,9 +1395,10 @@ def main(argv):
   # spec/smoke.test.sh -> smoke
   test_name = os.path.basename(test_file).split('.')[0]
 
-  allowed = opts.oils_failures_allowed
+  multiplier = 2 if opts.oils_cpp_bin_dir else 1
+  allowed = opts.oils_failures_allowed * multiplier
   all_count = stats.Get('num_failed')
-  oils_count = stats.Get('oils_num_failed')
+  oils_count = stats.Get('oils_num_failed') * multiplier
```
andychu, line 1401: I think we should drop the multipler here, and then say `%d oils python failures`

Because Python and C++ can fail in different ways -- they won't always fail at the same time!

(the alternative is to accurately track osh-cpp and ysh-cpp failures in a separate count, and print both)

andychu (commented):
Great, thanks for doing this!

This points to the fact that we should really assert the spec-cpp number of failures like Python.

We got everything passing earlier this year, so we haven't done that yet ...

test/sh_spec.py
momiji: Even if not perfect this just allows to prevent FATAL while executing. Most of tests, so spend less time to count am ually ;)
I agree there should be a perfect way to do, but we can do it later, iteratively ?

andychu:
OK I changed the accounting a slightly different way

Thanks for noticing this!  That was indeed a bit confusing

PR #2013, opened 2024-06-28
shopt -p and -o compatibility with bash

momiji:
-o changes which settings are displayed
-p display commands instead of a table

In hope this is now the correct behavior.

### DIFF
--- a/builtin/pure_osh.py
+++ b/builtin/pure_osh.py
@@ -169,7 +169,7 @@ class Set(vm._Builtin):
         # 'set -o' shows options.  This is actually used by autoconf-generated
         # scripts!
         if arg.show_options:
-            self.exec_opts.ShowOptions([])
+            self.exec_opts.ShowOptions([], False)
             return 0
 
         # Note: set -o nullglob is not valid.  The 'shopt' builtin is preferred in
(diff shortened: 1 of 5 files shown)


andychu:
Hm are you relying on this behavior anywhere? 

I agree that we are little incompatible with other shells, because we don't print this format

But we DO print  an unambiguous format that you can parse


```
andy@hoover:~/git/oilshell/oil$ bash -c 'set -o' |head
allexport       off
braceexpand     on
emacs           off
errexit         off
errtrace        off
functrace       off
hashall         on
histexpand      off
history         off
ignoreeof       off
andy@hoover:~/git/oilshell/oil$ zsh -c 'set -o' |head
noaliases             off
aliasfuncdef          off
allexport             off
noalwayslastprompt    off
alwaystoend           off
appendcreate          off
noappendhistory       off
autocd                off
autocontinue          off
noautolist            off
andy@hoover:~/git/oilshell/oil$ mksh -c 'set -o' |head
Current option settings
allexport      off                                                                         markdirs       off                                                                         privileged     off
asis           off                                                                         monitor        off                                                                         restricted     off
bgnice         off                                                                         noclobber      off                                                                         sh             off

```

momiji:
Yes I agree, it just to mimic the behavior of other for the table display.
The objective here is to get closer to bash for being a "near bash replacement".

As mentioned, all shells did not print the same, bash using space+tab while the others don't.

It doesn't really matter as most of the time it's just for printing, and in scripts most probably used with `awk` or `grep off$`.
And regarding the idea to display both shopt and set options, maybe we can add a `-a` for all ;)

andychu:
Well I'm suggesting we do something that's better, not just what bash is, e.g. in `shopt --tsv` or something

I don't see any evidence people rely on the bash-like format

I guess I like that we have 1 format now -- and we may have one more

I don't want to end up with 3 formats, since it's a maintenance hazard

Let's defer this until we sort out the other changes

PR #2014, opened 2024-06-28
errtrace shell option, 3rd attempt

momiji:
Let's continue discussion here...
There are still some tests and probably code fix regarding return and exit in err trap.


osh/cmd_eval.py
```diff
@@ -1896,7 +1896,12 @@ def ExecuteAndCatch(self, node, cmd_flags=0):
         """
         if cmd_flags & Optimize:
             node = self._RemoveSubshells(node)
-            self._NoForkLast(node)  # turn the last ones into exec
+            # Using _NoForkLast() hides exit code for ERR trap, temporarily comment it
```
andychu, line 1899: Oh NoForkLast actually breaks  a bunch of other stuff

I think we need `shopt --unset noforklast` or something like that

https://www.oilshell.org/blog/2024/03/release-0.21.0.html#starship-prompt-bug-pipelines-crash-after-using-it


I would leave it off this change

The way to work around it is to add the `;true` -- that's fine

Let's tackle this separately, since it's a crosscutting issue

I definitely welcome help on it though!!

I think this came up in one of @melvinw PR's too.  It is a problematic thing that all shells implement, slightly differently

spec/builtin-trap-bash.test.sh
```diff
@@ -655,3 +655,175 @@ after=1
...
+## STDOUT:
+ERR
+## END
+
+#### trap ERR shadowing with errtrace (-E)
+# ERR trap is not inherited by shell functions, sub processes, sub commands, ... unless the -o errtrace option is set
+set -E
+trap "echo ERR" ERR
+false ; trap "echo ERRx" ERR ; false ; trap "echo ERR" ERR
+cat <( trap "echo ERR2" ERR ; date X )
```
andychu, line 746: I think it's easier to follow if we break some of these cases up

I broke them up in to TWENTY cases in `spec/builtin-trap-err`

core/state.py
```diff
@@ -231,6 +232,31 @@ def __exit__(self, type, value, traceback):
...
+        # type: () -> None
+        pass
+
+    def __exit__(self, type, value, traceback):
+        # type: (Any, Any, Any) -> None
+        if self.errtrace: return
+        # errtrace is disabled, restore err trap (that was removed earlier)
+        # unless it has been set meanwhile
+        if self.err_handler and not self.trap_state.GetHook('ERR'):
+            self.trap_state.AddUserHook('ERR', self.err_handler)
```
andychu, line 257: There are some boolean checks now in `self._MaybeRunErrTrap()` --- can we use that style instead of a context manager?

andychu (commented):
Thanks for working on this - some initial comments

core/state.py
momiji: Not sure, as I found that err trap defined in {} is not restored on }, which make sense. This is why I dedicated a context just for that because of the "unless it has been set meanwhile"... maybe should I said replaced or overwritten?

spec/builtin-trap-bash.test.sh
momiji: Yes we can, it's just it takes much more time... and all 4 tests are near exactly the same, with and without errrtace, the same with shadowing...
This is just cosmetic we can do later once everything is working fine...

osh/cmd_eval.py
momiji: For the moment I was not able to find a way to get the exit code of the last command, unless I remove this option.
Once unset, all issues magically disappear, except one regarding pipeline I need to work on.

But if course, if there is another way to get the exit code and maybe call the err trap, then we should be able to remove it.

momiji: I think a summary of the state of this PR can be interesting, although I understand there are still some open topics you've raised and that are not closed yet.

Analysing the CI tests results, there's a very few issues remaining, most of them being not real, or related to other fixes we can work on.

**test/runtime-errors.sh test-errexit-multiple-processes**

Is due to the errors display, this can be fixed in another specific branch, I have an idea on how to fix it to have only one single error displayed, because I guess the idea is to use verbose_errexit to show only one single error, and not many of them. Currently 3 errors are displayed instead of only 1 :

```
  ls | false | wc -l
       ^~~~~
[ -c flag ]:1: errexit PID 354396: command.Simple failed with status 1
  ls | false | wc -l
  ^~
[ -c flag ]:1: errexit PID 354395: command.Simple failed with status 141
0
  ls | false | wc -l
       ^~~~~
[ -c flag ]:1: errexit PID 354392: command.Pipeline failed with status 1
```

The first because it's the original error (and the first one), the second should be hidden as it is a SIGPIPE, so we might know we're in sigpipe to prevent, or maybe always prevent sigpipe errors, don't know. The third because the result of the pipeline is in error.

The idea is while parsing the AST, to identify which parts are allowed to display errexit errors and which should not, and once the first error is displayed,, just stop displaying anything. This is the idea... Maybe add a field to the mem object, or in mutable_opts?

**test/spec-py.sh ysh-all-serial**

The NoForkLast change impacted the way the xtrace is displaying things, this is just the test to be updated to reflect the change.

**test/spec.sh builtin-trap-bash -r 19 -v**

This one is expected failing, but I think I got the reason why now : the "dbg 3" is printed on stdout, which is eaten by the pipeline of `false | false | false`. To verify this, just do `trap 'echo dbg $LINENO > /dev/stderr' DEBUG` and here comes the 3 `dbg 3`.

I guess bash and other shells are doing tricks to make the debug to go the the stdout of the pipeline and not the sub commands. Probably this another hidden issue we should also fix, as it may have other side effects not yet identified.
A dedicated branch for this one too?

**test/stateful.sh job-control-all -r 10**

This one is the only remaining issue I think there is. I didn't started yet, as I was looking to other subjects today.

momiji:
I think a summary of the state of this PR can be interesting, although I understand there are still some open topics you've raised and that are not closed yet.

Analysing the CI tests results, there's a very few issues remaining, most of them being not real, or related to other fixes we can work on.

**test/runtime-errors.sh test-errexit-multiple-processes**

Is due to the errors display, this can be fixed in another specific branch, I have an idea on how to fix it to have only one single error displayed, because I guess the idea is to use verbose_errexit to show only one single error, and not many of them. Currently 3 errors are displayed instead of only 1 :

```
  ls | false | wc -l
       ^~~~~
[ -c flag ]:1: errexit PID 354396: command.Simple failed with status 1
  ls | false | wc -l
  ^~
[ -c flag ]:1: errexit PID 354395: command.Simple failed with status 141
0
  ls | false | wc -l
       ^~~~~
[ -c flag ]:1: errexit PID 354392: command.Pipeline failed with status 1
```

The first because it's the original error (and the first one), the second should be hidden as it is a SIGPIPE, so we might know we're in sigpipe to prevent, or maybe always prevent sigpipe errors, don't know. The third because the result of the pipeline is in error.

The idea is while parsing the AST, to identify which parts are allowed to display errexit errors and which should not, and once the first error is displayed,, just stop displaying anything. This is the idea... Maybe add a field to the mem object, or in mutable_opts?

**test/spec-py.sh ysh-all-serial**

The NoForkLast change impacted the way the xtrace is displaying things, this is just the test to be updated to reflect the change.

**test/spec.sh builtin-trap-bash -r 19 -v**

This one is expected failing, but I think I got the reason why now : the "dbg 3" is printed on stdout, which is eaten by the pipeline of `false | false | false`. To verify this, just do `trap 'echo dbg $LINENO > /dev/stderr' DEBUG` and here comes the 3 `dbg 3`.

I guess bash and other shells are doing tricks to make the debug to go the the stdout of the pipeline and not the sub commands. Probably this another hidden issue we should also fix, as it may have other side effects not yet identified.
A dedicated branch for this one too?

**test/stateful.sh job-control-all -r 10**

This one is the only remaining issue I think there is. I didn't started yet, as I was looking to other subjects today.

osh/cmd_eval.py
```diff
@@ -1896,7 +1896,13 @@ def ExecuteAndCatch(self, node, cmd_flags=0):
         """
         if cmd_flags & Optimize:
             node = self._RemoveSubshells(node)
-            self._NoForkLast(node)  # turn the last ones into exec
+            # Using _NoForkLast() hides exit code for ERR trap, but only in non-interactive mode!
+            # Example:
+            # - ./bin/osh -c 'trap "echo ERR" ERR ; date X' - does not show ERR
+            # - ./bin/osh -c 'trap "echo ERR" ERR ; date X ; exit' - does show ERR
+            # This is why it is only called interactive mode
+            if self.mutable_opts.Get(option_i.interactive):
```
andychu, line 1904: OK I think this is issue 

https://github.com/oilshell/oil/issues/1853

I think we should do this separately, since it's not strictly related to `trap ERR`

I think it should be another option

- `shopt --unset nolastfork` -- not sure if it should be on or off by default - it depends on performance.  there is a Zulip thread about it I will ping
- it should use `exec_opts` not `mutable_opts`

osh/cmd_eval.py
```diff
@@ -2068,8 +2074,8 @@ def _MaybeRunErrTrap(self):
             return
 
         # bash rule - affected by set -o errtrace
-        if self.mem.InsideFunction():
-            return
+        # if self.mem.InsideFunction():
```
andychu, line 2077: why comment this out?  I'm surprised it didn't make a test fail ...

But it also didn't make `builtin-trap-err` pass

core/executor.py
```diff
@@ -189,7 +189,8 @@ def _MakeProcess(self, node, inherit_errexit=True):
                                         node,
                                         self.trap_state,
                                         self.multi_trace,
-                                        inherit_errexit=inherit_errexit)
+                                        inherit_errexit,
+                                        inherit_errtrace)
```
andychu, line 193: Or is it possible to pass `self.exec_opts.errtrace()` here directly, instead of having all the params?  I think it's always the same value?

core/state.py
andychu: Hm not sure what you mean, though as mentioned I suspect this isn't working right because test case 12 in `builtin-trap-err` still fails

osh/cmd_eval.py
```diff
@@ -811,7 +811,8 @@ def _DoSimple(self, node, cmd_st):
             # shells aren't consistent.
             # self.mem.SetLastArgument('')
 
-        run_flags = executor.DO_FORK if node.do_fork else 0
+        # NOTE: ERR trap will not be called if not forked
+        run_flags = executor.DO_FORK if node.do_fork or self.trap_state.GetHook('ERR') else 0
```
andychu, line 815: Hmm I'm still a bit confused by this one ...  also I'm not sure why test case 10 doesn't pass

andychu (commented):
Hm I think the main thing is that I expected test case 10 and 12 to pass with this change

http://op.oils.pub/github-jobs/7374/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap-err.html

one of them tests shell functions, and the other tests subprocesses/"subprograms"

Let me see why they are failing

andychu:
OK looks like test case 12 is failing because our line numbers are different

```
case: 12

[osh stdout]
Expected 'line=14\nnow with errtrace\nline=4\nline=10\nline=20\nok\n'
Got      'line=10\nnow with errtrace\nline=4\nline=10\nline=10\nok\n'

osh stdout:
line=10
now with errtrace
line=4
line=10
line=10
ok

```

Let me see if I can make the test insensitive to that detail

andychu:
test case 10 is like this

```
Expected 'line=6\nsubshell\nline=12\nasync\nline=14\nok\n'
Got      'line=6\nsubshell\nline=10\nline=12\nasync\nline=14\nok\n'

osh stdout:
line=6
subshell
line=10
line=12
async
line=14
ok

```

I'm not sure if that's OK ... I might have to play with it

spec/builtin-trap.test.sh
```diff
@@ -1,4 +1,4 @@
-## compare_shells: dash bash mksh ash
+## compare_shells: dash bash mksh
```
andychu, line 1: ash should not be deleted here.

andychu:
I split up `spec/builtin-trap-err` cases and merged master ... let's see the results

andychu:
Hm I think we should do something like this -- it's a lot simpler -- just a single if statement

https://github.com/oilshell/oil/pull/2014/commits/9fe9fe1f1264718e5f630fbd54024bcbf6d7175f

But it made one of the new tests fail ... it was printing ERR7 2 times instead of 3 .  So I reverted it

----


I think that test case could be split up -- since it's a bit hard to follow.  

I guess is tickling something that my test case isn't, because mine still passes ... That's in `spec/builtin-trap-err`

andychu:
Another reason to split it up is that we can add busybox ash, and see which ones match, like here:

http://op.oils.pub/github-jobs/7385/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap-err.html

In this case you can only see the result vs. bash, and it's not clear which ones agree

http://op.oils.pub/github-jobs/7385/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap-bash.html

there is significant disagreement among shells, so we try not to "just implement bash bugs"

---

I am not sure but I think the thing that failed when I simplified the code could be a bash quirk, not present in busybox ash?


We should try to use the simple if statement if possible, instead of another `ctx_HideErrTrap`

andychu:
Hm I also think to avoid getting stuck, and to make things easier to review, there are least 2 changes here

- add errtrace option, and  fix the "function" behavior
- fix the "inherit" behavior

or it could be the other way around:

- add errtrace options, and fix the "inherit" behavior
- and  fix the "function" behavior


This are is VERY tricky, so it also helps to have separate commits , in case we need to bisect

---

I think it's good to do "everything at once" to explore -- that is what I often do, to make sure I have an understanding of the problem

But then I try to break up any "risky" changes separately, which is often useful for testing regressions, bisecting, etc.

---

 `_NoLastFork` is also somewhat of a crosscutting issue -- I cc'd you on this thraed

https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/shopt.20--set.20noforklast


It affects not just the ERR trap, but also the RETURN trap.  There are **new** bash bugs related to fork optimizations.

---

So  that is one reason we don't "just implement bash" -- because bash has bugs and it changes substantially from version to version

We try to make an "executable spec", and in this case we can base it off bash AND busybox ash

andychu:
So basically I expect that after this change we will have to look at any other issues caused by `noforklast`

- running traps
- `shopt -s verbose_errexit` -- sometimes we **lose** error messages because of the fork optimizations
- I ran into a problem with Oils instrumentation - we are trying to track all commands run, and maybe get GC stats per process, etc.
- We might want to implement `set +m` to disable job control (a POSIX feature) which came up the thread on Thorsten's blog  post

there were some TODOs there

https://oilshell.zulipchat.com/#narrow/stream/266575-blog-ideas/topic/Hidden.20Optimizations.20in.20Unix.20Shell.2C.20Job.20Control.2C.20zed.2Fzsh.20bug

andychu:
Merged this as

https://github.com/oils-for-unix/oils/commit/a2f0ae0f1ac0acff89f8ef2016dd9cfea0681765

and

https://github.com/oils-for-unix/oils/commit/d8572e6616641955b7adb8c0cb44f28485baec90


Any more manual testing and feedback is welcome - you can download a tarball here - http://travis-ci.oilshell.org/github-jobs/7588/

Thanks for diving deep into the code!

PR #2018, opened 2024-07-11
Add spec tests for eval with argv/vars bindings

PossiblyAShrub:


spec/ysh-builtin-eval.test.sh
```diff
@@ -99,24 +99,42 @@ TODO
 ## END
 
 #### eval with argv bindings
-eval ^[echo "$@"] (argv=:| foo bar baz |)
-eval ^[pp line (:| $1 $2 $3 |)] (argv=:| foo bar baz |)
-
-## status: 0
+eval (^(echo "$@")) (pos_args=:| foo bar baz |)
```
andychu, line 102: I think this should be a comma, like `^(echo hi), pos_args= :|foo|`

BTW I think we might remove `"$@"` from YSH ...

I think if the user wants they can bind `ARGV` separately, as we mentioned

By default they only get `$1`

In the regex case, it's not argv but `Match=>groups()` (not yet implemented)

andychu (commented):
Thanks looks good!  I will merge this when CI passes

we can iterate on the details

spec/ysh-builtin-eval.test.sh
PossiblyAShrub: Ah, good catch. I'll fix that in the implementation PR then.

Removing `"$@"` in favor of (`@ARGV`) is sensible. Would we hard remove it or just make it non-idiomatic? Might be good to define what `eval (^(echo "$@"), pos_args=[...])` should mean.

andychu: I think we are going to do this "subset of a superset" thing (analogy to C++ evolution), so yeah we can have `shopt` options to remove `"$@"`

It is easy to make it a parse error in YSH

PR #2019, opened 2024-07-12
Test framework

justinpombrio:
Pure YSH test framework. Lacks the ability to print the body of the assertion that failed, which is important!

### DIFF
--- a/spec/ysh-stdlib-testing.test.sh
+++ b/spec/ysh-stdlib-testing.test.sh
@@ -5,28 +5,28 @@
 source --builtin testing.ysh
 setglobal _test_use_color = false
 
-test_suite "three bad tests" {
-    test_case "the assertion is false" {
+test-suite "three bad tests" {
+    test-case "the assertion is false" {
         assert [1 > 0]
         assert [1 > 1]
         assert [1 > 2]
     }
 
-    test_case "there is an exception while evaluating the assertion" {
+    test-case "there is an exception while evaluating the assertion" {
         assert ["Superman" > "Batman"]
     }
 
-    test_case "there is an exception outside of any assertion" {
+    test-case "there is an exception outside of any assertion" {
         error "oops!"
     }
 }
 
-test_case "one good test case" {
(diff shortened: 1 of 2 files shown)


andychu:
Thanks for sending this out !   

I've had some thoughts from working on OSH testing, and I was thinking about this

I posted them on this thread I pinged you on  - https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Interior.20vs.2E.20Exterior.20testing

related to:

- test discovery across OSH and YSH (and Python and C++) -- shell is inherently very polyglot
- table-based testing - for test parameters like ASAN/UBSAN, Clang/GCC, `bin/osh` vs. `_bin/cxx-opt/osh`
- test cases as regular procs
- 2-level structure vs. flat structure

Let me paste and OSH test in the next comment

This is an "ad hoc test" style we've been using for a long time

andychu:
OK I have been converting our OWN tests to a mini test framework `stdlib/osh/no-quotes.sh`.

It's only 135 lines now, although there is also a test discovery part

The constraint is that these tests must run **under bash**, and then you can run it under OSH seamlessly.

For example we cannot write

```
nq-run (&status) {
  echo hi
}
```

we have to write

```
nq-run status \
  echo hi

```

But it is starting to seem like we should make them consistent ... e.g. you can upgrade your OSH tests to YSH with some simple  syntactic transformation

https://github.com/oilshell/oil/commit/ceffed0c944d25463a97a550caf8f26f65904550

----

We are motivated to do this ourselves because we have like 20K lines of shell tests... we are going to keep many of them in bash/OSH to avoid circular dependencies when bootstrapping (if YSH breaks, we still want to test YSH !)

But it also seems like we should make the OSH and YSH test frameworks roughly consistent

I kinda like "no-quotes" as a name because git's tests look like this

```
test_expect_success '
  log hi   # shell string with potential quoting issues
' 
```

Whereas we avoid that in OSH and YSH, with the `nq-run` style with either `\` or `{ }`

PR #2020, opened 2024-07-15
[devtools] Make build/dev-shell.sh executable

ellen364:
Other commonly used shell scripts like `build/py.sh` are executable, so this one probably should be too.

(This was a minor bit of friction encountered while doing a fresh dev build and following the instructions in [Oil Dev Cheat Sheet](https://github.com/oilshell/oil/wiki/Oil-Dev-Cheat-Sheet).)


andychu:
Ah sorry the correct command is

```
. build/dev-shell.sh  # source it, which doesn't require executable bit
```

This page had a typo due to a contributor - https://github.com/oilshell/oil/wiki/Oil-Dev-Cheat-Sheet

I copied the contributors changes over to - https://github.com/oilshell/oil/wiki/Contributing


Now it looks like

- Requirements / fiddly bits
- Build Oils Python and C++ (5-10 minutes
- "Minimal build" (2 minutes)

(The reason the "minimal build" has been  demoted is because most people don't have python2 on their machine any more.  The complete build makes a python2 "wedge" for you.)

Let me know if you have any troubles!  

The contributor build is now tested more in CI, so I think it should be smoother ... and I think the instructions are slightly better / more readable now

ellen364:
Sourcing the file has worked, thanks.

I also noticed the `.envrc` and have got it working with `direnv`, which seems like a nice workflow.

Closing the PR as not needed.

PR #2021, opened 2024-07-16
[mycpp] share if-statement special cases with all visitors

melvinw:
The `SimpleVisitor` base class was missing these before. This was causing `ir_pass` to run into various collisions that `cppgen_pass` wasn't. Fixing this allows us to make `HeapObjectMember` more strict, which simplifies the next datalog patch.

### DIFF
--- a/mycpp/util.py
+++ b/mycpp/util.py
@@ -4,7 +4,8 @@ util.py
 from __future__ import print_function
 
 import sys
-from mypy.nodes import CallExpr, IfStmt, Block, Expression, MypyFile
+from mypy.nodes import (CallExpr, IfStmt, Block, Expression, MypyFile,
+                       MemberExpr, IntExpr, NameExpr, ComparisonExpr)
 from mypy.types import Instance, Type
 
 from typing import Any, Sequence, Optional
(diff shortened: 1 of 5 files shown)


mycpp/util.py
```diff
@@ -110,3 +111,57 @@ def ShouldSkipPyFile(node: MypyFile) -> bool:
...
+        return True
+
+    if isinstance(cond, NameExpr) and cond.name == 'TYPE_CHECKING':
+        # Omit if TYPE_CHECKING blocks.  They contain type expressions that
+        # don't type check!
+        return True
+
+    if isinstance(cond, MemberExpr) and cond.name == 'CPP':
+        # just take the if block
+        if hasattr(visitor, 'def_write_ind'):
```
melvinw, line 141: these `hasattr()` checks aren't the cutest, but hopefully the usage is obvious enough given the small number of users...

andychu: Hmm yeah this seems weird ... because the new function doesn't just do a check if you should skip, it also visits.

Is it possible to 

1) share the first 3 if statements?
2) not share the def_write_ind() part?    


I remember having trouble with the visitor, so I won't be surprised if this is hard / annoying ...

But I don't see why not right now


----

Is it because def_write_ind() is in multiple passes?

melvinw: Yeah, it's pretty tricky. `def_rite_ind()` is only called in `cppgen_pass` IIRC.

Sharing the first 3 statements would still require this function to accept the visitor as an arg (so the `if 0:` case can visit the else block). If limit ourselves to the `TYPE_CHECKING` and` __name__` cases we could rename this to `ShouldSkipIf(expr) -> bool`, though. However, this quickly starts to defeat the purpose of this helper. I think that means it's trying to do too many things and should be split up.

In retrospect, the act of visiting has no business being in the helper, it's pretty hard to mess up :P The hard part is making sure we agree that a given block should be visited (the details of printing can be left to cppgen_pass). I just pushed a commit that splits this up into three separate helpers. What do you think?
- `ShouldVisitIfExpr(stmt) -> bool`
- `ShouldVisitIfBody(stmt) -> bool`
- `ShouldVisitElseBody(stmt) -> bool`

melvinw:
Just double-checked: the diff of `_gen/bin/oils_for_unix.mycpp.cc` is empty between this patch and master

andychu:
OK great, yeah this looks more reasonable

It is more verbose but I think being explicit is better.  Thanks!

andychu:
Hm I imagine if we have our own IR, I guess we would do a transformation where

- `if CPP` blocks replaced with "if" branch
- `if PYTHON` blocks replaced with "else" branch, if any
- if TYPE_CHECKING blocks deleted
- if `__name__ ...` blocks deleted


i.e. we would create new DATA rather than changing the way we iterate over the MyPy data

But I realize that is hard because everything is so tightly coupled to MyPy right now

---

Having explicit decoupled functions in `util.py` will help us move to that style I think

Now I think the downside of visitors is that everything becomes very tightly coupled to that API , which is the problem we are having ...

PR #2025, opened 2024-07-18
[WIP] Move `procs` to variable namespace

PossiblyAShrub:
I haven't fully thought through the repercussions of this, but I'd like to see what fails in CI first.

Issue #1085, opened 2022-01-30
Introspection for everything


andychu:
One nice thing about Python is that you can look at `dir()` and `__doc__` for everything.  It has a uniform object model.

Shell isn't like that ... But maybe we can add something similar.

Many people are saying the docs aren't sufficient ... but I think it's also true that a good REPL would "take pressure" off them.

It's true that the builtins are not very discoverable.  Related to #388

andychu:
Could put it under the `pp` builtin:


- `pp shopt` -- option state and STACKS
- `pp redir` -- current redirects?  not sure if that's a good idea
- `pp vars` -- complement to `shvar`
  - show PATH and IFS and so forth?
- `pp proc-stack` -- show the stae
  - `pp proc` already prints procs and docstrings

PR #2028, opened 2024-07-20
Move procs to variable namespace

PossiblyAShrub:
Supersedes: #2025

Redoing the implementation now that we know more about the design and possible issues.

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -2324,7 +2324,7 @@ class Procs:
         # type: (str, value.Proc) -> None
         self.procs[name] = proc
 
-    def GetProc(self, name):
+    def Get(self, name):
         # type: (str) -> value.Proc
         vars = self.mem.var_stack[0]
         if name in vars:
@@ -2337,10 +2337,10 @@ class Procs:
 
         return None
 
-    def DelProc(self, to_del):
+    def Del(self, to_del):
         # type: (str) -> None
         """If a proc/sh-func with name `to_del` has been defined, undefined it"""
-        if not self.GetProc(to_del):
+        if not self.Get(to_del):
             return
 
         if to_del in self.procs:
@@ -2348,7 +2348,7 @@ class Procs:
         else:
             mylib.dict_erase(self.mem.var_stack[0], to_del)
 
-    def GetProcNames(self):
+    def GetNames(self):
         # type: () -> List[str]
         """Returns a *sorted* list of all proc names"""
         names = list(self.procs.keys())
--- a/core/state.py
+++ b/core/state.py
@@ -2326,8 +2326,9 @@ class Procs:
 
     def GetProc(self, name):
         # type: (str) -> value.Proc
-        if name in self.mem.var_stack[0]:
-            maybe_proc = self.mem.var_stack[0][name]
+        vars = self.mem.var_stack[0]
+        if name in vars:
+            maybe_proc = vars[name]
             if maybe_proc.val.tag() == value_e.Proc:
                 return cast(value.Proc, maybe_proc.val)
 
@@ -2350,9 +2351,11 @@ class Procs:
     def GetProcNames(self):
         # type: () -> List[str]
         """Returns a *sorted* list of all proc names"""
-        names = list(self.procs)
+        names = list(self.procs.keys())
 
-        for name, cell in self.mem.var_stack[0].items():
+        vars = self.mem.var_stack[0]
+        for name in vars:
+            cell = vars[name]
             if cell.val.tag() == value_e.Proc:
                 names.append(name)
 
--- a/core/state.py
+++ b/core/state.py
@@ -2309,6 +2309,34 @@ class Mem(object):
         return self.ctx_stack.pop()
 
 
+class Procs:
+
+    def __init__(self, mem):
+        # type: (Mem) -> None
+        self.mem = mem
+        self.procs = {}  # type: Dict[str, value.Proc]
+
+    def SetProc(self, name, proc):
+        # type: (str, value.Proc) -> None
+        self.mem.var_stack[0][name] = Cell(False, False, False, proc)
+
+    def SetShFunc(self, name, proc):
+        # type: (str, value.Proc) -> None
+        self.procs[name] = proc
+
+    def GetProc(self, name):
+        # type: (str) -> value.Proc
+        if name in self.mem.var_stack[0]:
+            maybe_proc = self.mem.var_stack[0][name]
+            if maybe_proc.val.tag() == value_e.Proc:
+                return cast(value.Proc, maybe_proc.val)
+
+        if name in self.procs:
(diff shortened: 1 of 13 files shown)


mycpp/gc_list.h
```diff
@@ -342,7 +342,7 @@ T List<T>::pop(int i) {
   len_--;
 
   // Shift everything by one
-  memmove(slab_->items_ + i, slab_->items_ + (i + 1), len_ * sizeof(T));
+  memmove(slab_->items_ + i, slab_->items_ + (i + 1), (len_ - i) * sizeof(T));
```
PossiblyAShrub, line 345: We might want to cherry-pick this out of here. I ran into this issue as described in the commit https://github.com/oils-for-unix/oils/pull/2028/commits/2bc8a856fec952c8a4522b7d824e8da4f301689f. But this is not necessary for the current form of this PR.

PossiblyAShrub:
This is ready for a review.

core/state.py
```diff
@@ -2309,6 +2309,59 @@ def PopContextStack(self):
         return self.ctx_stack.pop()
 
 
+class Procs:
+
+    def __init__(self, mem):
+        # type: (Mem) -> None
+        self.mem = mem
+        self.procs = {}  # type: Dict[str, value.Proc]
```
andychu, line 2317: self.procs could be called self.sh_funcs?

core/state.py
```diff
@@ -2309,6 +2309,59 @@ def PopContextStack(self):
...
+            maybe_proc = vars[name]
+            if maybe_proc.val.tag() == value_e.Proc:
+                return cast(value.Proc, maybe_proc.val)
+
+        if name in self.procs:
+            return self.procs[name]
+
+        return None
+
+    def Del(self, to_del):
```
andychu, line 2340: Is this function only used by `unset -f` ?  If so then maybe it can only call `mylib.dict_erase()` on `self.sh_funcs` ?

I think the way to delete a proc should now be the same way as we delete any other variable.

----

Or I guess if it makes tests fail, we could do this in a separate step

mycpp/gc_list.h
andychu: Oh thanks for finding this!   

I wouldn't mind if it was a separate PR with a unit test --  I guess we are missing some coverage

spec/ysh-proc.test.sh
```diff
@@ -451,3 +465,27 @@ p word (42, n=99) {
 (Int)   99
 Block
 ## END
+
+#### unset and procs
```
andychu, line 469: Maybe clarify the description here?   Now regular `unset` can unset a proc, it doesn't have to be `unset -f`?

That seems right

spec/ysh-proc.test.sh
```diff
@@ -238,7 +238,7 @@ p
 ## STDOUT:
 ## END
 
-#### procs are in same namespace as shell functions
+#### declare -F prints procs and shell-funcs
```
andychu, line 241: I wonder if we should say that `declare -F` is the "old OSH shell world" and have it only print shell functions

But we can think about that, this is great progress

core/state.py
```diff
@@ -2309,6 +2309,59 @@ def PopContextStack(self):
...
+    def SetProc(self, name, proc):
+        # type: (str, value.Proc) -> None
+        self.mem.var_stack[0][name] = Cell(False, False, False, proc)
+
+    def SetShFunc(self, name, proc):
+        # type: (str, value.Proc) -> None
+        self.procs[name] = proc
+
+    def Get(self, name):
+        # type: (str) -> value.Proc
```
andychu, line 2328: Maybe clarify in the docstring that procs are found before shell functions, since that can matter to the user

andychu (commented):
This looks great, thanks!

I am wondering if we should draw more of a separation between "OSH shell funcs" and "YSH procs", but this is progress toward that

I think it should unblock us with the `eval ()` change, which we wanted for `split` and flag parsing I believe

Yes we wanted to hide some names in `args.ysh`

mycpp/gc_list.h
PossiblyAShrub: Okay, I'll revert it here and send up another PR.

spec/ysh-proc.test.sh
PossiblyAShrub: Good point. I guess we could use `pp proc`... but I'll leave it as `declare -F` for now.

core/state.py
PossiblyAShrub: You're right, this can be simplified to just `mylib.dict_erase()`.

PossiblyAShrub:

---

Btw, I did find that my changes have this behavior:

```sh
$ hello() {
>   echo hello
> }
$ var hello = 'hi'
$ hello
"hello"
```

I'm not sure if this is what we want to do here. Should this be correct?

andychu:
> This is ready for another review.
> 
> Btw, I did find that my changes have this behavior:
> 
> ```shell
> $ hello() {
> >   echo hello
> > }
> $ var hello = 'hi'
> $ hello
> "hello"
> ```
> 
> I'm not sure if this is what we want to do here. Should this be correct?

OK I guess what's happening is that 

- it's looked up in the var namespace, but it's not a Proc, it's a string
- then it's looked up in the "sh func" namespace, and found


Hm I guess the exact rules are up for debate ...

You could argue that if the name is found in the var namespace, it could show an error "not a proc"

----

This would possibly have some annoying consequences like:

```
var ls = 42
ls   # external ls should still work ???   
```

---


I don't think the behavior you showed is bad though, I guess because it doesn't make sense to call the string "hi" anyway

I think we can leave it as is for now, and perhaps iterate

PossiblyAShrub:
> I think we can leave it as is for now, and perhaps iterate

Sounds good.

Are there any other blockers on this PR before we merge?

andychu:
Looks good, thanks!  

I merged, though now that I think about it, if `Del()` is only for `sh_funcs`, then I think `GetNames()` should be too

I think we should change `declare -F` and `compgen -A function` to only list shell functions


procs are now something "different" ... I think we might have `compgen -A proc` and so forth


----

Actually `declare -F` like is `dir()` !  I remember I wanted to write a blog post about how I used `dir()` and `getattr()` to learn Python, **not** reading the docs

We want that property too

- #1085 


It also reminds me that our object model is too weak ... maybe we should be able to have

- dir() on any object?
- dir() on stack frames?   So then you know how to `getVar()` and so forth


----

If you agree about excluding procs out of GetNames(), please send a follow-up

andychu:
I am also thinking of things about our "meta-object model", like Python has

- `__call__`
- `__dir__` (I think)
- `__get__` and `__getattr__`


I think we might want something like `__invoke__`

All these things can be "invoked" like a command:

- shell functions
- procs
- maybe modules

PR #2029, opened 2024-07-21
[builtins/dict] add -> erase() method

ellen364:
An oddity is that `erase` works with both `->` and `=>` despite only mutating, not transforming. But this behaviour matches other Dict and List methods, so the implementations are at least consistent with each other.

### DIFF
--- a/doc/ref/chap-type-method.md
+++ b/doc/ref/chap-type-method.md
@@ -343,7 +343,7 @@ Similar to `keys()`, but returns the values of the dictionary.
 
 ### erase()
 
-Accepts a key and deletes the relevant key-value pair. The method is idempotent.
+Ensures that the given key does not exist in the dictionary.
 
     var book = {
       title: "The Histories",
@@ -352,6 +352,11 @@ Accepts a key and deletes the relevant key-value pair. The method is idempotent.
     = book
     # => (Dict)   {title: "The Histories", author: "Herodotus"}
 
+    call book->erase("author")
+    = book
+    # => (Dict)   {title: "The Histories"}
+
+    # repeating the erase call does not cause an error
     call book->erase("author")
     = book
     # => (Dict)   {title: "The Histories"}
--- a/doc/ref/chap-type-method.md
+++ b/doc/ref/chap-type-method.md
@@ -343,6 +343,19 @@ Similar to `keys()`, but returns the values of the dictionary.
 
 ### erase()
 
+Accepts a key and deletes the relevant key-value pair. The method is idempotent.
+
+    var book = {
+      title: "The Histories",
+      author: "Herodotus",
+    }
+    = book
+    # => (Dict)   {title: "The Histories", author: "Herodotus"}
+
+    call book->erase("author")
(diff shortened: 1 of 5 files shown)


spec/ysh-methods.test.sh
```diff
@@ -393,6 +393,15 @@ pp line (en2fr => values())
 (List)   ["bonjour","ami","chat"]
 ## END
 
+#### Dict -> erase()
+var book = {title: "The Histories", author: "Herodotus"}
+call book->erase("author")
```
andychu, line 398: I would add another call to test that we do nothing when the key doesn't exist

andychu (commented):
Looks great!  Thanks for doing this

Glad the mycpp error message existed!   (Sometimes we don't check for the subset of Python we use)

spec/ysh-methods.test.sh
ellen364: New commit has the extra call.

Doing nothing when the key doesn't exist has really grown on me. Wasn't sure about it at first (felt like a silent failure) but realised that a lot of time I want a dictionary in a particular way and don't care whether the key was there or not, as long as it isn't now.

doc/ref/chap-type-method.md
```diff
@@ -343,6 +343,19 @@ Similar to `keys()`, but returns the values of the dictionary.
 
 ### erase()
 
+Accepts a key and deletes the relevant key-value pair. The method is idempotent.
```
andychu, line 346: How about saying: "erase() ensures that the given key doesn't not exist in the dictionary"

This means either deleting it, or doing nothing if it doesn't exist (which implies that erase() is idempotent)

---

FWIW I believe this same  issue is discussed in Ousterhout's book.  He regretted that for the Tcl language, they made "delete key that does not exist" an error, like it is in Python!

He thinks they should have redefined it to "ensure key does not exist", which is what we're doing here!

https://blog.pragmaticengineer.com/a-philosophy-of-software-design-review/

andychu (commented):
small doc suggestion, sorry I didn't see this the first time

I also hint at one reason we differ from Python, which seems to be nice, based on my experience in Oils

andychu:
I'm probably making a release tomorrow ... I may merge this with some tweaks to the docs, to get it in

Thanks for doing this!

doc/ref/chap-type-method.md
ellen364: The new commit has your suggested phrasing + an additional example. The docs basically repeat the tests now, but I think the extra example helps reinforce what behaviour to expect.

And A Philosophy of Software Design has just gone onto the reading list :slightly_smiling_face:

andychu:
Oh also please keep us updated on Zulip about any code you write with YSH!

Or if you are looking for other things to do, it would be cool to implement `Dict->get() inc() accum()`, which are marked with a red X

I think those are mostly similar to `erase()`, and I'm sure many people will use them!

https://www.oilshell.org/release/0.22.0/doc/ref/toc-ysh.html

PR #2032, opened 2024-07-25
Fix pp tabular

justinpombrio:
Fix a bug in pretty printing tabular alignment (it must force elements to be flat).

The new test case is robust against differing values of `_DEFAULT_MAX_TABULAR_WIDTH`.

### DIFF
--- a/data_lang/pretty.py
+++ b/data_lang/pretty.py
@@ -76,6 +76,8 @@ maximum line width.
 #
 # IfFlat(a, b) prints a if in flat mode or b otherwise.
 #
+# Flat(a) prints a in flat mode. You should generally not need to use it.
+#
 # ~ Measures ~
 #
 # The algorithm used here is close to the one originally described by Wadler,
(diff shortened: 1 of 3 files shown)


data_lang/pretty.py
```diff
@@ -76,6 +76,8 @@
 #
 # IfFlat(a, b) prints a if in flat mode or b otherwise.
 #
+# Flat(a) prints a in flat mode. You should generally not need to use it.
```
andychu, line 79: Hm I guess this reminds me that  I still want to test the policy/mechanism split, with the ASDL pretty printer

I guess that would also use _Tabular, but it doesn't necessarily use IfFlat / Flat directly either

And there is a bit of dependency issue because we are maintaining a separation of ASDL from Oils

e.g. the `value_t -> doc_t` transformation is ideally separate, and then we also have an ASDL `hnode_t -> doc_t` transform

andychu (commented):
Great thanks for fixing this ... I am a little confused about what's "public", and I think we have a good way to test that

PR #2042, opened 2024-08-04
add +/- annotations to jobs builtin

melvinw:
A small quality of life improvement for job control. @andychu brought this up in https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Job.20control.20hard.20to.20understand

```
[I] osh-0.22.0$ cat
^Z
[PID 489245] Stopped with signal 20
[I] osh-0.22.0$ cat
^Z
[PID 489246] Stopped with signal 20
[I] osh-0.22.0$ cat
^Z
[PID 489247] Stopped with signal 20
[I] osh-0.22.0$ jobs
%1 489245 Stopped [process] cat
%2- 489246 Stopped [process] cat
%3+ 489247 Stopped [process] cat
```

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -928,8 +928,8 @@ class Job(object):
         self.job_id = -1
         self.in_background = False
 
-    def DisplayJob(self, job_id, f, style):
-        # type: (int, mylib.Writer, int) -> None
+    def DisplayJob(self, job_id, f, style, extra):
+        # type: (int, mylib.Writer, int, str) -> None
         raise NotImplementedError()
 
     def State(self):
@@ -1014,12 +1014,12 @@ class Process(Job):
 
         return self.pid
 
-    def DisplayJob(self, job_id, f, style):
-        # type: (int, mylib.Writer, int) -> None
+    def DisplayJob(self, job_id, f, style, extra):
+        # type: (int, mylib.Writer, int, str) -> None
         if job_id == -1:
             job_id_str = '  '
         else:
-            job_id_str = '%%%d' % job_id
+            job_id_str = '%%%d%s' % (job_id, extra)
         if style == STYLE_PID_ONLY:
             f.write('%d\n' % self.pid)
         else:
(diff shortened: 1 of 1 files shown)


core/process.py
```diff
@@ -1777,9 +1777,17 @@ def DisplayJobs(self, style):
         #      24510                       | sleep 5 &
 
         f = mylib.Stdout()
+        current, previous = self.GetCurrentAndPreviousJobs()
         for job_id, job in iteritems(self.jobs):
             # Use the %1 syntax
```
andychu, line 1782: this comment can move to _DisplayJob()

core/process.py
```diff
@@ -1777,9 +1777,17 @@ def DisplayJobs(self, style):
...
-            job.DisplayJob(job_id, f, style)
+            extra = ''
```
andychu, line 1783: I think it should default to one space so it's aligned?

andychu (commented):
Thanks!  

I wonder if we can add a spec test like

```
$ bin/osh -c 'sleep 0.01 & sleep 0.02 & sleep 0.03 & jobs'
%1 3423 Running [subprog] command.Simple
%2 3424 Running [subprog] command.Simple
%3 3425 Running [subprog] command.Simple

```

(Although we really need to print code correctly)

...

I also noticed the pipeline case is a bit weird

we are not printing a job num

```
$ bin/osh -c 'sleep 0.1 | wc -l & jobs'
   3440 Running [subprog] command.Simple
```

bash and zsh do

```
$ bash -c 'sleep 0.1 | wc -l & jobs'
[1]+  Running                 sleep 0.1 | wc -l &
```

core/process.py
melvinw: Yes! Nice catch

core/process.py
melvinw: Done

melvinw:
Let me see about adding a spec test

andychu:
Hm I think we should land this?  Let me look at what happened

PR #2044, opened 2024-08-07
Eval with vars

PossiblyAShrub:


core/state.py
```diff
@@ -2403,11 +2403,9 @@ def Get(self, name):
         First, we search for a proc, and then a sh-func. This means that procs
         can shadow the definition of sh-funcs.
         """
-        vars = self.mem.var_stack[0]
-        if name in vars:
-            maybe_proc = vars[name]
-            if maybe_proc.val.tag() == value_e.Proc:
-                return cast(value.Proc, maybe_proc.val)
+        maybe_proc = self.mem.GetValue(name)
```
PossiblyAShrub, line 2406: It is not good enough just to check `self.mem.var_stack[0]` for the `parseArgs` eval test. I think this may also fix issues like:

```
proc my-proc {
  var toCall = someOtherProc
  toCall foo bar baz
}
```

---

I should also look at GetNames because now it's doing something different

PossiblyAShrub: This should be good now. We will leave `GetNames` as-is after the discussion here: https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/value.2EProc.20in.20the.20var.20namespace.

PossiblyAShrub:
This is ready for a review.

core/state.py
```diff
@@ -1136,6 +1136,53 @@ def _MakeArgvCell(argv):
...
+
+    def __init__(self, mem, dollar0, pos_args, vars):
+        # type: (Mem, Optional[str], Optional[List[str]], Optional[Dict[str, value_t]]) -> None
+        self.mem = mem
+        self.pushed_pos_args = False
+        self.pushed_dollar0 = False
+        self.pushed_vars = False
+
+        # $0 needs to have lexical scoping. So we store it with other locals.
+        # As "0" cannot be parsed as an lvalue, we can safely store arg0 there.
```
andychu, line 1150: s/arg0/dollar0/

core/state.py
```diff
@@ -1136,6 +1136,53 @@ def _MakeArgvCell(argv):
...
 
+class ctx_Eval(object):
+    """Push temporary variable frame and override $0, $1, $2, etc."""
+
+    def __init__(self, mem, dollar0, pos_args, vars):
+        # type: (Mem, Optional[str], Optional[List[str]], Optional[Dict[str, value_t]]) -> None
+        self.mem = mem
+        self.pushed_pos_args = False
+        self.pushed_dollar0 = False
+        self.pushed_vars = False
```
andychu, line 1147: style nitpick - I think it's clearer to save

```
self.dollar0 = dollar0
self.pos_args = pos_args
self.vars = vars
```

and then in the `__exit__`  you can use the same conditions as in the constructor

Since they are all `is not None` I think it will read clearly

core/state.py
```diff
@@ -1136,6 +1136,53 @@ def _MakeArgvCell(argv):
...
+
+        # $0 needs to have lexical scoping. So we store it with other locals.
+        # As "0" cannot be parsed as an lvalue, we can safely store arg0 there.
+        if dollar0 is not None:
+            self.pushed_dollar0 = True
+            assert mem.GetValue("0", scope_e.LocalOnly).tag() == value_e.Undef
+            self.lval = LeftName("0", loc.Missing)
```
andychu, line 1154: I might call this `self.dollar0_lval` or something ... `self.lval` is a bit generic

spec/ysh-proc.test.sh
```diff
@@ -524,3 +524,28 @@ grep
 ## STDOUT:
 sh-func grep
 ## END
+
+#### proc resolution changes with the local scope
```
andychu, line 528: I'm not 100% sure why we need this

It seems reasonable and consistent, but maybe add some comments about what goes wrong if we don't have it?

andychu (commented):
Thanks, this looks great!

PossiblyAShrub:

spec/ysh-proc.test.sh
```diff
@@ -524,3 +524,35 @@ grep
...
+proc inner {
+  var foo = bar
+  foo  # Will now reference `proc bar`
+}
+
+foo
+inner
+foo  # Back to the global scope, foo still references `proc foo`
+
+# Without this behavior, features like `eval(b, vars={ flag: __flag })`, needed
```
andychu, line 548: Hm actually I think we may need something more like `ctx_Shvar`

https://github.com/oils-for-unix/oils/blob/master/builtin/pure_ysh.py#L32

It does NOT push a new stack frame, like we are doing here.

Instead it temporarily replaces variable of the given names on SAME stack frame.

The reason we do it that way is because procs don't have dynamic scope.  The problem with pushing a new stack frame is that you will "lose" your current local variables ...  You will ONLY have the new ones.

I think it may come up in `Str => replace()` too -- are the current variables in the frame available for replacement?  It could go either way, but I think for `eval()` for `lines` and `split-by` etc. you will want to be able to use OTHER variables in the same stack frame.

(This was something I learned the hard way -- originally I had pushed a new stack frame, but then that broke some things)

---

Let me see if I can come up with a test case for this ... I will merge this, but then I think we probably will run into this issue, and can revise it

andychu:
Oh another thing is that I recently added `_io->eval(b)` ... that is for pure funcs to run arbitrary shell

We should synchronize that with plain `eval (b)` eventually.  But let's look at the other issue first

spec/ysh-proc.test.sh
PossiblyAShrub: Actually, in this case, isn't this good? Previously you could:

```
proc foo(;;; block) {
  var bar = 123
  eval (block)
}

foo { = bar }
# => 123
```

We would leak variables from the `eval` frame into the inner block. With `vars`, you can control which variables are visible (although, you won't get mutation.)

andychu: The `ctx_Shvar` will save AND restore

But I tried to reproduce the issue I was talking about, and couldn't

https://github.com/oils-for-unix/oils/commit/59ece78259d0eeed24d1f7b51aeec952c365becd

I guess we can move on now

If we hit it, we can fix it then ... I'm pretty sure it's an issue but I wasn't able to come up with an example, and then I got distracted trying to print the stack in a readable format!  That could be useful for tests

PR #2045, opened 2024-08-07
[builtins] Add Dict => get() method

ellen364:
Use verbose variable name `default_value` because `default` breaks mycpp, which seems to confuse it for the default case in a switch statement. (Going by an initial read of the errors. I haven't dug into it.)

This implementation makes the default value a required argument, but I can make it optional (more Pythonic) if preferred.

Still a bit confused about `=>` vs `->` because all `Dict` methods, including this one, seem to work with both.

### DIFF
--- a/builtin/method_dict.py
+++ b/builtin/method_dict.py
@@ -61,3 +61,20 @@ class Erase(vm._Callable):
 
         mylib.dict_erase(dictionary, key)
         return value.Null
+
+
+class Get(vm._Callable):
+
+    def __init__(self):
+        # type: () -> None
+        pass
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+
+        dictionary = rd.PosDict()
+        key = rd.PosStr()
+        default_value = rd.PosValue()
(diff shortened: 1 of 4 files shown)


andychu:
This looks great, thank you!

- Yes `default` will conflict with C++ ... sometimes I use `default_`, but `default_value` works too
- We are almost certainly going to change `=>` to `.` !
  - thread here - https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/.3D.3E.20is.20unwieldy.2Funfamiliar.3B.20object.20model.20too.20weak

Right now `->` and `=>` are synonyms

The idea was to "try out" the syntax, and see if it feels right ... And that was probably good, because we found it is a little odd!

So this is great progress, and then we can change it to `d.get(k, 0)` like Python later

----

On the second arg being required ... that is something I'm thinking about too

Right now we have a correspondence between:

- named args are always optional
- positional args are always required

But in Python, `d.get(k)` has an optional AND positional arg

We may rethink that too, but that's separate from this change.

Thanks a lot!

PR #2047, opened 2024-08-09
Don't push a new frame in `eval ()`

PossiblyAShrub:
This test captures the effect of this change:

```
#### eval with vars follows same scoping as without
proc local-scope {
  var myVar = "foo"
  eval (^(echo $myVar), vars={ someOtherVar: "bar" })
  eval (^(echo $myVar))
}

# In global scope
var myVar = "baz"
eval (^(echo $myVar), vars={ someOtherVar: "bar" })
eval (^(echo $myVar))

local-scope
## STDOUT:
baz
baz
foo
foo
## END
```

### DIFF
--- a/builtin/pure_ysh.py
+++ b/builtin/pure_ysh.py
@@ -3,61 +3,22 @@ builtin/pure_ysh.py - YSH builtins that don't do I/O.
 """
 from __future__ import print_function
 
-from _devbuild.gen.runtime_asdl import (cmd_value, scope_e)
+from _devbuild.gen.runtime_asdl import cmd_value
 from _devbuild.gen.syntax_asdl import command_t, loc, loc_t
-from _devbuild.gen.value_asdl import (value, value_e, value_t, LeftName)
+from _devbuild.gen.value_asdl import value, value_e, value_t
 from core import error
 from core import state
 from core import vm
 from frontend import flag_util
-from frontend import location
 from frontend import typed_args
 from mycpp import mylib
 from mycpp.mylib import tagswitch
 
-from typing import TYPE_CHECKING, cast, Any, Dict, List, Tuple
+from typing import TYPE_CHECKING, cast, Any, Dict, List
 
 if TYPE_CHECKING:
     from display import ui
     from osh.cmd_eval import CommandEvaluator
-
-
-class ctx_Shvar(object):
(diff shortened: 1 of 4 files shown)


core/state.py
```diff
@@ -1161,14 +1162,33 @@ def __enter__(self):
...
 
         if self.pos_args is not None:
             self.mem.argv_stack.pop()
 
         if self.dollar0 is not None:
             self.mem.SetLocalName(self.dollar0_lval, value.Undef)
 
+    # Note: _Push and _Pop are separate methods because the C++ translation
+    # doesn't like when they are inline in __init__ and __exit__.
+    def _Push(self, pairs):
```
andychu, line 1175: How about creating free functions like:

```
state.PushVars(mem, restore)
state.PopVars(mem, restore)
```

and then using it from both `ctx_Shvar` and `ctx_Eval` ?

---

Also, I think where this is going is that we are going to move it up to the YSH level:


```
push-vars ({a: 42, b: 'foo'} {
  echo $a $b
}
```

There is a thread about "unifying stacks with context managers" on Zulip

andychu (commented):
Great thanks!  Glad it was easy to find this

And I'm glad it came up at the same time as the Hay bug ...

core/state.py
PossiblyAShrub: What about using `ctx_Eval` instead of `ctx_Shvar` for `shvar`?

andychu: Oh yes, that makes sense!  I wasn't sure how similar they were

builtin/pure_ysh.py
```diff
@@ -80,7 +41,7 @@ def Run(self, cmd_val):
             # But should there be a whitelist?
             raise error.Usage('expected a block', loc.Missing)
 
-        pairs = []  # type: List[Tuple[str, value_t]]
+        vars = {}  # type: Dict[str, value_t]
```
andychu, line 44: Oh I think we might need `NewDict()` here

This is a trick/hack to preserve dict ordering in Python (underneath it uses OrderedDict)

In mycpp, our `Dict<K, V>` is defined to be ordered

(I hit this the other day too)

andychu (commented):
Oh yes, good idea!

andychu:
Awesome, thank you!  Now my question is if we should also use `ctx_Eval` for TWO places in Hay

I'll post on Zulip

PR #2048, opened 2024-08-10
Implement `string.split()`

PossiblyAShrub:
See design discussion on zulip: https://oilshell.zulipchat.com/#narrow/stream/447377-api-design/topic/API.20design.20-.20shSplit.28.29.20split.28.29.20splitlines.28.29

TODOs:
- [x] Write spec tests
- [x] Implementation
- [x] Update docs

### DIFF
--- a/builtin/method_str.py
+++ b/builtin/method_str.py
@@ -491,17 +491,13 @@ class Split(vm._Callable):
         s.split(sep, count=-1)
 
         Count behaves like in replace() in that:
-        - A negative `count` => ignore
-        - A positive `count` => there will be at most `count` splits
-        - Exception: `count = 0` is an error
+        - `count` <  0 -> ignore
+        - `count` >= 0 -> there will be at most `count` splits
         """
         string = rd.PosStr()
         sep = rd.PosStr()
         count = mops.BigTruncate(rd.NamedInt("count", -1))
         rd.Done()
-
-        if count == 0:
-            raise error.Structured(3, "count cannot be zero", rd.LeftParenToken())
 
         if len(sep) == 0:
             raise error.Structured(3, "sep must be non-empty", rd.LeftParenToken())
--- a/builtin/method_str.py
+++ b/builtin/method_str.py
@@ -2,7 +2,7 @@
 
 from __future__ import print_function
 
-from _devbuild.gen.syntax_asdl import loc_t
+from _devbuild.gen.syntax_asdl import loc, loc_t
 from _devbuild.gen.value_asdl import (value, value_e, value_t, eggex_ops,
                                       eggex_ops_t, RegexMatch)
 from core import error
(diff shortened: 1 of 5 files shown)


PossiblyAShrub:
This is ready for a review.

builtin/method_str.py
```diff
@@ -477,3 +477,53 @@ def Call(self, rd):
...
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+        """
+        s.split(sep, count=-1)
+
+        Count behaves like in replace() in that:
+        - A negative `count` => ignore
+        - A positive `count` => there will be at most `count` splits
+        - Exception: `count = 0` is an error
```
andychu, line 496: Hm I think zero and negative should be the same case -- ignored, which is equivalent to zero splits



CPython behaves this way too


```
ysh andy@hoover:~/git/oilshell/oil$ = 'foo'.replace('o', 'x', count=1)
(Str)   'fxo'
ysh andy@hoover:~/git/oilshell/oil$ = 'foo'.replace('o', 'x', count=2)
(Str)   'fxx'
ysh andy@hoover:~/git/oilshell/oil$ = 'foo'.replace('o', 'x', count=0)
(Str)   'foo'
ysh andy@hoover:~/git/oilshell/oil$ = 'foo'.replace('o', 'x', count=-1)
(Str)   'fxx'

```

doc/ref/chap-type-method.md
```diff
@@ -256,6 +256,26 @@ The `%start` or `^` metacharacter will only match when `pos` is zero.
...
+
+Split a string by a `Str` separator `sep` into a `List` of chunks.
+
+    pp ('a;b;;c'.split(';'))       # => ["a", "b", "", "c"]
+    pp ('a<>b<>c<d'.split('<>'))   # => ["a","b","c<d"]
+    pp ('🌞🌝🌞🌝🌞'.split('🌝'))  # => ["🌞", "🌞", "🌞"]
+
+Optionally, provide a `count` to split on `sep` at most count times. A `count`
+less than zero will split on all occurrences of `sep`. `count` cannot equal
+zero.
```
andychu, line 269: ditto here, don't need zero special case

spec/ysh-builtin-eval.test.sh
```diff
@@ -109,7 +109,7 @@ foo bar baz
 #### eval lines with argv bindings
 proc my-split (;;; block) {
   while read --raw-line {
-    var cols = _reply => split()
+    var cols = split(_reply)
```
andychu, line 112: oh yeah I forgot we have both a free function and a method

now that we are getting rid of `=>` in favor of `.`, there's a one char difference

```
split(s)
s.split()
```

We can decide on whether we need both separately

I suppose it is inconsistent if we have `split()` and not others

---

I think `shSplit()` should possibly be a free function ...   it was related to glob(), because the traditional shell word eval algorithm  uses IFS splitting and globbing

andychu (commented):
Looks great, thanks!

builtin/method_str.py
PossiblyAShrub: Hmm, yeah, I guess we should allow `count=0`. Although, we probably shouldn't "ignore" like a negative count.

For reference, JS and Python do this:

```
>>> "a;b;c".split(";", maxsplit=0)  # Python
['a;b;c']
> "a;b;c".split(';', 0)  // JS
[]
```

JS doesn't include the "remainder" chunk, so it sort of agrees with Python:

```
> "a,b,c".split(',', 1)
[ 'a' ]
> "a,b,c".split(',', 2)
[ 'a', 'b' ]
```

But we follow Python's style of including a "remainder" elsewhere, so I'll do that for `count=0` too:

```
ysh$ = 'a;b;c'.split(';', count=0)
(List)   ["a;b;c"]
```

spec/ysh-builtin-eval.test.sh
PossiblyAShrub: > I think `shSplit()` should possibly be a free function ... it was related to glob(), because the traditional shell word eval algorithm uses IFS splitting and globbing

Yeah, I agree that those are fine as free functions for that reason.

However, `split` really should only be a method. Let's leave it in for now though as otherwise there is no way to get a python style `string.split()` (split on all runs of whitespace.)

PossiblyAShrub:

andychu:
Merged, thank you!

andychu:
BTW I am thinking that `shSplit()` is "discouraged" -- because the algorithm is fiddly and possibly slow -- and the regex is more general than IFS

So `split(s)` can be an alias for `s.split()` for convenience

and without any second arg, it will do the whitespace splitting

I think whitespace splitting as a regex has some unicode issues, but it might just work, and if not we can do non-Unicode to start (since I think this function will be extremely commonly used)

PR #2062, opened 2024-08-28
Add spec test for ArithSub evaluation bug in autotools configure

PossiblyAShrub:


andychu:
FWIW I'm thinking we should add the cases we came up on Zulip as a "reminder" ...

Those were cases with

- `x=1+2; echo $(( x * 3 ))`, 
- `$(( '1+2' + 3 ))`
- `$(( "1+2" + 3 ))`

We don't need to fix it now though, if the other change is enough to unblock autoconf (and I think it is)

PossiblyAShrub:
Good idea. I opened https://github.com/oils-for-unix/oils/pull/2070.

PR #2064, opened 2024-08-28
Fix autotools inf loop bug

PossiblyAShrub:


andychu:
Thanks for tracking this down!  I ran the spec test and see we are erroneously getting `as_val=0` rather than `as_val=1`

I didn't debug it fully, but my first thought is if `shopt -s eval_unsafe_arith` fixes it?   Let me try

andychu:
Second thought is maybe `$(( $* ))` is treating `0 + 1` as 1 token and not 3?

Not sure

PossiblyAShrub:
> Second thought is maybe `$(( $* ))` is treating `0 + 1` as 1 token and not 3?

Yep, that's exactly the issue. Previously we just tried to parse it as an int and only after failing that would we parse it as a proper expression. I'm hoping that by just invoking the arith parser first, we can fix the whole issue.

andychu:
BTW this snippet exhibits the same bug

```
 
 as_fn_arith2() {
   as_val=$(( "$1" ))
 }
 
 as_fn_arith2 0+1
 echo as_val=$as_val
```

it is 0 with OSH but 1 with bash

So at least it is not specific to `$*` -- it still happens with `$1`.  That's good because if it were specific to `$*` it would probably be more annoying to fix


...

So it's probably worth testing that this case is also fixed too

andychu:
I tested a bit more and these both exhibit the same bug too

```
 as_fn_arith2() {
   myvar=$1
   as_val=$(( $myvar ))
 }

 as_fn_arith2() {
   myvar=$1
   as_val=$(( myvar ))  # no $ which is allowed inside arithmetic
 }
 
```

So yeah I think it is a very general pattern!

I think what happened is that we had a very strict version, which wasn't good enough, and then expanded it in a way that was also not covering this case

The code structure is probably pretty suboptimal

PossiblyAShrub:
Yeah, the primary sign of the issue is this:

```
osh$ echo $(( "0 + 1" ))
```

But not this:

```
osh$ echo $(( 0 + 1 ))
```

The root cause is that `_StringToBigInt` is used for *both* dynamic parsing of complex expressions (eg. `a + b`) and integer literals. That substitution required dynamically parsing '0 + 1' using `_StringToBigInt`, which would see the leading zero and assume the entire string is an octal constant. You could trick it with a leading `0x` or `n#...` as well.

So far, I've made the easy change; I just reordered the sequence of events. Before it was:

- Try to parse an int literal
- If that fails, parse an entire string using the arith parser

Now it is:
- First invoke the arith parser
- If the arith parser says the input string is just a word, parse that word literal (using the same code as before)
- Otherwise, it's a more complex expression, so we evaluate it recursively

This result is still not optimal as we end up having to construct a new arith parser for each term:

https://github.com/oils-for-unix/oils/blob/71116ac5e990d2eb6e56aa0ada6cde9d86e9ea0c/osh/sh_expr_eval.py#L403-L411

(each call of _StringToInt calls `parse_ctx.MakeArithParser`.)

A better solution should fully resolve the expression string, parse it and then finally evaluate. (More like a traditional compiler.) Unfortunately, I think the change will be a little messy. Currently we allow "heterogeneous" expressions like `$(( '1 + 2' + 3 ))` which is both statically and dynamically parsed!

---

Actually, I see there is another correctness issue here!

```
osh-0.23.0$ bash -c 'echo $(( "1 + 2" * 3 ))'
7
osh-0.23.0$ echo $(( "1 + 2" * 3 ))
9
```

Here's the parse tree for reference:

```
osh-0.23.0$ osh -nc 'echo $(( "1 + 2" * 3 ))'
(command.Simple
  blame_tok: 
    (Token
      id: Id.Lit_Chars
      length: 4
      col: 0
      line: (SourceLine line_num:1 content:"echo $(( \"1 + 2\" * 3 ))" src:(source.CFlag))
    )
  more_env: []
  words: [
    (CompoundWord parts:[...0x122e])
    (CompoundWord
      parts: [
        (word_part.ArithSub
          left: (Token id:Id.Left_DollarDParen length:3 col:5 line:...0x1236)
          anode: 
            (arith_expr.Binary
              op: (Token id:Id.Arith_Star length:1 col:17 line:...0x1236)
              left: 
                (CompoundWord
                  parts: [
                    (DoubleQuoted
                      left: (Token id:Id.Left_DoubleQuote length:1 col:9 line:...0x1236)
                      parts: [(Token id:Id.Lit_Chars length:5 col:10 line:...0x1236)]
                      right: (Token id:Id.Right_DoubleQuote length:1 col:15 line:...0x1236)
                    )
                  ]
                )
              right: (CompoundWord parts:[(Token id:Id.Lit_Digits length:1 col:19 line:...0x1236)])
            )
          right: (Token id:Id.Right_DollarDParen length:1 col:22 line:...0x1236)
        )
      ]
    )
  ]
  is_last_cmd: F
)
```

osh/sh_expr_eval.py
```diff
@@ -386,57 +385,63 @@ def _StringToBigInt(self, s, blame_loc):
...
+            integer = self.EvalToBigInt(node2)
+        else:
+            # BoolEvaluator doesn't have parse_ctx or mutable_opts
+            assert self.mutable_opts is not None
+
+            # We don't need to flip _allow_process_sub, because they can't be
+            # parsed.  See spec/bugs.test.sh.
+            with state.ctx_Option(self.mutable_opts,
+                                  [option_i._allow_command_sub], False):
+                integer = self.EvalToBigInt(node2)
```
andychu, line 444: On inspection, I think this could be problematic because we are evaluating `node2` when   `eval_unsafe_arith` is OFF

----

The reason eval_unsafe_arith exists is because of this design flaw in shell, pointed out by the Fedora security team, and which I re-discovered quite a few years ago (and reported to OpenBSD  and they fixed !!!   bash still has the bug)

https://github.com/oils-for-unix/blog-code/tree/main/crazy-old-bug

----

I am thinking we might want something like

rename:

match.LooksLikeInteger -> match.LooksLikeDecimalInt   (also search for LOOKS_LIKE_INTEGER)

and then add

```
match.LooksLikeHexInt
match.LooksLikeOctInt
match.LooksLikeShellInt  # arbitrary base
```

with a regular expression.

----

The idea is to detect `0123` but not `0 + 1`     , which I think is the bug


I am not sure if this is the best approach but it seems consistent with what we already do


and it avoids the issue of "try parsing" and then evaluate

I guess it also avoids invoking the arith parser many times?

The regular language checks should be fast.   (they don't allocate!)

(Side note: the way the code is laid out with LOOKS_LIKE_INTEGER etc. is a little tortured, but I think you can just follow the pattern.  The reason it looks weird is because we have this whole `fastlex` Python extension ... in theory we can get rid of it, now that Oils is translated to C++.    There can be just 2 implementations/translations of the lexer, not 2.5 or 3!)

andychu (commented):
Thanks, this looks very promising!  I posted my initial thoughts here, let me know what you think, or if any other approaches seem better

andychu:
Upon sleeping on it, another strategy I realized may fit is to use a regex at runtime, which we do in exactly 2 places now

https://github.com/oils-for-unix/oils/blob/master/frontend/consts.py#L353

This is instead of the `match.LooksLikeInteger`

The reason for that would so you can use submatch extraction ,e.g.

```
0x123
  ^^^ extract this
64#a
^^ ^  # extract two things
```

It is a minor difference though ...

(re2c does support submatch extraction, but our wrappers don't, basically because historically it would only be used %1 of the time or less)

osh/sh_expr_eval.py
```diff
@@ -372,11 +377,11 @@ def _StringToBigInt(self, s, blame_loc):
...
                     digit = int(ch)
                 else:
-                    e_strict('Invalid digits for numeric constant %r' % digits,
-                             blame_loc)
+                    # Unreachable per the regex validation above
+                    raise AssertionError()
 
                 if digit >= base:
-                    e_strict(
+                    e_die(
```
PossiblyAShrub, line 384: Changed this from `e_strict` to `e_die` because other shells already raise an error here.

Previously:

```
osh-0.23.0$ bash -c 'echo $(( 2#A ))'
bash: line 1: 2#A: value too great for base (error token is "2#A")
osh-0.23.0$ osh -c 'echo $(( 2#A ))'
0

osh-0.23.0$ osh --version
Oils 0.23.0		https://www.oilshell.org/

git commit = 30f7d6a33c655e9e32b97aabc9240ace42a11d8e
```

PossiblyAShrub:

osh/sh_expr_eval.py
```diff
@@ -337,29 +337,34 @@ def _StringToBigInt(self, s, blame_loc):
         bare word: variable
         quoted word: string (not done?)
         """
-        if s.startswith('0x'):
+        m = util.RegexSearch(r'^\s*0x([0-9A-Fa-f]+)\s*$', s)
```
andychu, line 340: Hm actually I am not sure `\s` works, because libc does POSIX ERE, and `\s` is Perl/Python style

It is maybe a little surprising that no spec tests fail because of this, I guess  we are missing coverage

---

I suggest what we should do is write a helper like `NumberMatch(s, '0x([0-9...])')`  or `MatchNumber` which implicitly adds the whitespace rule, and then we use it for ALL 3 cases

And ALSO unit test it. 

I often treat regexes as code and unit test them ...  e.g.  there is a unit test for ASSIGN_ARG_RE

```
class RegexTest(unittest.TestCase):

    def testSplitAssignArg(self):
        CASES = [
            # var name, op, value
            ('s', ['s', '', '']),
            ('value', ['value', '', '']),

```

---

Or it might be possible to simply tweak some spec test cases for all 3 -- add a space, see if they fail, and then handle the spaces

Although it is tricky because this logic is somewhat hidden in a whole bunch of if statements

---

It is definitely annoying that we are using 2 different regex dialects in Oils -- Python compiled to re2c, and POSIX ERE ...  I would like it to be all one thing.

The reason I suggested the latter is because of the submatch capture, which does seem to make things easier ... in theory we can add this to our re2c stuff, which may or  may not be worth it right now


(In some future world we would use eggex for everything, although that introduces a circular dependency!)

andychu (commented):
Thanks for doing this!

andychu:
Side note: another thing that is annoying that you may have run into is that  you can't do

```
MY_RE = '^' + pattern + '$'
```

at the top level because of a mycpp limitation

You can only do that in `frontend/consts_gen.py`, and have our tools eagerly evaluate it at compile time, and then bake in the regex!  It is a bit convoluted



---

In theory mycpp could be a "real language", but for now we will have to work around these issues



---

But actually it is not without reason, because we don't want to do any work at startup in `oils-for-unix`.  We are definitely trying to REDUCE startup time (it is higher than other shells now)

osh/sh_expr_eval.py
PossiblyAShrub: Good catch. `\s` is not (strictly) POSIX! But we *have* a spec test which covers the `$(( " 42 " ))` case: https://ci.oilshell.org/uuu/github-jobs/7841/ovm-tarball.wwz/_tmp/spec/osh-py/whitespace.html.

Somewhat annoyingly, this bug would have slipped past our testing because glibc (and musl libc) supports the `\s` character class shorthand with `REG_EXTENDED`... (From glibc's perspective, this is still technically POSIX compliant: "[the] interpretation of an ordinary character preceded by an unescaped <backslash> ( '\\' ) is undefined..." [1])

<details>
<summary>
Test C program validating this and links to libc sources where <code>\s</code> is parsed.
</summary>

- glibc: https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/regcomp.c;h=5380d3c7b9d5139d544f06bbaaba717dc36e14e7;hb=HEAD#l1879
- musl: https://git.musl-libc.org/cgit/musl/tree/src/regex/regcomp.c#n416

```c
#include <regex.h>
#include <stdio.h>

int main() {
  regex_t pat;
  int error;
  const int errbufz = 128;
  char errbuf[errbufz];

  int cflags = REG_EXTENDED;
  int eflags = 0;

  const char *regex = "\\s+";
  const char *test = "foo \t test";

  error = regcomp(&pat, regex, cflags);
  if (error) {
    goto error;
  }

  regmatch_t match;
  error = regexec(&pat, test, 1, &match, eflags);
  if (error) {
    goto error;
  }

  printf("Found match %d:%d\n", match.rm_so, match.rm_eo);

  regfree(&pat);
  return 0;

 error:
  regerror(error, &pat, errbuf, errbufz);
  fprintf(stderr, "Error: %s\n", errbuf);
  regfree(&pat);
  return 1;
}
```
</details>

Do we test OSH with musl or other libcs? I wonder if we have any other (POSIX) portability issues.

Anyways, I like the idea that we should unit test this. Currently we are missing coverage on `$(( " 0xFF " ))` and `$(( 2#0110 ))`. I'll also update the regexs so they are POSIX compliant.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_05

andychu: Oh I see, wow, I totally did not know that!  

I just know that I never use `\s` with `egrep`, which uses the libc syntax

But now I noticed it works ... I wonder when that happened ...

---

Yeah so I guess even unit testing doesn't really catch it, because our unit tests run with GNU libc -- well at least right now they do

I think the place where it might break is OS X and BSD !   There was someone interested in setting up OpenBSD CI on sourcehut, but we don't have that unfortunately ...

PossiblyAShrub:

andychu:
Looks great, thanks for this thorough fix!

PR #2069, opened 2024-09-06
[builtin/trap] Reset condition on unsigned-integer

jasom:
The POSIX specification requires this.

I also added a spec test.

Fixes 2055

### DIFF
--- a/builtin/trap_osh.py
+++ b/builtin/trap_osh.py
@@ -19,6 +19,7 @@ from frontend import signal_def
 from frontend import reader
 from mycpp import mylib
 from mycpp.mylib import iteritems, print_stderr
+from mycpp import mops
 
 from typing import Dict, List, Optional, TYPE_CHECKING
 if TYPE_CHECKING:
@@ -152,6 +153,15 @@ class TrapState(object):
         return len(self.traps) != 0 or len(self.hooks) != 0
 
 
+def _IsUnsignedInteger(s):
+    # type: (str) -> bool
+
+    try:
+        intval = mops.FromStr(s)
+    except ValueError:
+        return False
+    return not mops.Greater(mops.ZERO, intval)
+
 def _GetSignalNumber(sig_spec):
     # type: (str) -> int
 
@@ -262,7 +272,10 @@ class Trap(vm._Builtin):
             return 1
 
         # NOTE: sig_spec isn't validated when removing handlers.
-        if code_str == '-':
+        # Per POSIX, if the first argument to trap is an unsigned integer
+        # then reset every condition
+        # https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/V3_chap02.html#tag_18_28
+        if code_str == '-' or _IsUnsignedInteger(code_str):
             if sig_key in _HOOK_NAMES:
                 self.trap_state.RemoveUserHook(sig_key)
                 return 0
--- a/spec/builtin-trap.test.sh
+++ b/spec/builtin-trap.test.sh
@@ -288,3 +288,14 @@ on exit
 status=0
 ## END
 
+#### Remove trap with an unsigned integer
+
+trap 'echo noprint' EXIT
+trap 1 EXIT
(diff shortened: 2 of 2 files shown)


builtin/trap_osh.py
```diff
@@ -262,7 +272,7 @@ def Run(self, cmd_val):
             return 1
 
         # NOTE: sig_spec isn't validated when removing handlers.
-        if code_str == '-':
+        if code_str == '-' or _IsUnsignedInteger(code_str):
```
andychu, line 275: Can you add a link to the POSIX doc from the bug?  So readers understand why it is here

spec/builtin-trap.test.sh
```diff
@@ -288,3 +288,14 @@ on exit
 status=0
 ## END
 
+#### Remove trap with an unsigned integer
+
+trap 'echo noprint' EXIT
+trap 1 EXIT
+echo printed
```
andychu, line 295: One thing I realized is that we're not respecting multiple traps - `trap 'echo hi' INT EXIT`

But that's orthogonal to this change, so we can leave it out for now



```
$ bash -c 'trap "echo hi" EXIT INT; trap -p; echo ; trap "echo z" INT; trap -p; trap 1 EXIT INT; trap -p'
trap -- 'echo hi' EXIT
trap -- 'echo hi' SIGINT

trap -- 'echo hi' EXIT
trap -- 'echo z' SIGINT
```

builtin/trap_osh.py
```diff
@@ -152,6 +152,16 @@ def ThisProcessHasTraps(self):
         return len(self.traps) != 0 or len(self.hooks) != 0
 
 
+def _IsUnsignedInteger(s):
+    # type: (str) -> bool
+
+    try:
+        if int(s) >= 0:
```
andychu, line 159: Minor thing - for user facing strings, we use `mops.FromStr(s)` -- this is because when mycpp translates it, `int()` operates on C `int`, while `mops.FromStr(s)` is the 64-bit type, which may become big integers


And I would perhaps wrap that in its own `try/except` so it's clear where ValueError is raised

andychu (commented):
Great, thanks for doing this!

builtin/trap_osh.py
jasom: Done

builtin/trap_osh.py
jasom: I *think* I got this right; I'm not at all familiar with mops, but I made some guesses until the tests passed again.

spec/builtin-trap.test.sh
jasom: Yeah, I noticed that too when linking to the spec; should I open an issue (and/or make a failing spec test) or are you going to?

builtin/trap_osh.py
```diff
@@ -152,6 +153,15 @@ def ThisProcessHasTraps(self):
...
 
 
+def _IsUnsignedInteger(s):
+    # type: (str) -> bool
+
+    try:
+        intval = mops.FromStr(s)
+    except ValueError:
+        return False
+    return not mops.Greater(mops.ZERO, intval)
```
andychu, line 163: I think this would need `mops.Equal(mops.ZERO) or` ...  for `trap 0 INT`

Or probably a better idea is to do something like

```
util.RegexSearch('^[0-9]+$')
```

we could also use something like `match.LooksLikeInteger()`, without the negative number check, but that is a more fiddly change

spec/builtin-trap.test.sh
andychu: A failing spec test would be appreciated!

and then update `oils_failures_allowed:` at the top of the file

builtin/trap_osh.py
jasom: The Regex search is probably better, but I believe the current version is correct because I use NOT(GREATER(x,y)) which is the same as GREATER_OR_EQUAL(y,x).

PR #2070, opened 2024-09-10
Add failing arith-dynamic spec tests

PossiblyAShrub:
These cases were extracted from the discussion on Zulip

(#oil-dev>ArithSub evaluation bug in autotools configure)

### DIFF
--- a/spec/arith-dynamic.test.sh
+++ b/spec/arith-dynamic.test.sh
@@ -1,0 +1,95 @@
+## compare_shells: bash dash mksh zsh
+## oils_failures_allowed: 3
+
+# Various tests for dynamic parsing of arithmetic substitutions.
+
+#### Double quotes
+echo $(( "1 + 2" * 3 ))
+echo $(( "1+2" * 3 ))
+## STDOUT:
+7
+7
+## END
+
+## N-I dash status: 2
+## N-I dash STDOUT:
+## END
+
+## N-I mksh status: 1
+## N-I mksh STDOUT:
+## END
+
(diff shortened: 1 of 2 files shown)


andychu:
Looks good!  For some reason the CI failed though 

http://op.oilshell.org/uuu/github-jobs/7854/ovm-tarball.wwz/_tmp/soil/logs/osh-spec.txt

```
done: all-tests-to-html
Error reading line from file: _tmp/spec/osh-py/arith-dynamic.stats.txt
```


----

Oh probably because there needs to be a stub in `test/spec.sh` ?  (I have been meaning to get rid of that file somehow, though it does provide auto-completion for each "task")

andychu:
Oh unrelated but I also remembered we want to add the same behavior to `replace()` that we did with `split()` -- there is that potential infinite loop now

PossiblyAShrub:
> Oh probably because there needs to be a stub in `test/spec.sh` ? (I have been meaning to get rid of that file somehow, though it does provide auto-completion for each "task")

Yep, I think that was it.

re: removing this file, it would be nice if you could do `test/spec-py.sh run-file spec/arith-dynamic.test.sh`. Then you could rely on the file-system for auto-completion. Right now that errors with `No such file or directory: 'spec/spec/arith-dynamic.test.sh.test.sh'`.

andychu:
Yes true, we should just use file system completion!

Originally all those functions had a list of shells and `--failures-allowed` , but now that's in the file itself

--

Thanks for the spec tests!

PR #2071, opened 2024-09-11
Fix issues in `Str => replace(eggex, *)`

PossiblyAShrub:
This PR fixes the following issues:
- Infinite loop if we have a zero-width match
- Crash if the string contains a NUL byte

### DIFF
--- a/doc/ref/chap-type-method.md
+++ b/doc/ref/chap-type-method.md
@@ -143,6 +143,13 @@ The following matrix of signatures are supported by `replace()`:
     s => replace(eggex_val, subst_str)
     s => replace(eggex_val, subst_expr)
 
+Replacing by an `Eggex` has some limitations:
+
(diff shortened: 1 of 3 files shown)


doc/ref/chap-type-method.md
```diff
@@ -143,6 +143,13 @@ The following matrix of signatures are supported by `replace()`:
     s => replace(eggex_val, subst_str)
     s => replace(eggex_val, subst_expr)
 
+Replacing by an `Eggex` has some limitations:
+
+- If a `search()` results in an empty string match, eg.
+  `'abc'.split(/ space* /)`, then we raise an error to avoid an infinite loop.
+- The string to replace on cannot contain NUL bytes because we use the libc
```
andychu, line 150: I wonder if our binding is the issue ... in the future maybe we can relax this

andychu (commented):
Looks great, thank you!

PR #2075, opened 2024-09-15
Correct misleading comment in `_build/oils.sh`

meator:
The header of this shell script states that the compiler can be either `cxx` or `clang` (but it can be anything) and it states that the default `VARIANT` is `'dbg'`, which is [disproved few lines below](https://github.com/oils-for-unix/oils/blob/master/build/ninja_main.py#L133):

```sh
  local variant=${2:-opt}    # default is optimized build
```


andychu:
Great, thank you!

PR #2076, opened 2024-09-16
add mycpp-souffle translator

melvinw:
This commit adds a new translator called `mycpp-souffle`. It translates
files using mycpp with optimizations enabled.

You can build the translation examples with souffle optimizations
enabled with the same targets you normally would, but with `mycpp`
replaced by `mycpp-souffle` in the suffix. For example, you build the
intermediate C++ and the `cxx-asan` binary for
`mycpp/examples/gc_stack_roots.py` respectively by running the following
commands:
```
ninja _gen/mycpp/examples/gc_stack_roots.mycpp-souffle.cc
ninja _bin/cxx-asan/mycpp/examples/gc_stack_roots.mycpp-souffle
```

You can also build OSH and YSH with this new variant. Their targets are
named `_bin/<variant>/mycpp-souffle/osh` and
`_bin/<variant>/mycpp-soufle/ysh` respectively. Currently, only the
`cxx-opt`, `cxx-asan`, and `cxx-asan+gcalways` variants are enabled for
the new translator. The binaries translated with vanilla mycpp are still
available at their existing paths, e.g. `_bin/cxx-dbg/osh`.

The translated C++ for the entire program with optimizations is named
`_gen/bin/oils_for_unix.mycpp-souffle.cc`.

### DIFF
--- a/benchmarks/osh-parser.sh
+++ b/benchmarks/osh-parser.sh
@@ -60,12 +60,19 @@ parser-task() {
   local times_out="$out_dir/$host.$job_id.times.csv"
 
   local shell_name
-  shell_name=$(basename $sh_path)
+  case $sh_path in
+    _bin/*/mycpp-souffle/*)
+      shell_name=osh-native-souffle
+      ;;
+    *)
+      shell_name=$(basename $sh_path)
+      ;;
+  esac
 
   # Can't use array because of set -u bug!!!  Only fixed in bash 4.4.
   extra_args=''
   case "$shell_name" in
-    osh|oils-for-unix.*)
+    osh*|oils-for-unix.*)
       extra_args='--ast-format none'
       ;;
   esac
@@ -106,7 +113,14 @@ cachegrind-task() {
   mkdir -p $out_dir/$cachegrind_out_dir
 
   local shell_name
-  shell_name=$(basename $sh_path)
+  case $sh_path in
+    _bin/*/mycpp-souffle/*)
+      shell_name=osh-native-souffle
+      ;;
+    *)
+      shell_name=$(basename $sh_path)
+      ;;
+  esac
 
   local script_name
   script_name=$(basename $script_path)
@@ -117,7 +131,7 @@ cachegrind-task() {
   # Can't use array because of set -u bug!!!  Only fixed in bash 4.4.
   extra_args=''
   case "$shell_name" in
-    osh|oils-for-unix.*)
+    osh*|oils-for-unix.*)
       extra_args="--ast-format none"
       ;;
   esac
@@ -202,7 +216,8 @@ measure() {
   local provenance=$1
   local host_job_id=$2
   local out_dir=${3:-$BASE_DIR/raw}
-  local osh_cpp=${4:-$OSH_CPP_BENCHMARK_DATA}
+  shift 3
+  local -a osh_cpp=( "${@:-$OSH_CPP_BENCHMARK_DATA}" )
 
   local times_out="$out_dir/$host_job_id.times.csv"
   local lines_out="$out_dir/$host_job_id.lines.csv"
@@ -222,7 +237,7 @@ measure() {
     > $times_out
 
   local tasks=$BASE_DIR/tasks.txt
-  print-tasks $provenance "${SHELLS[@]}" $osh_cpp > $tasks
+  print-tasks $provenance "${SHELLS[@]}" "${osh_cpp[@]}" > $tasks
 
   # Run them all
   cat $tasks | xargs -n $NUM_TASK_COLS -- $0 parser-task $out_dir
@@ -232,7 +247,8 @@ measure-cachegrind() {
   local provenance=$1
   local host_job_id=$2
   local out_dir=${3:-$BASE_DIR/raw}
-  local osh_cpp=${4:-$OSH_CPP_BENCHMARK_DATA}
+  shift 3
+  local -a osh_cpp=( "${@:-$OSH_CPP_BENCHMARK_DATA}" )
 
   local cachegrind_tsv="$out_dir/$host_job_id.cachegrind.tsv"
   local lines_out="$out_dir/$host_job_id.lines.tsv"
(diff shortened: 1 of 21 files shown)


andychu:
Hm looks like an R failure

```
*<c2><a0>44, 45
*<c2><a0>35, 36
*<c2><a0>50, 51
*<c2><a0>53, 54
*<c2><a0>17, 18
Backtrace:
    x
 1. +-global main(commandArgs(TRUE))
 2. | \-global MyCppReport(in_dir, out_dir)
 3. |   \-... %>% arrange(`C++ : Python`)
 4. +-dplyr::arrange(., `C++ : Python`)
 5. +-dplyr::mutate(., `C++ : Python` = `C++`/Python)
 6. +-tidyr::spread(., key = impl, value = user_ms)
 7. \-tidyr:::spread.data.frame(., key = impl, value = user_ms)
 8.   \-cli::cli_abort(...)
 9.     \-rlang::abort(...)
Execution halted
```


Unfortunately the scripts don't have good error messages

I suspect probably some file or row was generated, or two files mismatched because the output file wasn't created ?

Feel free to DM me if you are having build system issues - this stuff is very fiddly

the "pea" stuff was only partially done I think

melvinw:
Ah, yes. I noticed that the other day. Spending some time on this tonight. Hopefully it's not too hairy

benchmarks/osh-parser.sh
```diff
@@ -60,12 +60,19 @@ parser-task() {
   local times_out="$out_dir/$host.$job_id.times.csv"
 
   local shell_name
-  shell_name=$(basename $sh_path)
+  case $sh_path in
+    _bin/*/mycpp-souffle/*)
+      shell_name=osh-native-souffle
```
andychu, line 65: Hm the relationship between sh_path and shell_name is a smell I remember juggling ... 

I think there is a place in R where we have a heuristic for `osh-native` as well

hm

benchmarks/osh-parser.sh
```diff
@@ -60,12 +60,19 @@ parser-task() {
...
+    *)
+      shell_name=$(basename $sh_path)
+      ;;
+  esac
 
   # Can't use array because of set -u bug!!!  Only fixed in bash 4.4.
   extra_args=''
   case "$shell_name" in
-    osh|oils-for-unix.*)
+    osh*|oils-for-unix.*)
```
andychu, line 75: OK I see the * is for `osh-native-souffle`

benchmarks/report.R
```diff
@@ -65,6 +65,8 @@ GetOshLabel = function(shell_hash, prov_dir) {
       label = 'osh-ovm'
     } else if (length(grep('bin/osh', lines)) > 0) {
       label = 'osh-cpython'
+    } else if (length(grep('_bin/.*/mycpp-souffle/osh', lines)) > 0) {
```
andychu, line 68: OK yeah I see here is the heuristic I was thinking of

I think there is some existing smell where we should clarify the relationship between shell output and R output

it is a bit fuzzy now

andychu (commented):
OK wow I didn't realize you were going through all the benchmarks!   that's cool that it all works, and there is some good movement on many of them

Seems like the "CPU-bound" ones, e.g. in benchmarks/compute

I do remember the smell about sh_path and shell_name .... let me think of if there is a good way to improve that

Some of the benchmarks have definitely gotten a bit messy, and they are not entirely consistent with each other

I am also going to summarize a few things on Zulip

build/ninja_main.py
```diff
@@ -129,16 +130,23 @@ def ShellFunctions(cc_sources, f, argv0):
 main() {
   ### Compile oils-for-unix into _bin/$compiler-$variant-sh/ (not with ninja)
 
-  local compiler=${1:-cxx}   # default is system compiler
-  local variant=${2:-opt}    # default is optimized build
-  local skip_rebuild=${3:-}  # if the output exists, skip build'
+  local compiler=${1:-cxx}        # default is system compiler
+  local variant=${2:-opt}         # default is optimized build
+  local translator=${3:-mycpp}    # default is the translator w/o optimizations
```
andychu, line 135: the Usage string at the top of the generated file should be updated with the TRANSLATOR? argument

and note that it can be mycpp or mycpp-souffle

It seems like this will also help if we ever do a migration off MyPy

andychu:
BTW I would just fix the Usage string

I will deal with the `sh_path / shell_name` stuff separately -- that is a smell that was already there

It's cool that this is working!  what do we want to do before making it the default?  

(Julian Brown has stress tested our C++ build pretty effectively!   The bugs were more like "oversights" than fundamental design problems )

melvinw:
Sounds good! I just pushed the usage fix.

I think we can make `mycpp-souffle` the default translator when at least the following conditions are satisfied:
1. We hook `mycpp-souffle` up to the `spec-cpp` and `interactive` jobs
2. We try out some other optimizations to make sure that this framework works well for things other than just the stack roots solver.

mycpp/pass_state.py
```diff
@@ -568,6 +568,9 @@ def ComputeMinimalStackRoots(cfgs: dict[str, ControlFlowGraph],
     that can be queried by cppgen_pass.
     """
     DumpControlFlowGraphs(cfgs, facts_dir=facts_dir)
+    # The facts files can be pretty large. Sync them first to avoid reading
+    # truncated files from the solver.
+    subprocess.run('sync {}/*.facts'.format(facts_dir), shell=True)
```
andychu, line 573: hm this seems weird

I don't think there is any situation where after calling `f.close()` you can't read the right thing back in a different process

that would be either  a Python bug or a kernel bug

andychu (commented):
I will merge this for now but I think the latest change is weird

mycpp/pass_state.py
melvinw: Oh you know what. I bet this is actually a case of parallel instances of mycpp-souffle clobbering each other's output. I was seeing this failure in one of the CI jobs that builds cxx-opt and cxx-asan? This could be addressed by giving each one a unique tempdir for the fact outputs. I'll confirm that and tidy this up later

andychu:
If you build two variants like `cxx-opt` and `cxx-asan`, the `build.ninja` should only create one copy of the original file

It shouldn't run mycpp twice

Unless there is a dependency graph bug (which is not hard to introduce)

melvinw:
Oh good point. I'll spend some time figuring out what else could be going on here

PR #2077, opened 2024-09-18
Define procs in the current scope, not the global one

PossiblyAShrub:
Experiment for [#language-design>Metaprogramming in ysh, sort of](https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Metaprogramming.20in.20ysh.2C.20sort.20of) on Zulip.

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -2390,7 +2390,7 @@ class Procs:
 
     def SetProc(self, name, proc):
         # type: (str, value.Proc) -> None
-        self.mem.var_stack[0][name] = Cell(False, False, False, proc)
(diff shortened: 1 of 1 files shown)


andychu:
Thanks, I think this makes sense as a language change

How about add a spec test for now?  

(We also need to document this somewhere, although it might be OK to work out Hay and all that at the same time)

andychu:
I'm working in this area now, so I merged

I think it will affect some of the spec tests I'm working on, so I'll take care of that

https://oilshell.zulipchat.com/#narrow/stream/384942-language-design/topic/Hay.2C.20ctx.2C.20eval.20APIs.2C.20value.2EObj.2C.20modules.20are.20all.20related

PR #2079, opened 2024-09-25
use tempdir for souffle facts and outputs

melvinw:
This should keep us from crashing when two instances of mycpp-souffle are running at the same time. This a a stopgap of sorts until the we figure out the build dependency issue that's causing this problem.


mycpp/pass_state.py
```diff
@@ -538,8 +544,16 @@ def DumpControlFlowGraphs(cfgs: dict[str, ControlFlowGraph],
...
-    fact_files = {}
+
     os.makedirs(facts_dir, exist_ok=True)
+    # Open files for all facts that we might emit even if we don't end up having
+    # anything to write to them. Souffle will complain if it can't find the file
+    # for anything marked as an input.
+    fact_files = {
+        fact_type.name():
+        open('{}/{}.facts'.format(facts_dir, fact_type.name()), 'w')
+        for fact_type in Fact.__subclasses__()
```
andychu, line 555: This feels like a lot of indirection for what is just a bunch of strings ... I think it could be just a list of strings:

```
fact_subdirs = ['assign', 'bind', ...]
```

And then I think you could write `assert len(fact_subdirs) == Fact.__subclasses__()` if really necessary

I try to avoid reflection like `__subclasses__` in the main code -- since those idioms don't exist in say C++

mycpp/mycpp_main.py
```diff
@@ -370,7 +371,12 @@ def main(argv):
     log('\tmycpp pass: DATAFLOW')
     stack_roots = None
     if opts.minimize_stack_roots:
-        stack_roots = pass_state.ComputeMinimalStackRoots(cfgs)
+        souffle_dir = tempfile.TemporaryDirectory()
```
andychu, line 374: I think it could be useful to view the temp dir when debugging, rather than automatically cleaning it up

Not sure if that's feasible though

In any case, It would be nice to have a comment showing what the directory layout is

melvinw: I can add an environment var (an option would be cumbersome since it would require adding an extra translator) to let you specify the output dir when debugging

melvinw: Will also add a comment about the directory layout

mycpp/pass_state.py
melvinw: I dunno. There's zero risk of this file getting translated to C++, right? I think it would be much more (relatively speaking lol) annoying to have to update this list every time we add a new fact. If we'd be using `__subclasses__()` in the assert to remind ourselves to come update the list, why not just use it directly?

mycpp/mycpp_main.py
melvinw: Added `MYCPP_SOUFFLE_DIR`. Let me know what you think

mycpp/pass_state.py
andychu: Well I still think we should move mycpp to use our own ASDL  schema, "collapsing" the MyPy AST a bit

In that case, it won't be that far from C++

And mycpp is too slow, so there is some motivation for that :)

PR #2086, opened 2024-10-01
Allow nested procs/funcs

PossiblyAShrub:
We chatted about this on Zulip. Since procs/funcs are defined in the local scope, they no longer have the strange dynamic scoping behaviour which led to this being an error anyways.


andychu (commented):
Looks great, thanks for doing this!  

I think this will let us change the counter example to `counter_proc = makeCounter()`, which is a slightly less awkward dance

I will do some examples at some point, maybe with `__call__` 

It's not as syntactically short as JS closures, but at least it will now be possible to do closure-like things if you need them, and we can always add things based on feedback/usage

PR #2087, opened 2024-10-02
[WIP] Replace ctx with a pure YSH definition

PossiblyAShrub:


Melkor333:
I just pushed my changes required to make hay in ysh work... :)
the `get` method might not be necessary, but was useful for my pure ysh `hay` implementation.

andychu:
Hm looks interesting!  Glad you were able to use `evalInFrame()`

Is there a demo of Hay in YSH? 

Another thing to try would be to replace the `stdlib/ysh/args.ysh`  ctx with this one

At least try and post a message to let us know what breaks :)  If it works that will be miraculous

PR #2090, opened 2024-10-05
[builtin] Add pexpect and spec tests for bind built-in

KingMob:
Hi Andy, here's the initial PR for tests of `bind`, as we discussed in Zulip [here](https://oilshell.zulipchat.com/#narrow/stream/121539-oil-dev/topic/bind.20built-in.20overview) and [here](https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/GNU.20readline.20.2F.20bind.20builtin).

The sole non-interactive test is in `builtin-bind.test.sh`. (Essentially just `bind -l`).

Everything else in `spec/stateful/bind.py`. It tests several important `bind` flags, including -m, -x, -q, -r, and -u. I think that will cover most _programmatic_ uses of bind in scripts (but not as much user-facing interactive aspects like .inputrc, or listing out bindings). It could definitely be expanded.

I've verified that bash passes all the tests in my VM and osh fails them all.

I played with `$SH -i -c`, but couldn't get around some weird outputs I kept seeing, so I went with pexpect.


andychu:
Cool, thanks!    Our CI creates HTML reports here:

https://op.oilshell.org/uuu/github-jobs/7983/

So yeah this looks good, it passes with bash but fails with OSH

https://op.oilshell.org/uuu/github-jobs/7983/interactive.wwz/_tmp/spec/stateful/index.html

---

I think we have to add a flag like `--oils-failures-allowed` to make it "pass", and then we can "clamp down" on that when more tests pass


Also there is a test that is erroneously failing on Python 3, let me fix that

andychu:
OK I made a couple trivial changes, and CI passes

Thanks this is great!  Keep us updated / ask questions on Zulip :-)  I still only have a fuzzy idea of how `bind` works

KingMob:
Great! 

I'll dive into the Python/readline aspects next. No idea how long that'll take.

PR #2098, opened 2024-10-20
Flag changes required by `zoxide` when sourcing other shell’s startup files

nisbet-hubbard:
https://github.com/ajeetdsouza/zoxide#installation

As other similar adjustments may be required depending on the software one has installed under previous shell, it can be helpful to collect compatible configurations in a file under `/etc/profile.d` and source it from `oshrc` and keep shell-dependent ones such as `zoxide`’s in shell-specific files like `.bashrc` and `oshrc`.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -52,7 +52,7 @@ It's very hard to tell when and if `/etc/profile`, `~/.bashrc`,
 `~/.bash_profile`, etc. are executed.
 
 OSH and YSH intentionally avoid this.  If you want those files, simply `source`
-them in your `oshrc` (see [OSH Compatibility Tips](https://github.com/oils-for-unix/oils/wiki/OSH-Compatibility-Tips) to resolve possible incompatibilities).
+them in your `oshrc`.
 
 [mess]: https://shreevatsa.wordpress.com/2008/03/30/zshbash-startup-files-loading-order-bashrc-zshrc-etc/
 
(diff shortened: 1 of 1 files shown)


andychu:
Thanks for getting this to work and documenting it! 

I think we may want to have a sentence like:

For compatibility with programs like `zoxide`, see the wiki:

https://github.com/oils-for-unix/oils/wiki/Shell-Programs-That-Run-Under-OSH

or maybe we make a new page "OSH Compatibility Tips" 

Also, I didn't know what zoxide was, so it would be useful to include a link

I think we should put "Starship" on the wiki too.    That one seems reasonably popular

doc/getting-started.md
```diff
@@ -52,7 +52,7 @@ It's very hard to tell when and if `/etc/profile`, `~/.bashrc`,
 `~/.bash_profile`, etc. are executed.
 
 OSH and YSH intentionally avoid this.  If you want those files, simply `source`
-them in your `oshrc`.
+them in your `oshrc` (see [OSH Compatibility Tips](https://github.com/oils-for-unix/oils/wiki/OSH-Compatibility-Tips) to resolve possible incompatibilities).
```
andychu, line 55: Let's leave this change out, since the page is linked a few lines below

doc/getting-started.md
```diff
@@ -73,6 +73,7 @@ OSH](https://github.com/oilshell/oil/wiki/How-To-Test-OSH).
 - On Arch Linux and other distros,`$LANG` may not get set without
   `/etc/profile`.  Adding `source /etc/profile` to your `oshrc` may solve this
   problem.
+- See [OSH Compatibility Tips](https://github.com/oils-for-unix/oils/wiki/OSH-Compatibility-Tips) to configure programs that rely on `eval` to initialize (e.g., starship, zoxide).
```
andychu, line 76: no comma after e.g.

andychu (commented):
Thanks!

PR #2099, opened 2024-10-20
Extend args.ysh's flag proc with Float, Str support

quexxon:
Additionally, now that basic support for type objects has been added (ref: 9db992cc), refactor to use type objects for flag types. This provides a cleaner, more strict interface when describing flags (e.g., `flag -c --count (Int)`).

### DIFF
--- a/stdlib/ysh/args.ysh
+++ b/stdlib/ysh/args.ysh
@@ -10,11 +10,11 @@ const __provide__ = :| parser flag arg rest parseArgs |
 # parser (&spec) {
 #   flag -v --verbose (help="Verbosely")  # default is Bool, false
 #
-#   flag -P --max-procs ('int', default=-1, doc='''
+#   flag -P --max-procs (Int, default=-1, doc='''
 #     Run at most P processes at a time
 #     ''')
 #
-#   flag -i --invert ('bool', default=true, doc='''
+#   flag -i --invert (Bool, default=true, doc='''
 #     Long multiline
 #     Description
 #     ''')
@@ -33,7 +33,6 @@ const __provide__ = :| parser flag arg rest parseArgs |
 # TODO: See list
 # - It would be nice to keep `flag` and `arg` private, injecting them into the
 #   proc namespace only within `Args`
-# - We need "type object" to replace the strings 'int', 'bool', etc.
 # - flag builtin:
 #   - handle only long flag or only short flag
 #   - flag aliases
@@ -45,7 +44,7 @@ proc parser (; place ; ; block_def) {
   ##
   ##   # NOTE: &spec will create a variable named spec
   ##   parser (&spec) {
-  ##     flag -v --verbose ('bool')
+  ##     flag -v --verbose (Bool)
   ##   }
   ##
   ##   var args = parseArgs(spec, ARGV)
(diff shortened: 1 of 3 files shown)


stdlib/ysh/args.ysh
```diff
@@ -80,26 +79,46 @@ proc parser (; place ; ; block_def) {
...
 
-  # TODO: validate `type`
+  # Bool has a default of false, not null
+  if (type is Bool and default === null) {
+    setvar default = false
+  }
 
-  # TODO: Should use "trimPrefix"
-  var name = long[2:]
+  var name = long => trimStart('--')
```
andychu, line 121: FYI I changed "transforming" method calls to be `long.trimStart('--')` recently

(And mutating method calls are `call mylist->append(42)`)

But `=>` is still allowed for method calls ... 

in addition to being the "function chain" operator

I'm interested in feedback on that, e.g. whether we allow

```
long => trimStart('--') => split('x') => join()  
```

where split() and join() are free functions .  Or we only allow



```
long.trimStart('--') => split('x') => join()  
```

which seems OK ... then there is "one way to write it"

stdlib/ysh/args.ysh
```diff
@@ -80,26 +79,46 @@ proc parser (; place ; ; block_def) {
...
 
-proc flag (short, long ; type='bool' ; default=null, help=null) {
+const kValidTypes = [Bool, Float, Int, Str]
+const kValidTypeNames = []
+for vt in (kValidTypes) {
+  call kValidTypeNames->append(vt.name)
+}
+
+func isValidType (type) {
+  try {
```
andychu, line 89: I think the  `try` is unnecessary here?


---

Actually I think this is another interesting argument for defining `obj1 === obj2` as identity

because then we can replace the entire `isValidType()` function with `mytype in kValidTypes`

stdlib/ysh/args.ysh
```diff
@@ -80,26 +79,46 @@ proc parser (; place ; ; block_def) {
...
+  # Bool has a default of false, not null
+  if (type is Bool and default === null) {
+    setvar default = false
+  }
 
-  # TODO: Should use "trimPrefix"
-  var name = long[2:]
+  var name = long => trimStart('--')
 
   ctx emit flags ({short, long, name, type, default, help})
```
andychu, line 123: Actually I also realized it is possible for us to store `type.name` in the "spec", so we have a string rather than a type object ... though maybe that doesn't actually make anything easier

andychu (commented):
This looks excellent!  Thank you

The main comment is about a `try { }`, and there are some other fuzzy questions/thoughts

stdlib/ysh/args.ysh
quexxon: The `try` is necessary because `is` raises an error on mismatched types:
```
ysh-0.23.0$ = 'int' is Int
  = 'int' is Int
          ^~
[ interactive ]:1: fatal: Mismatched types
```
So if a caller uses a non-type-object value for the type argument to `flag`, they will get a confusing error.

This behavior was actually a little surprising to me. I think there are more circumstances when I would knowingly want to use `is` with potentially mismatched types (e.g., validating input parameters) than circumstances when I would be glad that it caught me accidentally using the wrong type. In other words, I would have expected `false` rather than an error for a mismatched type.

andychu: Hmmm yeah I think we should remove that check!   At least, it is inconsistent with Python

And JavaScript it seems

stdlib/ysh/args.ysh
andychu: Great, thanks for the feedback!

andychu:
Merged, thank you!   

Great feedback on the language too -- I have been mentioning to Aidan that we are "testing" the language by writing the stdlib, and this is a great example of that

PR #2101, opened 2024-10-22
[ysh] Lex '&&', '||' for better errors in expr mode

PossiblyAShrub:
Lex '&&', '||' in expression mode for better error messages:

```
  = a && b
      ^~
[ -c flag ]:1: Use 'and' in expression mode

  = a || b
      ^~
[ -c flag ]:1: Use 'or' in expression mode
```

Eggex has `Expr_Bang` which means we cannot check this case the same way:

```
  = !a
    ^
[ -c flag ]:1: Syntax error in expression (near Id.Expr_Bang)
```

(I don't think it's possible to cleanly check this specific error unless we moved away from pgen2.)

### DIFF
--- a/doc/error-catalog.md
+++ b/doc/error-catalog.md
@@ -179,29 +179,33 @@ Examples:
 Incorrect:
 
     # Expression mode
-    if (a || b) {
-      echo yes
+    if (!a || b && c) {
+      echo no
     }
 
     # Command mode
-    if test --dir a or test --dir b { ... }
+    if not test --dir a or test --dir b and test --dir c {
+      echo no
+    }
 
 Correct:
 
     # Expression mode
-    if (a or b) {
+    if (not a or b and c) {
       echo yes
     }
 
     # Command mode
-    if test --dir a || test --dir b { ... }
+    if ! test --dir a || test --dir b && test --dir c {
+      echo yes
+    }
 
 In general, code within parentheses `()` is parsed as Python-like expressions
 -- referred to as [expression mode](command-vs-expression-mode.html). The
 standard boolean operators are written as `a and b`, `a or b` and `not a`.
 
 This differs from [command mode](command-vs-expression-mode.html) which uses
-`||` for "OR", `&&` for "AND" and `!` for "NOT".
+shell-like `||` for "OR", `&&` for "AND" and `!` for "NOT".
 
 ## Runtime Errors - Traditional Shell
 
--- a/doc/error-catalog.md
+++ b/doc/error-catalog.md
@@ -174,6 +174,35 @@ Examples:
       echo yes
     }
 
+### OILS-ERR-15
+
+Incorrect:
+
+    # Expression mode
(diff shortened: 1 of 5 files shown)


andychu:
OK looks great! 

And yeah too bad about `!` ... we can probably check it in arithmetic, but it would have to use a different strategy, not this "over-lexing" one.  It would have to be aware of parsing

How about adding a  OILS-ERR- code in `doc/error-catalog.md` ?  Because not everyone will know what "expression mode" is

I think they could use a format like:

```
No:
   if (a || b) {
     echo yes
   }

Yes:
  if (a or b) {
    echo yes
  }

```

and then say that Python-like expressions generally occur within `()` 


Actually it can link to `doc/command-vs-expression-mode.html`

andychu:
I guess we can also have something like:

```
No:
  if test --dir d1 and test --dir d2 { ... }
  if (test --dir d1 and test --dir d2) { ... }

Yes:
  if test --dir d2 && test --dir d2 { ... }
```

doc/error-catalog.md
```diff
@@ -174,6 +174,35 @@ Examples:
...
+
+Incorrect:
+
+    # Expression mode
+    if (a || b) {
+      echo yes
+    }
+
+    # Command mode
+    if test --dir a or test --dir b { ... }
```
andychu, line 187: I would use the same style of 

```
if test --dir a || test --dir b { 
  echo yes
}

```

for the two examples ... just to make it consistent, and because `...` isn't literal YSH

doc/error-catalog.md
```diff
@@ -174,6 +174,35 @@ Examples:
       echo yes
     }
 
+### OILS-ERR-15
+
+Incorrect:
+
+    # Expression mode
+    if (a || b) {
```
andychu, line 182: maybe add `if (a || b && c)`

and contrast with `if (a or b and c)`

to show both operators

I guess we can also show negation

```
if (!a || b && c)

if (not a or b and c)

```

doc/error-catalog.md
```diff
@@ -174,6 +174,35 @@ Examples:
...
+
+    # Command mode
+    if test --dir a || test --dir b { ... }
+
+In general, code within parentheses `()` is parsed as Python-like expressions
+-- referred to as [expression mode](command-vs-expression-mode.html). The
+standard boolean operators are written as `a and b`, `a or b` and `not a`.
+
+This differs from [command mode](command-vs-expression-mode.html) which uses
+`||` for "OR", `&&` for "AND" and `!` for "NOT".
```
andychu, line 204: could mention that this is shell-like, rather than Python-like

I think we often say that shell <=> commands <=> procs    and Python <=> expressions <= funcs

andychu (commented):
Looks good!

PossiblyAShrub:

andychu:
Looks good!  one of the jobs failed with a Docker error, so I kicked it

PR #2102, opened 2024-10-22
[ysh] Implement 1..<5 and 1..=5 open/closed range syntax

PossiblyAShrub:
BREAKING: The `a..b` syntax has now been removed:

```
  = 1..2
     ^~
[ -c flag ]:1: Use 1..<5 for an open range, or 1..=5 for a closed range
```

Depends on https://github.com/oils-for-unix/oils/pull/2101 to prevent merge conflicts (assuming https://github.com/oils-for-unix/oils/pull/2101 will be merged before this PR.)


PossiblyAShrub:
This is ready for a review.

ysh/expr_eval.py
```diff
@@ -1260,6 +1260,10 @@ def _EvalExpr(self, node):
                 i2 = _ConvertToInt(self._EvalExpr(node.upper),
                                    'Range end should be Int', node.op)
 
+                # i2 is a big int so we don't have to worry about overflow
```
andychu, line 1263: yeah it should be a BigInt, but isn't yet.

The first step may be to raise a hard error for overflow, which is OK

Also unfortunately the ranges are still truncated in the lines below, I think we should fix that, though not sure if it should come before or after the BigInt changes mentioned

andychu (commented):
This is excellent, thanks for tracking it all down!

I was thinking if we want to pretty print the ranges separately, like

```
$ = 1 ..= 3
(Range) 1 ..= 3
```

But I think the way it is may be more transparent in a sense - people can see the equivalence between the two range syntaxes.  It is printed in one canonical form

----

So it looks good other than some tiny doc tweaks

PossiblyAShrub:

ysh/expr_eval.py
PossiblyAShrub: Okay, I've removed the comment and left the TODO below it.

Where are we at with a BigInt implementation? I remember there were a few Zulip threads a while back.

PR #2103, opened 2024-10-23
In $LIB_YSH/args.ysh, make flag/arg/rest private to parser

quexxon:
Simplify and clarify the module interface by making the `flag`, `arg`, and `rest` procs private to the `parser` DSL.


andychu:
This is great, thank you!

PR #2105, opened 2024-10-24
[stdlib/args] Support flags that remember all values passed, with List[T] syntax

quexxon:
Following the implementation of parameterized type objects in 4cbfa15f6, it is now possible to extend the set of supported flag types with List variants to provide multi-value flags.

### DIFF
--- a/stdlib/ysh/args-test.ysh
+++ b/stdlib/ysh/args-test.ysh
@@ -319,6 +319,34 @@ proc test-vs-python3-argparse {
   #assert [expected === r.stdout]
 }
 
+proc test-multi-value-flags {
+  parser (&spec) {
+    flag -f --float (List[Float])
+    flag -i --int (List[Int])
+    flag -s --str (List[Str])
+  }
+
+  var args = parseArgs(spec, :| -f 1.0 -s one -i 0 --str two --int 1 -s three |)
+
+  assert [type(args.float) === 'List']
+  assert [len(args.float) === 1]
+  assert [floatsEqual(args.float[0], 1.0)]
+
+  call args->erase('float') # remove List[Float] value for subsequent equality check
+
+  var expected = {
+    "int": [0, 1],
+    "str": :| one two three |,
+  }
+
+  assert [expected === args]
+
+  try { call parseArgs(spec, :| -f not_a_float |) }
+  assert [2 === _error.code]
+  try { call parseArgs(spec, :| -i not_an_int |) }
+  assert [2 === _error.code]
+}
+
 if is-main {
   byo-maybe-run
 }
--- a/stdlib/ysh/args.ysh
+++ b/stdlib/ysh/args.ysh
@@ -191,7 +191,7 @@ func parseArgs(spec, argv) {
               error "Expected Int after '$arg'" (code=2)
             }
 
-            setvar value = args[flag.name] if (flag.name in args) else []
+            setvar value = get(args, flag.name, [])
             try { call value->append(int(argv[i])) }
             if (_status !== 0) {
               error "Expected Int after '$arg', got '$[argv[i]]'" (code=2)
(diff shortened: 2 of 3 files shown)


stdlib/ysh/args.ysh
```diff
@@ -80,18 +80,17 @@ proc parser (; place ; ; block_def) {
   call place->setValue(p)
 }
 
-const kValidTypes = [Bool, Float, Int, Str]
+const kValidTypes = [Bool, Float, List[Float], Int, List[Int], Str, List[Str]]
 const kValidTypeNames = []
 for vt in (kValidTypes) {
-  call kValidTypeNames->append(vt.name)
+  var name = vt.name if ('name' in propView(vt)) else vt.unique_id
```
andychu, line 86: ok this is interesting ... I was wondering if they should be the same field name, but I sorta wanted to distinguish between "primitive types" and "parameterized types that are computed from primitive types"

But I guess this is only used for printing strings, so it might be OK

----

I also wonder if we should get rid of the `propView()` everywhere, by treating Obj more like Dict

BTW I also have this idea of turning `prototype()` and `propView()` into `first() rest()`.  I think those names kinda make more sense

I took the "prototype" name from JS, but a linked list naming convention  might be better

stdlib/ysh/args-test.ysh
```diff
@@ -319,6 +319,34 @@ proc test-vs-python3-argparse {
...
+proc test-multi-value-flags {
+  parser (&spec) {
+    flag -f --float (List[Float])
+    flag -i --int (List[Int])
+    flag -s --str (List[Str])
+  }
+
+  var args = parseArgs(spec, :| -f 1.0 -s one -i 0 --str two --int 1 -s three |)
+
+  assert [type(args.float) === 'List']
```
andychu, line 331: I also wonder if we should change `type(args.float)` to return a type object now, instead of a string

I think that relies on the Obj equality change we were talking about!   So both of these will work:

`type(args.float) is List`

`type(args.float) == List`

?

Something to think about

stdlib/ysh/args.ysh
```diff
@@ -195,13 +206,32 @@ func parseArgs(spec, argv) {
             if (_status !== 0) {
               error "Expected Float after '$arg', got '$[argv[i]]'" (code=2)
             }
+          } elif (flag.type is List[Float]) {
+            setvar i += 1
+            if (i >= len(argv)) {
+              error "Expected Float after '$arg'" (code=2)
+            }
+
+            setvar value = args[flag.name] if (flag.name in args) else []
```
andychu, line 215: Oh how about `get(args, flag.name, [])` everywhere?   I think that is equivalent and shorter

It is like Python's `args.get(flag.name, [])`, except that we can't use a method, so we use the free function get()

stdlib/ysh/args.ysh
```diff
@@ -195,13 +206,32 @@ func parseArgs(spec, argv) {
...
+            try { call value->append(float(argv[i])) }
```
andychu, line 216: oh actually this reminds me that I want to make an accum() method on dicts:

```
call mydict->accum(flag.name, float(argv[i]))
```

which will do the empty list logic

I find that pattern to be quite common.  I think Perl  might have something like this, but Python/JS don't

And it is a trivial method ... not sure if accum() is the best name

----

Although maybe this was summing an integer, not appending to a list ... I'll start a thread on #language-design about it

andychu (commented):
This is awesome, thank you!  Reading the code inspired a bunch of tangential thoughts

It is indeed useful for testing the language

---


I think the main one is that the free function `get()` might be shorter than the if expressions

stdlib/ysh/args.ysh
quexxon: Oh, yeah! `get` is _much_ better. I'll make that change :+1:

stdlib/ysh/args.ysh
```diff
@@ -185,6 +185,17 @@ func parseArgs(spec, argv) {
...
               error "Expected Int after '$arg', got '$[argv[i]]'" (code=2)
             }
+          } elif (flag.type is List[Int]) {
+            setvar i += 1
+            if (i >= len(argv)) {
+              error "Expected Int after '$arg'" (code=2)
+            }
+
+            setvar value = get(args, flag.name, [])
+            try { call value->append(int(argv[i])) }
```
andychu, line 195: I kinda want to "compress" this whole `try { ... }; if ...` in each of the 3 blocks, as well as some other things, but since this has good tests, it should be easy to do

PR #2109, opened 2024-10-26
implement pretty printing for value.Obj

melvinw:
This commit adds a quick implementation of `pp value` (and by extension `=`) for objects. It generates pleasing output, but might not be optimal. I figure we can check this in and tweak it as we find cases that yield wonky outputs. What do folks think?


andychu:
OK cool, I played with it  a bit, and it seems like it wraps pretty well

PR #2119, opened 2024-11-06
Add dependency installation script for arch

Kaonashie:
Small change in the dependency script to add Arch support. 

Also adds a check for whether or not you are in the right directory for running the file.

### DIFF
--- a/build/deps.sh
+++ b/build/deps.sh
@@ -22,7 +22,23 @@
 #
 #     rm -r -f ~/wedge  # would be better
 
+
+# Check if we're in the right directory
+if [[ ! -d "stdlib/osh" ]]; then
+    echo "Error: This script must be run from the root of the Oil project directory"
+    echo "Please cd to the root directory and try again"
+    exit 1
+fi
+
 : ${LIB_OSH=stdlib/osh}
+if [[ ! -f "$LIB_OSH/bash-strict.sh" ]] || [[ ! -f "$LIB_OSH/task-five.sh" ]]; then
+    echo "Error: Required source files not found in $LIB_OSH/"
+    echo "Expected files:"
+    echo "  - $LIB_OSH/bash-strict.sh"
+    echo "  - $LIB_OSH/task-five.sh"
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


andychu:
Great thanks!

Let us know if you have any problems installing things.  I recall that people have successfully developed Oils on Arch in the past

---

We have an automated dev setup test  for debian, alpine, and fedora.   (Actually Fedora passes but some packages don't build, which should be fixed)

http://op.oilshell.org/uuu/sourcehut-jobs/git-5eaf087a737085f643d41715eb7bb7fbfcdfbb5f/

We could do something similar for Arch, but maybe later!

Kaonashie:
Thank you so much for the quick reply! I managed to get everything setup for development on my Arch machine without any issues ! I would love to help out with the Arch env tests if you decide to do it at some point :)

andychu:
Yes I think we should do that at some point, though it's probably better to do it "on demand" when  we have any breakages, or more Arch contributors.  As long as it works for you, that seems OK now

---

As far as other things to do, I maintained this list in the past ... some of them are old

https://github.com/oils-for-unix/oils/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

There are some example commits if you search Zulip `#help-wanted` for "example", e.g.

https://oilshell.zulipchat.com/#narrow/channel/417617-help-wanted/topic/Good.20example.20of.20Dict.20.3D.3E.20values.28.29

i.e. small commits that show how we fill out the standard library;

http://op.oilshell.org/uuu/github-jobs/8225/

There are some stdlib functions with a red X here - feel free to ask any questions about them on https://oilshell.zulipchat.com/ (which is where all the discussion happens!)

http://op.oilshell.org/uuu/github-jobs/8225/ovm-tarball.wwz/_release/VERSION/doc/ref/toc-ysh.html

Thanks again for the PR!

Kaonashie:
It's a pleasure! 

Thank you so much for taking the time to list out the details and guiding me through. Means the world. I'll start going through the issues and see what I can do 🙂.

PR #2132, opened 2024-11-24
Add some YSH list methods

leungbk:


andychu:
Thank you!  I'm happy that you were able to find all the code AND the docs, and update them

This is an area where we really need help!

I answered on Zulip about accum() too.

I'd be curious to hear thoughts on using YSH as well, since feedback from people using it helps a lot, e.g. the `=>` issue

PR #2133, opened 2024-11-24
[build/deps] Add the "libasan" package to the Fedora dependencies

akinomyoga:
Without this package, `ninja` fails.


akinomyoga:
Hmm, this might also be related to the `root` files.

andychu:
Great thanks!

akinomyoga:
Thanks. I've modified the codes based on the `master` branch, but should I create PRs on top of `soil-straging`?

andychu:
Yes I merge to `soil-staging`, and  then the CI merges to `master` after tests pass

It doesn't matter too much since I just change it for all contributors

akinomyoga:
Thanks

PR #2134, opened 2024-11-24
Add a basic configuration for the clangd language server

leungbk:


leungbk:
This configuration seems to work well for most of the C++, but by itself it doesn't help with the C in the vendored Python.

I noticed that I could get `clangd` to mostly work in the Python source tree when I entered `Python-2.7.13/` and within that directory ran `./configure` and `bear -- make`, where `bear` is a different tool that generates a `compile_commands.json`. The problem is that manually running this separate build generated a lot of changes in `Python-2.7.13/` that would persist in `git status`.

I'm not sure how to get `clangd` to work in `Python-2.7.13/` without this additional step, so feel free to close this PR if you don't find it useful.

andychu:
OK cool, if this helps navigate the generated C++ code, I think it's useful

I am not using `clangd` right now, but I think many people use it with vim?  which editor do you use?

----

I used `bear` a long time ago -- I haven't used it recently

I find myself needing to look at python code rather than C code in   `Python-2.7.13/` ... there are slight patches (OVM_MAIN), but it is mostly unmodified

I want to delete a lot of this old CPython/OVM stuff, since it is less important, now that we have a totally separate C++ build/implementation

PR #2135, opened 2024-11-24
[refactor] Add "core/bash_impl" for array operations

akinomyoga:
https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/2024-11-22.20Meeting/near/484115604

This creates a file `core/bash_impl.py` and includes the `Length` operation for BashArray/BashAssoc/SparseArray as an example.

### DIFF
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -1,0 +1,34 @@
+"""bash_impl.py - implements operations on Bash data structures"""
+
+from _devbuild.gen.value_asdl import value
+
+
+#------------------------------------------------------------------------------
+# All BashArray operations depending on the internal
+# representation of SparseArray come here.
+
+def BashArray_Length(array_val):
+    # type: (value.BashArray) -> int
+
+    # There can be empty placeholder values in the array.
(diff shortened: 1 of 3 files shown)


andychu:
Great thank you!

And thanks for finding `pea/oils-typecheck`

That is updated with `build/dynamic-deps.sh pea-hack` I think, which may not be documented ... that is probably one thing we should smooth over!

andychu:
Oh BTW that file is not really essential I think, it only for an experimental rewrite of mycpp

The real file that determines type checking is  this one

```
$ head _build/NINJA/bin.oils_for_unix/typecheck.txt 
_devbuild/gen/arg_types.py
_devbuild/gen/grammar_nt.py
_devbuild/gen/help_meta.py
_devbuild/gen/hnode_asdl.py
```

It is generated when you do `./NINJA-config.sh`

And then `devtools/types.sh check-oils` reads it

So technically when we add files, we need to run ./NINJA-config.sh to not miss type checking

---

But thanks for also finding the pea one!

PR #2138, opened 2024-11-24
[osh/sh_expr_eval] Support `[[ -v a[-1] ]]` for BashArray

akinomyoga:
In the current master branch, `[[ -v a[-1] ]]` always returns exit status 1 even when there are elements and `"${a[-1]}"` is valid.

### DIFF
--- a/spec/array.test.sh
+++ b/spec/array.test.sh
@@ -853,6 +853,9 @@ e=()
 
 ## status: 1
 ## STDERR:
+  [[ -v e[-1] ]] && echo 'e has -1'
+        ^
+[ stdin ]:2: fatal: -v got invalid negative index -1
 ## END
 
 ## OK bash STDERR:
--- a/spec/array.test.sh
+++ b/spec/array.test.sh
@@ -824,3 +824,41 @@ bash: line 2: a[-1]: bad array subscript
 ## N-I mksh STDERR:
 ## END
 
+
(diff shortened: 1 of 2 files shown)


akinomyoga:
When the negative index is out of range, Bash produces an error while `osh` silently ignores it. Is it intentional that `osh` is more permissive?

akinomyoga:
> So I think we just remove the `strict_word_eval` check to match bash and mksh?

OK, I added a new commit: 89db9b959e71be8c1076336211f78a4a8167b391

spec/array.test.sh
```diff
@@ -824,3 +824,44 @@ bash: line 2: a[-1]: bad array subscript
...
+
+#### Negative out-of-range index in [[ -v a[index] ]]
+e=()
+[[ -v e[-1] ]] && echo 'e has -1'
+
+## status: 1
+## STDERR:
+  [[ -v e[-1] ]] && echo 'e has -1'
+        ^
+[ stdin ]:2: fatal: -v got invalid negative index -1
```
andychu, line 858: It might be nice to print the length of the array too, which gives an indication of why it's invalid

But we can add that later if we think it's helpful

akinomyoga: Thank you. I created a PR: https://github.com/oils-for-unix/oils/pull/2140.

Issue #661, opened 2020-03-15
 declare -A dict=() not allowed in Oil


andychu:
> Also what's the difference between declare -A dict and declare -A dict=() in bash?

There are three different states for variables in Bash. In addition, the "set" state can be categorized into two sub-states:

1. Not declared (slot is not allocated)
2. Declared but unset (slot is allocated, but no value is set)
3. Set (slot is allocated, and a value is set)
   1. Empty
   2. Non-empty

See the following test case for the behavior differences:

```bash
# test5.sh

function is-declared() { declare -p "$1" &>/dev/null; }
function is-enumerated-by-declare() { declare | grep -q "^$1="; }
function is-enumerated-by-varname-expansion() { eval "printf '%s\n' \"\${!$1@}\"" | grep -q "^$1\$"; }
function is-enumerated-by-compgen () { compgen -v | grep -q "^$1\$"; }
function is-enumerated-by-compgen-arrayvar () { compgen -A arrayvar | grep -q "^$1\$"; }
function show-status {
  local -a msg=(yes)
  local f
  for f in $(compgen -A function -- is-); do
    "$f" "$1"
    echo "$f: ${msg[$?]:-no}"
  done
}

echo "[undeclared]"
show-status dict
echo "[declared]"
declare -A dict
show-status dict
echo "[set]"
dict=()
show-status dict
```

```console
$ bash test5.sh
[undeclared]
is-declared: no
is-enumerated-by-compgen: no
is-enumerated-by-compgen-arrayvar: no
is-enumerated-by-declare: no
is-enumerated-by-varname-expansion: no
[declared]
is-declared: yes
is-enumerated-by-compgen: no
is-enumerated-by-compgen-arrayvar: no
is-enumerated-by-declare: no
is-enumerated-by-varname-expansion: no
[set]
is-declared: yes
is-enumerated-by-compgen: yes
is-enumerated-by-compgen-arrayvar: yes
is-enumerated-by-declare: yes
is-enumerated-by-varname-expansion: yes
```

Related to this, in the help-bash mailing list, I have heard about some script library that gives a default array content when the definition is not given by the application. I'm not sure if it is a good design, but in that library, the empty array is one of the possible configurations, so one needs to distinguish *the unset state* (`declare -A dict`) from *the empty state* (`declare -A dict=()`). Also, if the array is associative, the library wants to declare the associative array attribute in advance. Such a script can be achieved in the following way:

```bash
# library
declare -A dict
[...]

PR #2139, opened 2024-11-24
[builtin/assign_osh] Support `declare -p` for SparseArray

akinomyoga:
This PR implements `declare -p sp` for SparseArray

### Representation ([index]=value)

In this PR, I currently use the form `([index]=value)` for `declare -p`, but this form was exclusively used for the associative arrays before, AFAICR. However, I also remember that we now have osh and ysh where osh is more compatible with Bash. What is the current situation for the representation of `([index]=value)`? Can I use this form for the sparse-array representation (assuming that it will be implemented in the future)?

Or do we need to stick with `(); name[index]=value; ...`? There are still other places that want to use the string representation of a sparse-array value, but the usage of the form `(); name[index]=value` is limited as the second part `; ...` is only valid as "a command".

Another related discussion would be the mixed form `([begin_index]=val1 val2 val3 ...)`. I notice that there is a failing spec test (allowed failure) for it.

### Better function name?

I named the serialization function for the sparse array as `ToStrForAssignment`, but it might be better to name it as `ToStrShellEncode`, etc. I can follow if you have existing naming for this kind of function.

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -11,6 +11,7 @@ from _devbuild.gen.runtime_asdl import (
 from _devbuild.gen.value_asdl import (value, value_e, value_t, LeftName)
 from _devbuild.gen.syntax_asdl import loc, loc_t, word_t
 
+from core import bash_impl
 from core import error
 from core.error import e_usage
 from core import state
@@ -131,7 +132,7 @@ def _PrintVariables(mem, cmd_val, attrs, print_flags, builtin=_OTHER):
         if flag_x == '+' and cell.exported:
             continue
 
-        if flag_a and val.tag() != value_e.BashArray:
+        if flag_a and val.tag() not in [value_e.BashArray, value_e.SparseArray]:
             continue
         if flag_A and val.tag() != value_e.BashAssoc:
             continue
@@ -145,7 +146,7 @@ def _PrintVariables(mem, cmd_val, attrs, print_flags, builtin=_OTHER):
                 flags.append('r')
             if cell.exported:
                 flags.append('x')
-            if val.tag() == value_e.BashArray:
+            if val.tag() in [value_e.BashArray, value_e.SparseArray]:
                 flags.append('a')
             elif val.tag() == value_e.BashAssoc:
                 flags.append('A')
@@ -206,6 +207,10 @@ def _PrintVariables(mem, cmd_val, attrs, print_flags, builtin=_OTHER):
                 body.extend(["[", key_quoted, "]=", value_quoted])
             if len(body) > 0:
                 decl.extend(["=(", ''.join(body), ")"])
+
+        elif val.tag() == value_e.SparseArray:
+            sparse_val = cast(value.SparseArray, val)
+            decl.extend(["=", bash_impl.SparseArray_ToStrForShellPrint(sparse_val)])
 
         else:
             pass  # note: other types silently ignored
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -2,6 +2,12 @@
 
 from _devbuild.gen.value_asdl import value
 
+from data_lang import j8_lite
+from mycpp import mops
+from mycpp import mylib
+
+from typing import List
+
 
 #------------------------------------------------------------------------------
 # All BashArray operations depending on the internal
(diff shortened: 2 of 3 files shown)


andychu:
Yeah I think we should try to be more compatible with bash

I don't remember all the reasons, but I think there is nothing constraining us now, fundamentally

---

For the name, I might call `SparseArray_ShellPrint` or something ?    Because it is a shell format `declare -p`, meant to be fed to  `eval`

This is distinct from "pretty printing" of `=` operator and `pp value` and  `pp test_` and `pp asdl_` and so forth

I would choose whatever names help you develop.  Sometimes I change them for "global" consistency, but for now it's OK to be locally consistent

akinomyoga:
> I don't remember all the reasons, but I think there is nothing constraining us now, fundamentally

OK.

What I'm now afraid of is how the code change would affect the YSH side since there seems to be an overlapping region in the implementation. Anyway, I'll look into the codebase to understand it.

> For the name, I might call `SparseArray_ShellPrint` or something ? Because it is a shell format `declare -p`, meant to be fed to `eval`

How about `SparseArray_ToStrShellPrint` or something? If we simply call it `ShellPrint`, it appears to be going to output something to a stream. The function I want to add is just to convert the value to a string. I want to use this function in other places which are not necessarily for printing.

Or it might be something like `SparseArray_ToStr(type=string_rep_e.ShellPrint)` since there seem to be even other forms of representing the value (such as in `data_lang/j8.py` and in `display/pp_value.py`).

> This is distinct from "pretty printing" of `=` operator and `pp value` and `pp test_` and `pp asdl_` and so forth

Yeah, these ones.

> I would choose whatever names help you develop. Sometimes I change them for "global" consistency, but for now it's OK to be locally consistent

OK. I'll use `ToStrForShellPrint` for now.

> I remember there was this issue
> 
> ```
> declare -a a=()
> declare -A A=()
> ```

Yeah, it was #661.

> We can probably change that now so it's like bash. But I don't have a strong idea of how to do it cleanly

I still think of an approach similar to `std::initializer_list`. I suggested a hybrid version of the existing semantics [ `()` is BashArray `([k]=v)` is BashAssoc] and an initializer list in https://github.com/oils-for-unix/oils/issues/661#issuecomment-599334537, but ideally, all `()` would be uniformly typed as `InitializerList(List[Tuple[str, bool, str]])` where triplet contains `key, append, value`. For example,
- `()` would be `[]`
- `(["key"]=value)` would be `[("key", False, "value")]`
- `(1 2 3)` would be `[(None, False, "1"), (None, False, "2"), (None, False, "3")]`
- `([10]+=a [20]+=b c d ["key"]=e)` would be `[("10", True, "a"), ("20", True, "b"), (None, False, "c"), (None, False, "d"), ("key", False, "e")]`

Then, LHS can be initialized using the given initializer list. The operation can be implemented as a function `bash_impl.BashArray_AssignList(BashArray lhs, InitializerList rhs)`, etc. This is consistent with the Bash behavior (except for unset variables; in Bash, the array types are actually attributes of the variable placeholder, so unset variables can also have types).

> We can talk about it I guess, though I think it's better to focus on SparseArray first?

I agree. Actually, I still think it's mostly orthogonal because the operation that we now try to implement in `bash_impl` is anyway useful, and the codes for the assignments simply call methods in `bash_impl`.

> We might have to add `-a -A` flags, which we don't have. Right now we only have `-r -x -n` flags. I think that is the fundamental issue

We already have `-a` and `-A`. If you mean that the `-a`/`-A` flags in the variable placeholders (instead of the value itself), I don't think it is needed actually. We can use the LHS type to dispatch operation for a given initializer list. However, if we want to make even the behavior for unset variables the same as Bash, we might need to introduce the type in the variable placeholders, or we may simply introduced "*typed None*".

akinomyoga:
6c7c3d2db1f715b5278d23d4c9e76d9c43e1b26a I renamed it to `SparseArray_ToStrForShellPrint`. We can adjust the name later if necessary.

andychu:
Yeah I am not sure, but I am thinking of cases where you change the type of a variable.

Sometimes it is not allowed

```
$ bash -c  'declare -a z=''; z+=1; declare -p z; declare -A z; declare -p z'
declare -a z=([0]="1")
bash: line 1: declare: z: cannot convert indexed to associative array
declare -a z=([0]="1")

```

but then here it is kinda allowed

```
$ bash -c  'declare z=''; z+=1; declare -p z; declare -A z; declare -p z'
declare -- z="1"
declare -A z=([0]="1" )
```


I would prefer not having to remember `-a -A`, to keep things simple

But if it turns out to be necessary then we could do it

---

But yeah I still don't want to implement bash bugs that are not documented

andychu:
Thanks, I merged this!   tests look good

PR #2140, opened 2024-11-24
[osh/sh_expr_eval] Tell valid index range for out-of-range `[[ -v a[i] ]]`

akinomyoga:
https://github.com/oils-for-unix/oils/pull/2138#discussion_r1855493290

### DIFF
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -1030,7 +1030,7 @@ class BoolEvaluator(ArithEvaluator):
                 if index < 0:
                     index += n
                     if index < 0:
-                        e_die('-v got invalid negative index %s' % index_str,
+                        e_die('-v got index %s, which is out of bounds for array of length %d' % (index_str, n),
                               blame_loc)
                         return False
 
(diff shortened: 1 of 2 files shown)


akinomyoga:
If you have a better idea about the way to include the information in the message, please tell me.

osh/sh_expr_eval.py
```diff
@@ -1030,7 +1030,7 @@ def _IsDefined(self, s, blame_loc):
                 if index < 0:
                     index += n
                     if index < 0:
-                        e_die('-v got invalid negative index %s' % index_str,
+                        e_die('-v got invalid negative index %s (expect: index >= -array_length = %d)' % (index_str, -n),
```
andychu, line 1033: How about just `array length = %d` or `array length is %d` ? 

I think that will help the user enough

I am a little confused it says `-array_length`, hard for me to evaluate in my head :)   

---

Also I think we can say "index out of bounds" to be more specific than "invalid"

So perhaps

```
-v got index %d, which is out of bounds for array of length %d
```

?

I think that is shorter and conveys the info clearly

akinomyoga: Thank you for the suggestion. I updated the message: [`d23bf7c3..3b5b5ce`](https://github.com/oils-for-unix/oils/compare/d23bf7c351b6c1682f9d3c7e53fc5a8178e0f2ec..3b5b5ce019271348597aba5180782deebcf3a1d9).

andychu:
Thanks!

PR #2143, opened 2024-11-29
[test/spec] Add failing globstar tests

leungbk:
Related to #232.

Following what Ellen did in #1757, I had to use some restrictive naming (foo.md) in my test cases to work around the `_tmp` (?) subdirectory. I don't understand what the testing environment looks like: if indeed the current directory during test execution is the repo root, I would expect some of my globs to pick up, say, `osh/bool_stat.py`.

### DIFF
--- a/spec/globstar.test.sh
+++ b/spec/globstar.test.sh
@@ -22,9 +22,9 @@ shopt -s globstar
 mkdir -p c/subdir
 touch {leaf.md,c/leaf.md,c/subdir/leaf.md}
 
-echo **/*.* | tr ' ' '\n'
+echo **/*.* | tr ' ' '\n' | sort
 echo
-echo **/**/*.* | tr ' ' '\n'
+echo **/**/*.* | tr ' ' '\n' | sort
 ## STDOUT:
 c/leaf.md
 c/subdir/leaf.md
@@ -54,7 +54,7 @@ shopt -s globstar
 mkdir -p c/subdir
 touch c/subdir/leaf.md
 
-echo {**/*.*,} | sort -u | sed 's/[[:space:]]*$//'
+echo {**/*.*,} | sort | sed 's/[[:space:]]*$//'
 ## STDOUT:
 c/subdir/leaf.md
 ## END
@@ -66,11 +66,11 @@ mkdir directory
 touch leaf.md
 touch directory/leaf.md
 
-echo **/*.* | sort -u
-echo directory/**/*.md | sort -u
-echo d**/*.md | sort -u
-echo **y/*.md | sort -u
-echo d**y/*.md | sort -u
+echo **/*.* | sort
+echo directory/**/*.md | sort
+echo d**/*.md | sort
+echo **y/*.md | sort
+echo d**y/*.md | sort
 ## STDOUT:
 directory/leaf.md leaf.md
 directory/leaf.md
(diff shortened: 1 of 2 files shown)


andychu (commented):
This is great, thanks!

What happens if we add `compare_shells: zsh`?    Can we get some tests to pass?

If possible, we try to implement behavior accepted by more than one shell, because bash has many quirks that are different across versions

andychu:
other questions

- how do bash/zsh interpret say `***` and `****`?
- Regarding the difference between `**` and `**/**`, I found some weird zsh behavior that bash doesn't agree with

```
andy@hoover:~/git/oilshell/oil$ zsh -c 'echo */*.z'
_tmp/a.z

andy@hoover:~/git/oilshell/oil$ zsh -c 'echo **/*.z'
_tmp/a.z _tmp/z/a.z

andy@hoover:~/git/oilshell/oil$ zsh -c 'echo **/**/*.z'
_tmp/a.z _tmp/a.z _tmp/z/a.z _tmp/z/a.z _tmp/z/a.z

andy@hoover:~/git/oilshell/oil$ bash -c -c 'shopt -s globstar; echo **/**/*.z'
_tmp/a.z _tmp/z/a.z

```

leungbk:
> how do bash/zsh interpret say *** and ****?

From a quick grep of `bash`, I cannot find anything in its source tree about *** or ****. GPT-4o believes they're both treated the same as *, and that only ** is treated as a globstar.

In `zsh`, I have read in its NEWS file that *** has something to do with symlinks, though I haven't found the corresponding tests. Emacs' `eshell` globbing, which is described as mimicking that of `zsh`, [apparently handles symlinks as well](https://git.savannah.gnu.org/cgit/emacs.git/tree/test/lisp/eshell/em-glob-tests.el?h=8184f64ab0025afb0b901b9cae3917e6f722974b#n221), though I don't yet understand its behavior.

> ```
> andy@hoover:~/git/oilshell/oil$ zsh -c 'echo **/**/*.z'
> _tmp/a.z _tmp/a.z _tmp/z/a.z _tmp/z/a.z _tmp/z/a.z
> ```

Here's my guess for how these five outputs come about:
- The first one occurs by having the first `**/` in the `**/**/*.z` pattern match `tmp/` and having the second one match zero additional subdirectories.
- The second occurs by having the first `**/` match zero subdirectories and the second `**/` match `tmp/`
- The last three occur by having the first (second) `**/` match {0,1,2} ({2,1,0}) subdirectories, with the subdirectories being `_tmp` and `_tmp/z`

> ```
> andy@hoover:~/git/oilshell/oil$ bash -c -c 'shopt -s globstar; echo **/**/*.z'
> _tmp/a.z _tmp/z/a.z
> ```

Here, as before, `**/` can match zero or more subdirectories, but I guess `bash`'s globbing lumps together the {0,1,2} outcomes I described above.

leungbk:
OK, I've updated the spec to include `zsh`. I dropped a test confirming that `**` would behave as `*` when globstar is disabled since `zsh` does not seem to have a way of disabling globstar.

Since `bash` appears not to treat `***` the same way `zsh` does, do you want any tests for this?

andychu:
Oh we should keep the other tests -- the way to do it is to add 

```
case $SH in zsh) exit ;; esac   # don't run anything for zsh
```

at the top

and then

```
N-I zsh STDOUT:
END
```


The goal is just to take a survey at first, without committing to any behavior in OSH

Thanks!

leungbk:
OK, I've restored that test from before, and I've also added a test for `***`.

andychu:
Ah looks like this hit my breakage on soil-staging

http://op.oilshell.org/uuu/github-jobs/

(the staging branch can be broken, but master should not be)

I will try re-running this, thank you!

andychu:
Great, thank you!  I'll start a thread about how to implement this

We learned that there are some odd differences between bash and zsh

( I removed the `sort -u` from the test cases, since we want to see those)

PR #2145, opened 2024-11-29
[refactor] Add `Bash{Array,Assoc}_ToStrForShellPrint` for `declare -p` and `set -x`

akinomyoga:
Note: This changes the result for an empty `BashAssoc`. We have been using `declare -A assoc` to mean an empty associative array while allowing `declare -A assoc=()` for *compatibility*. Now, this produces `declare -A assoc=()` for `declare -p assoc` to be consistent with Bash's output. **edit**: To explain the background, we want to call the ToStr function in [this form consistent with `BashArray`](https://github.com/oils-for-unix/oils/pull/2145/files#diff-828bd8642431550aea162b9540190291ce9f0cd52ce6b3c1041a09f585adbbe8R166), where the result is appended to `var_name=`. To append something to `declare -A assoc=` for an empty associative array, we need to use `()`.

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -163,50 +163,11 @@ def _PrintVariables(mem, cmd_val, attrs, print_flags, builtin=_OTHER):
 
         elif val.tag() == value_e.BashArray:
             array_val = cast(value.BashArray, val)
-
-            # mycpp rewrite: None in array_val.strs
-            has_holes = False
-            for s in array_val.strs:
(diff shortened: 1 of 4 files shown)


akinomyoga:
OK, there was actually a test case for `declare -A a` for an empty BashAssoc. I added a commit to update the test case.

andychu:
Thanks!

akinomyoga:
Thanks!

PR #2146, opened 2024-11-29
[osh] Implement BashAssoc `assoc+=()`

akinomyoga:
https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/2024-11-22.20Meeting/near/485167138

### DIFF
--- a/osh/cmd_eval.py
+++ b/osh/cmd_eval.py
@@ -231,8 +231,19 @@ def PlusEquals(old_val, val):
                 raise AssertionError()  # parsing should prevent this
 
         elif case(value_e.BashAssoc):
-            # TODO: Could try to match bash, it will append to ${A[0]}
-            pass
+            if tag == value_e.Str:
+                e_die("Can't append string to array")
+
+            elif tag == value_e.BashAssoc:
+                assoc_lhs = cast(value.BashAssoc, UP_old_val)
+                assoc_rhs = cast(value.BashAssoc, UP_val)
+
+                for key in assoc_rhs.d.keys():
+                    assoc_lhs.d[key] = assoc_rhs.d[key]
+                val = assoc_lhs
+
+            else:
+                raise AssertionError()  # parsing should prevent this
 
         else:
             e_die("Can't append to value of type %s" % ui.ValType(old_val))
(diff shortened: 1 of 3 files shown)


osh/cmd_eval.py
```diff
@@ -231,8 +231,19 @@ def PlusEquals(old_val, val):
                 raise AssertionError()  # parsing should prevent this
 
         elif case(value_e.BashAssoc):
-            # TODO: Could try to match bash, it will append to ${A[0]}
-            pass
+            if tag == value_e.Str:
+                e_die("Can't append string to array")
```
akinomyoga, line 235:
```suggestion
                e_die("Can't append string to associative arrays")
```

Oh, the error message should be "associative arrays"...

osh/cmd_eval.py
```diff
@@ -231,8 +231,19 @@ def PlusEquals(old_val, val):
...
 
         elif case(value_e.BashAssoc):
-            # TODO: Could try to match bash, it will append to ${A[0]}
-            pass
+            if tag == value_e.Str:
+                e_die("Can't append string to associative arrays")
+
+            elif tag == value_e.BashAssoc:
+                assoc_lhs = cast(value.BashAssoc, UP_old_val)
+                assoc_rhs = cast(value.BashAssoc, UP_val)
```
andychu, line 239: Do we need to consider `tag == value_e.BashArray` ?


```
assoc+=([k]=v)  # right now we consider this BashAssoc

assoc+=(x y) # I think we consider this an array

assoc+=str
```


I guess we can add an assert for now?

akinomyoga: > Do we need to consider `tag == value_e.BashArray` ?

I just followed the current code for BashArray.

https://github.com/oils-for-unix/oils/blob/0df89bf3d19c70a0003db71881c45e8637199ef5/osh/cmd_eval.py#L216-L230

osh/cmd_eval.py
```diff
@@ -231,8 +231,19 @@ def PlusEquals(old_val, val):
...
+
+                for key in assoc_rhs.d.keys():
+                    assoc_lhs.d[key] = assoc_rhs.d[key]
+                val = assoc_lhs
+
+            else:
+                raise AssertionError()  # parsing should prevent this
```
akinomyoga, line 246: > We could add an assert for now -- just disallow the combination

Yeah, but I think that that is this line.

andychu:
Ah OK, good point!

akinomyoga:
Thank you!

PR #2152, opened 2024-11-30
[core] Fix silent out-of-range error in unsetting BashArray element

akinomyoga:
While Bash causes an error for `unset` with the invalid negative index,

```console
$ bash -c 'a=(1); unset -v "a[-2]"'
bash: line 1: unset: [-2]: bad array subscript
$ echo $?
1
```

the current `master` of `osh` succeeds silently:

```console
$ bin/osh -c 'a=(1); unset -v "a[-2]"'
$ echo $?
0
```

This patch adds the check and outputs an error message for the invalid negative index.

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -2292,12 +2292,16 @@ class Mem(object):
                 index = lval.index
                 if index < 0:
                     index += n
+                    if index < 0:
+                        e_die(
+                            "Index %d is out of bounds for array of length %d"
+                            % (lval.index, n))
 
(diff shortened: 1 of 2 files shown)


akinomyoga:
61849c90b9e3ad229c71a15e9d96fc383d9f0d48 is a minor fix for styling. I included it in this PR, but if we need a separate PR for an independent fix, please let me know.

andychu:
Great thank you!

PR #2154, opened 2024-11-30
[mylib] Fix a code comment

akinomyoga:
`mylib.BigIntSort()` is introduced to sort the keys of `SparseArray`. It was first introduced to implement the methods in `func_misc.SparseOp`.  It is related to `SparseArray` rather than `BashArray`.  This patch fixes the code comment for `BigIntSort`.


andychu:
OK this is fine, but I guess at some point we will s/SparseArray/BashArray/ everywhere

PR #2153, opened 2024-11-30
[osh] Fix crash with `${a[@]: -(2*${#a[@]}+1)}`

akinomyoga:
The current `master` crashes with the following command:

```console
$ bin/osh -c 'a=(1 2 3); echo "${a[@]: -7}"'
Traceback (most recent call last):
  File "/home/murase/.mwg/git/oilshell/oil/bin/oils_for_unix.py", line 201, in <module>
    sys.exit(main(sys.argv))
  File "/home/murase/.mwg/git/oilshell/oil/bin/oils_for_unix.py", line 170, in main
    return AppBundleMain(argv)
  File "/home/murase/.mwg/git/oilshell/oil/bin/oils_for_unix.py", line 140, in AppBundleMain
    return shell.Main('osh', arg_r, environ, login_shell, loader, readline)
  File "/home/murase/.mwg/git/oilshell/oil/core/shell.py", line 1213, in Main
    cmd_flags=cmd_eval.IsMainProgram)
  File "/home/murase/.mwg/git/oilshell/oil/core/main_loop.py", line 375, in Batch
    is_return, is_fatal = cmd_ev.ExecuteAndCatch(node, cmd_flags)
  File "/home/murase/.mwg/git/oilshell/oil/osh/cmd_eval.py", line 2091, in ExecuteAndCatch
    status = self._Execute(node)
  File "/home/murase/.mwg/git/oilshell/oil/osh/cmd_eval.py", line 1890, in _Execute
    status = self._Dispatch(node, cmd_st)
  File "/home/murase/.mwg/git/oilshell/oil/osh/cmd_eval.py", line 1728, in _Dispatch
    status = self._ExecuteList(node.children)
  File "/home/murase/.mwg/git/oilshell/oil/osh/cmd_eval.py", line 1967, in _ExecuteList
    status = self._Execute(child)
  File "/home/murase/.mwg/git/oilshell/oil/osh/cmd_eval.py", line 1890, in _Execute
    status = self._Dispatch(node, cmd_st)
  File "/home/murase/.mwg/git/oilshell/oil/osh/cmd_eval.py", line 1617, in _Dispatch
    status = self._DoSimple(node, cmd_st)
  File "/home/murase/.mwg/git/oilshell/oil/osh/cmd_eval.py", line 851, in _DoSimple
    allow_assign=True)
  File "/home/murase/.mwg/git/oilshell/oil/osh/word_eval.py", line 2301, in EvalWordSequence2
[...]

PR #2155, opened 2024-11-30
[osh] Fix last element `${a[@]: -1}` after `unset -v a[-1]`

akinomyoga:
In the current master, after unsetting the last element `unset -v a[-1]` when there is a hole before the last element, the subsequent call of array slice `${a[@]: -1}` fails to handle negative offset properly.

```console
$ bash -c 'a=(1); a[9]=x; unset -v "a[-1]"; echo "${a[@]: -1}"'
1
$ bin/osh -c 'a=(1); a[9]=x; unset -v "a[-1]"; echo "${a[@]: -1}"'

$
```

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -2302,6 +2302,8 @@ class Mem(object):
                     # Special case: The array SHORTENS if you unset from the end.  You
                     # can tell with a+=(3 4)
                     strs.pop()
+                    while len(strs) > 0 and strs[-1] is None:
+                        strs.pop()
                 elif index < last_index:
                     strs[index] = None
                 else:
--- a/spec/array.test.sh
+++ b/spec/array.test.sh
@@ -931,3 +931,24 @@ begin=-5 -> ()
 
 ## N-I mksh STDOUT:
 ## END
+
+
+#### array length after unset
+case $SH in mksh) exit ;; esac
+
+a=(x)
+a[9]=y
+echo "len ${#a[@]};"
+
+unset -v 'a[-1]'
+echo "len ${#a[@]};"
+echo "last ${a[@]: -1};"
+
+## STDOUT:
+len 2;
+len 1;
+last x;
+## END
+
+## N-I mksh STDOUT:
+## END


core/state.py
```diff
@@ -2297,6 +2297,8 @@ def Unset(self, lval, which_scopes):
                     # Special case: The array SHORTENS if you unset from the end.  You
                     # can tell with a+=(3 4)
```
andychu, line 2298: Hm I don't recall this behavior

What does bash do if you `unset a[-2]` or `a[-3]`?    Is it documented anywhere?

andychu: I wonder if we have test coverage of `a[-2]` and a[-len]` and so forth??  If so maybe it is a good idea to add it?

akinomyoga: > What does bash do if you `unset a[-2]` or `a[-3]`? Is it documented anywhere?

The problem that this PR tries to fix is unrelated to the behavior for `unset a[-2]` in the present case. Anyway, `a[-2]` and `a[-3}` removes an element `a[n-2]` and `a[n-3]`, respectively, with `n` being the index of the maximum set element plus one.

The present problem is that after `a[9]=x; unset -v "a[-1]"`, the array does not become the original state in the current `master` of `osh`. With `a[9]`, the array becomes `(1 None None None None None None None None x)`. Then, with `unset -v "a[-1]"`, the array should become `(1)`. However, in the current `master` of `osh`, it results in `(1 None None None None None None None None)`. Then, after that, `${a[@]: -1}` in `osh` picks up the last element of the filler `None`. This doesn't make sense.

akinomyoga: > I wonder if we have test coverage of `a[-2]` and `a[-3]` and `a[-len]` and so forth?? If so maybe it is a good idea to add it?

I think that is covered by the test added by #2153.

akinomyoga: Maybe I have confused you by showing the comparison with Bash. Even if we do not compare the behavior with Bash, the current `osh` behavior is absolutely strange.

```bash
a[0]=x
a[2]=y
unset -v 'a[2]'
# Here, we expect the state to be equivalent to a=(x)
echo "${a[@]: -1}"  # --> This produces no words.
```

akinomyoga:
Thank you. I rebased it.

andychu:
Great thanks!

akinomyoga:
Thanks!!

PR #2157, opened 2024-12-01
[core/state] Implement SparseArray `a[i]=v`

akinomyoga:
BigInt operations for comparison are done through `bash_impl.BigInt_Less` as suggested in Ref. [1]

[1] https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/2024-11-22.20Meeting/near/485124692

## How should I return an error status?

For the result of `SparseArray_GetElement`, there are three possible cases:

- An element is found at the specified index. In this case, we return the value.
- No element is found at the specified index. In this case, we return `None`.
- The index is invalid. In particular, this happens when the index is too negative as `((index < -n))` (where `n` is the max index plus 1). I would like to return an error status this case, and let the caller determine how to handle it.

~~In the current PR,~~ In the planned PR for `SparseArray_GetElement`, I return a tuple `(str result, int error_code)` as the simplest approach, but this can be implemented in various ways.

- Another way is to return `(str result, bool has_error)` to just tell the existence of an error,
- Or reverting the meaning of `bool` from `has_error` to `success`, `(str result, bool success)`.
- Or something like `(str result, get_array_element_error_code_e error_code)` with `get_array_element_error_code_e` being an enum type to represent the error.
- One might even define a type as `GetElementResult = (str result, int error_code)`.
- Or is there already a way established in the Oils codebase?

**edit**: Oh, this PR is not about `SparseArray_GetElement` but about `SparseArray_SetElement`. Anyway, I want to discuss the same thing for `SparseArray_SetElement`.

For this PR, I now return `int error_code` for `SparseArray_SetElement`, but we have options in the same way as `SparseArray_GetElement` as discussed above.

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -1997,10 +1997,12 @@ class Mem(object):
                         error_code = bash_impl.SparseArray_SetElement(
                             lhs_sp, mops.BigInt(lval.index), rval.s)
                         if error_code == 1:
-                            n = bash_impl.SparseArray_Length(lhs_sp)
+                            n_big = mops.Add(
+                                bash_impl.SparseArray_MaxIndex(lhs_sp),
+                                mops.ONE)
                             e_die(
-                                "Index %d is out of bounds for array of length %d"
-                                % (lval.index, n), left_loc)
+                                "Index %d is out of bounds for array of length %s"
+                                % (lval.index, mops.ToStr(n_big)), left_loc)
                         return
 
                 # This could be an object, eggex object, etc.  It won't be
--- a/core/state.py
+++ b/core/state.py
@@ -20,6 +20,7 @@ from _devbuild.gen.value_asdl import (value, value_e, value_t, Obj, sh_lvalue,
                                       sh_lvalue_e, sh_lvalue_t, LeftName,
                                       y_lvalue_e, regex_match, regex_match_e,
                                       regex_match_t, RegexMatch)
+from core import bash_impl
 from core import error
 from core.error import e_usage, e_die
 from core import num
@@ -1989,6 +1990,17 @@ class Mem(object):
                             for i in xrange(n):
                                 strs.append(None)
                             strs[lval.index] = rval.s
+                        return
+
+                    elif case2(value_e.SparseArray):
+                        lhs_sp = cast(value.SparseArray, UP_cell_val)
+                        error_code = bash_impl.SparseArray_SetElement(
+                            lhs_sp, mops.BigInt(lval.index), rval.s)
+                        if error_code == 1:
+                            n = bash_impl.SparseArray_Length(lhs_sp)
+                            e_die(
+                                "Index %d is out of bounds for array of length %d"
+                                % (lval.index, n), left_loc)
                         return
 
                 # This could be an object, eggex object, etc.  It won't be
--- a/spec/ble-idioms.test.sh
+++ b/spec/ble-idioms.test.sh
@@ -431,3 +431,25 @@ declare -a sp1=([10]=a [20]=b [99]=c [100]=1 [101]=2 [102]=3)
 
 ## N-I bash/zsh/mksh/ash STDOUT:
 ## END
+
+
+#### SparseArray: a[i]=v
+case $SH in bash|zsh|mksh|ash) exit ;; esac
+
+sp1[10]=a
+sp1[20]=b
+sp1[30]=c
+var sp1 = _a2sp(sp1)
+declare -p sp1
+sp1[10]=X
+sp1[25]=Y
+sp1[90]=Z
+declare -p sp1
+
+## STDOUT:
+declare -a sp1=([10]=a [20]=b [30]=c)
+declare -a sp1=([10]=X [20]=b [25]=Y [30]=c [90]=Z)
+## END
+
+## N-I bash/zsh/mksh/ash STDOUT:
+## END
--- a/spec/ble-idioms.test.sh
+++ b/spec/ble-idioms.test.sh
@@ -453,3 +453,71 @@ declare -a sp1=([10]=X [20]=b [25]=Y [30]=c [90]=Z)
 
 ## N-I bash/zsh/mksh/ash STDOUT:
 ## END
+
+
+#### SparseArray: Negative index with a[i]=v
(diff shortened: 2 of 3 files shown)


andychu (commented):
I think what you've done with the error codes makes sense

Sometimes we raise exceptions directly from the "inside", e.g. we could raise it from CanonicalizeIndex directly.  I think we have enough info

But in general we have a mix of

1. raising exceptions
2. boolean values

and if you need 3 cases, then a boolean isn't enough, so an integer error is OK

andychu:
I also wonder if we can reuse the same test cases for BashArray and SparseArray?

e.g. we can have

```
a=(a b)
do-test-case

setvar a = _a2sp(a)
do-test-case
```


and then the outside behavior will be exactly the same, and we don't have to write the same test cases twice

And then when we remove BashArray, we can simply delete those lines?   we will run the same test case just once, instead of twice

spec/ble-idioms.test.sh
```diff
@@ -431,3 +431,25 @@ declare -a sp1=([10]=a [20]=b [99]=c [100]=1 [101]=2 [102]=3)
...
+sp1[10]=X
+sp1[25]=Y
+sp1[90]=Z
+declare -p sp1
+
+## STDOUT:
+declare -a sp1=([10]=a [20]=b [30]=c)
+declare -a sp1=([10]=X [20]=b [25]=Y [30]=c [90]=Z)
+## END
+
```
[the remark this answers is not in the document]
akinomyoga: Yeah, I'll add a test. Thank you for the suggestion.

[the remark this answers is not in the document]
akinomyoga: I added three tests: 685b111fb7180a7cdec198f2d264929c3dd4221d

core/state.py
```diff
@@ -1991,6 +1992,17 @@ def SetValue(self, lval, val, which_scopes, flags=0):
                             strs[lval.index] = rval.s
                         return
 
+                    elif case2(value_e.SparseArray):
+                        lhs_sp = cast(value.SparseArray, UP_cell_val)
+                        error_code = bash_impl.SparseArray_SetElement(
+                            lhs_sp, mops.BigInt(lval.index), rval.s)
```
akinomyoga, line 1998:
```suggestion
                            lhs_sp, mops.BigInt(lval.index), rval.s)
```

When should we use `mops.BigInt(int)` and when `mops.IntWiden(int)`?

akinomyoga:
> I think what you've done with the error codes makes sense

OK, thanks. I'll use the current approach for now.

> Sometimes we raise exceptions directly from the "inside", e.g. we could raise it from CanonicalizeIndex directly. I think we have enough info

I see. In the present case, let me avoid using the exception because I'm going to use `_CanonicalizeIndex ` for other functions in the later PRs, and the out-of-bound index doesn't necessarily cause an error. Throwing an exception has a large cost, so I want to avoid involving the exception in the normal execution (without errors).

> I also wonder if we can reuse the same test cases for BashArray and SparseArray?

We can finally do that after I complete the SparseArray implementation. Or it is technically possible, but it's not practically useful in implementing `SparseArray` because

- I need to look up and identify the tests in `spec/*.test.sh` that correspond to the implemented feature of SparseArray and turn on the tests for SparseArray. It is troublesome for me to scan all the files for every single new feature of SparseArray.
- The current test cases for BashArray are not written in the way of using the shell function `do-test-case`, so we need to rewrite all the test cases to perform the test in a shell function. However, after we replace the implementation of the arrays from BashArray to SparseArray, that complication with the shell function `do-test-case` is not necessary anymore.
- A test usually does not use a single feature of BashArray.
  1. The test needs to construct an array
  2. The test checks the behavior of the target operation.
  3. The test needs to check the state of the array after the operation.

  Steps 1 and 3 use other features of the array, but SparseArray is currently not fully implemented, so the required features in steps 1 and 3 are not necessarily implemented at the time the target feature is added. If we need to use the existing test cases for the new implementation of SparseArray, it poses a strong and non-trivial restriction on the order I implement SparseArray. In particular, to avoid conflicts, I will have to implement the features of SparseArray in the exact same order as done for `BashArray` in the codebase of Oil in the past. However, investigating the history of BashArray would require another big effort.

spec/ble-idioms.test.sh
```diff
@@ -431,3 +431,93 @@ declare -a sp1=([10]=a [20]=b [99]=c [100]=1 [101]=2 [102]=3)
...
+  ^~~~
+[ stdin ]:6: fatal: Index -11 is out of bounds for array of length 10
+## END
+
+## N-I bash/zsh/mksh/ash status: 0
+## N-I bash/zsh/mksh/ash STDERR:
+## END
+
+
+#### SparseArray: Negative out-of-bound index with a[i]=v (2/2)
```
andychu, line 503: This seems like the same test case as the one above?  

It doesn't hurt to have duplicates, but sometimes it makes it harder to find 

Also I wonder if we should not make so many errors fatal, and instead make them recoverable ... Then you can put them in the same test case block

We are perhaps not that consistent about this, and bash isn't either

andychu (commented):
OK thanks!

spec/ble-idioms.test.sh
akinomyoga: > This seems like the same test case as the one above?

Yeah, if one knows the implementation, it feels a duplicate. However, there was a case #2153, where `-21` caused an error, but `-20` did not for an array of length 10.

> Also I wonder if we should not make so many errors fatal, and instead make them recoverable ... Then you can put them in the same test case block

Yeah, maybe some of the failures can be recoverable. However, in the present case, Bash does die.

akinomyoga:
Thanks.

@andychu Do you have a comment on https://github.com/oils-for-unix/oils/pull/2157#discussion_r1864903200?

core/state.py
akinomyoga: I received a reply in Zulip: https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/mops.2EBigInt.28i.29.20vs.2E.20mops.2EIntWiden.28i.29/near/485491386

PR #2159, opened 2024-12-01
[core/dev] Fix xtrace output for `a+=(v)` and `declare a+=(v)`

akinomyoga:
In the current `master`, xtrace outputs wrong commands for `a+=(v)` and `declare a+=(v)` in respective ways. Both are two different bugs. This PR fixes both.

### "`declare a=()`" for `declare a+=()`

For `declare a+=()`, xtrace does not consider the type of the assignment `=` vs `+=`.

```console
$ bin/osh -c 'set -x; declare a+=(v)'
+ declare a=(v)
```

829dddec0 fixes this problem.

### "`a+=(1 2)`" for `a+=(2)`

For `a+=(v)`, xtrace outputs the state of the array after the assignment as the value on the right-hand side.

```console
$ bin/osh -c 'a=(1); set -x; a+=(2)'
+ a+=(1 2)
```

8a586ab2d fixes this.

### DIFF
--- a/spec/xtrace.test.sh
+++ b/spec/xtrace.test.sh
@@ -384,3 +384,16 @@ declare a+=(2)
 ## END
 ## N-I dash/mksh STDERR:
 ## END
+
+
+#### Regression: xtrace for "a+=(v)"
+case $SH in dash|mksh) exit ;; esac
+
+a=(1)
+set -x
+a+=(2)
+## STDERR:
++ a+=(2)
+## END
+## N-I dash/mksh STDERR:
+## END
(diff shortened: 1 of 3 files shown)


spec/xtrace.test.sh
```diff
@@ -367,3 +367,33 @@ ok
...
+## END
+## OK bash STDERR:
++ a+=('2')
++ declare a
+## END
+## N-I dash/mksh STDERR:
+## END
+
+
+#### Regression: declare -a a+=(v)
```
andychu, line 389: shouldn't this be named without `declare` ?     They are 2 different cases

akinomyoga: Ah, right! I fixed it.

PR #2161, opened 2024-12-01
[core refactor] Refactor the code for BashArray/BashAssoc `unset -v a[i]`

akinomyoga:
This moves the codes for `unset -v 'a[i]'`.

### DIFF
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -198,6 +198,12 @@ def BashAssoc_SetElement(assoc_val, key, s):
     assoc_val.d[key] = s
 
 
+def BashAssoc_UnsetElement(assoc_val, key):
+    # type: (value.BashAssoc, str) -> None
+
+    mylib.dict_erase(assoc_val.d, key)
+
+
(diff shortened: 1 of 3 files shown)


akinomyoga:
### Two different array lengths

One is the number of elements obtained by `${#arr[@]}`. The other is the max index plus one, which is used to locate the element specified by a negative index. Currently, the former is obtained by `BashArray_Length`. I would like to add a function to get the latter. How should I name these functions? Or, to ask differently, how should we call these two types of array lengths by distinct names?

andychu:
Oh sorry I had to delete the soil-staging branch for a different reason, and that meant PRs were deleted I think

akinomyoga:
I think #1958 is also a PR that was automatically closed by GitHub. I reopened it.

akinomyoga:
I'll rework on this PR after #2163 is settled.

andychu:
Thanks!

akinomyoga:
Thanks

PR #2163, opened 2024-12-02
[refactor] Rename `${#arr[@]}` from `_Length` to `_Count`

akinomyoga:
I here suggest changing the name `_Length` to `_Count`. This is because there are two types of "lengths" in arrays.  One is the count of elements.  The other is the max index plus one.  They differ for sparse arrays.  For example, for an array constructed by `a=(); a[0]=x a[9]=y`, the former is 2, while the latter is 10.  This PR suggests calling the former "count", and the latter "length".

Ref: original discussion: https://github.com/oils-for-unix/oils/pull/2161#issuecomment-2510146764

This PR also contains the fix for the error message of a negative out-of-bound index. I added the "array length" to the error message of the negative array index in [\#2160 `core/state.py`](https://github.com/oils-for-unix/oils/pull/2160/files#diff-947c57d83e952ecdb670a97c89273b980cf665a12fd551ef0dc2e2d401593a90R1975), but that "length" was actually "count". Commit d9ca1a0d8c89543d43bccf22b334145b4f33d36d fixes it so that it uses the proper "length".

### DIFF
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -26,7 +26,7 @@ def BigInt_Less(a, b):
 # representation of SparseArray come here.
 
 
-def BashArray_Length(array_val):
+def BashArray_Count(array_val):
     # type: (value.BashArray) -> int
 
     # There can be empty placeholder values in the array.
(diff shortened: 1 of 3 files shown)


andychu:
Thanks

akinomyoga:
Thank you

PR #2165, opened 2024-12-02
[osh] Check negative out-of-bound index for `${a[i]}` and `$((a[i]))`

akinomyoga:
In the current master, they do not cause any error or warning.

```console
$ bin/osh -c 'a=(1); echo ${a[-2]}; echo $((a[-2]))'

0
```

while Bash produces warning messages (though they are just messages, and it does not cause any error):

```console
$ bash -c 'a=(1); echo ${a[-2]}; echo $((a[-2]))'
bash: line 1: a: bad array subscript

bash: line 1: a: bad array subscript
0
$
```

This PR checks the negative out-of-bound index in two cases `${a[i]}` and `$((a[-2]))` and outputs warning messages.

### DIFF
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -45,6 +45,7 @@ from _devbuild.gen.value_asdl import (
     RegexMatch,
 )
 from core import alloc
+from core import bash_impl
 from core import error
 from core.error import e_die, e_die_status, e_strict, e_usage
 from core import num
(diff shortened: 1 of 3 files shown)


osh/sh_expr_eval.py
```diff
@@ -730,7 +734,17 @@ def Eval(self, node):
...
                             small_i = mops.BigTruncate(
                                 self.EvalToBigInt(node.right))
-                            s = word_eval.GetArrayItem(array_val.strs, small_i)
+                            s, error_code = word_eval.GetArrayItem(
+                                array_val.strs, small_i)
+                            if error_code == 1:
+                                # Note: Bash outputs warning but does not make
+                                # it a real error.  We follow the Bash behavior
+                                # here.
+                                self.errfmt.Print_(
```
andychu, line 743: Another thing we can do is make a hard error when `shopt --set strict_array`, which we already have

There is an `e_strict()` pattern that we sometimes use for that

Though that increases the test matrix, so maybe we can do that in another pass

akinomyoga: Moved to #2168.

PR #2166, opened 2024-12-02
[osh refactor] Refactor the code for BashArray/BashAssoc `[[ -v a[i] ]]`

akinomyoga:


osh/sh_expr_eval.py
```diff
@@ -1026,24 +1027,18 @@ def _IsDefined(self, s, blame_loc):
...
-                        return False
-
-                if index < n:
-                    return val.strs[index] is not None
-
-                # out of range
-                return False
+                result, error_code = bash_impl.BashArray_HasElement(val, index)
+                if error_code == 1:
+                    e_die(
```
andychu, line 1032: Hm the other change was just a warning, and this one is an error

```
$ bash -c 'test -v a[-1]; echo status=$?'
bash: line 1: a: bad array subscript
status=1
```

I think this argues for the `e_strict()` style then ... we can match bash EXACTLY if we want, but then OSH should have `shopt -s strict_array`, which is CONSISTENT

i.e. this part would not change, but the previous `$((a[-1]))` and `${a[-1]}` should change to be consistent, I think

akinomyoga: Moved to #2168.

PR #2167, opened 2024-12-02
[core/state] Implement SparseArray `unset -v sp[i]`

akinomyoga:


core/bash_impl.py
```diff
@@ -300,6 +300,27 @@ def SparseArray_SetElement(sparse_val, index, s):
...
+
+    index, ok = _SparseArray_CanonicalizeIndex(sparse_val, index)
+    if not ok:
+        return 1  # error_code: out-of-range index
+    mylib.dict_erase(sparse_val.d, index)
+
+    # update max_index
+    if mops.Equal(index, sparse_val.max_index):
+        sparse_val.max_index = mops.MINUS_ONE
+        for index in sparse_val.d.keys():
```
andychu, line 318: I think you can remove `keys()` here -- it's actually less efficient in Python 2, and in mycpp

i.e. `for index in sparse_val.d` will just iterate over the dictionary keys

while  `.keys()` makes a copy of it, and then iterates over it

core/bash_impl.py
```diff
@@ -300,6 +300,27 @@ def SparseArray_SetElement(sparse_val, index, s):
...
 
+def SparseArray_UnsetElement(sparse_val, index):
+    # type: (value.SparseArray, mops.BigInt) -> int
+    """A non-zero return value represents an error code. If this
+    function returns 1, the specified index is out of range.
+
+    """
+
+    index, ok = _SparseArray_CanonicalizeIndex(sparse_val, index)
+    if not ok:
```
andychu, line 311: It might be better to use integer error codes for both, rather than bool, and int

Also I think it would be nice to have

`if error == bash_impl.IndexOutOfRange`

instead of hard-coding `1`  .    Then we don't need the comment, and it is more readable

core/state.py
```diff
@@ -2288,6 +2288,15 @@ def Unset(self, lval, which_scopes):
                         raise error.Runtime(
                             "%s[%d]: Index is out of bounds for array of length %d"
                             % (var_name, lval.index, n))
+                elif val.tag() == value_e.SparseArray:
+                    val = cast(value.SparseArray, UP_val)
+                    error_code = bash_impl.SparseArray_UnsetElement(
+                        val, mops.IntWiden(lval.index))
+                    if error_code == 1:
```
andychu, line 2295: Yeah since we are using this error code a lot, I think it's good to call it `bash_impl.IndexOutOfRange`

core/bash_impl.py
akinomyoga: Yeah, so we go back to the discussion at https://github.com/oils-for-unix/oils/pull/2157#issue-2708501765. I'm happy to refactor them using named constants.

There are already many codes using the current approach with `int`, so I'd rather do it globally in a separate PR than including the global refactoring in this PR (which is for `unset -v sp[i]`).

core/bash_impl.py
akinomyoga: Oh, I haven't thought that `Dict` itself would iterate the keys. Thank you. Also, I've naively thought `.keys()` would return an iterator. I'll fix it later.

core/bash_impl.py
akinomyoga: I fixed it and rebased the PR on top of the latest master.

core/bash_impl.py
akinomyoga: I fixed it and rebased the PR on top of the latest master.

akinomyoga:
Thanks

core/bash_impl.py
akinomyoga: I created PR #2175.

Issue #824, opened 2020-09-03
language semantics shouldn't depend on the compiler


andychu:
Perl even documents this:

https://perldoc.perl.org/perlnumber.html#Storing-numbers

```
Perl can internally represent numbers in 3 different ways: as native integers, as native floating point numbers, and as decimal strings. Decimal strings may have an exponential notation part, as in "12.34e-56" . Native here means "a format supported by the C compiler which was used to build perl".
```

I think integer and floating point semantics are reasonably well defined though?

Issues:

- Oil uses 32 bit int (even on 64 bit) while other shells use 64 bits (long int)
- `printf` had an issue, with negative numbers I think

andychu:
Integer overflow is undefined behavior in C.

When undefined behavior is invoked, the behavior depends on the compiler.  So we should avoid overflow for user operations.

Instead we can use these compiler extensions:

https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

https://clang.llvm.org/docs/LanguageExtensions.html

----

Also TODO:

- document portability
  - we use the EXTGLOB extension for GNU glob
  - and eventually these overflow builtins
  - otherwise it's ISO C++11 and POSIX

PR #2173, opened 2024-12-05
[spec/ble-idioms] Workaround for bash bug on certain platforms

andychu:


andychu:
The funny thing about this  is that we learned that  bash does not NOT support 64-bit array indices reliably, even the latest version of bash

there are 2 more point release here:

https://ftp.gnu.org/gnu/bash/


```
[ ]	[bash-5.2.32.tar.gz](https://ftp.gnu.org/gnu/bash/bash-5.2.32.tar.gz)	2024-08-02 11:45 	10M	 
[ ]	[bash-5.2.32.tar.gz.sig](https://ftp.gnu.org/gnu/bash/bash-5.2.32.tar.gz.sig)	2024-08-02 11:45 	95 	 
[ ]	[bash-5.2.37.tar.gz](https://ftp.gnu.org/gnu/bash/bash-5.2.37.tar.gz)	2024-09-23 18:02 	11M	 
[ ]	[bash-5.2.37.tar.gz.sig](https://ftp.gnu.org/gnu/bash/bash-5.2.37.tar.gz.sig)	2024-09-23 18:03 	95
```

But even if those fix it, than it will be a long time before they are rolled out to distros

---

So you could actually argue that OSH does NOT need 64 bit indices.  Because bash doesn't really have it.

If somebody wrote a script with OSH with 64 bit indices, and then tried to run it on bash, then it may not work.  But that is true in other cases too.

But probably there is no other option, so we should have it.  32-bit indices is not good enough either ....

akinomyoga:
What I'm unsure about is the condition that triggers the Bash bug.

- Is the problem always reproduced within Debian 10?
- Doesn't the problem arise in any other systems?

In particular, signed integer overflow is an undefined behavior (UB) in C++. This means that the compiler can optimize the code to generate random behaviors, which is likely related to environmental dependence. If this is the case, the behavior is unrelated to the operating system, but rather the compiler/library version, its default set of optimization options, the CPU type that the compiler detects, etc., which is really unpredictable. Even if we identify the exact explanation of the present environmental dependence, it would still be difficult to predict the result for the test deterministically. With the approach of this PR, we may pass the tests on the current CI system and our current local machines, but it doesn't ensure they pass the tests on other users' machines. For example, suppose a user downloads the source code of Oils for Unix and runs the tests, it's not ensured to pass the tests even though we know that there is an environment-dependent behavior of Bash 5.2.

> The funny thing about this is that we learned that bash does not NOT support 64-bit array indices reliably, even the latest version of bash

The situation is actually the same for the current implementation of OSH. Since OSH doesn't check the integer overflow either, it is actually causing UB now. Then, in the C++ perspective, the result is unpredictable. Currently, we haven't observed the unexpected behvior of OSH in the tests, but it's not because OSH is reliably implemented. It's working just by accident. In particular, the current code causes the integer overflow twice, and the result becomes consistent after the second overflow by accident.

> So you could actually argue that OSH does NOT need 64 bit indices. Because bash doesn't really have it.

We could argue Bash doesn't reliably implement it for an extreme case, but I don't see the reason to argue that we should specifically restrict ourselves to 32-bit indices. We can just avoid using an extreme index like `INTMAX_MAX`. If we would extend the argument to avoid something related together, the conclusion would largely depend on what we count as related. In the same logic as "we should not use 64-bit integer", we could argue "we should not use arrays because Bash's array is not reliable in an extreme case", or we could argue "we should not use integers at all in shell because Bash behaves unreliably in an extreme case", or even "we should not use Bash at all".

andychu:
Yeah I agree, we should have the 64-bit integers

And OSH also has a bug, but I mentioned that we definitely want to add overflow detection to `gc_mops.h`, and I added some test code for that recently

Just pointing out that it is  funny corner case

Also I think it's good we have enough tests to catch this, and similar bugs

andychu:
Also this is related to

- https://github.com/oils-for-unix/oils/issues/824

I think bash on 32-bit machines has 32-bit math.

But OSH on 32-bit machines should have 64-bit math, since we use `mops::BigInt` and `int64_t` everywhere

---

Though a problem is if we use the compiler builtins to check overflow, it is not available on some compilers

But I think it is available on 99% of compilers, and I looked into implementing overflow detection ourselves, and I think that would be a distraction from the rest of the project.   And the compiler developers can probably do it better/faster than we can,

So for at least for the near future the 1% of people will not have overflow checking, maybe they have to pass `./configure --unsafe-integer-math` or something

andychu:
Also I should mention there is a chance that this bug is related to our build process, not necessarily one in bash

we copy bash binaries across different OSes, which is not necessarily a valid thing to.    the libc ABI changes rarely

---

However I think I solved that problem recently:

There is a Debian 10 "bootstrap" Docker image, and a Debian 12 one:

https://github.com/oils-for-unix/oils/blob/master/deps/Dockerfile.wedge-bootstrap-debian-10

And I think everything should match.

---

But it is possible that there is a bug on our side.  The last time I re-built this Docker image was in June, when upgrading to bash 5.2

https://github.com/oils-for-unix/oils/blob/master/soil/host-shim.sh#L63

---

BUT I also observed this bug on my `lenny` Debian 10 machine.  And that has nothing to do with Docker.

So I am leaning toward it being a bash bug, since integer overflow bugs are not uncommon

andychu:
>  With the approach of this PR, we may pass the tests on the current CI system and our current local machines, but it doesn't ensure they pass the tests on other users' machines. For example, suppose a user downloads the source code of Oils for Unix and runs the tests, it's not ensured to pass the tests even though we know that there is an environment-dependent behavior of Bash 5.2.

(responding to this part)  Having the end users run tests is something I would like to do eventually, though unfortunately we are VERY far from it

---

Right now the tests are only for developers and CI.

They are coupled to Debian / Ubuntu -- differences like `/bin/ls` vs. `/usr/bin/ls` will break them, etc.

i.e. there are many cases where the tests will fail, but there is no bug in the shell.  

Also the tests depend on python2 and python3 I think

They are not pure POSIX, and I don't think POSIX even specifies enough for it

---

I would like to fix this -- for example, I would like to run on Alpine Linux with musl libc, and also FreeBSD or OpenBSD:

https://github.com/oils-for-unix/oils/blob/master/test/alpine.sh

If the "multiple outputs" change would help with that, I'd be open to it.  But I'm almost certain we are going to hit all sorts of other problems ifrst.

spec/ble-idioms.test.sh
```diff
@@ -1036,15 +1055,5 @@ echo "[${a[@]: -4}][${a[*]: -4}]"
...
 
-# Note: Bash behavior depends on the version and the environment.  In some
-# conditions, the index 0x7FFFFFFFFFFFFFFF hits an integer overflow bug.  Then
-# all slicing with a negative index results in empty lists:
-## BUG bash STDOUT:
-[][]
-[][]
-[][]
-[][]
-## END
```
akinomyoga, line 1047: Can we leave this part as a code comment? It's still good to keep in a comment what would be the possible output in the Bash with signed integer overflow.

akinomyoga (commented):
OK, it's good to skip the check in the environments confirmed to have a problem.

akinomyoga (approved):

PR #2174, opened 2024-12-06
Fix `bind -r` support with older readline fns

KingMob:
### About 
OK, this was pretty easy, as I suspected. The older method of unbinding still works; it's essentially like "binding" the keyseq to a null pointer, which readline interprets as wanting to remove the binding.

This is based off bash code added in 4.1 from 2011. The newer version is from bash 5.0 from 2019.

IIUC, this is slightly less efficient code, but no big deal. I left the newer code in place, but commented out. LMK if you want to remove it. Otherwise, I thought I'd delete it once the final pieces of bind -x are complete, and we're all done.

----

### Tests

I added a non-interactive test for `bind -r`. We can use `bind -q` to see what readline thinks the bindings are, but this can't test that the binding behavior is actually removed.

I incremented oils_cpp_failures_allowed, since I'm assuming the new test won't succeed, but lmk if that's wrong.

Since the interactive bind -r test uses a custom shell function, it's dependent on bind -x behavior, and so, still fails.

### DIFF
--- a/builtin/readline_osh.py
+++ b/builtin/readline_osh.py
@@ -138,10 +138,8 @@ class Bind(vm._Builtin):
                 if arg.u is not None:
                     readline.unbind_rl_function(arg.u)
 
-                if 0:
-                    # disabled until we fix error with rl_function_of_keyseq_len()
-                    if arg.r is not None:
-                        readline.unbind_keyseq(arg.r)
+                if arg.r is not None:
+                    readline.unbind_keyseq(arg.r)
 
                 if arg.x is not None:
                     self.errfmt.Print_("warning: bind -x isn't implemented",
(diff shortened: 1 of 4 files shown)


andychu:
Cool thanks!

I noticed the new spec tests passes, but the `spec/stateful` test still fails?

```
test/stateful.sh bind-quick -r 1
```

```

case    osh     bash
Ignoring osh-cpp failure: 1 test bind -r for removing bindings
1       FAIL    ok      test bind -r for removing bindings
        0/5 ok
spec/stateful/bind.py: Expected 2 failures, got 1

```

both in CI and on my machine 

Is this a problem with the test, or the code?

andychu:
Oh sorry, I see you already answered that!   

Yes let's talk about `bind -x` next, I feel like the callback may be tricky ... I will start a thread on Zulip

Issue #2168, opened 2024-12-03
Revisit array warnings/errors with `shopt --set strict_array`


akinomyoga:
_Originally posted by @andychu in https://github.com/oils-for-unix/oils/pull/2165#discussion_r1866867535_

> Another thing we can do is make a hard error when `shopt --set strict_array`, which we already have
>
> There is an `e_strict()` pattern that we sometimes use for that
>
> Though that increases the test matrix, so maybe we can do that in another pass

I don't have the capacity to handle this in parallel with the currently ongoing bug fixes, refactoring, and SparseArray implementation of arrays.

Let's create an issue before forgetting about this.

akinomyoga:
The above is for `${a[-1]}` and `$((a[-1]))`. The following is about `test -v a[-1]`.

_Originally posted by @andychu in https://github.com/oils-for-unix/oils/pull/2166#discussion_r1866873427_

> Hm the other change was just a warning, and this one is an error
> 
> ```
> $ bash -c 'test -v a[-1]; echo status=$?'
> bash: line 1: a: bad array subscript
> status=1
> ```
> 
> I think this argues for the `e_strict()` style then ... we can match bash EXACTLY if we want, but then OSH should have `shopt -s strict_array`, which is CONSISTENT
> 
> i.e. this part would not change, but the previous `$((a[-1]))` and `${a[-1]}` should change to be consistent, I think

There is also `a[-1]=v`. In addition, each of them has codes for BashArray, BashAssoc, and SparseArray.

- [ ] Are there any other places?
- [ ] How does Bash behave?
- [ ] What would be the default `osh` behavior for each, and how would it be changed under `strict_array`?

andychu:
OK yes, we can do this after SparseArray

I think my basic viewpoint is

- bash is inconsistent about errors.  But sometimes it is easier to just copy bash, since it makes the spec tests pass.
  -   if bash gives a warning, give a warning.  if it is a fatal error, then fatal error.
- However, `shopt --set strict_array` should make ALL index of bounds a fatal error.   That is more consistent and easier to explain/document.  I think it's just a better experience for the user

So as another pass, we just `grep IndexOutOfBounds */*.py`, and make sure that if a fatal error is NOT thrown, then we throw it with `strict_array`

andychu:
I noticed that bash only complains about out of bounds **negative** indices, but not positive

That doesn't make sense to me

```
[...]

PR #2175, opened 2024-12-06
[core/bash_impl] Use `error_code_t` to return error

akinomyoga:
> https://github.com/oils-for-unix/oils/pull/2167#discussion_r1866878249
>
> Also I think it would be nice to have
> 
> `if error == bash_impl.IndexOutOfRange`
> 
> instead of hard-coding `1` . Then we don't need the comment, and it is more readable

This introduces an enum type to return the error status of a negative out-of-bound index of indexed arrays. To add the enum, I added `error_code = OK | IndexOutOfRange` in `runtime.asdl` (which may be extended to include other types of errors in the future). If there is a better way, please let me know.

### DIFF
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -1,5 +1,6 @@
 """bash_impl.py - implements operations on Bash data structures"""
 
+from _devbuild.gen.runtime_asdl import error_code_e, error_code_t
 from _devbuild.gen.value_asdl import value
 
 from data_lang import j8_lite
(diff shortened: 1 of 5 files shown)


andychu:
This looks great, thank you!  ASDL is nice to use like this

Given what I wrote in:

https://github.com/oils-for-unix/oils/issues/2168

I think eventually we will have

```
OK | BadNegativeIndex | BadPositiveIndex
```

because bash happens to behave differently in these cases.

But I guess we can do that when we get to it

PR #2180, opened 2024-12-07
[refactor] Use `Optional[str]` instead of `str` when `None` is possible

akinomyoga:
https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/2024-11-22.20Meeting/near/486625799

There are two declarations of the form `v = None  # type: str` and `v = None  # type: Optional[str]`. This PR unifies them to `v = None  # type: Optional[str]`.

- 35481d91f1b476f248dacda352e157197d91c3e7 is fixes for the codes I contributed.
- 9493b3b1d is the fixes for the trivial case: rewriting from `= None  # type: str` to `= None  # type: Optioanl[str]`

I tried to find non-trivial cases where the variable is first assigned with an object and then with `None`, but it seems in most cases, the object assigned first has a class type. I haven't checked all the files, but I decided not to fix these non-trivial cases within this PR.

### DIFF
--- a/builtin/func_misc.py
+++ b/builtin/func_misc.py
@@ -650,7 +650,7 @@ class SparseOp(vm._Callable):
         #i = mops.BigTruncate(rd.PosInt())
         op_name = rd.PosStr()
 
-        no_str = None  # type: str
+        no_str = None  # type: Optional[str]
 
         if op_name == 'len':  # ${#a[@]}
             rd.Done()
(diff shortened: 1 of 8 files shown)


andychu:
OK thanks, I noted on Zulip that in **theory** we could turn on MyPy's strict optional checking

this will do static checking of None

In practice there were 600+ errors, and I'm not sure if the cost is worth the benefit (although I could be convinced!)

But yeah I think these changes can help the readability of the code, and make it more consistent, even if it's not checked

PR #2184, opened 2024-12-08
[osh] Implement SparseArray `${a[@]:offset:count}`

akinomyoga:
This PR actually largely rewrites the handling of the BashArray slicing together.

The second commit adds the support for BigInt *offset* in the construct `${a[@]:offset:count}`.

### DIFF
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -22,6 +22,18 @@ def BigInt_Less(a, b):
     return mops.Greater(b, a)
 
 
+def BigInt_GreaterEq(a, b):
+    # type: (mops.BigInt, mops.BigInt) -> bool
+
+    return not mops.Greater(b, a)
+
+
+def BigInt_LessEq(a, b):
+    # type: (mops.BigInt, mops.BigInt) -> bool
+
+    return not mops.Greater(a, b)
+
+
 #------------------------------------------------------------------------------
 # All BashArray operations depending on the internal
 # representation of SparseArray come here.
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -367,38 +367,91 @@ def _PerformSlice(
             substr = s[byte_begin:byte_end]
             result = value.Str(substr)  # type: value_t
 
-        elif case(value_e.BashArray):  # Slice array entries.
-            val = cast(value.BashArray, UP_val)
+        elif case(value_e.BashArray,
+                  value_e.SparseArray):  # Slice array entries.
             # NOTE: This error is ALWAYS fatal in bash.  It's inconsistent with
             # strings.
             if has_length and length < 0:
                 e_die("Array slice can't have negative length: %d" % length,
                       loc.WordPart(part))
 
-            orig = bash_impl.BashArray_GetValues(val)
+            offset = mops.IntWiden(begin)
+            if bash_impl.BigInt_Less(offset, mops.ZERO):
+                # ${@:-3} starts counts from the end
+                if val.tag() == value_e.BashArray:
+                    val = cast(value.BashArray, UP_val)
+                    array_length = mops.IntWiden(
+                        bash_impl.BashArray_Length(val))
+                elif val.tag() == value_e.SparseArray:
+                    val = cast(value.SparseArray, UP_val)
+                    array_length = bash_impl.SparseArray_Length(val)
+                else:
+                    raise AssertionError()
 
-            # Quirk: "begin" for positional arguments ($@ and $*) counts $0.
-            if arg0_val is not None:
-                new_list = [arg0_val.s]
-                new_list.extend(orig)
-                orig = new_list
+                # The array length counts $0 for $@ and $*
+                if arg0_val is not None:
+                    array_length = mops.Add(array_length, mops.ONE)
 
-            n = len(orig)
-            if begin < 0:
-                i = n + begin  # ${@:-3} starts counts from the end
+                offset = mops.Add(offset, array_length)
+
+            if bash_impl.BigInt_Less(offset, mops.ZERO):
+                strs = []  # type: List[str]
             else:
-                i = begin
-            strs = []  # type: List[str]
-            if i >= 0:
-                count = 0
-                while i < n:
-                    if has_length and count == length:  # length could be 0
-                        break
-                    s = orig[i]
-                    if s is not None:  # Unset elements don't count towards the length
-                        strs.append(s)
-                        count += 1
-                    i += 1
+                # Quirk: "begin" for positional arguments ($@ and $*) counts $0.
+                prepends_arg0 = False
+                if arg0_val is not None:
+                    if bash_impl.BigInt_Greater(offset, mops.ZERO):
+                        offset = mops.Sub(offset, mops.ONE)
+                    elif not has_length or length >= 1:
+                        prepends_arg0 = True
+                        length = length - 1
+
+                if has_length and length == 0:
+                    strs = []
+
+                elif val.tag() == value_e.BashArray:
+                    val = cast(value.BashArray, UP_val)
+                    orig = bash_impl.BashArray_GetValues(val)
+                    n = len(orig)
+
+                    strs = []
+                    i = mops.BigTruncate(offset)
+                    count = 0
+                    while i < n:
+                        if has_length and count == length:  # length could be 0
+                            break
+                        s = orig[i]
+                        if s is not None:  # Unset elements don't count towards the length
+                            strs.append(s)
+                            count += 1
+                        i += 1
+
+                elif val.tag() == value_e.SparseArray:
+                    val = cast(value.SparseArray, UP_val)
+
+                    # TODO: We may optimize this by finding the first index
+                    # using the binary search.  Furthermore, the sorting by
+                    # SparseArray_GetKeys can be replaced with the heap sort so
+                    # that we only extract the first LENGTH elements of the
+                    # indices greater or equal to OFFSET.
+                    i = 0
+                    for index in bash_impl.SparseArray_GetKeys(val):
+                        if bash_impl.BigInt_GreaterEq(index, offset):
+                            break
+                        i = i + 1
+
+                    if has_length:
+                        strs = bash_impl.SparseArray_GetValues(val)[i:i+length]
+                    else:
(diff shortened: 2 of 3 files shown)


akinomyoga:
Test [ble-idioms\#31](http://op.oilshell.org/uuu/github-jobs/8666/ovm-tarball.wwz/_tmp/spec/osh-py/ble-idioms.html) is failing for Bash somehow. The test passes in my environment. I currently don't have an idea. Is Bash on the CI configured in a special way? For example, is `intmax_t` (which is used as an arithmetic integer type by Bash) 32-bit in Bash on the CI?

In addition, the number of failing tests differs between the Python and C++ versions:

> http://op.oilshell.org/uuu/github-jobs/8666/ovm-tarball.wwz/_tmp/soil/logs/osh-spec.txt
>
> ```bash
> ...
> ble-idioms: FATAL: Got 1 total failures, but 0 are allowed
> ...
> ```
>
> http://op.oilshell.org/uuu/github-jobs/8666/cpp-spec.wwz/_tmp/soil/logs/osh-all.txt
>
> ```bash
> ...
> ble-idioms: FATAL: Got 2 Oils C++ failures, but 0 are allowed
> ...
> ```

andychu:
Hm weird, I can confirm it passes on my machine too

I checked out your branch

I can run it inside the CI docker containers to see what's wrong

I'll look at it after looking at the other PRs

andychu:
OK I can reproduce this locally

This shouldn't happen -- I guess there is a bug lower down in our C++, or in the translation

I will look at it!

```
$ ./NINJA-config.sh
$ ninja

$ test/spec-cpp.sh run-file ble-idioms -r 31 -v
mkdir: created directory '_tmp/spec/osh-cpp'

ble-idioms.test.sh
 31     895     pass    FAIL    SparseArray: ${a[@]:BigInt}
case: 31

[osh-cpp stdout]
Expected '[x][x]\n[y x][y x]\n[z y x][z y x]\n[z y x][z y x]\n'
Got      '[x][x]\n[y x][y x]\n[z y x][z y x]\n[][]\n'


```

andychu:
Oh I suspect this must be a 32-bit vs. 64-bit math issue?  Let me see

andychu:
Oh I think it is a bug in the indexing operation in C++!  After

```
a[0x7FFFFFFFFFFFFFFF]=x
```

then the length is still 3, not 4

I am debugging in GDB a bit


```
(gdb) p array_length 
$3 = 140737488342784
(gdb) n
45521           if (arg0_val != nullptr) {
(gdb) p array_length 
$4 = 3
(gdb) p offset
$5 = -4

```

andychu:
Ah OK I'm almost certain the bug is that we are using `int` here, not `BigInt`

https://github.com/oils-for-unix/oils/blob/master/core/value.asdl#L52

and then that means the statement

```
a[0x7FFFFFFFFFFFFFFF]=x
```

works in Python, but not C++!  Because in C++ the int overflows ...

Maybe for this change we can comment out those large integers, and add a TODO

Since it's not strictly related

---

But that's a definite bug, so thanks for writing good test cases and finding it !!

andychu:
Actually one thing we can do is 

- add a separate failing test case for array indices over 32 bits
- allow the failure with an annotation at the top of the file

```
$ grep cpp_failures */*.test.sh
spec/bugs.test.sh:# oils_cpp_failures_allowed: 1
```

then the CI will pass

---


so then we can fix that separately from this PR

akinomyoga:
> * add a separate failing test case for array indices over 32 bits
> * allow the failure with an annotation at the top of the file

I added these in commit fb8e9fe7bc74a840ddc3e7c152ca0582d5de9715.

akinomyoga:
The C++ test failure for `${@:offset:length}` was just because of the difference in the value of `$0`. `$0` in the Python version is just `osh`, while `$0` in the C++ version is an absolute path. Is it because of the way how both commands are called by the test script? Anyway, this wasn't a translation issue. I modified the test case to cover both cases: [`fb8e9fe..0450ebe`](https://github.com/oils-for-unix/oils/compare/fb8e9fe7bc74a840ddc3e7c152ca0582d5de9715..0450ebea60502d3204c0177071348ee9a01cf363).

akinomyoga:
Bash still fails on the CI for [case ble-idioms \#35](http://op.oilshell.org/uuu/github-jobs/8682/ovm-tarball.wwz/_tmp/spec/osh-py/ble-idioms.html#details-35-bash).

andychu:
Oh weird **bash** is also failing in the CI !  But not on my machine

there are a bunch of errors here lol

http://op.oilshell.org/uuu/github-jobs/8682/ovm-tarball.wwz/_tmp/spec/osh-py/ble-idioms.html

This test case is hard for shells to run :)

Let me see why bash is failing

andychu:
Oh weird there is a difference between bash 4.4 and bash 5.2

But our CI is supposed to **pin** the versions

```
$ bash -c "$(cat _tmp/35)"
[][]
[][]
[][]
[][]
andy@lenny ~/git/oilshell/oil (SparseArray-slice)$ bash --version
GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu)
```

versus


```
$ bash -c "$(cat _tmp/35)"
[x][x]
[y x][y x]
[z y x][z y x]
[z y x][z y x]
andy@hoover:~/git/oilshell/oil$ bash --version
GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)

```
```

**Comment by andychu** [comment]
Oh this is weird, our prebuilt version of bash has this problem too

```
$ ~/wedge/oils-for-unix.org/pkg/bash/5.2.21/bin/bash -c "$(cat _tmp/35)"
[][]
[][]
[][]
[][]

```

**Comment by andychu** [comment]
This is very weird, I think this is a bash bug ... the same code shouldn't behave in 2 different ways compiled on 2 different machines

machine info

```
andy@hoover:~/git/oilshell/oil$ hostname; cat /etc/debian_version; ldd ~/wedge/oils-for-unix.org/pkg/bash/5.2.21/bin/bash
hoover
12.0
        linux-vdso.so.1 (0x00007ffe5cd77000)
        libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007f5079467000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5079286000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f50795ef000)

```

and

```
$ hostname; cat /etc/debian_version; ldd ~/wedge/oils-for-unix.org/pkg/bash/5.2.21/bin/bash
lenny
buster/sid
        linux-vdso.so.1 (0x00007fff7f7ea000)
        libtinfo.so.5 => /lib/x86_64-linux-gnu/libtinfo.so.5 (0x00007f5d9935b000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f5d99157000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5d98d66000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f5d998be000)

```

**Comment by andychu** [comment]
Those are two of my desktop machines, and then you can see  the Docker image OS info here:

http://op.oilshell.org/uuu/github-jobs/8682/ovm-tarball.wwz/_tmp/soil/logs/os-info.txt

```
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```


So on the Debian buster OS images, bash 5.2.21 is producing the `[][]` output, not the other output

Seems like a bash bug to me

----

I guess one thing we can do is do

```
case $SH in bash) exit ;; esac
```

It is a bit unfortunate that we won't have the comparison, but I think it's fine.



---

I wonder if this is actually the **same bug** in bash as OSH -- are they using `int` rather than `int64_t` ???

But why would it be different on Debian buster?    Different GCC defaults for integer sizes ??    Or some other type that is `typedef` ???

akinomyoga:
OK, I tried it locally and found that behavior changes depending on the Bash version. Actually, as far as I tried, the behavior has changed to the expected one in 5.2, though it might also depend on some factors in the environment and the compiler as you pointed out.

I slightly modified the test case so that it does not use `0x7FFFFFFFFFFFFFFF` (`INTMAX_MAX`), then the behavior became consistent in all the Bash versions I tested. This means that this is an `intmax_t` overflow issue.

Actually, the reason that I included the test with `INTMAX_MAX` is exactly that I anticipated a possible bug in handling the integer overflow. The "length" of the array is *max index plus one* so if *max index* is already the maximum representable integer, it may cause an issue. Then, the test hit the overflow in Bash instead of one in OSH. It is not surprising because traditional C programs typically haven't cared about the integer overflow so much.

Does OSH care about integer overflows in general? If one starts to care about overflow, one needs to perform the boundary check for every single integer operation, but as far as I remember, I haven't seen these codes to check integers in the OSH codebase. For example, the operations in `mops.py` and `gc_mops.h` do not seem to include the boundary checks.

akinomyoga:
> I guess one thing we can do is do
> 
> ```
> case $SH in bash) exit ;; esac
> ```

I can technically exclude Bash from the test targets, but I added the `BUG` section for Bash in commit 68cf14a48541a18000955f1c5bf0d4770423b058. This makes the test on the CI succeed, but it fails locally.

Currently, the test framework (`test/sh_spec.py`) seems to require the shell output to match one possibility exactly, but can I extend it to match one of multiple possible outputs?

andychu:
I responded about big integers on - https://oilshell.zulipchat.com/#narrow/channel/449378-big-integer/topic/Scope.20cut.3A.20let's.20use.20compiler.20builtins.20to.20check.20for.20overflow

Thanks for changing the tests!

We **can** assert different output for bash 4.4 and bash 5.2!  You can see some examples where we have `compare_shells: bash-4

However this appears to be an environment issue, not a version issue!  So I think what we have is OK for now

And thanks for anticipating the boundary condition bug -- those are good to test!  And we found bugs in OSH and bash, which is great

PR #2185, opened 2024-12-08
[osh] Implement BashAssoc `${a[@]@Q}`

akinomyoga:


akinomyoga:
Spec test [`var-op-bash`](http://op.oilshell.org/uuu/github-jobs/8667/ovm-tarball.wwz/_tmp/spec/osh-py/var-op-bash.html#details-16-osh) seems to fail, but I realized that the current master produces `''` for `${unset_variable@Q}`:

```bash
$ bash -c 'echo ${unset_variable@Q}'

$ bin/osh -c 'echo ${unset_variable@Q}'
''
$
```

Before this PR, this behavior of OSH should be fixed first. Let me make this PR Draft PR until it's fixed.

akinomyoga:
This again fails only in the C++ version, though the [Results](http://op.oilshell.org/uuu/github-jobs/8715/cpp-spec.wwz/_tmp/spec/osh-cpp/compare.html) page linked from [Jobs for run 8715](http://op.oilshell.org/uuu/github-jobs/8715/#job-cpp-spec) says "wwz: 404 Not Found".

andychu:
Hm yes there is a bug for some reason

the last part of the **log** is

http://op.oilshell.org/uuu/github-jobs/8715/cpp-spec.wwz/_tmp/soil/logs/osh-all.txt

```
assoc failed with status 1

*** 1 tests FAILED
test/spec-version.sh: line 71: zsh: command not found

Results: file:///home/uke/oil/_tmp/spec/osh-cpp/index.html
```

andychu:
ok and then I go to the index.html instead of compare.html

http://op.oilshell.org/uuu/github-jobs/8715/cpp-spec.wwz/_tmp/spec/osh-cpp/index.html

and then I see

http://op.oilshell.org/uuu/github-jobs/8715/cpp-spec.wwz/_tmp/spec/osh-cpp/assoc.html


I am not sure why it differs, I can look at it a bit

the spec test framework could probably show a nicer diff, or the test could be rewritten to show a nicer diff

andychu:
OK I bet this a **hash table ordering** bug

`['()<>&|' '"' $'\'\'' '[]*?']`

`[$'\'\'' '"' '()<>&|' '[]*?']`

In Python


```
a = {}  # this is UNORDERED
```



we have to do something like


```
a = NewDict()   # to make it ORDERED, it translates into the right C++ container
```


I'm not sure where this is, but I can look into it

andychu:
FWIW I think I fixed the issue where the comparison HTML is not written on failure

https://github.com/oils-for-unix/oils/commit/ec1ae69e0dfb4d796f994a8bf64d36933fa99b45

akinomyoga:
> In Python
> 
> ```
> a = {}  # this is UNORDERED
> ```
> 
> we have to do something like
> 
> ```
> a = NewDict()   # to make it ORDERED, it translates into the right C++ container
> ```
> 
> I'm not sure where this is, but I can look into it

Hmm, I probably don't understand what you mean because f84ef8358249107f7296117a8b90a4d31de1e3a0 made the C++ version not compile:

```
[1/4] cxx asan '' _gen/bin/oils_for_unix.mycpp.cc _build/obj/cxx-asan/_gen/bin/oils_for_unix.mycpp.o
FAILED: _build/obj/cxx-asan/_gen/bin/oils_for_unix.mycpp.o
build/ninja-rules-cpp.sh compile_one cxx asan '' _gen/bin/oils_for_unix.mycpp.cc _build/obj/cxx-asan/_gen/bin/oils_for_unix.mycpp.o _build/obj/cxx-asan/_gen/bin/oils_for_unix.mycpp.o.d
_gen/bin/oils_for_unix.mycpp.cc: In function ‘value_asdl::value_t* assign_osh::_ReconcileTypes(value_asdl::value_t*, bool, bool, syntax_asdl::word_t*)’:
_gen/bin/oils_for_unix.mycpp.cc:11059:40: error: ‘NewDict’ was not declared in this scope; did you mean ‘NewList’?
11059 |         return Alloc<value::BashAssoc>(NewDict());
      |                                        ^~~~~~~
      |                                        NewList
_gen/bin/oils_for_unix.mycpp.cc: In member function ‘virtual int assign_osh::NewVar::Run(runtime_asdl::cmd_value::Assign*)’:
_gen/bin/oils_for_unix.mycpp.cc:11251:44: error: ‘NewDict’ was not declared in this scope; did you mean ‘NewList’?
11251 |             rval = Alloc<value::BashAssoc>(NewDict());
      |                                            ^~~~~~~
      |                                            NewList
Command exited with non-zero status 1
_build/obj/cxx-asan/_gen/bin/oils_for_unix.mycpp.o { elapsed: 1.86, max_RSS: 310944 }
```

andychu:
hm can you try:

```
tmp = NewDict()  # type: Dict[str, str]
return value.BashAssoc(tmp)
```

Sorry this was an existing bug, and also our mycpp translator is limited

It doesn't propertly understand `NewDict()` as an expression, maybe because there is no place to put the type, like `NewDict[str,str]()` etc.   That is a Python 3 idiom but we don't use it anywhere, not sure if it works in Python 2

The translator is pretty limited and specific, although I did spend the last 10 days making the code a lot nicer, so we can improve it

andychu:
I tried to enhance the mycpp translator to handle `BashAssoc(NewDict())`

Unfortunately the type it infers from  the inner expression is

```
DICT TYPE builtins.dict[builtins.str, Any]
```

I am not sure why MyPy gives it an Any type, so for now I think we have to use the workaround

---

I tried the `tmp` variable and it works

(Long term I would like to have our own type checker, not rely on an old version of MyPy, but that work is a bit far away)

akinomyoga:
Thank you. [`f84ef835..fb13a2ed`](https://github.com/oils-for-unix/oils/compare/f84ef8358249107f7296117a8b90a4d31de1e3a0..fb13a2edcb806c42826b8ee5ee42ee7364b77b19) seems to have fixed it.

spec/assoc.test.sh
```diff
@@ -812,3 +812,24 @@ orange is orange
...
+echo "[${a[*]@Q}]"
+
+## STDOUT:
+[$'\'\'' '"' '()<>&|' '[]*?']
+[$'\'\'' '"' '()<>&|' '[]*?']
+## END
+
+## OK bash STDOUT:
+['[]*?' ''\'''\''' '"' '()<>&|']
+['[]*?' ''\'''\''' '"' '()<>&|']
```
andychu, line 834: hm bash's quoting is very odd!   It contains extra empty strings

It could just be `\'\'`

andychu:
thank you!

spec/assoc.test.sh
akinomyoga: I thought you knew this simple quoting method. For an arbitrary string, one can single-quote it by replacing all the occurrences of `'` to `'\''` and surrounding it by `'...'`. To express the algorithm in a shell script, for an arbitrary string `$str`, one can generate a quoted value by `quoted="'${str//$q/$Q}'"`, where `q=\' Q="'\''"`. Didn't know this?

andychu: Oh yes that's true ... I did know that idiom, but I didn't recognize it there

So yes it is a correct and simple algorithm, but I guess this is one of the worst cases for it

---

In Oils we usually prefer to print the string on a single line, even if it contains `\n`, which means you have to use the `$''` kind of string ...   We might want more control over that, in people want to print POSIX strings, but so far there doesn't seem like demand, and bash sometimes uses `$''` strings as well

akinomyoga:
Thanks!

spec/assoc.test.sh
akinomyoga: > Oh yes that's true ... I did know that idiom, but I didn't recognize it there

Ah, yeah, It's not so obvious that this is generated by that idiom. I probably got used to it too much so that I can immediately recognize it subconsciously.

> (In Oils we usually prefer to print the string on a single line, even if it contains `\n`, which means you have to use the `$''` kind of string ...

Yeah, I was searching QSN to check whether you have mentioned the idiom in some of the pages of QSN.

andychu: Oh "QSN" is now "J8 Notation", although they are virtually the same for strings, it's C-style `\`  escapes, which POSIX shell doesn't have

I think Oils could better support/document the pure POSIX idiom -- we probably should have a builtin function, even though it is trivial to write it in OSH/YSH 

The idiom produces POSIX shell strings, but not J8 strings, which don't have the "word joining"

(Actually "word joining" like `foo\'bar` in YSH may be disallowed, except in limited cases like `--foo='bar'` ... basically because there are so many ways of quoting, most users aren't used to reading it, and C-style quoting is powerful to express every string)

PR #2192, opened 2024-12-10
[test/sh_spec] Support multiple possible results

akinomyoga:
7b95b2c6091cb0124e6969ffb95dfd106de55f5a This fixes the problem of environment-dependent confirmed results in https://github.com/oils-for-unix/oils/pull/2184#issuecomment-2530164337. The current `master` only allows the possible combination of result levels to be one of `PASS | FAIL`, `OK | FAIL`, `N-I | FAIL`, and `BUG | FAIL` for each shell for each test case. This PR allows an arbitrary combination of possible result levels for each shell and each test case. For example, in the https://github.com/oils-for-unix/oils/pull/2184#issuecomment-2530164337 case, when Bash produces the expected results, it shows `PASS`. When Bash produces the confirmed buggy behavior, it shows `BUG`. Otherwise, it shows `FAIL`.

4f2768e76accdc45bbb774752afa6b1c8667ad4e This additionally supports registering multiple possible values for each result level.

### DIFF
--- a/test/sh_spec.py
+++ b/test/sh_spec.py
@@ -193,17 +193,17 @@ def AddMetadataToCase(case, qualifier, shells, name, value, line_num):
         if shell not in case:
             case[shell] = {}
 
-        # Check a duplicate specification
-        name_without_type = re.sub(r'-(json|repr)$', '', name)
-        if (name_without_type in case[shell] or
-                name_without_type + '-json' in case[shell] or
-                name_without_type + '-repr' in case[shell]):
-            raise ParseError('Line %d: duplicate spec %r for %r' %
-                             (line_num, name, shell))
-
         if qualifier not in case[shell]:
             case[shell][qualifier] = {}
-        case[shell][qualifier][name] = value
+
+        if name not in case[shell][qualifier]:
+            case[shell][qualifier][name] = value
+        elif not isinstance(case[shell][qualifier][name], list):
+            case[shell][qualifier][name] = [
+                    case[shell][qualifier][name], value
+            ]
+        else:
+            case[shell][qualifier][name].append(value)
 
 
 # Format of a test script.
@@ -397,7 +397,11 @@ def CreateStringAssertion(d, key, assertions, qualifier=False):
 
     exp_json = d.get(key + '-json')
     if exp_json is not None:
-        exp = json.loads(exp_json, encoding='utf-8')
+        if isinstance(exp, list):
+            exp = [json.loads(v, encoding='utf-8') for v in exp_json]
+        else:
+            exp = json.loads(exp_json, encoding='utf-8')
+
         a = EqualAssertion(key, exp, qualifier=qualifier)
         assertions.append(a)
         found = True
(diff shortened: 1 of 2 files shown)


andychu:
Hm is this to fix the bash `[][]` error?   

OK yeah I see the problem is that `spec/ble-idioms` fails on my machine locally now

Because we changed the assertion to match the CI

---

I think the idea behind spec tests and wedges is that bash 5.2 should **always** produce the same result, and bash 4.4 should always produce the same result

But that is not true

hmmm

andychu:
Hm actually let me try detecting /etc/debian_version as a special case ... that might help

akinomyoga:
The original reason that I came up with this is for the environment-specific behavior, but I think even without it I think this is useful.

akinomyoga:
and this detected inconsistent or broken test specifications: 047554441c410777c64d20b42207909f7150b70d

test/sh_spec.py
```diff
@@ -760,29 +818,46 @@ def RunCases(cases, case_predicate, shells, env, out, opts):
...
+                        # If any failed, then the result is FAIL.
+                        # If any are OK, but none are FAIL, the result is OK.
+                        cell_result = min(cell_result, result)
+                        if msg:
+                            messages.append(msg)
+
+                    qualifier_level = max(
+                        [a.QualifierLevel() for a in assertions])
+
+                    # We choose the results of the assertion set with the
```
andychu, line 842: Hm I would have to think about the "mildest failure" logic ... where else would we use that?

andychu (commented):
Thanks for looking deeply into this!    I can sort of see the logic in doing this

But I wonder how many usages we would have of multiple results?

I think this is a reasonable approach for the problem we just encountered:

https://github.com/oils-for-unix/oils/pull/2173

http://op.oilshell.org/uuu/github-jobs/8698/ovm-tarball.wwz/_tmp/spec/osh-py/ble-idioms.html


It is very explicit, which calls out the fact that bash has an unusual bug.  It is NOT just an integer overflow bug in bash 4.4 or bash 5.2, it is a bug in bash 5.2 that only happens on certain machines

----

Thank you for finding the inconsistencies ... we can definitely merge those

I am a little concerned that the multiple results adds a lot of logic to test/sh_spec.py, but it may not solve many problems.  

The other problems I would like to solve are related to say turning on bash 5.2 on `spec/assoc` and other files.  Because we are kinda stuck on bash 4.4 -- it was a lot of work to upgrade, so I didn't finish it


We may need this in the future, but if it's only one example, then I think we should try to be simple.  The Oils codebase is already 8.5 years old so sometimes I have problems maintaining all the different parts !!   e.g. upgrading all the spec tests

But there are some valuable fixes here, which definitely helps

test/sh_spec.py
akinomyoga: I mean, if the test result (status/stdout/stderr) satisfies e.g. both the "OK" requirements and the "N-I" requirements, the final result we show in the table would be "OK". By "mild", I mean the ordering of `PASS > OK > N-I > BUG > FAIL`.

andychu:
I mentioned on the other thread that if this change lets end users run spec tests on systems like Alpine Linux, FreeBSD,  or OpenBSD, then I'd be open to it


https://github.com/oils-for-unix/oils/pull/2173#issuecomment-2536567818

But I think we are VERY far from it right now unfortunately

I think that that the simpler change in #2173 will get us past this problem, and then maybe we can talk about this larger change after we migrate SparseArray and so forth ?

Thank you for looking into this deeply

akinomyoga:
OK, I'll break a part of this (i.e., the check for inconsistent or broken test cases) into an independent PR.

akinomyoga:
Closing this for now.

PR #2193, opened 2024-12-10
[test/sh_spec] Fix error message for C++ failures

akinomyoga:
This is a trivial fix. In the current master, the error message for the error count shows a different number for the allowed C++ failures.


andychu:
Thanks!

Issue #610, opened 2020-02-23
Implement dotglob


andychu:
Related to #609 and #552 

This should just be a libc flag change.  Maybe wait until after translation is done?

andychu:
There's a failing spec test for this in `spec/glob.test.sh`

```
$ test/spec.sh glob  -r 35 
glob.test.sh
case    line    dash    bash    mksh    ash     osh
 35     293     N-I     pass    N-I     N-I     FAIL    dotglob (bash option that dashglob is roughly consistent with)

                dash    bash    mksh    ash     osh
        pass    0       1       0       0       0
        N-I     1       0       1       1       0
        FAIL    0       0       0       0       1
        total   1       1       1       1       1

```

andychu:
Demo

```
$ mkdir tmp
$ cd tmp
$ touch a b .bashrc
$ echo *
a b
$ shopt -s dotglob
$ echo *
a b .bashrc
```

andychu:
man glob

```

       GLOB_PERIOD
              Allow a leading period to be matched by metacharacters.  By default, metacharacters can't match a leading period.

```

andychu:
OK interesting

```
osh-0.21.0$ = glob('*')
(List 0x7fb9cd964af0)   []
osh-0.21.0$ = glob('.*')
(List 0x7fb9cd964c30)   [".",".."]
```

I didn't know this was the behavior of glob

I think there are two different things here

- bash-compatible `shopt -s dotglob`
- optional args to glob() function in YSH

jtb20:
Test case:

[dotglob.ysh.txt](https://github.com/user-attachments/files/15666204/dotglob.ysh.txt)

HTH!

rdw20170120:
1. I launch YSH with `ysh`
2. Oils executes my `~/.config/oils/yshrc`
3. My `yshrc` contains `shopt --set ysh:all` (just to be sure)
4. Then it executes `source $_this_dir/shopt.ysh`
5. My `shopt.ysh` repeats all the individual `shopt` commands that I had captured via `shopt -p`, but adds overrides for a couple that I change
6. Among those commands is `shopt -u dotglob` (reasserting the SAME setting)
7. Execution of that command results in the output:

> osh warning: GLOB_PERIOD wasn't found in libc, so 'shopt -s dotglob' won't work

It seems to me that one should ONLY get that warning from `shopt --set dotglob`, but NOT from `shopt --unset dotglob`.

[...]

PR #2194, opened 2024-12-11
Support shopt -s dotglob

leungbk:
Related to #610.

The Python build seems okay, but I could use some help getting the ninja build to pass; my attempt at defining GLOB_PERIOD doesn't seem to work.

```
In file included from /home/brian/oils/cpp/preamble.h:6:
_gen/bin/oils_for_unix.mycpp.cc: In member function ‘int glob_::Globber::_Glob(BigStr*, List<BigStr*>*)’:
_gen/bin/oils_for_unix.mycpp.cc:41233:22: error: expected unqualified-id before ‘(’ token
41233 |       flags |= libc::GLOB_PERIOD;
      |                      ^~~~~~~~~~~
```


leungbk:
OK, I managed to resolve my problem from before: `mycpp/TEST.sh test-translator` now passes on my machine.

But I cannot tell from the GitHub interface where specifically to look to investigate the failing jobs; clicking on any single job takes me to a `publish-html` step that doesn't appear to expose anything helpful.

andychu:
Cool, I will look at this!

Oh there is a URL hidden in the logs.

https://github.com/oils-for-unix/oils/actions/runs/12284550355/job/34280567011?pr=2194

I switched from SSH to HTTP copying, and it introduced a lot of spew, which has hidden it

I can fix that

----

In the mean time I go to 

which is linked on https://www.oilshell.org/    and then  look for the CI results


---

Thanks for the help!

andychu:
Oh actually you can see a few more failures here

https://op.oilshell.org/uuu/github-jobs/8712/

https://op.oilshell.org/uuu/github-jobs/8712/dev-setup-debian.wwz/_tmp/soil/logs/py-all-and-ninja.txt

```
.......
======================================================================
ERROR: testGlob (__main__.LibcTest)
----------------------------------------------------------------------
  File "pyext/libc_test.py", line 177, in testGlob
    print(libc.glob('\\'))
RuntimeError: unknown problem

----------------------------------------------------------------------
Ran 12 tests in 0.001s

FAILED (errors=1)
```



I am not sure what happened, I guess the default behavior changed somehow?

andychu:
Wow this looks great, thank you!  Thanks for tracking down all the Python and C++ stuff

I think it can be OK to just pass `0` everywhere for the glob() flags ...  I don't think we have any optional args in Python right now.

---

If we don't support globskipdots, then maybe we can tweak the test so it doesn't pass?  I guess we are missing test coverage?

leungbk:
OK, I have updated the globstar test to make sure it continues to fail.

I also corrected my comment in `glob_.py`: filtering out only `.` and `..` in the manner that I did does not always match the behavior of `shopt -s globskipdots`. I tried to look at how `bash` implemented it but couldn't quite follow all the casework.

andychu:
Great, thank you!

Now I remember I didn't start a Zulip thread about how to implement `globstar`!  I was wondering how it behaves, and the tests really helped, particularly knowing that it's different in bash and zsh

PR #2195, opened 2024-12-18
[test/sh_spec] Handle duplicate/conflicting test spec definitions

akinomyoga:
This PR picks a part of #2192 to add checks for inconsistent spec tests.

fa625e262

- Check duplicate test spec definitions. For example, when the `STDOUT` section is specified multiple times for a specific shell, we produce an error message.
- Properly handle multiple levels of failure (`OK`, `BUG`, `N-I`, etc.). In the current master, when both `BUG` and `OK` requirements are specified for a specific shell, and when the `OK` requirement is specified later, the result shown in the table becomes `OK`. In this commit, we change it to choose the worst failure level. For example, when the shell satisfies the `BUG` and `OK` requirements, we show `BUG` in the table.
- This commit also modifies existing inconsistent test spec definitions

08cd83827

- In the current master, the filename is not printed when the parse error happens, which makes it difficult to identify the error location in the spec tests. This commit outputs the filename on `ParseError`.

ab7a324d4

- This commit also checks duplicates of `stdout`, `stdout-json`, and `stdout-repr`, or similar ones for `stderr` and `status`.

### DIFF
--- a/spec/arith.test.sh
+++ b/spec/arith.test.sh
@@ -859,7 +859,7 @@ echo ---
 ---
 ## END
 
-## OK bash/dash/mksh/zsh STDOUT:
+## OK bash/dash/zsh STDOUT:
 10
 5
 -9223372036854775808
--- a/spec/assign-extended.test.sh
+++ b/spec/assign-extended.test.sh
@@ -698,7 +698,7 @@ None
 2
 3
 ## END
-## OK mksh status: 0
+## BUG mksh status: 0
 
 #### syntax error in array assignment
 a=x b[0+]=y c=z
--- a/spec/ble-features.test.sh
+++ b/spec/ble-features.test.sh
@@ -550,7 +550,7 @@ v=tempenv1 f4_unset global,tempenv1
 [global,tempenv1,tempenv2,tempenv3] v: (unset) (unset 4)
 ## END
 
-## OK zsh/ash/dash/mksh/yash STDOUT:
+## OK zsh/ash/dash/mksh STDOUT:
 # value-unset, mksh-unset, tempenv-value-unset?
 [global,tempenv1,tempenv2,tempenv3] v: tempenv3
 [global,tempenv1,tempenv2,tempenv3] v: (unset) (unset 1)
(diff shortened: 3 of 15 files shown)


andychu:
Thank you!    I like that this caught the conflicting assertions

---

I'm not sure I understand this part:

> Properly handle multiple levels of failure (OK, BUG, N-I, etc.). In the current master, when both BUG and OK requirements are specified for a specific shell, and when the OK requirement is specified later, the result shown in the table becomes OK. In this commit, we change it to choose the worst failure level. For example, when the shell satisfies the BUG and OK requirements, we show BUG in the table.


Does this change now disallow


```
OK bash stdout-json: ""
BUG bash STDOUT:
END
```


i.e. what's an example of when have both OK and BUG, but we do NOT issue a ParseError ?

akinomyoga:
> Does this change now disallow
> 
> ```
> OK bash stdout-json: ""
> BUG bash STDOUT:
> END
> ```

No, it doesn't change the behavior for the above case. The above test spec is still allowed and results in `BUG` in the result table when the shell satisfies all the requirements.

An example case where the result changes is this:

```
BUG bash STDOUT:
END
OK bash stdout-json: ""
```

This pattern can be found for example in the following place:

https://github.com/oils-for-unix/oils/blob/fa2dae4944f73c55471f2e771682932c255bf6b6/spec/assign-extended.test.sh#L696-L701

In the current master, the table shows `OK` for this combination because `OK` overwrites the previously specified `BUG`. However, it should show `BUG` in the table. This PR compares `OK` and `BUG` and chooses `BUG` (instead of just overwriting `BUG` with `OK`).

andychu:
Oh I see ... I think this spec should be changed to

```
BUG bash STDOUT:
END
BUG bash status: 0
```

Is it possible to detect the inconsistent qualifiers and raise a ParseError ?  

I think there may only be a few of them, and we should change them, like the others are changed

I didn't intend to have bash or mksh be both BUG and OK

akinomyoga:
OK, I actually have commit 8cb3ddd51 to do that. I pushed it.

andychu:
Looks great, thank you!

PR #2197, opened 2024-12-19
`bind -x` support

KingMob:
Here's the WIP on bind -x support for running arbitrary code


KingMob:
Marking as "Ready for review". 

Still need to add READLINE env var support and fix unicode errors. Will remove debug comments later.

builtin/readline_osh.py
```diff
@@ -142,9 +184,10 @@ def Run(self, cmd_val):
                     readline.unbind_keyseq(arg.r)
 
                 if arg.x is not None:
-                    self.errfmt.Print_("warning: bind -x isn't implemented",
-                                       blame_loc=cmd_val.arg_locs[0])
-                    return 1
+                    if self.bindx_cb is None:
```
andychu, line 187: How about initializing `bindx_cb` in `core/shell.py`, then passing it to the constructor, and then make it a member?

That is generally the style we use

pyext/line_input.c
```diff
@@ -1050,6 +1068,265 @@ PyDoc_STRVAR(doc_unbind_keyseq,
...
+    } else {
+        debug_print("No termcap 'ce', using newline");
+        rl_crlf();
+    }
+}
+
+
+/* Save readline state to Python variables */
+static int
+save_readline_state(void) {
```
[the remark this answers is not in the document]
andychu: If you're not sure if these functions will be used yet, you can wrap them in `#if 0` to make it clear that right now they don't need to be compiled

frontend/py_readline.py
```diff
@@ -140,7 +140,34 @@ def print_shell_cmd_map(self):
...
+                "%s: missing double-quotes around the key sequence" % keyseq)
+        keyseq = keyseq[1:-1]
+
+        cmd = cmdseq_split[1]
+        line_input.bind_shell_command(keyseq, cmd)
+
+    def set_bind_shell_command_hook(self, hook):
+        # type: (Callable[[str, str, int], Tuple[int, str, str]]) -> None
+
+        if hook is None:
```
andychu, line 167: Should this be an assert like `assert hook is not None, "..."` ?

builtin/readline_osh.py
```diff
@@ -42,19 +46,57 @@ def __exit__(self, type, value, traceback):
...
+
+        return (status, line_buffer, str(point))
+
 
 class Bind(vm._Builtin):
     """Interactive interface to readline bindings"""
 
-    def __init__(self, readline, errfmt):
-        # type: (Optional[Readline], ui.ErrorFormatter) -> None
+    def __init__(self, readline, errfmt, mem, evaluator):
```
andychu, line 87: It doesn't seem like `mem` is used now

pyext/line_input.c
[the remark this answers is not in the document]
KingMob: OK. I'll use the preprocessor to filter them out. And I'll see how much I can move to Python. Not the calls to readline proper, but maybe the rest. I wrote this as a minimal Oils translation of how bash does it, but if Python is better, I'll figure something out.

pyext/line_input.c
```diff
@@ -1050,6 +1068,265 @@ PyDoc_STRVAR(doc_unbind_keyseq,
...
+    // const char *kseq;
+    // const char *cmd;
+    char *kseq;
+    char *cmd;
+    Keymap kmap, cmd_xmap;
+
+    // if (!PyArg_ParseTuple(args, "ss:bind_shell_command", &kseq, &cmd)) {
+    //     return NULL;
+    // }
+    if (!PyArg_ParseTuple(args, "eses:bind_shell_command", "utf-8", &kseq, "utf-8", &cmd)) {
```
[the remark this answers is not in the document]
KingMob: They display different behavior in the presence of non-ascii chars, but I need to rerun the tests to check.

I don't recall "eses" actually being what we want, though I don't think there's a difference when used with pure ascii.

I'll know more when I dive into the unicode stuff.

frontend/py_readline.py
```diff
@@ -140,7 +140,34 @@ def print_shell_cmd_map(self):
...
+            raise ValueError("%s: missing colon separator" % bindseq)
+
+        # Below checks prevent need to do so in C, but also ensure rl_generic_bind
+        # will not try to incorrectly xfree `cmd`/`data`, which doesn't belong to it
+        keyseq = cmdseq_split[0].rstrip()
+        if len(keyseq) <= 2:
+            raise ValueError("%s: empty/invalid key sequence" % keyseq)
+        if keyseq[0] != '"' or keyseq[-1] != '"':
+            raise ValueError(
+                "%s: missing double-quotes around the key sequence" % keyseq)
```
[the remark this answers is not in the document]
KingMob: I'm open to suggestions. I just adopted this from the surrounding style.

frontend/py_readline.py
KingMob: Sure, np. I loosely based it off `set_completion_display_matches_hook`, which does something similar, but unlike completion, the bind-x callback should always be present (tho we may want to defer setting it until actually used)

builtin/readline_osh.py
KingMob: Yes, but I think I'll need it for the READLINE_* env vars.

andychu:
Thanks for the replies ... I think it's OK to make the edits on the stuff we're sure about, and then we can land this because new tests pass 

And then we can investigate the issues we're not sure about, like the Unicode

And we can also worry about C++ separately.  There is a lot of C code right now, but the first step is to make it work ... then refactor it, assuming the tests will catch breakages

The most likely thing is that we will have some helper library like we do with J8 notation - that is shared between C++ and Python/C

KingMob:
OK, I'll make the edits in the next week, and let you know when it's ready for re-review.

----

> ...would it make more sense to put some of this error-checking higher-up in the call stack in that case, i.e. readline_osh.py?

☝🏻 Can you weigh in on this one? It's a bigger change?

andychu:
Yes, we should probably reduce the code in frontend/py_readline.py too, because it's not translated

It's meant to be a trivial wrapper

pyext/line_input.c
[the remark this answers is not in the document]
KingMob: Done

core/shell.py
```diff
@@ -654,8 +654,9 @@ def Main(
                                          fd_state, tracer, errfmt, loader)
     b[builtin_i.source] = source_builtin
     b[builtin_i.dot] = source_builtin
-    b[builtin_i.eval] = meta_oils.Eval(parse_ctx, exec_opts, cmd_ev, tracer,
-                                       errfmt, mem)
+    evaluator = meta_oils.Eval(parse_ctx, exec_opts, cmd_ev, tracer, errfmt,
+                               mem)
+    b[builtin_i.eval] = evaluator
```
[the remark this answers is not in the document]
KingMob: Done

builtin/readline_osh.py
KingMob: OK, I took a closer look at this. I wrote it so that the callback's initialized only on demand; since I'd bet the vast majority of users will never use the functionality at all, why pay the price?

But the rest of oils doesn't seem to do that afaict, so I'll do as you suggested. LMK if you want me to revert it tho.

frontend/py_readline.py
KingMob: Done

pyext/line_input.c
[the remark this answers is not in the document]
KingMob: So, the issue was, afaict, the shell command needed a copy made, since readline stores the pointer until unbound. (Not needed for the keyseq; readline converts it and stores a copy in its own format.)

Once I switched from 'es' back to parsing with 's', and used strdup to copy the cmd, I stopped getting weird errors. At the same time, unicode characters started working, so this may have killed two birds with one stone.

frontend/py_readline.py
```diff
@@ -140,7 +140,34 @@ def print_shell_cmd_map(self):
...
     def unbind_keyseq(self, keyseq):
         # type: (str) -> None
-        line_input.unbind_keyseq(keyseq)
+        line_input.unbind_keyseq(keyseq)    
+
+    def bind_shell_command(self, bindseq):
+        # type: (str) -> None
+        cmdseq_split = bindseq.strip().split(":", 1)
+        if len(cmdseq_split) != 2:
+            raise ValueError("%s: missing colon separator" % bindseq)
```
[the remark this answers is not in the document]
KingMob: Done. Moved up to readline_osh.py.

builtin/readline_osh.py
KingMob: Done.

builtin/readline_osh.py
```diff
@@ -79,16 +122,16 @@ def Run(self, cmd_val):
...
                         + ", -".join(self.exclusive_flags),
                         blame_loc=cmd_val.arg_locs[0])
                     return 1
                 else:
                     found = True
         if found and not arg_r.AtEnd():
             self.errfmt.Print_(
-                "error: cannot mix bind commands with the following flags: -" +
-                ", -".join(self.exclusive_flags),
+                "error: Too many arguments. Check your quoting. Also, you cannot mix normal bindings with the following flags: -"
```
andychu, line 133: I'm not sure what "normal bindings" are

And I would remove "check your quoting", since it is not specific, and that is not unique to bind

andychu (commented):
OK looks good!

I made some comment about the error text, but it's not essential to address now

builtin/readline_osh.py
KingMob: @andychu 

"Normal" bindings are the ones that map a key code to a named readline fn. They don't use flags at all (it's the default behavior of bind). This is in contrast to command bindings that execute arbitrary shell code, and use `-x`.

If you recall, bash bind allows us to mix a variety of flags and normal bindings, but the issue was, the order of operations is completely unintuitive and hidden (it's not their order at the CLI). We decided to make it so osh bind could only do one thing at a time, and see if anyone complained.

---

I'll remove the bit about "check your quoting". 

One thing that is unique to bind, however, is the bind -x quoting syntax. Bash bind _requires_ two levels of nested quotes to work right: `'"\C-o": do something --here'`. And in fact, bash bind considers it a bug if the inner quotes around the keyseq aren't specifically _double_ quotes... don't ask. 😠

PR #2201, opened 2024-12-25
[osh] Perform indirect expansion with `${!a[@]...}`

akinomyoga:
- **PR(a)** I wanted to submit a PR for SparseArray `${sp[@]...}`
  - **PR(b)** However, the behaviors of existing BashArray/BashAssoc `${a[@]@Q}` and `${a[@]@a}` are broken, and `${a[@]@P}` is not implemented. So I first need to fix the behaviors of BashArray/BashAssoc `${a[@]@?}`.
    - **PR(c)** However, before fixing the above behavior, BashArray/BashAssoc `${a@a}` and `${a[0]@a}` need to be fixed because it implements something different from Bash's.
      - **PR(d)** Then, I tried to fix the above. However, I realized that even another incomplete behavior conflicts. `${!a[@]...}` should perform the indirect expansion, though `osh` tries to generate the keys of `a`.

This PR is **PR(d)** in the above chain. In the current master, osh interprets e.g. `${!a[@]...}` in a completely different way compare to Bash.

```bash
$ bash -c 'a=(var); var=value; echo "${!a[@]:0}"'
value
$ bin/osh -c 'a=(var); var=value; echo "${!a[@]:0}"'
0
```

Bash considers `${!a[@]:0}` as `Slice(IndirectExpansion(a[@]))` while osh considers it as `Slice(Keys(a))`. It should be noted that Bash's `${!a[@]}` treated as `Keys(a)` is an exceptional case, and in all the other cases, `!a[@]` is treated as an indirect expansion.

### DIFF
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -887,18 +887,21 @@ class AbstractWordEvaluator(StringWordEvaluator):
 
             elif case(value_e.Str):
                 val = cast(value.Str, UP_val)
-                bvs_part = self.unsafe_arith.ParseVarRef(val.s, blame_tok)
-                return self._VarRefValue(bvs_part, quoted, vsub_state,
-                                         vtest_place)
+                var_ref_str = val.s
 
             elif case(value_e.BashArray):  # caught earlier but OK
-                e_die('Indirect expansion of array')
+                val = cast(value.BashArray, UP_val)
+                var_ref_str = ' '.join(bash_impl.BashArray_GetValues(val))
 
             elif case(value_e.BashAssoc):  # caught earlier but OK
-                e_die('Indirect expansion of assoc array')
+                val = cast(value.BashAssoc, UP_val)
+                var_ref_str = ' '.join(bash_impl.BashAssoc_GetValues(val))
 
             else:
                 raise error.TypeErr(val, 'Var Ref op expected Str', blame_tok)
+
+        bvs_part = self.unsafe_arith.ParseVarRef(var_ref_str, blame_tok)
+        return self._VarRefValue(bvs_part, quoted, vsub_state, vtest_place)
 
     def _ApplyUnarySuffixOp(self, val, op):
         # type: (value_t, suffix_op.Unary) -> value_t
@@ -1465,18 +1468,10 @@ class AbstractWordEvaluator(StringWordEvaluator):
 
             elif part.prefix_op.id == Id.VSub_Bang:
                 if (part.bracket_op and
-                        part.bracket_op.tag() == bracket_op_e.WholeArray):
+                        part.bracket_op.tag() == bracket_op_e.WholeArray and
+                        not suffix_op_):
                     # undef -> empty array
                     val = self._ProcessUndef(val, part.name_tok, vsub_state)
-
-                    # ${!a[@]-'default'} is a non-fatal runtime error in bash.
-                    # Here it's fatal.
-                    if suffix_op_ and suffix_op_.tag() == suffix_op_e.Unary:
-                        unary_op = cast(suffix_op.Unary, suffix_op_)
-                        if consts.GetKind(unary_op.op.id) == Kind.VTest:
-                            e_die(
-                                'Test operation not allowed with ${!array[@]}',
-                                unary_op.op)
 
                     # ${!array[@]} to get indices/keys
                     val = self._Keys(val, part.name_tok)
@@ -1490,8 +1485,8 @@ class AbstractWordEvaluator(StringWordEvaluator):
                     vtest_place.name = None
                     vtest_place.index = None
 
-                    val = self._EvalVarRef(val, part.name_tok, quoted, vsub_state,
-                                           vtest_place)
+                    val = self._EvalVarRef(val, part.name_tok, quoted,
+                                           vsub_state, vtest_place)
 
             else:
                 raise AssertionError(part.prefix_op)
@@ -1503,8 +1498,8 @@ class AbstractWordEvaluator(StringWordEvaluator):
             with tagswitch(suffix_op_) as case:
                 if case(suffix_op_e.Nullary):
                     op = cast(Token, UP_op)
-                    val, quoted2 = self._Nullary(val, op, var_name, part.name_tok,
-                                                 vsub_state)
+                    val, quoted2 = self._Nullary(val, op, var_name,
+                                                 part.name_tok, vsub_state)
 
                 elif case(suffix_op_e.Unary):
                     op = cast(suffix_op.Unary, UP_op)
@@ -1520,7 +1515,8 @@ class AbstractWordEvaluator(StringWordEvaluator):
 
                     else:
                         # Other suffix: value -> value
-                        val = self._ProcessUndef(val, part.name_tok, vsub_state)
+                        val = self._ProcessUndef(val, part.name_tok,
+                                                 vsub_state)
                         val = self._ApplyUnarySuffixOp(val, op)
 
                 elif case(suffix_op_e.PatSub):  # PatSub, vectorized
--- a/spec/var-ref.test.sh
+++ b/spec/var-ref.test.sh
@@ -1,4 +1,5 @@
 ## compare_shells: bash-4.4
+## oils_failures_allowed: 1
 
 # Var refs are done with ${!a}
 #
@@ -78,8 +79,9 @@ A_nobrackets=0
 
 #### ${!a[@]-'default'} is illegal
 
-# bash disallows this when a is an array.  We make it an error because [@]
-# implies it's an array.
+# bash disallows this when a is an array.  We originally made it an error
+# because [@] implies it's an array, but we now changed to follow Bash when the
+# array has a single element.
 
 argv.py "${!a[@]-default}"
 echo status=$?
@@ -89,6 +91,8 @@ argv.py "${!a[@]-default}"
 echo status=$?
 ## status: 1
 ## STDOUT:
+['default']
+status=0
 ## END
 ## BUG bash status: 0
 ## BUG bash STDOUT:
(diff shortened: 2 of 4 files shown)


andychu:
Oh wow this is crazy, I didn't know that

Now that I look at `doc/ref/toc-osh`, I see that `word-lang` is missing BOTH of these operations

I think we should have

```
op-indirect ${!x} ${!x[@]:0}   
op-keys ${!assoc[@]}
```

and then add those two topics as

```
### op-indirect

### op-keys

```

in `doc/ref/chap-word-lang.md`

---

I again wonder if this is documented in the bash manual ...

andychu:
Also, I think it is ok if initially `${sp[@]...}` only does what OSH does  correctly now ...

but it is also OK to fix it in the other order I guess

andychu:
To build the reference you can do `build/doc.sh all-ref`, and then look in `_release/VERSION`

You can just add something very short for the behavior we are  correcting here, it doesn't necessarily have to document the whole thing (because I know op-indirect is very complex)

akinomyoga:
### Bash Reference Manual

> I again wonder if this is documented in the bash manual

This is technically documented in Bash's manual, though one needs to "derive" this fact based on several descriptions. Although Bash's reference manual is often ambiguous or often has an undocumented part, this *specific* part is well-defined in Bash's reference manual.

First of all, in the Bash Reference Manual, the indirection is described as a part of *parameter* (i.e., kind of a "variable name"):

> [3.5.3 Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion) / ¶4
>
> If the first character of *parameter* is an exclamation point (`!`), and *parameter* is not a nameref, it introduces a level of indirection. [...] The exceptions to this are the expansions of `${!prefix*}` and `${!name[@]}` described below. [...]

Then, after the descriptions, parameter expansion constructs follow:

> [3.5.3 Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion) / ¶7
>
> **${parameter:-word}** [...]
> **${parameter:=word}** [...]
> **${parameter:?word}** [...]
> **${parameter:+word}** [...]
> **${parameter:offset}**
> **${parameter:offset:length}** [...]

where *parameter* is used, so this explicitly includes indirection of the form `!a[@]`, etc.

Then, the exceptions are described:

> [3.5.3 Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion) / ¶7
>
> <b>${!prefix*}</b>
> **${!prefix@}** [...]
> **${!name[@]}**
> **${!name[*]}** [...]

where *prefix* and *name* are used instead of *parameter* which includes `!a[@]`.

akinomyoga:
> I think we should have
> 
> ```
> op-indirect ${!x} ${!x[@]:0}   
> op-keys ${!assoc[@]}
> ```
> 
> and then add those two topics as
> 
> ```
> ### op-indirect
> 
> ### op-keys
> ```
> 
> in `doc/ref/chap-word-lang.md`

Hmm, since the indirection is considered a kind of variable name by Bash, it might not be ideal to describe "indirection" and "keys" as if they are concepts of an equal level, i.e., "op"s. We may still try to interpret `!` as an indirection operator in the parameter expansion syntax, but `${!name[@]}` and `${!prefix@}` might be described as constructs independent of the normal variable expansion syntax with operators.

akinomyoga:
I added `op-indirect` in `doc/ref/chap-word-lang.md` (96bac620b). The description corresponding to `op-keys` was put on a different level.

spec/var-ref.test.sh
```diff
@@ -78,8 +79,9 @@ A_nobrackets=0
 
 #### ${!a[@]-'default'} is illegal
 
-# bash disallows this when a is an array.  We make it an error because [@]
-# implies it's an array.
+# bash disallows this when a is an array.  We originally made it an error
```
andychu, line 82: I'm having trouble with part ...  

I'm not sure if this is the right change -- we are differing from bash in at least some case

And if so, I think the title of the spec test should be updated

Maybe it can be split into two test cases -- when `a` is an array, and when it is not?  Or when it has a single element?

andychu (commented):
Hmm I added `## compare_shells: bash-4.4 bash` at the top of the file

So we test against bash 5

And then I get some more differences



```
case    line    bash-4  bash    osh
  4      80     BUG     FAIL    pass    ${!a[@]-'default'} is illegal
case: 4
bash-4 stdout:
['default']
status=0
status=1

bash-4 stderr:
bash-4.4: line 5: x y z: bad substitution


case: 4
[bash stdout] Expected "['default']\nstatus=0\nstatus=1\n", got 'status=1\nstatus=1\n'
bash stdout:
status=1
status=1

bash stderr:
bash: line 1: a[@]: invalid indirect expansion
bash: line 5: x y z: invalid variable name

```

andychu:
Are we sure we need to do PR(b) PR(c) PR(d) before SparseArray?

I ran the tests with bash 4 and 5, and bash doesn't agree with itself

which makes it very hard for OSH to do something "right".  Also I don't quite understand the rationale for deviating from bash 

---

It might help to prioritize by what gets `ble.sh` to run.  As far as I understand, the SparseArray is essential to make it run, because otherwise it uses too much memory.

But are PR(b) PR(c) PR(d) necessary to get it to run?  If not, maybe we can put these off

I think updating to bash 5 is a good idea in general, but I don't want to get too distracted, since it could take a long time to figure out

---

I think it's OK to implement **less** than bash

For example, can we simply **disallow**    `${!a[@]:0}`   ?    And then we can wait until a script needs to use it to implement it

It's not good to have a difference vs. bash, I agree, but it is OK to **reject** the input (until someone needs it)

I'm not sure if ble.sh uses it

andychu:
Thanks for updating the docs - that helps - I didn't know  that some of the `!` are special cases ...

andychu:
I think it should always be possible to rewrite   `${!a[@]:0}` in another way too, no ?

Of course if say Gentoo or Nix  uses that, then "nobody" knows how to rewrite it.  So sometimes we have to implement it, and indeed we have to implement more than I thought

But so far I have not seen `${!a[@]:0}`   ...

spec/var-ref.test.sh
akinomyoga: > I'm not sure if this is the right change -- we are differing from bash in at least some case

Ah, No. Actually, the statement "bash disallows this when a is an array" is false, and the change in this PR makes osh more consistent with Bash actually. The remaining difference is that osh dies (i.e., `e_die()`) with more than one element, while Bash returns `1` and continues the execution of the subsequent commands.

I'll update the sentence "bash disallows ...".

> And if so, I think the title of the spec test should be updated

Yeah, I think that the title of the spec test doesn't practically reflect the Bash behavior. I'll update it.

> Maybe it can be split into two test cases -- when `a` is an array, and when it is not? Or when it has a single element?

Actually, we already have separate test cases. This existing spec test is for the case where there are more than one element, and the other test case I newly added is the test for the case with a single element.

akinomyoga:
> Are we sure we need to do PR(b) PR(c) PR(d) before SparseArray?

If I was asked whether it is *needed*, it isn't mandatory *technically*. However, if the changes are applied in the reverse order, I need to effectively *re-implement* the things applied earlier since they are overlapping and the best code structure depends on the overall feature set. This multiplies the amount of work unnecessarily, and I don't want to waste time.

Another possibility is to determine the final OSH design in a way different from Bash so that the implementation can be naturally made in the order PR(a) -> PR(b) -> PR(c) -> PR(d). However, determining the behavior based on the implementation order seems strange.

> I ran the tests with bash 4 and 5, and bash doesn't agree with itself

Since Bash also had bugs and some of them were fixed in Bash 5. When Bash 4 has a bug, the present test framework requires Bash to reproduce the bug, so the test fails even when Bash behaves in the correct way. This was actually one of the cases #2192 tried to resolve.

> which makes it very hard for OSH to do something "right".

We should basically follow Bash 5, though there might be new regressions in Bash 5. I later need to check the cases one by one.

> Also I don't quite understand the rationale for deviating from bash

It's the opposite. The current master version deviates from Bash, and this PR makes it more consistent with Bash, though there is still a subtle difference in how it fails ("e_die" vs "exit 1 and continue").

> It might help to prioritize by what gets `ble.sh` to run. As far as I understand, the SparseArray is essential to make it run, because otherwise it uses too much memory.
> 
> But are PR(b) PR(c) PR(d) necessary to get it to run? If not, maybe we can put these off

This is technically possible, but that means I need to precisely reproduce osh's strange behaviors in BashArray/BashAssoc for SparseArray. Some of the behaviors are nontrivial to implement if one needs to preserve the current osh's behavior. I already have stubs of PR(d), PR(c), and PR(b) locally. What blocks merging PR(d), PR(c), and PR(b) (though I haven't yet described what they fix and how they do it in detail)?

> I think updating to bash 5 is a good idea in general, but I don't want to get too distracted, since it could take a long time to figure out

I agree with it. I'll process all of them. I don't think there are many differences from Bash 5.

> I think it's OK to implement **less** than bash
> 
> For example, can we simply **disallow** `${!a[@]:0}` ? And then we can wait until a script needs to use it to implement it

This is also possible, but this actually makes the codebase less clean. It doesn't simplify the codebase.

> It's not good to have a difference vs. bash, I agree, but it is OK to **reject** the input (until someone needs it)
> 
> I'm not sure if ble.sh uses it

As far as I know, `ble.sh` doesn't use these specific behaviors. However, the reason that I want to push those changes is to keep the structure codebase simple. Nevertheless, the behavioral changes visible to users should be made in dedicated commits with confirmation, which is the reason that I don't submit a big PR that changes many behaviors here and there all at once.

akinomyoga:
> Hmm I added `## compare_shells: bash-4.4 bash` at the top of the file

The current test framework doesn't seem to allow us to specify the expected behavior for bash-4 and bash-5 separately. If I write `## BUG bash-4 STDOUT: ...`, it fails to parse the test spec file. `## BUG bash-4.4 STDOUT:` doesn't work either. Actually, this could have been handled with #2192 by allowing two possible results for Bash.

akinomyoga:
I checked the differences in Bash 5.

```
var-ref.test.sh
case    line    bash    osh
  0      10     pass    pass    var ref ${!a}
  1      16     pass    pass    ${!ref-default}
  2      32     FAIL    pass    ${!undef:-}
```

The above FAIL is caused because Bash 5 made it `${!undef}` an explicit error. The comment of the spec test also says "bash gives empty string, but I feel like this could be an error", so Bash actually changed to follow your feeling. The OSH code also says this is weird behavior:

https://github.com/oils-for-unix/oils/blob/28c2f5c4eef5a6d4ea48cad53a13f9df14698d27/osh/word_eval.py#L885-L887

```
  3      50     FAIL    pass    comparison to ${!array[@]} keys (similar SYNTAX)
```

This is the issue with the hash ordering. The test case is flawed in the sense that it depends on the hash ordering.

```
  4      80     FAIL    pass    ${!a[@]-'default'} is legal
```

This is caused by the same behavioral change as case 2.

```
 20     387     FAIL    pass    Bizarre tilde expansion in array index
```

This has been fixed in Bash. The `FAIL` is because the spec test requires Bash to fail. We can just remove the BUG section for Bash.

```
 29     618     FAIL    pass    Array indirect expansion with replacements
 30     701     FAIL    FAIL    Array indirect expansion with @? conversion
```

These are also bugs of Bash 4. We can remove the BUG sections. For case 30, we need to add an `OK` section because of the difference in the quoting rule with `@Q`.

spec/var-ref.test.sh
akinomyoga: 810389985

akinomyoga:
I updated `spec/var-ref.py` from bash-4 to bash-5 in [branch `array-indirect-bash5`](https://github.com/oils-for-unix/oils/compare/array-indirect-suffix...array-indirect-bash5).

andychu:
OK, well if you think it's easier to do it before SparseArray, that's OK

I was worried about piling up a lot of problems in front of that

---

I guess it's a bug that we are not able to assert both bash 4 and bash 5 ???  

But I think it is OK to update to bash 5, drop bash 4, as you say


----

(Although funny thing is that I did just start using a new web host OpalStack, and they are using an old version of Red Hat Linux.  And then I ran into the bash 4.3 bug where empty arrays are confused with unset arrays, with `set -u`.   I remember I hit that bug a long time ago and it made me angry with bash.  And then I hit it again after 8 years maybe.  So the old bugs still matter ...)

andychu:
i.e in theory we should be able to test both bash 4 and bash 5, I thought some spec tests did that ... but maybe they are either all bash 4, or all bash 5

But in practice it might be better to just "move foward" to bash 5, even though sometimes old bugs matter.

(I know bash has all these shopt -s compat43 stuff, which we definitely don't want !!! :-) )

akinomyoga:
> And then I ran into the bash 4.3 bug where empty arrays are confused with unset arrays, with `set -u`. I remember I hit that bug a long time ago and it made me angry with bash. And then I hit it again after 8 years maybe. So the old bugs still matter ...)

Yeah, I think that is a 4.2 bug fixed in 4.3. That also troubles `bash-completion` very much. To work around this bug, `bash-completion` has many `if ((${#array[@]})); do "do something using ${array[@]}"; done` instead of simply `"do something using ${array[@]}"`.

> i.e in theory we should be able to test both bash 4 and bash 5,

Yeah, it is also an option to treat `bash-4` and `bash-5` as distinct shells in the test framework.

> I thought some spec tests did that ... but maybe they are either all bash 4, or all bash 5
> 
> But in practice it might be better to just "move foward" to bash 5, even though sometimes old bugs matter.

[branch `array-indirect-bash5`](https://github.com/oils-for-unix/oils/compare/array-indirect-suffix...array-indirect-bash5) switches from `bash-4` to `bash-5`. I'll submit it after this PR is merged.

> (I know bash has all this `shopt -s compat43` stuff, which we definitely don't want !!! :-) )

`compatXX` is prepared for the intentional change in the "correct behavior". The bug fixes are not reverted by the "compatXX" options; i.e., compatXX cannot be used to restore "buggy" behaviors in old Bash.

osh/word_eval.py
```diff
@@ -887,19 +887,22 @@ def _EvalVarRef(self, val, blame_tok, quoted, vsub_state, vtest_place):
 
             elif case(value_e.Str):
                 val = cast(value.Str, UP_val)
-                bvs_part = self.unsafe_arith.ParseVarRef(val.s, blame_tok)
-                return self._VarRefValue(bvs_part, quoted, vsub_state,
-                                         vtest_place)
+                var = val.s
```
andychu, line 890: How about calling this `bvs_str` or something?  I think that is what it is, it is a string that follows the syntax of what's inside `${...}`, the BracedVarSub

spec/var-ref.test.sh
```diff
@@ -76,10 +77,13 @@ A_keys=0
...
-#### ${!a[@]-'default'} is illegal
+#### ${!a[@]-'default'} is legal but fails with more than one element
 
-# bash disallows this when a is an array.  We make it an error because [@]
-# implies it's an array.
+# bash allows this construct, but the indirection fails when the array has more
+# than one element because the variable name contains a space.  OSH originally
+# made it an error unconditionally because [@] implies it's an array, so the
+# behavior has been different from Bash when the array has a single element.
+# We now changed it to follow Bash even when the array has a single element.
```
andychu, line 86: BTW is is possible now to fail in the middle of a word, and then `return 1` for the command that the word belongs to

That is done with FailGlob

I think that is what bash is doing

I am still a bit confused about the test case, but maybe we can wait for the bash 5 change to see if we want to do that or not

osh/word_eval.py
```diff
@@ -887,19 +887,22 @@ def _EvalVarRef(self, val, blame_tok, quoted, vsub_state, vtest_place):
...
 
             elif case(value_e.BashArray):  # caught earlier but OK
-                e_die('Indirect expansion of array')
+                val = cast(value.BashArray, UP_val)
+                var = ' '.join(bash_impl.BashArray_GetValues(val))
```
andychu, line 894: So this creates an INVALID bvs_str whenever the array or assoc has more than one value?  

Because something like `${a b c}` is never valid in bash/OSH

That part was confusing me.  I think that deserves a comment here

---

Actually a funny thing is this may be valid in YSH - I want to use the syntax

```
var y = ${x %2.f}
```

to replace `printf -v y '%.2f' $x`

andychu (commented):
OK I actually didn't understand the space issue, now I see, but I think the code needs a comment

Whenever I see `' '.join()` with data that's going to be parsed, I think it will be a string injection bug.   Usually that is not correct, and you need quoting before joining.

If it were a command, that would lead to injection bugs.  But sinec it is parsed like `${}`, I guess it is OK, and that's what bash does.

osh/word_eval.py
akinomyoga: Hmm, it needs to be a variable name of the form `name` or an array element `name[...]`. It cannot be like `a:-xx`, `a:offset:count`, or `a%.*`. How about `var_name_str` or `var_ref_str`?

andychu: var_ref_str sounds good, thanks!

akinomyoga: Thanks! I updated the variable name: [`8cfe898..6ff4b1b`](https://github.com/oils-for-unix/oils/compare/8cfe8981cae3f6d2e67edce2245686a429e2d9f0..6ff4b1b22ace07b659bb9d98499836e5b07b0ef0).

akinomyoga:
Hmm, I was actually still working on the other comments from you... I'll include additional changes to the next PR for upgrading test from bash-4 to bash-5.

osh/word_eval.py
akinomyoga: #2204

spec/var-ref.test.sh
akinomyoga: #2204

PR #2200, opened 2024-12-24
[osh] Fix regression of `${undef@P}` and `${undef@Q}` with `nounset`

akinomyoga:
This PR refactors and simplifies the existing code for checking `nounset`. This also fixes the regression of `${undef@P}` and `${undef@Q}` mentioned in #2186:

> https://github.com/oils-for-unix/oils/pull/2186#issuecomment-2537311633 @akinomyoga 
>
> > then change the test to
> > ```
> >            if not vsub_state.has_test_op and not vsub_state.has_nullary_op:
> >                 val = self._EmptyStrOrError(val, part.token)
> > ```
> >   
> > and then for `@Q` and `@P`, handle `case(value_e.Undef)` in the normal way, producing empty string
> 
> This change breaks `osh -uc "${undef@P}"` and `osh -uc "${undef@Q}"`. The `nounset` check is skipped.
>
> https://github.com/oils-for-unix/oils/pull/2186#issuecomment-2537349028 @andychu 
>
> Ahh we are missing test coverage for `-u` ? OK thank you, I will look at it
>
> https://github.com/oils-for-unix/oils/pull/2186#issuecomment-2537384972 @akinomyoga 
>
> As I mentioned, in my local repository, I already have commits to fix everything by refactoring the calls of `_EvalBracketOp`.
>
> https://github.com/oils-for-unix/oils/pull/2186#issuecomment-2537419550 @andychu 
>
> Ah OK, yes definitely feel free to fix it along with the other issues!

PR #2202, opened 2024-12-25
[syntax refactor] rename "BracedVarSub.{token => name_tok}"

akinomyoga:
https://github.com/oils-for-unix/oils/pull/2200#discussion_r1897040894

### DIFF
--- a/frontend/syntax.asdl
+++ b/frontend/syntax.asdl
@@ -137,7 +137,7 @@ module syntax
 
   BracedVarSub = (
       Token left,        # in dynamic ParseVarRef, same as name_tok
-      Token token,       # location for the name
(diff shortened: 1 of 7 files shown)


andychu:
Thank you!

----

You might already have a good way of doing this, but what I do is 

1. change the ASDL
2. build/py.sh all   (actually sometimes I use bulid/py.sh py-codegen which is faster)
3. devtools/types.sh check-types | tee worklist
4. vim worklist, then jump everywhere with `gF`

I only started doing that in the middle of Oils development ... 

---

Or sometimes I just `sed`, there are some old functions in `devtools/refactor.sh`

akinomyoga:
Thanks!

PR #2203, opened 2024-12-26
[osh] Follow the Bash change in `${!undef[@]-}`

akinomyoga:
This PR upgrades the spec tests in `var-ref` from bash-4 to bash-5 (dfc8fd8cc).

In addition, we change the behavior of `${!undef[@]-}` to match the behavior fixed in Bash 5 (7baa3a74d).

I wrote in https://github.com/oils-for-unix/oils/pull/2201#issuecomment-2563150818 that I would work on https://github.com/oils-for-unix/oils/pull/2201#discussion_r1897976186 and https://github.com/oils-for-unix/oils/pull/2201#discussion_r1897976947, but I decided to separate it from this PR because it seems to affect a wider range of behavior for `${!ref}` in general.

### DIFF
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -883,7 +883,9 @@ class AbstractWordEvaluator(StringWordEvaluator):
         UP_val = val
         with tagswitch(val) as case:
             if case(value_e.Undef):
-                return value.Undef  # ${!undef} is just weird bash behavior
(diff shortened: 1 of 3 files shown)


andychu:
OK great, thanks for updating all the tests with comments!

akinomyoga:
Thansk!

PR #2205, opened 2024-12-27
[ysh] Fix is-main when combined with use

PossiblyAShrub:
Previously, is-main would succeed in a module sourced via use


andychu:
Thank you for fixing this!

PR #2207, opened 2024-12-29
[build] Check if GLOB_PERIOD exists

Melkor333:
As discussed briefly in Zulip. Not sure what a "peoper" solution is, defining `GLOB_PERIOD` as `0` is probably wrong :)


andychu:
OK thanks -- actually I think it should  never be 0  in real life, so `0` is a good test of whether it exists!  Then we can print a warning on `shopt -s dotglob`

I also did the same for FNM_EXTMATCH, the extended glob issue

PR #2209, opened 2024-12-31
[ysh] Implement SparseArray `$[sp1 === sp2]`, `append ... (sp)`, and `$[bool(sp)]`

akinomyoga:
Each commit corresponds to `$[sp1 === sp2]`, `append ... (sp)`, or `$[bool(sp)]`.

### DIFF
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -471,6 +471,21 @@ def SparseArray_UnsetElement(sparse_val, index):
     return error_code_e.OK
 
 
+def SparseArray_Equals(lhs, rhs):
+    # type: (value.SparseArray, value.SparseArray) -> bool
+
+    len_lhs = len(lhs.d)
+    len_rhs = len(rhs.d)
+    if len_lhs != len_rhs:
+        return False
+
+    for index in lhs.d:
+        if index not in rhs.d or rhs.d[index] != lhs.d[index]:
+            return False
+
+    return True
+
+
 def SparseArray_ToStrForShellPrint(sparse_val):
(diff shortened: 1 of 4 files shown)


core/bash_impl.py
```diff
@@ -471,6 +476,21 @@ def SparseArray_UnsetElement(sparse_val, index):
...
 
+def SparseArray_Equals(lhs, rhs):
+    # type: (value.SparseArray, value.SparseArray) -> bool
+
+    len_lhs = len(lhs.d)
+    len_rhs = len(rhs.d)
+    if len_lhs != len_rhs:
+        return False
+
+    for index in lhs.d.keys():
```
andychu, line 487: We can remove .keys() here, since it will make a copy

In Python dictionaries iterate over their keys by default (which is unlike Go, which may be confusing!)

akinomyoga: Ah, yes! Thank you for checking these in detail! This is the same as https://github.com/oils-for-unix/oils/pull/2167#discussion_r1866875583 and #2169. I created the initial version of the commits in this PR even before opening #2167 and #2169, but I forgot to update the commits after #2167 and #2169. I'll update it.

akinomyoga: [`1121bfd..6726ba3`](https://github.com/oils-for-unix/oils/compare/1121bfd4837e976b2f6ef6ba0ccceb2e6e5a4bb0..6726ba3f16206970b47a2e4334c8f1207e687715)

andychu:
Thanks!

PR #2210, opened 2024-12-31
[osh] Implement `${a[@]@P}` and make `${a[@]@a}` and `${a[@]@Q}` consistent with Bash

akinomyoga:
This corresponds to **PR(b)** explained in #2201.

### The behavior for `@a`

* In Bash, for the same reason as described in #2208, `${a[@]@a}` generates `a a ... a` and `A A ... A` for indexed and
  associative arrays, respectively.
* In osh, for the same reason as described in #2208, `${a[@]@a}` generates single words `a` and `A` for indexed and
  associative arrays, respectively, because osh returns the type of the entire array.

### Others

This PR also includes the modification to `@Q`. Also, `@P` was only supported for a scalar variable, but now this PR implements it for BashArray/BashAssoc. 03cf713b2 is a related refactoring commit.

### DIFF
--- a/osh/prompt.py
+++ b/osh/prompt.py
@@ -242,22 +242,18 @@ class Evaluator(object):
 
         return ''.join(ret)
 
-    def EvalPrompt(self, UP_val):
-        # type: (value_t) -> str
+    def EvalPrompt(self, s):
+        # type: (str) -> str
         """Perform the two evaluations that bash does.
 
         Used by $PS1 and ${x@P}.
         """
-        if UP_val.tag() != value_e.Str:
-            return ''  # e.g. if the user does 'unset PS1'
-
-        val = cast(value.Str, UP_val)
 
         # Parse backslash escapes (cached)
-        tokens = self.tokens_cache.get(val.s)
+        tokens = self.tokens_cache.get(s)
         if tokens is None:
-            tokens = match.Ps1Tokens(val.s)
-            self.tokens_cache[val.s] = tokens
+            tokens = match.Ps1Tokens(s)
+            self.tokens_cache[s] = tokens
 
         # Replace values.
         ps1_str = self._ReplaceBackslashCodes(tokens)
@@ -304,7 +300,12 @@ class Evaluator(object):
         # Now try evaluating $PS1
         ps1_val = self.mem.env_config.GetVal('PS1')
         #log('ps1_val %s', ps1_val)
-        return self.EvalPrompt(ps1_val)
+        UP_ps1_val = ps1_val
+        if UP_ps1_val.tag() == value_e.Str:
+            ps1_val = cast(value.Str, UP_ps1_val)
+            return self.EvalPrompt(ps1_val.s)
+        else:
+            return ''  # e.g. if the user does 'unset PS1'
 
 
 PROMPT_COMMAND = 'PROMPT_COMMAND'
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -1129,7 +1129,18 @@ class AbstractWordEvaluator(StringWordEvaluator):
                     if cell.nameref:
                         chars.append('n')
 
-            result = value.Str(''.join(chars))
+            count = 1
+            with tagswitch(val) as case:
+                if case(value_e.Undef):
+                    count = 0
+                elif case(value_e.BashArray):
+                    val = cast(value.BashArray, UP_val)
+                    count = bash_impl.BashArray_Count(val)
+                elif case(value_e.BashAssoc):
+                    val = cast(value.BashAssoc, UP_val)
+                    count = bash_impl.BashAssoc_Count(val)
+
+            result = value.BashArray([''.join(chars)] * count)
 
         else:
             e_die('Var op %r not implemented' % lexer.TokenVal(op), op)
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -1078,7 +1078,10 @@ class AbstractWordEvaluator(StringWordEvaluator):
                     self._ProcessUndef(val, vsub_token, vsub_state)
 
                     # For unset variables, we do not generate any quoted words.
-                    result = value.Str('')
+                    if vsub_state.array_ref is not None:
+                        result = value.BashArray([])
+                    else:
+                        result = value.Str('')
 
                 elif case(value_e.Str):
                     str_val = cast(value.Str, UP_val)
@@ -1100,7 +1103,7 @@ class AbstractWordEvaluator(StringWordEvaluator):
                         # TODO: should use fastfunc.ShellEncode
                         j8_lite.MaybeShellEncode(s) for s in values
                     ]
-                    result = value.Str(' '.join(tmp))
+                    result = value.BashArray(tmp)
                 else:
                     e_die("Can't use @Q on %s" % ui.ValType(val), op)
 
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -1048,7 +1048,7 @@ class AbstractWordEvaluator(StringWordEvaluator):
         return val
 
     def _Nullary(self, val, op, var_name, vsub_token, vsub_state):
-        # type: (value_t, Token, Optional[str], Token, VarSubState) -> Tuple[value.Str, bool]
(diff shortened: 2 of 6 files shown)


osh/word_eval.py
```diff
@@ -1060,10 +1060,28 @@ def _Nullary(self, val, op, var_name, vsub_token, vsub_state):
                     result = value.Str('')
```
andychu, line 1060: this might need `type: value_t` because it's the first one

That is, it infers the specific type `value.Str`, but you may have to declare the parent type

osh/prompt.py
```diff
@@ -242,22 +242,18 @@ def _ReplaceBackslashCodes(self, tokens):
 
         return ''.join(ret)
 
-    def EvalPrompt(self, UP_val):
+    def EvalPrompt(self, s):
         # type: (value_t) -> str
```
andychu, line 246: This should be `type: (str) -> str` now, although I wonder if it will cause any other type errors

andychu (commented):
Thanks!  It looks like there are a few type errors, which can be checked with

devtools/types.sh check-oils

See log at the end here:

http://op.oilshell.org/uuu/github-jobs/8824/cpp-coverage.wwz/_tmp/soil/logs/compare-gcc-clang.txt

---

Happy New Year!

osh/prompt.py
akinomyoga: Yeah, I forgot to update the type annotation. I fixed it. It didn't cause any other type errors because I already updated all the other occurrences of the calls of `EvalPrompt` in commit 03cf713b280749558ab33f2dc1b47df464b86b64.

osh/word_eval.py
akinomyoga: Yeah, thanks.

akinomyoga:
I fixed type annotations in [`be89909..0a18208`](https://github.com/oils-for-unix/oils/compare/be899098be94b88a3df34414335a3dc27c4d27e5..0a18208c175dc53fc4a374060c7354c2586a3616), and the type check passes in my local environment. However, I got the following C++ compilation error:

```cpp
_gen/bin/oils_for_unix.mycpp.cc: In member function ‘BigStr* prompt::Evaluator::EvalFirstPrompt()’:
_gen/bin/oils_for_unix.mycpp.cc:41915:34: error: ‘class value_asdl::value_t’ has no member named ‘s’
41915 |     return this->EvalPrompt(val->s);
      |                                  ^
```

The corresponding Python code is this:

https://github.com/oils-for-unix/oils/blob/0a18208c175dc53fc4a374060c7354c2586a3616/osh/prompt.py#L303-L305

Why does `val` still `value_t` even though I cast it to `value.Str` by `val = cast(value.Str, ps1_val)`?

akinomyoga:
> Why is `val` still `value_t` even though I cast it to `value.Str` by `val = cast(value.Str, ps1_val)`?

I thought you once told me that `val = cast(...)` would declare a new variable in the local scope so the type would be different from the outer scope, but it now seems to interfere with the type of `val` in the outer scope. If I change the variable name to `val2`, the problem doesn't happen.

Anyway, I decided to use the convention of `UP_ps1_val` here: [`0a18208..c38fa23`](https://github.com/oils-for-unix/oils/compare/0a18208c175dc53fc4a374060c7354c2586a3616..c38fa237c0f74416967b1d6a1cfd8f58a102efb1).

akinomyoga:
I updated tests because now OSH implements the expected behavior: [`c38fa23..8e8271e`](https://github.com/oils-for-unix/oils/compare/c38fa237c0f74416967b1d6a1cfd8f58a102efb1..8e8271e492c7a2300dca4e20faf3e2b96dcb6bce).

spec/var-op-bash.test.sh
```diff
@@ -300,10 +300,11 @@ status=0
 status=0
 ## END
 ## OK osh STDOUT:
-status=1
+a b c
+status=0
 a $'b\\nc'
 status=0
-a
+a a a
```
andychu, line 307: I can see that OSH differs from bash because of a quoting difference, which is OK

But it also seems to differ with `a a a` vs. `''` ?

Or maybe that is addressed in a follow-up?


TBH I don't understand the reasons for the evaluation logic - in these cases, my review falls back to "does OSH match bash?"

spec/var-op-bash.test.sh
```diff
@@ -328,7 +329,8 @@ A - A
 status=0
 ## END
 ## OK osh STDOUT:
-status=1
+- y
```
andychu, line 332: One trick we could use to make bash and OSH match is to do

```
sq=\'
s/y/${sq}y${sq}/
```

or vice versa

And then add a comment that this "papers over" an irrelevant or acceptable difference

---

I am not saying we need to do this, only if you think it makes the tests more understandable

I think if the only difference is quoting, then we should feel free to "adjust" the output, as long as it doesn't "hide" any other bugs!

(We can also use strings like  `y55 y56 y57` etc. to reduce the changes of accidentally hiding a bug, i.e. add a random number)

spec/var-op-bash.test.sh
akinomyoga: Ah! Thanks for pointing out this. This is just my oversight. OSH should generate the same result as Bash. Let me think about how we should distinguish array-origin BashArray from `$@`-origin BashArray.

andychu: It is OK if we fix that later (or even not at all if nobody runs into it, and bash does not document it)

Sometimes when I encounter a quirk like this, I split it into a separate test case, and then mark it `oils_failures_allowed += 1`


(Basically because there seem to be infinite bash implementation details!   I think using `@` with a type query is EXTREMELY obscure)

So that way the first test case have bash STDOUT === osh STDOUT, which is a nice proprety

spec/var-op-bash.test.sh
akinomyoga: Does 0653c5e29 work for you? To avoid a false positive, I match the entire line for replacement.

andychu: yes looks good thanks!

spec/var-op-bash.test.sh
akinomyoga: Could you wait for a while? If it turns out to be easy to fix, I can fix it within this PR. If not, I'll handle it in a separate PR later.

akinomyoga: I decided to defer it. The best solution could be to introduce a new type to represent the list coming from `$@`, but one might relate it to the forthcoming switching from BashArray to SparseArray. We can discuss that later.

akinomyoga: Now this is an allowed OSH failure: fda3cd43d785c0c2d71ff9084091ee5ee63555d2

spec/var-op-bash.test.sh
```diff
@@ -472,3 +458,81 @@ a
...
+['a a a a a a a a a']
+['A', 'A', 'A', 'A']
+['A A A A']
+[]
+['']
+## END
+
+## OK bash STDOUT:
+@Q
+["'1'", "'2'", "'3'", "'4'", "'5'", "'6'", "'7'", "'8'", "'9'"]
```
andychu, line 518: This is a big "OK bash" block, but our intention is to make it basically the same as bash

If I read it over, I can see the difference comes from 2 sources

1. quoting
2. ordering of assoc arrays


So in general I think we can try to

1. split up the test cases, e.g. for `[] ['']` which are identical
2. "adjust" the difference with sed, like we've already done

Then we will have more GREEN, which is easier to read/verify later IMO ... it matches our intention better


In some tests, I use `echo ${A[@]} | sort` and so forth to mask the difference in order

andychu (commented):
OK sounds good!   I made a comment about minimizing "OK bash" blocks when our intention is to agree with bash, but this is a matter of judgement

There will always be some of these, but IMO it is easier to maintain later if we don't have to read and edit the large assertion blocks, because it's easy to make mistakes, i.e. copy/paste bugs where we didn't actually fix things

spec/var-op-bash.test.sh
akinomyoga: #2217

PR #2215, opened 2025-01-02
[osh] Test `${a[@]:-}`, `${a[*]:-}`, etc. based on the string result

akinomyoga:
This follows the behavioral change from Bash 4.4 to 5.0.


osh/word_eval.py
```diff
@@ -684,14 +685,31 @@ def _ApplyTestOp(
...
-                is_falsey = len(val.d) == 0
+                if tok.id in (Id.VTest_ColonHyphen, Id.VTest_ColonEquals,
+                              Id.VTest_ColonQMark, Id.VTest_ColonPlus):
+                    if vsub_state.join_array:
+                        sep_width = len(self.splitter.GetJoinChar())
+                    else:
+                        sep_width = 1  # we use ' ' for a[@]
+
+                    if sep_width == 0:
+                        is_falsey = all(len(s) == 0 for s in strs)
```
akinomyoga, line 706: Hmm, OK. I forgot that `all(...)` doesn't translate in mycpp, and I think you turned it into a translation error recently.

akinomyoga: [`9145e4f..fba1079`](https://github.com/oils-for-unix/oils/compare/9145e4f395bace2e7927445da6ca985abf9aa78e..fba10790d8e4ac3ab6923780ccb7e5e158cb0bda)

osh/word_eval.py
```diff
@@ -684,14 +685,35 @@ def _ApplyTestOp(
...
+                        sep_width = 1  # we use ' ' for a[@]
+
+                    if sep_width == 0:
+                        is_falsey = True
+                        for s in strs:
+                            if len(s) != 0:
+                                is_falsey = False
+                                break
+                    else:
+                        is_falsey = len(strs) == 0 or (len(strs) == 1 and
```
andychu, line 712: Hm this seems worth documenting in `doc/ref/chap-word-lang.md` under `op-test` ?

I didn't know this !

spec/var-op-test.test.sh
```diff
@@ -540,3 +540,154 @@ echo "${undef-\e}"
...
+
+# Bash 2.0..4.4 has a bug that "${a[@]:-xxx}" produces an empty string.  It
+# seemed to consider a[@] and a[*] are non-empty when there is at least one
+# element even if the element is empty.  This was fixed in Bash 5.0.
+#
+# ## BUG bash STDOUT:
+# a   : 'no-colon' 'with-colon'
+# a[0]: 'no-colon' 'with-colon'
+# a[@]: 'no-colon' 'with-colon'
+# a[*]: 'no-colon' 'with-colon'
```
andychu, line 591: I think it could be useful to split this test up like the other one, since bash agrees with OSH on the first part

Although it is a bit annoying because you have to share `test-hyphen`

I suppose that could be `source spec/testdata/var-op-test-helper.sh` or something

Or I guess it is OK to copy and paste too, since the function is short

andychu (commented):
OK thank you, it looks like this improves bash compatibility so it's good

Although now I am sort of interested in  "${@:-}" and so forth because it's a POSIX construct

---

I wonder if you can also improve test case 18, because it looks like we are missing coverage there.  I think it may hit these same code paths

Or add another test case below 18


---

I am a bit surprised by the difference between

```
set -- ""
set -- "" ""
```

and also zsh is different



```
#### $@ and - and +
echo argv=${@-minus}
echo argv=${@+plus}
echo argv=${@:-minus}
echo argv=${@:+plus}
```

osh/word_eval.py
akinomyoga: This is actually a combination of `$*` / `$@`, double quoting, and `IFS`. I just calculated the width of the resulting length without actually calculating the string, which may look non-trivial. I'll add documentation and also a code comment here.

akinomyoga: Hmm, I realized that OSH has another divergence from Bash:

```console
$ bash -c 'IFS=; set -- "1x2" "3x4"; echo $*'
1x2 3x4
$ bin/osh -c 'IFS=; set -- "1x2" "3x4"; echo $*'
1x23x4
```

akinomyoga: I added 228ea8040.

andychu: OK interesting, yeah that looks like a bug

Regarding IFS=, we have to rewrite word splitting ... long story, but you can see there are multiple bugs reported over the years in `test/spec.sh word-split`

This is one of the biggest remaining areas of divergence

I am not sure if this is the same bug

---

For `$*`,  is the problem simply that we should ignore IFS??  We should always use a space rather than the join char?  Not sure

```
$ dash -c 'IFS=x; set -- a b; echo $*'
a b

$ dash -c 'IFS=; set -- a b; echo $*'
a b
```

andychu: Basically I am not sure that we need to see if it's double quoted or not, for this case?

But I know that is an issue with word splitting in general.  We have a complex set of data structures like `Piece` that know about quoting

spec/var-op-test.test.sh
akinomyoga: 361ac8ca3

akinomyoga:
> I wonder if you can also improve test case 18, because it looks like we are missing coverage there. I think it may hit these same code paths
> 
> Or add another test case below 18

I added a set of tests in commit 655d14d09.

osh/word_eval.py
akinomyoga: > Basically I am not sure that we need to see if it's double quoted or not, for this case?

I think so. POSIX says (list formatting is added by me):

> **`$*`**
>
>Expands to the positional parameters, starting from one, initially producing one field for each positional parameter that is set.
>
> - When the expansion occurs in a context where field splitting will be performed, any empty fields may be discarded and each of the non-empty fields shall be further split as described in [2.6.5 Field Splitting](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_06_05).
> - When the expansion occurs in a context where field splitting will not be performed, the initial fields shall be joined to form a single field with the value of each parameter separated by the first character of the IFS variable if IFS contains at least one character, or separated by a <space> if IFS is unset, or with no separation if IFS is set to a null string.

So, `IFS` is used only in the context where *field splitting* (word splitting) is inactive.

> But I know that is an issue with word splitting in general. We have a complex set of data structures like `Piece` that know about quoting

I think we can include the context information in `SimpleVarSub` and `BracedVarSub` at the parser or lexer stage (instead of analyzing the AST on the evaluation time); i.e., `$*` in `echo "$*"` and `echo $*` would be treated as different constructs in the parser.

akinomyoga: I think a simple solution may be to convert `$*` to `$@` when `$*` is used in a context where field splitting is active. It seems `$*` and `$@` are functionally equivalent in the context where field splitting is performed.

osh/word_eval.py
```diff
@@ -684,14 +685,45 @@ def _ApplyTestOp(
...
 
-            elif case(value_e.BashAssoc):
-                val = cast(value.BashAssoc, UP_val)
-                is_falsey = len(val.d) == 0
+                if tok.id in (Id.VTest_ColonHyphen, Id.VTest_ColonEquals,
+                              Id.VTest_ColonQMark, Id.VTest_ColonPlus):
+                    # The first character of IFS is used as a separator only
+                    # for the double-quoted "$*", or otherwise, a space " " is
+                    # used (for $*, $@, and "$@").
+                    # TODO: We current do not check whether the current $* is
```
andychu, line 703: Hm I noticed this function has a "quoted" param, we might be able to use that

spec/var-op-test.test.sh
```diff
@@ -399,6 +399,76 @@ argv=minus
...
+echo argv=${*+plus}
+echo argv=${*:-minus}
+echo argv=${*:+plus}
+## STDOUT:
+argv=
+argv=plus
+argv=
+argv=plus
+## END
+## BUG mksh/osh STDOUT:
```
andychu, line 451: I usually let OSH fail and adjust allowed_failures

i.e. we don't usually have BUG osh, which is yellow - we make it RED 

This is basically a matter of convention

doc/ref/chap-word-lang.md
```diff
@@ -189,6 +189,59 @@ or second forms, it is interpreted as the third form of the variable name
 surrounded by operators.
 
 
+### op-bracket
+
+The suffix operator of the form `[key]` is used to reference inner strings of
+the operand value.  For an indexed array `${array[index]}` can be used to
+specify an array element at offset `index`, where `index` undergoes arithmetic
+evaluation.  For an associated array, `${assoc[key]}` can be used to specify a
```
andychu, line 197: typo associated -> associative

doc/ref/chap-word-lang.md
```diff
@@ -189,6 +189,59 @@ or second forms, it is interpreted as the third form of the variable name
...
+    <3>
+    $ printf '<%s>\n' ${a[@]}
+    <1>
+    <2>
+    <3>
+
+When the variable substituion is inside double quotations, the `[*]` form joins
+the elements by the first character of `IFS`:
+
+    $ printf '<%s>\n' "${a[*]}"
```
andychu, line 220: I think these cases can more concisely explained by a table, with 4 entries or 8 entries, like:

$@
"$@"
$*
"$*"

andychu (commented):
OK thanks, this looks good ... I guess the main point is that is an array with 1 value of empty string is considered empty

---


Thanks for adding the docs

I am going to adjust them a bit, and also adjust the test failures

andychu:
OK thanks

I also fixed the ${*:-} bug on top of this - https://github.com/oils-for-unix/oils/commit/31ce9f63777c2e2ca85228ae7711f1bb3bae8d62

Thanks for finding it and writing the tests - that is the hard part!

osh/word_eval.py
akinomyoga: Oh, OK. I haven't noticed that it corresponds to the condition I wanted!

spec/var-op-test.test.sh
akinomyoga: I actually wondered if it would be possible to add another type of confirmed failure (e.g. "TODO") that produces red characters (or characters of another color) in the table. With the current strategy,

- When the number of allowed failures is increased by some changes, we cannot identify which failure was the increased one. We need to check out an older commit, build it, and compare the test results manually. Or one needs to open the github-jobs website, looking up the corresponding page, and compare the results.
- Even when the problem is fixed in the process of implementing another feature, the failure can be left to be FAIL if the result does not *exactly* match the expected result (which probably describes the Bash behavior).

akinomyoga:
Thank you!

spec/var-op-test.test.sh
andychu: Hm well technically ALL the ~100 failures now are TODO !   :-)   Some of them will be classified as "OK osh", but never "BUG osh" I think ... our bugs are just things we haven't done yet

And notice there are ZERO FAIL on bash/dash/mksh/zsh ... so FAIL is really reserved for OSH

and BUG is reserved for the other shells

At least that is the convention I follow, but we can talk about it more if it causes a problem

akinomyoga: I'm not sure if I could call it "a problem", but at least the following situation can be improved:

> * When the number of allowed failures is increased by some changes, we cannot identify which failure was the increased one. We need to check out an older commit, build it, and compare the test results manually. Or one needs to open the github-jobs website, looking up the corresponding page, and compare the results.
> * Even when the problem is fixed in the process of implementing another feature, the failure can be left to be FAIL if the result does not _exactly_ match the expected result (which probably describes the Bash behavior).

PR #2216, opened 2025-01-02
[builtin/completion_osh] Support SparseArray COMPREPLY and COMP_ARGV

akinomyoga:


andychu:
OK thank you, this looks like good progress

Issue #24, opened 2017-08-05
cd fails when called without a parameter


lheckemann:
```
$ ./_bin/oil.ovm osh
osh$ cd
Traceback (most recent call last):
  File "/home/andy/git/oil/Python-2.7.13/Lib/runpy.py", line 174, in _run_module_as_main
  File "/home/andy/git/oil/Python-2.7.13/Lib/runpy.py", line 72, in _run_code
  File "/home/andy/git/oil/bin/oil.py", line 440, in <module>
  File "/home/andy/git/oil/bin/oil.py", line 427, in main
  File "/home/andy/git/oil/bin/oil.py", line 410, in OilMain
  File "/home/andy/git/oil/bin/oil.py", line 311, in OshMain
  File "/home/andy/git/oil/bin/oil.py", line 122, in InteractiveLoop
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 892, in Execute
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 856, in _Execute
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 674, in _Dispatch
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 885, in _ExecuteList
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 856, in _Execute
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 584, in _Dispatch
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 473, in _RunSimpleCommand
  File "/home/andy/git/oil/bin/../core/cmd_exec.py", line 214, in _RunBuiltin
  File "/home/andy/git/oil/bin/../core/builtin.py", line 406, in _Cd
IndexError: list index out of range
```

andychu:
should use $HOME

andychu:
Thanks for noticing the bug, fixed!

PR #2217, opened 2025-01-02
[spec/var-op-bash] Split #24 into respective tests for `@Q`, `@P`, and `@a`

akinomyoga:
https://github.com/oils-for-unix/oils/pull/2210#discussion_r1901061335

I split the test into three, but it seems non-trivial to normalize the results so that the expected STDOUT can be shared. It is because, to normalize the results (particularly for the hash ordering), one needs to assume the form of the results, which is actually the test target. In this PR, I just split the tests and reduced the OK sections for the tests with the common results. I think this is sufficient because it is now easy to compare these smaller OK sections with the corresponding PASS sections.

### DIFF
--- a/spec/var-op-bash.test.sh
+++ b/spec/var-op-bash.test.sh
@@ -460,12 +460,10 @@ A
 ## END
 
 
-#### Array expansion with nullary var ops
-
+#### Array expansion with nullary var op @Q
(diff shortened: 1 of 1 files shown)


andychu:
Thanks, this looks better!

PR #2219, opened 2025-01-04
[refactor] Use `bash_impl` to print arrays

akinomyoga:
This PR refactors the printing codes (j8 and pp_value) to use `bash_impl` instead of directly manipulating the internal representation of BashArray/SparseArray/BashAssoc.


andychu:
Great thanks!

akinomyoga:
Thank you!

PR #2220, opened 2025-01-04
[mycpp] Fix constructor with template arguments deleted in C++20

akinomyoga:
The constructor declaration/definition of the form `template<typename T> class C { C<T>(); }` is removed in C++20. This causes many compiler warnings in recent compilers. One can always declare it without template arguments (in any C++ versions): `template<typename T> class C { C(); }`.


andychu:
OK thanks, I didn't know this!

PR #2229, opened 2025-01-12
Stateful spec tests for `bind -x` behavior

KingMob:
These are the initial stateful tests for `bind -x` behavior.

I'm not clear yet on how to test changes to the READLINE_* envvars. I'm somewhat sure that READLINE_LINE changes would be reflected in the stream for pexpect, but afaict, pexpect has no concept of a cursor to even begin to test changes to READLINE_POINT.

I'm ignoring for now, since those are less important use cases.


andychu:
Thanks, sorry for mixing this up with the other PR!

PR #2230, opened 2025-01-14
[doc] Fix promptVal example

wbob:
was testing the docs to set myself a custom prompt as described in ~/.config/oils/yshrc and ran into:
```
<Runtime error: Attribute 'promptval' not found on Obj>
```


andychu:
Oh great thank you!  Sorry for the delay

andychu:
Let us know if you have any other feedback on the prompt, or otherwise

Thanks for trying it!

Issue #2235, opened 2025-01-23
osh: bash prompt special characters


simonLeary42:
This very well may be out of scope, feel free to close the issue if it is.

https://www.gnu.org/software/bash/manual/html_node/Controlling-the-Prompt.html

```sh
$ for symbol in a d e h H j l n r s t T @ A u v V w W '!' '#' '$'; do PS1="\\$symbol" osh; done

^D
<Error: \d is invalid or unimplemented in $PS1>
^D

^D
wheatley
^D
wheatley.local
^D
<Error: \j is invalid or unimplemented in $PS1>
^D
<Error: \l is invalid or unimplemented in $PS1>
^D


^D

^D
osh
^D
<Error: \t is invalid or unimplemented in $PS1>
^D
<Error: \T is invalid or unimplemented in $PS1>
^D
<Error: \@ is invalid or unimplemented in $PS1>
^D
11:11
^D
simon
^D
0.24.0
^D
<Error: \V is invalid or unimplemented in $PS1>
^D
~
^D
simon
^D
<Error: \! is invalid or unimplemented in $PS1>
^D
<Error: \# is invalid or unimplemented in $PS1>
^D
$
^D
```

the above example does not check `\D`, `\nnn`, `\[`, or `\]`, simply out of laziness on my part.

andychu:
The tests for the prompt are here:

https://oils.pub/release/0.26.0/test/spec.wwz/osh-py/prompt.html

```
$ test/spec.sh prompt 
prompt.test.sh
case    line    bash    osh
  0       3     pass    pass    sh -i
  1      14     pass    pass    \[\] are non-printing
  2      21     pass    pass    literal escapes
  3      26     pass    pass    special case for $
```

Writing failing tests is usually about half the work, and we merge failing tests ... if those exist, then other people may be motivated to make them go green (and that has happened before)

PR #2236, opened 2025-01-23
add tests for all bash prompt special characters

simonLeary42:
https://github.com/oils-for-unix/oils/issues/2235

### DIFF
--- a/spec/prompt.test.sh
+++ b/spec/prompt.test.sh
@@ -262,6 +262,7 @@ match=0
 
 
 #### \j for number of jobs
+set -m
 PS1='foo \j bar'
 echo "${PS1@P}" | egrep -q 'foo 0 bar'
 echo matched=$?
--- a/spec/prompt.test.sh
+++ b/spec/prompt.test.sh
@@ -247,11 +247,11 @@ echo "${PS1@P}" | egrep -q '^foo (bash|osh) bar$'
 echo match=$?
 
 PS1='foo \V bar'
-echo "${PS1@P}" | egrep -q '^foo [0-9]+.[0-9]+ bar$'
+echo "${PS1@P}" | egrep -q '^foo [0-9]+\.[0-9]+ bar$'
 echo match=$?
 
 PS1='foo \v bar'
-echo "${PS1@P}" | egrep -q '^foo [0-9]+.[0-9]+.[0-9]+ bar$'
+echo "${PS1@P}" | egrep -q '^foo [0-9]+\.[0-9]+\.[0-9]+ bar$'
 echo match=$?
 
 ## STDOUT:
--- a/spec/prompt.test.sh
+++ b/spec/prompt.test.sh
@@ -179,6 +179,15 @@ echo matched=$?
 matched=0
 ## END
 
+#### \d for date
+PS1='foo \d bar'
+echo "${PS1@P}" | egrep -q 'foo [A-Z][a-z]+ [A-Z][a-z]+ [0-9]+ bar'
+echo matched=$?
+
+## STDOUT:
+matched=0
+## END
+
 #### \D{%H:%M} for strftime
 PS1='foo \D{%H:%M} bar'
 echo "${PS1@P}" | egrep -q 'foo [0-9][0-9]:[0-9][0-9] bar'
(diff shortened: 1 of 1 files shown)


spec/prompt.test.sh
```diff
@@ -205,18 +241,78 @@ echo matched=$?
...
+
+## STDOUT:
+matched=0
+matched=0
+matched=0
+## END
+
+#### \l for TTY device basename
+PS1='foo \l bar'
+tty_device_basename="$(basename "$(tty)")"
```
simonLeary42, line 284: is it ok to rely on external programs `basename` and `tty`?

spec/prompt.test.sh
```diff
@@ -205,18 +241,78 @@ echo matched=$?
...
+echo matched=$?
+
+## STDOUT:
+matched=0
+matched=0
+## END
+
+#### \# for command number
+(
+osh -i <<EOF
```
simonLeary42, line 307: is it ok to start a subprocess here?

andychu:
Thanks, this looks very helpful!

I let the CI run, but the spec tests timed out after 15 minutes.  Then I cloned locally and noticed that test case 27 for `\j` hangs on my machine ?

If I do

```
test/spec.sh prompt -r 27 
```

then it hangs

---

Though I am not sure why, because when I run it manually it doesn't hang

```
$ cat _tmp/27
#### \j for number of jobs
PS1='foo \j bar'
echo "${PS1@P}" | egrep -q 'foo 0 bar'
echo matched=$?
sleep 999 &
echo "${PS1@P}" | egrep -q 'foo 1 bar'
echo matched=$?
kill %%
fg
echo "${PS1@P}" | egrep -q 'foo 0 bar'



andy@hoover:~/git/oils-for-unix/simonLeary42/oils$ bash _tmp/27
matched=0
matched=0
_tmp/27: line 9: fg: no job control

```

andychu:
Here is the CI dashboard

https://mb.oilshell.org/uuu/github-jobs/9010/

and then if you click through you can see a timeout after 15 minutes, which is the max we let it run:

https://mb.oilshell.org/uuu/github-jobs/9010/#job-ovm-tarball

---

The tests aren't really sandboxed, so we basically use a Docker container I build as a source of truth ... (could be improved at some point)

----

To make progress, I might just comment out test case 27 for now, and then we can see if the others pass?

If there are a bunch of green tests for bash, and a red ones for OSH, and it runs reliably in the CI, then that is good progress, and can be merged

---

I tend to break things up in steps, so we can do `\j` last I guess

andychu:
BTW the way I "extract" the test case is `test/spec.sh prompt -r 27 -p > _tmp/27` -- that prints it instead of running it

And I either run with `bash` on my machine, or a specific versions of `bash` that we have in `build/dev-shell.sh` , i.e. compiled from source

simonLeary42:
I'm struggling to come up with a good way to run the "command number" prompt test. I have to either find a reliable way to get the current "command number", which I'm not sure is even implemented in `osh`, and then just test that the next "command number" is the previous "command number" +1, or find a reliable way to spawn a subprocess of the current shell. I can't use `$0` because `-bash` isn't a valid path, I can't use `$SHELL` because `osh` doesn't set that variable as far as I can tell, and the "command number" doesn't reset when using parentheses. I could try and use other clues to figure out what the current shell is and then hard code the path to each shell, but I don't like that...

simonLeary42:
on my machine, all the prompt test cases pass for bash except 0 and 32.

andychu:
OK looks like 9 tests fail on OSH, and they pass on bash:

https://mb.oilshell.org/uuu/github-jobs/9023/ovm-tarball.wwz/_tmp/spec/osh-py/prompt.html

Thanks!

I pushed a change to add this metadata line, which lets the CI pass

```
## oils_failures_allowed: 9
```

and then we can "whittle down" the errors as we go

andychu:
I also added you as a collaborator, so you can run the CI yourself, and don't have to wait for me to hit "approve"

Most contributors directly modify a branch of `oils-for-unix/oils` for that reason, instead of having their own fork

It is a weird github limitation - https://github.com/oils-for-unix/oils/wiki/Contributing

andychu:
(The merged worked!  But from now on, please work in the upstream repo, so then you can see the CI results immediately without waiting for approval )

----


Thanks for the tests!   I was able to knock off `\v` and `\V` very quickly

https://github.com/oils-for-unix/oils/commit/b2da0dbda535b873ad9cc14ab265df66bdc599a8

And it worked in C++ transparently

Let me see about the others

andychu:
I think `\D{}` is actually more general than `\t \T \@ \d`

I mean I can understand why we should have them, but it annoys me to carry that over, since it is more readable to use `\D{}` and an explicit format IMO

rather than remembering what those mean

andychu:
I am not quite sure how to get history number or command number, and what the exact difference is

I think the history number should come from GNU readline -- if we already have an API, then it shouldn't be too hard to add

PR #2238, opened 2025-01-27
`sh_spec.py` check for range out of bounds

simonLeary42:
I ran into this little thing when I was running new test cases and was accidentally checked out to the wrong branch. It ran 0 cases and exited normally and I was confused.

### DIFF
--- a/test/sh_spec.py
+++ b/test/sh_spec.py
@@ -647,11 +647,11 @@ def RunCases(cases, case_predicate, shells, env, out, opts):
 
     #pprint.pprint(cases)
     if isinstance(case_predicate, spec_lib.RangePredicate) and (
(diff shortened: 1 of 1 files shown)


andychu:
Hm I tried to merge, but I forgot that this auth issue does block it:

```
Run soil/github-actions.sh publish-cpp-tarball
$OILS_GITHUB_KEY not set
```

I sent you an invite to write to the main repo -- can you try pushing this  to a branch there?

So it would be the `oils-for-unix/oils` repo, and then you can create the same branch

It might be something like


```
git remote add upstream git@github.com:oils-for-unix/oils.git
git push upstream
```

??

Sorry for the hassle

PR #2239, opened 2025-01-27
`sh_spec.py` check for range out of bounds

simonLeary42:
I ran into this little thing when I was running new test cases and was accidentally checked out to the wrong branch. It ran 0 cases and exited normally and I was confused.


simonLeary42:
https://github.com/oils-for-unix/oils/pull/2238

andychu:
Thank you!

Issue #2081, opened 2024-09-26
Incremental search overlaps long `PS1`


nisbet-hubbard:
When `PS1` is long enough, launching an incremental search doesn’t dismiss the prompt:

```bash
[root@longhostname root]# 
(i-search)`': name root]# 
```

The prompt goes away entirely only after the length of the search terms exceeds it.

Sometimes, it can linger on like this:

```bash
oils-for-unix-0.23.0]# lr _build/oils.shnix-0.23.0]# 
```

andychu:
Thanks for the report, I reproduced this ... I'm not sure why this is happening since we are basically just calling GNU readline with a prompt string

Maybe we are calling it the wrong way, or maybe it doesn't calculate the width correctly

andychu:
Hm this happens in both Python and C++ - `bin/osh` and `_bin/cxx-asan/osh`

the former calls `line_input()` directly

But it doesn't happen in `demo/cpython/readline_mod.py` , which also ends up calling `raw_input()`, in `frontend/py_readline.py`

that is odd

andychu:
Oh man I tracked this down to

```
core/comp_ui.py:    readline.parse_and_bind('set horizontal-scroll-mode on') 
```

disabling that restores the bash-like behavior

I added that in order to fix an issue with line wrapping, and our custom completion UI

Gah

we are trying to complete in a nicer way than bash, by not re-printing the prompt all the time, but it appears that GNU readline may not support that without bugs

I guess we could change the default back to something bash-like, and provide the current behavior as an option (which I like better)

andychu:
@melvinw Thanks for finding this bug!  I am not sure what to do about this, other than downgrade to the bash-like experience, which doesn't use "reuse the prompt line"

i.e. zsh and OSH reuse the prompt line, by printing ANSI sequences to go up, whereas bash doesn't

I feel that is nicer, but it causes this bug in OSH.  (zsh has their own line editor of course; they don't use GNU readline)

---

It could be that GNU readline only correctly supports the bash-like behavior, and that's my suspicion, although I didn't look into it that deeply

https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/incremental.20search.20bug

andychu:
Related issues

- https://github.com/oils-for-unix/oils/issues/257 - this is the main one, which caused `horizontal-scroll-mode`.  The root cause is our zsh-like behavior, as opposed to bash-like
[...]

Issue #257, opened 2019-03-19
Detect when the user types past the last column in the terminal


jyn514:
This has a symptom similar to https://github.com/oilshell/oil/issues/252 and that's how I found it, but the cause I think is different. When completing a line which wraps to the next line, the *completed* line overwrite the *suggestions*:

![overwrite](https://user-images.githubusercontent.com/23638587/54642014-fc835e00-4a69-11e9-8007-6da41920f73b.png)

When we backspace, it deletes the suggestions on the current line:

![deletes the suggestions on the current line](https://user-images.githubusercontent.com/23638587/54642070-23419480-4a6a-11e9-8435-4f4cd297bcb8.png).

If we try to tab-complete again, it gets stranger: the cursor goes to the right of the screen and parts of the prompt get written on the new line.

![image](https://user-images.githubusercontent.com/23638587/54642214-6dc31100-4a6a-11e9-8476-b3d19513773a.png)

This is probably caused by `core/comp_ui.py` not taking into account the line wrap.

jyn514:
Related to https://github.com/oilshell/oil/issues/240

andychu:
Yeah there are a lot of bugs in this area... I am going to fix some known ones, and then we'll have to do another round of testing to see what's left :-/

Right now OSH has --completion-display=nice or =basic.  I think I have to revert to the "basic" one unfortunately.  The "nice" one is more like zsh, but probably has more bugs.

Both of them have to known the terminal width, which is related to the SIGWINCH issue.  I noticed that the Python interpreter, OSH, and readline are all fighting over SIGWINCH.

andychu:
I think I fixed most of the low hanging display bugs, and I plan to make a release today or tomorrow, so we can do another round of testing.

However this one is somewhat fundamental.  To restate the issue clearly:

- I'm trying to make a nicer zsh- or fish- like UI where hitting tab doesn't scroll the screen every time.  I think this is a much better experience.
- We do this by calling display.EraseLines() after the user hits ENTER or Ctrl-C or Ctrl-D, etc.
- We're using GNU readline, and we don't get events on every keypress.  So we don't know when the user typed past the last column.
- I tested zsh and fish, and they both have some method to detect the last column.
  - zsh erases the next line once the cursor spills over onto it (hm maybe readline has an option for that?)
  - fish actually wraps the command onto the newline!

andychu:
[...]

PR #2241, opened 2025-01-31
make minimal completion display more bash-like

melvinw:
There are three commits here. One makes it so only the nice completion display sets `horizontal-scroll-mode`. This should help with #2081. The other changes the display style of the minimal completion display. Instead of only printing a subset of matches in a single column and saying that there are more, we now just print them all in the packed format. This behavior will be more similar to what bash users are familiar with and just generally more useful. The last change makes the minimal completion display the default. This should minimize friction for newcomers.

### DIFF
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -146,10 +146,13 @@ class MinimalDisplay(_IDisplay):
     without testing it.
     """
 
-    def __init__(self, comp_state, prompt_state, debug_f):
-        # type: (State, PromptState, _DebugFile) -> None
+    def __init__(self, comp_state, prompt_state, debug_f, term_width, signal_safe):
+        # type: (State, PromptState, _DebugFile, int, iolib.SignalSafe) -> None
         _IDisplay.__init__(self, comp_state, prompt_state, 10, mylib.Stdout(),
                            debug_f)
+
+        self.signal_safe = signal_safe
+        self.term_width = term_width
 
     def _RedrawPrompt(self):
         # type: () -> None
@@ -158,6 +161,19 @@ class MinimalDisplay(_IDisplay):
         self.f.write(self.prompt_state.last_prompt_str)
         self.f.write(self.comp_state.line_until_tab)
 
+    def _GetTerminalWidth(self):
+        # type: () -> int
+        if self.signal_safe.PollSigWinch():  # is our value dirty?
+            try:
+                self.term_width = libc.get_terminal_width()
+            except (IOError, OSError):
+                # This shouldn't raise IOError because we did it at startup!  Under
+                # rare circumstances stdin can change, e.g. if you do exec <&
+                # input.txt.  So we have a fallback.
+                self.term_width = 80
+
+        return self.term_width
+
     def _PrintCandidates(self, unused_subst, matches, unused_match_len):
         # type: (Optional[str], List[str], int) -> None
         #log('_PrintCandidates %s', matches)
@@ -165,22 +181,11 @@ class MinimalDisplay(_IDisplay):
         display_pos = self.comp_state.display_pos
         assert display_pos != -1
 
-        too_many = False
-        i = 0
-        for m in matches:
-            self.f.write(' %s\n' % m[display_pos:])
-
-            if i == self.num_lines_cap:
-                too_many = True
-                i += 1  # Count this one
-                break
-
-            i += 1
-
-        if too_many:
-            num_left = len(matches) - i
-            if num_left:
-                self.f.write(' ... and %d more\n' % num_left)
+        to_display = [m[display_pos:] for m in matches]
+        lens = [len(m) for m in to_display]
+        max_match_len = max(lens)
+        term_width = self._GetTerminalWidth()
+        _PrintPacked(to_display, max_match_len, term_width, -1, self.f)
 
         self._RedrawPrompt()
 
--- a/core/comp_ui.py
+++ b/core/comp_ui.py
@@ -30,6 +30,17 @@ DEFAULT_TERM_WIDTH = 80
 DEFAULT_MATCH_LINE_LIMIT = 10
 
 
+def _GetTerminalWidth():
+    # type: () -> int
+    try:
+        return libc.get_terminal_width()
+    except (IOError, OSError):
+        # This shouldn't raise IOError because we did it at startup!  Under
+        # rare circumstances stdin can change, e.g. if you do exec <&
+        # input.txt.  So we have a fallback.
+        return DEFAULT_TERM_WIDTH
+
+
 def _PromptLen(prompt_str):
     # type: (str) -> int
     """Ignore all characters between \x01 and \x02 and handle unicode
@@ -95,24 +106,13 @@ class _IDisplay(object):
         self.num_lines_cap = num_lines_cap
         self.f = f
         self.debug_f = debug_f
-        self.term_width = DEFAULT_TERM_WIDTH
-        try:
-            self.term_width = libc.get_terminal_width()
-        except (IOError, OSError):  # stdin not a terminal
-            pass
-
+        self.term_width = _GetTerminalWidth()
         self.signal_safe = signal_safe
 
     def _GetTerminalWidth(self):
         # type: () -> int
         if self.signal_safe.PollSigWinch():  # is our value dirty?
-            try:
-                self.term_width = libc.get_terminal_width()
-            except (IOError, OSError):
-                # This shouldn't raise IOError because we did it at startup!  Under
-                # rare circumstances stdin can change, e.g. if you do exec <&
-                # input.txt.  So we have a fallback.
-                self.term_width = DEFAULT_TERM_WIDTH
+            self.term_width = _GetTerminalWidth()
 
         return self.term_width
 
@@ -356,6 +356,8 @@ class NiceDisplay(_IDisplay):
 
     def ReadlineInitCommands(self):
         # type: () -> List[str]
+        # NOTE: This setting prevents line-wrapping from clobbering completion
+        # output. See https://github.com/oils-for-unix/oils/issues/257
         return ['set horizontal-scroll-mode on']
 
     def Reset(self):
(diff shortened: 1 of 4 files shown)


melvinw:
Here are some screencasts of how the new minimal mode behaves.

Line wrapping: https://asciinema.org/a/0kltGyy5yse6oYavHwRkite2y

Packed match printing: https://asciinema.org/a/iWcwC8m0XOa7lpnB2sRtKmzCy

core/comp_ui.py
```diff
@@ -142,41 +146,46 @@ class MinimalDisplay(_IDisplay):
...
+        self.term_width = term_width
+
     def _RedrawPrompt(self):
         # type: () -> None
         # NOTE: This has to reprint the prompt and the command line!
         # Like bash, we SAVE the prompt and print it, rather than re-evaluating it.
         self.f.write(self.prompt_state.last_prompt_str)
         self.f.write(self.comp_state.line_until_tab)
 
+    def _GetTerminalWidth(self):
```
andychu, line 164: I think we can move this _GetTerminalWidth method up to the base class _IDisplay?

core/comp_ui.py
```diff
@@ -142,41 +146,46 @@ class MinimalDisplay(_IDisplay):
...
+        # type: () -> int
+        if self.signal_safe.PollSigWinch():  # is our value dirty?
+            try:
+                self.term_width = libc.get_terminal_width()
+            except (IOError, OSError):
+                # This shouldn't raise IOError because we did it at startup!  Under
+                # rare circumstances stdin can change, e.g. if you do exec <&
+                # input.txt.  So we have a fallback.
+                self.term_width = 80
```
andychu, line 173: We can change this 80 to DEFAULT_TERMINAL_WIDTH I guess

core/shell.py
```diff
@@ -1084,19 +1086,22 @@ def Main(
...
             term_width = 0
-            if flag.completion_display == 'nice':
-                try:
-                    term_width = libc.get_terminal_width()
-                except (IOError, OSError):  # stdin not a terminal
-                    pass
+            try:
+                term_width = libc.get_terminal_width()
+            except (IOError, OSError):  # stdin not a terminal
+                pass
```
andychu, line 1092: I think we should set term_width to DEFAULT_TERM_WIDTH here, for BOTH 'nice' and 'minimal'


Then the logic below will be cleaner

core/shell.py
```diff
@@ -1106,7 +1111,8 @@ def Main(
 
         else:  # Without readline module
             display = comp_ui.MinimalDisplay(comp_ui_state, prompt_state,
-                                             debug_f)
+                                             debug_f, DEFAULT_TERM_WIDTH,
```
andychu, line 1114: without GNU readline, I think we can still use the real term width?

andychu (commented):
OK great, I looked at the code

Let me try out the behavior a bit

andychu:
I tried it out -- the main thing is if I hit TAB with nothing on the prompt, it writes over the entire screen in an undesirable way

Can we set max_lines to maybe 5 or 8 ?

core/shell.py
andychu: We can also get rid of `term_width = 0` then, since it will be initialized in both cases

core/comp_ui.py
```diff
@@ -340,6 +354,10 @@ def __init__(
         # hash of matches -> count.  Has exactly ONE entry at a time.
         self.dupes = {}  # type: Dict[int, int]
 
+    def ReadlineInitCommands(self):
+        # type: () -> List[str]
+        return ['set horizontal-scroll-mode on']
```
andychu, line 359: Please add a comment with this link, so we remember what it's for!

https://github.com/oils-for-unix/oils/issues/257

core/comp_ui.py
```diff
@@ -85,13 +88,37 @@ def __init__(self):
...
-    def __init__(self, comp_state, prompt_state, num_lines_cap, f, debug_f):
-        # type: (State, PromptState, int, mylib.Writer, _DebugFile) -> None
+    def __init__(self, comp_state, prompt_state, num_lines_cap, f, debug_f, signal_safe):
+        # type: (State, PromptState, int, mylib.Writer, _DebugFile, iolib.SignalSafe) -> None
         self.comp_state = comp_state
         self.prompt_state = prompt_state
         self.num_lines_cap = num_lines_cap
         self.f = f
         self.debug_f = debug_f
+        self.term_width = DEFAULT_TERM_WIDTH
```
andychu, line 98: I think we should have a free function `_GetTermWidth()` that does the try/except, and that will be called in `__init__` here

And then we will have a **method** `self._MaybeGetTermWidth()` that does the SIGWINCH thing

that will have a bit less duplication with DEFAULT_TERM_WIDTH

andychu (commented):
Thanks, looks good!

core/comp_ui.py
melvinw: Good idea. Pulled that logic out of the method. I left the method name alone, though. "Maybe" suggests that it might fail or something and return None/-1, which I thought was more confusing than the name collision (the "self." disambiguates this enough IMO)

core/comp_ui.py
melvinw: Nice catch! Added a comment

andychu:
Thanks, code looks good

I was about to merge and then I saw this failure!  

http://mb.oilshell.org/uuu/github-jobs/9044/

That's a little scary, but I think it is a mycpp bug.  I think we need a different name for the method and function?

Because in C++ you can call a method like `MyMethod()`, it's not `this->MyMethod()`  

I am not sure that is the cause but I think so ... we are doing a naive syntactic translation

----

Also I was wondering if it is weird to call a function in the constructor, in `Alloc<T>()`, but that might be OK






```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==15622==ERROR: AddressSanitizer: SEGV on unknown address 0x00000000001c (pc 0x55cb21d55fd3 bp 0x7fff089e6910 sp 0x7fff089e68f0 T0)
==15622==The signal is caused by a READ memory access.
==15622==Hint: address points to the zero page.
    #0 0x55cb21d55fd2 in iolib::SignalSafe::PollSigWinch() (/home/uke/oil/_bin/cxx-asan/oils-for-unix+0x13afd2)
    #1 0x55cb21e09948 in comp_ui::_IDisplay::_GetTerminalWidth() _gen/bin/oils_for_unix.mycpp.cc:21086
    #2 0x55cb21e09896 in comp_ui::_IDisplay::_IDisplay(comp_ui::State*, comp_ui::PromptState*, int, mylib::Writer*, util::_DebugFile*, iolib::SignalSafe*) _gen/bin/oils_for_unix.mycpp.cc:21081
    #3 0x55cb21e09f88 in comp_ui::MinimalDisplay::MinimalDisplay(comp_ui::State*, comp_ui::PromptState*, util::_DebugFile*, iolib::SignalSafe*) _gen/bin/oils_for_unix.mycpp.cc:21128
    #4 0x55cb22119b2f in comp_ui::MinimalDisplay* Alloc<comp_ui::MinimalDisplay, comp_ui::State*&, comp_ui::PromptState*&, util::_DebugFile*&, iolib::SignalSafe*&>(comp_ui::State*&, comp_ui::PromptState*&, util::_DebugFile*&, iolib::SignalSafe*&) (/home/uke/oil/_bin/cxx-asan/oils-for-unix+0x4feb2f)
    #5 0x55cb22005b76 in shell::Main(BigStr*, args::Reader*, Dict<BigStr*, BigStr*>*, bool, pyutil::_ResourceLoader*, py_readline::Readline*) _gen/bin/oils_for_unix.mycpp.cc:59043
    #6 0x55cb21d75732 in oils_for_unix::AppBundleMain(List<BigStr*>*) _gen/bin/oils_for_unix.mycpp.cc:10731
    #7 0x55cb21d75de5 in oils_for_unix::main(List<BigStr*>*) _gen/bin/oils_for_unix.mycpp.cc:10767
    #8 0x55cb220091d8 in main _gen/bin/oils_for_unix.mycpp.cc:59180
    #9 0x7f0aa4ce509a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2409a)
    #10 0x55cb21d53fd9 in _start (/home/uke/oil/_bin/cxx-asan/oils-for-unix+0x138fd9)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/home/uke/oil/_bin/cxx-asan/oils-for-unix+0x13afd2) in iolib::SignalSafe::PollSigWinch()
==15622==ABORTING

test/ysh-runtime-errors.sh: fatal: Should fail under YSH: expected status 3, got 1
FAIL  test-bug-2129

```

melvinw:
Oh weird... Pushed a fix to rename the method

melvinw:
Also, yeah I think it should be safe to call `_GetTerminalWidth()` in the constructor. It catches any exceptions that might otherwise bubble up and cause problems

melvinw:
Huh. Looks like the translation bug is fixed, but there's still this other failure in `cpp-spec`
```
...
done: all-tests-to-html
builtin-trap failed with status 1

*** 1 tests FAILED
test/spec-version.sh: line 71: zsh: command not found

Results: file:///home/uke/oil/_tmp/spec/osh-cpp/index.html

real	3m6.446s
user	5m12.757s
sys	3m40.627s
Comparison: file:///home/uke/oil/_tmp/spec/osh-cpp/compare.html
```

andychu:
I think the error might be because some of the members aren't initialized yet or something ... ASAN is catching it

But in any case we are lucky that it caught it

---

It looks like this test is failing in Python only, not sure why:

http://mb.oilshell.org/uuu/github-jobs/9045/cpp-spec.wwz/_tmp/spec/osh-cpp/builtin-trap.html

http://mb.oilshell.org/uuu/github-jobs/9045/cpp-spec.wwz/_tmp/spec/osh-cpp/builtin-trap.html#details-20-osh

hmmmm

andychu:
Hm I restarted the job, and it passed ... that is annoying, I haven't seen this flaky test before

Anyway it's merged now - thanks!

PR #2244, opened 2025-02-02
add OILS_COMP_UI

melvinw:
This will be more user-friendly than the existing command-line arg, since people can set it in their init files.

### DIFF
--- a/doc/ref/chap-special-var.md
+++ b/doc/ref/chap-special-var.md
@@ -383,9 +383,11 @@ Override the default YSH history location.
 
 ### OILS_COMP_UI
 
-Set which completion UI to use. Set to `minimal` for a UI that approximates the
-default behavior of readline. Set to `nice` for a UI with a fancy pager and a
-prompt with horizontal scrolling instead of wrapping. The default is `minimal`.
+Set which completion UI to use. Defaults to `minimal`.
+
+- `minimal` - a UI that approximates the default behavior of GNU readline.
+- `nice` - a UI with a fancy pager and a prompt with horizontal scrolling instead of wrapping.
+
 This variable is currently only checked once during shell initialization.
 
 ## cd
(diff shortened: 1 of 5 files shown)


andychu:
OK I think this is a good change

We should document it in:

- `doc/ref/toc-osh` - maybe near `OILS_VERSION`, add an `[Interactive]` section that lists OILS_COMP_UI
- `doc/ref/toc-ysh` - ditto - applies to both OSH and YSH
- `doc/ref/chap-special-var.md` - there should be a `## Interactive`, and then `### OILS_COMP_UI`

and I would mention that  it is only consulted at startup

(and then we can also document OILS_MAX_COMP_LINES in the same place)

----

It is possible to write a test for this?  I guess it is hard to test because it changes the TAB behavior only ...

I suppose the only way we have of testing that is `spec/stateful`, which uses pexpect

I guess we are going to run into the same issue with OILS_MAX_COMP_LINES - it is hard to test

hmmmm

melvinw:
Pushed some docs. Let me know if I should tweak the wording

doc/ref/chap-special-var.md
```diff
@@ -379,6 +379,15 @@ Override the default OSH history location.
 
 Override the default YSH history location.
 
+## Interactive
+
+### OILS_COMP_UI
+
+Set which completion UI to use. Set to `minimal` for a UI that approximates the
+default behavior of readline. Set to `nice` for a UI with a fancy pager and a
```
andychu, line 387: I would say "GNU readline" because some people might not know what "readline" is

I think it will nicer to read with a bulleted list, like

- `minimal` - ...
- `nice` - ...

andychu (commented):
Looks good!

doc/ref/chap-special-var.md
melvinw: 👍 done

andychu:
Great thanks!

I pushed a `test/manual.sh` for now ... I am wondering about this `ENV.OILS_COMP_UI` issue for YSH, which I started doing in November or so ...

( I am trying to untangle the confusion of global vars in shell ... which ones are from ENV and which ones aren't?  There are a bunch of quirks around PWD, PS1, etc. )

Issue #627, opened 2020-03-06
Incorrect behavior of $@ and $*  when IFS=''


xPMo:
```bash
set -- a 'b c'
IFS=''
printf '[%s]\n' $@ $*
```

In bash or dash, this prints:

```
[a]
[b c]
[a]
[b c]
```

In osh, this prints:

```
[ab c]
[ab c]
```

The same applies to bash arrays under `${a[@]}` and `${a[*]}`.
In general, an array without quotes should not be subject to joining. First substitute all words, *then* split them on IFS. (Exceptions: `[[` and `case`)

andychu:
Wow interesting, I'm not sure why this is yet, I verified that bash/mksh/zsh/busybox ash agree.

However dash 0.5.8 agrees with osh.

xPMo:
Dash 0.5.10 agrees with Bash, so does [whatever version of dash this is](https://tio.run/##S0kszvj/vzi1REFXVyFRQT1JIVmdy9Mt2JaroCgzryRNQT1atTg2Jk9dQcVBQUXrf3FGomFxaa6CflJmnn4KSDMA)

My sha1sum is `847e4cc4f63e83160d166574e5e312f5f85c9772`

xPMo:
> 2014-10-08  Herbert Xu <herbert@gondor.apana.org.au>

> 	* Split unquoted $@/$* correctly when IFS is set but empty.

[Changelog]https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/ChangeLog)

andychu:
OK thanks a lot!  That helps.  I verified this bug through the spec tests.

I think it should be pretty easy to fix but I'm not quite sure.  I think test case 8 here is related

https://www.oilshell.org/release/0.8.pre2/test/spec.wwz/assoc.html

xPMo:
>  I think test case 8 here is related

It is similar, but I don't think it's related. Unquoted `$*` or `$@`  shouldn't * join*, while quoted `"$*"` shouldn't  *split*:

```bash
$ set -- a 'b   c'
$ printf '[%s]\n' "$*"
[a b   c]    # osh prints "[a b c]"
```

xPMo:
Generalizing the behavior with an arbitrary string concatenated before or after the `${@}`/`${*}`:

**`"pre${@}post"`**
- Let `${N}` be the last element of `"$@"`.
- Substitute `"pre${1}" "${2}" ... "${N}post"`

[...]

Issue #48, opened 2017-10-18
make clean doesn't work in the tarball distribution


andychu:
This is because all the Python source is in the app bundle already.  Hm.

```
~/src/oil-0.1.0$ make clean
build/actions.sh clean
find: ‘Python-2.7.13/Lib’: No such file or directory
Makefile:89: recipe for target 'clean' failed
make: *** [clean] Error 1
```

andychu:
Fixed by this commit: https://github.com/oilshell/oil/commit/297593f1b31040b5b2d8900374f1bdc4515d21af

Issue #47, opened 2017-10-13
Improve error message when binary is stripped (Assertion error making OSH AUR package)


andychu:
Hm this is because the `./configure` step writes out the PREFIX to `_tmp/detected-config.sh`.  If the file isn't installed to that place, it won't run.

The core issue is that a Unix executable doesn't know where it is run from!  The executable has to be set as `PYTHONPATH` in order to find the bytecode.

https://stackoverflow.com/questions/4031672/without-access-to-argv0-how-do-i-get-the-program-name

```
$>   export OVM_VERBOSE=1
$>   oil.ovm 
ovm_path = ovm_path_buf (/usr/local/bin/oil.ovm)
ovm_path: /usr/local/bin/oil.ovm
# installing zipimport hook
import zipimport # builtin
# installed zipimport hook
oil.ovm: Modules/main.c:347: Ovm_Main: Assertion `sts != -1' failed.
Aborted (core dumped)
```

> In the PKGBUILD, I do execute ./configure and make to compile OSH. I do not, however, invoke the install script that ships with OSH. This is because the PKGBUILD convention dictates that you shouldn’t rely on an external script to actually do the placement of binaries on the system.

> The PKGBUILD convention prefers that any binaries that need to be installed on the system be placed within a specific subdirectory beneath wherever the PKGBUILD file is sitting (this is referred to as $pkgdir).

andychu:
Possible solutions:

- Instead of finding the .zip file dynamically with `PYTHONPATH` / open(), we can compile it into the binary.
- We could use two files instead of a concatenated file.  I don't like that distros put stuff in `/usr/local/share` though?  But that means that arch-independent and arch-dependent files are separate, which is good.

timetoplatypus:
I don't follow what you mean by "The executable has to be set as PYTHONPATH in order to find the bytecode". Can you clarify?

andychu:
This blog post has some details:

http://www.oilshell.org/blog/2017/05/05.html

But to summarize, I wanted to treat the fact that Oil is written in Python as an implementation detail.  So basically I created an app bundle with the compiled Python interpreter and a zip file of all the `.pyc` files.

However it uses a trick about zip files.. you can simply do this:

    cat python_interpreter app.zip > oil.ovm

[...]

Issue #12, opened 2017-03-10
Spec tests fail with ASCII locale ?


andychu:
This was reported on the mailing list and also here:

https://lobste.rs/s/obktxo/oil_shell_help_needed_if_you_have_debian/comments/akvbsk#c_akvbsk

sh_spec.py needs to read with a known encoding.  *.test.sh should be restricted to utf-8 ?

```
$ ./spec.sh all
./spec-runner.sh run-cases append 
./spec-runner.sh run-cases arith-context 
./spec-runner.sh run-cases arith 
./spec-runner.sh run-cases array 
./spec-runner.sh run-cases assign 
./spec-runner.sh run-cases assoc-zsh 
./spec-runner.sh run-cases assoc 
./spec-runner.sh run-cases blog1 
Traceback (most recent call last):
  File "./sh_spec.py", line 838, in <module>
    sys.exit(main(sys.argv))
  File "./sh_spec.py", line 766, in main
    tokens = Tokenizer(LineIter(f))
  File "./sh_spec.py", line 125, in __init__
    self.next()
  File "./sh_spec.py", line 129, in next
    self.cursor = self.it.__next__()
  File "./sh_spec.py", line 87, in LineIter
    for i, line in enumerate(f):
  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 133: ordinal not in range(128)
./spec-runner.sh run-cases brace-expansion 
./spec-runner.sh run-cases bugs 
./spec-runner.sh run-cases builtins 
./spec-runner.sh run-cases case_ 
./spec-runner.sh run-cases command-sub 
./spec-runner.sh run-cases comments 
*** Got 1 allowed osh failures, exit with status 0
./spec-runner.sh run-cases dbracket 
Traceback (most recent call last):
  File "./sh_spec.py", line 838, in <module>
    sys.exit(main(sys.argv))
  File "./sh_spec.py", line 766, in main
    tokens = Tokenizer(LineIter(f))
  File "./sh_spec.py", line 125, in __init__
    self.next()
  File "./sh_spec.py", line 129, in next
    self.cursor = self.it.__next__()
  File "./sh_spec.py", line 87, in LineIter
    for i, line in enumerate(f):
  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
[...]

PR #2248, opened 2025-02-17
[osh] Fix unquoted `$*`, `${*}`, `${a[*]}`, etc. under empty IFS

akinomyoga:
Please see the commit message for the details.

This fixes an issue discussed in DM (Koichi Murase, Andy Chu, 2025-02-16) about the behavior of `$*` with `IFS=`. A related discussion is https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/How.20other.20shells.20implement.20IFS.

By the way, in the testing, I found Bash bugs: `${!prefix*}` and `${!arr[*]}` doesn't follow the rule with `IFS=`, as others `$*` and `${a[*]}` (See the added test cases, spec/word-split 39 and 40).

### DIFF
--- a/core/runtime.asdl
+++ b/core/runtime.asdl
@@ -52,7 +52,7 @@ module runtime
 
     # "$@" or "${a[@]}" # never globbed or split (though other shells
     # split them)
-  | Array(List[str] strs)
+  | Array(List[str] strs, bool quoted)
     # only produced when EXTGLOB_FS flag is passed
   | ExtGlob(List[part_value] part_vals)
 
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -223,16 +223,16 @@ def _ValueToPartValue(val, quoted, part_loc):
 
         elif case(value_e.BashArray):
             val = cast(value.BashArray, UP_val)
-            return part_value.Array(bash_impl.BashArray_GetValues(val))
+            return part_value.Array(bash_impl.BashArray_GetValues(val), quoted)
 
         elif case(value_e.SparseArray):
             val = cast(value.SparseArray, UP_val)
-            return part_value.Array(bash_impl.SparseArray_GetValues(val))
+            return part_value.Array(bash_impl.SparseArray_GetValues(val), quoted)
 
         elif case(value_e.BashAssoc):
             val = cast(value.BashAssoc, UP_val)
             # bash behavior: splice values!
-            return part_value.Array(bash_impl.BashAssoc_GetValues(val))
+            return part_value.Array(bash_impl.BashAssoc_GetValues(val), quoted)
 
         # Cases added for YSH
         # value_e.List is also here - we use val_ops.Stringify()s err message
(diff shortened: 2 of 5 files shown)


andychu:
Thanks for looking at this!

Unfortunately it looks like test cases 11 and 17 fail now here --  maybe that is similar to what you noticed on the first try:


http://op.oilshell.org/uuu/github-jobs/9101/ovm-tarball.wwz/_tmp/spec/osh-py/toysh-posix.html

compared to the latest release:

https://oils.pub/release/latest/test/spec.wwz/osh-py/toysh-posix.html

They are actually cases where shells **disagree**

 I ran both of them locally ... It looks like a real issue, but I was not able to understand it immediately

But I will take another look, after the other PR

andychu:
Hm I reduced test case 11 into test case 12

It seems like bash and mksh match here

But dash disagrees, and now OSH is matching dash

I am not sure why yet

```
$ bash -c 'IFS=x; set -- a "" b; argv $@; for x in $@; do echo -$x-; done'
['a', '', 'b']
-a-
--
-b-

$ mksh -c 'IFS=x; set -- a "" b; argv $@; for x in $@; do echo -$x-; done'
['a', '', 'b']
-a-
--
-b-

$ bin/osh -c 'IFS=x; set -- a "" b; argv $@; for x in $@; do echo -$x-; done'
['a', 'b']
-a-
-b-

$ dash -c 'IFS=x; set -- a "" b; argv $@; for x in $@; do echo -$x-; done'
['a', 'b']
-a-
-b-

```

andychu:
For test case 18, there is also disagreement

But mksh agrees with us, and not with bash

This looks like it has to do with `for i in =$*=`, whereas test cases 11 and 12 are about `for i in $@` and argv `$@`

akinomyoga:
Yeah, I also noticed those behavioral differences between shells. I didn't know this Bash's behavior for `IFS=x`. I was checking Bash's source code to find out what was happening. It seems Bash implements it in a similar way as the current master, but it treats an empty `IFS` as a special case. For the default IFS case, a behavioral difference between the current master and the new approach in this PR hasn't been observed because of the special behavior for whitespace IFS.

andychu:
Oh great to see that this fixes many bugs in `spec/word-split`, I didn't notice that at first!


http://op.oilshell.org/uuu/github-jobs/9104/ovm-tarball.wwz/_tmp/spec/osh-py/word-split.html

So yeah this is almost certainly a good change, I am just wondering about `spec/toysh-posix` #11 and #12 now ...

akinomyoga:
I’ll later reply.

andychu:
OK just to clarify, the bug appears to be a combination of 3 features

- `IFS=x`
- $@ (unquoted)
- `''` empty string as argument

I wrote some test cases that I believe show that 

---

It might be OK to land this change anyway, because it fixes a lot of bugs, and we are matching dash at least ...  but not bash/mksh

andychu:
Actually the test cases I wrote, which is `spec/word-split #47 #48` show that 

```
set -- 'a b' c ''   # empty string at the end - NO difference
```

```
set -- one '' two   # empty string in the middle - dash/ash and now OSH differ
```

I am OK to leave it as is for now

I don't understand why there is a difference, but I think our code is consistent at least

----

We can  move on to other SparseArray stuff, and then fix other IFS bugs, and if we do all that, maybe this will become clearer (but I don't think it's that important because shells already disagree)

andychu:
OK I convinced myself this is not that important ... We still have to rewrite IFS a bit anyway for `IFS=\` bugs, so maybe this will fall out of that

I am more excited to turn on SparseArray :-)


Thanks for fixing this !!   This is bug #627 from 2020

PR #2249, opened 2025-02-17
[osh] Implement SparseArray `${!sp[0]}` and `${!sp[@]}`

akinomyoga:
This implements the following two expansions for SparseArray:

- The indirect expansions through a SparseArray element: `${!sp[0]}`
- The keys of a SparseArray: `${!sp[@]}`


andychu:
Great thank you!

PR #2252, opened 2025-02-18
[spec/word-split] Add more failing test cases for `IFS=x; set "" "" ""; $*`

akinomyoga:
These are follow-up fixes for #2248. This PR also contains a fix to another broken test case: 0187f3bb62f2b0375c789f089a899c988d554673, where `=` of an assignment was missing.

----

Bash's behavior `IFS=x` seems strange to me (see the added test case, where the number of arguments decreases by one every time resetting them). However, even if we report it to the upstream Bash, I guess Chet wouldn't finally change the behavior. This area is the one marked as "*unspecified*" by POSIX, and the behavior among the shells is anyway not consistent. In addition, as far as I look at the codebase of Bash, the fix wouldn't be simple. It would require rewriting multiple places in the Bash codebase. In addition, I guess the impact on the real use cases wouldn't be so significant.

Except for `bash` and `mksh`, all the shells produce no arguments for `$*` with `IFS=x; set "" "" "" "" ""`. Yash in the oils-for-unix test (yash-2.49) seems to produce a distinct result (which preserves the empty elements with `$*`), but Yash has already fixed its behavior to be consistent with the majority (which produces no arguments) at least in yash-2.57.

There would be two options for the design of osh, 1) to replicate Bash's behavior exactly, or 2) to follow the majority of different shells. In this PR, I added the tests as failing tests requesting them to be the same as Bash's, but it doesn't mean I request osh to be consistent with Bash in the future.

### DIFF
--- a/spec/toysh-posix.test.sh
+++ b/spec/toysh-posix.test.sh
@@ -202,8 +202,10 @@ myfunc one "" two
 ## N-I zsh STDOUT:
 ## END
 
-#### Behavior of IFS=x and unquoted $@ - reduction of case above
-case $SH in zsh) exit ;; esac
+#### IFS=x and '' and unquoted $@ - reduction of case above
+
+setopt SH_WORD_SPLIT
+#set -x
 
 set -- one "" two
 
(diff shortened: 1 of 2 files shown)


andychu:
OK thanks, this was helpful!

I chose the Almquist shell behavior, which OSH now matches -- to give 6 failures instead of 9


- On the one hand, dash/busybox ash are actually the same codebase.  So it really only counts as one shell, if we are doing "voting".
- bash/mksh match SOMETIMES - but your test cases showed that bash and mksh don't always match

But I think the dash/ash behavior is reasonable, so I changed it to that

I think our priority is to rewrite the IFS splitting, so we shouldn't worry about this specific case until that is done.  The main issue is the IFS=\ bugs, because I used `\` in a bad way.

Aidan has a PR open for that, but some other stuff broke ... I have to look at that PR again to see if it's the right strategy, although I certainly appreciate any thoughts/help you can offer

(YSH gets rid of word splitting, because it is a silly anachronism.  So I have less motivation to work on OSH word splitting, but we do need it)

---

I also copied a different IFS bug from toysh-posix, so now we have 7 failures in spec/word-split



Thanks!

PR #2255, opened 2025-02-19
[bash_impl] SparseArray: omit index in the printed form in the dense case

akinomyoga:
For a smoother transition of the indexed-array internal representation from `BashArray` to `SparseArray`, I here suggest switching the printed form of `SparseArray` (used by `declare -p` and other places) based on whether it is indeed "sparse" or not.

- The current behavior of `SparseArray`:
  - It always uses the designated array initializer form: `([0]=1 [1]=2 [4]=5 ...)`
- This PR modifies the behavior as follows:
  - When it is dense (i.e., when all the indices up to `max_index` are filled), it would use the same printed form as the present `BashArray`: `(1 2 3 4 5 ...)` 
  - When it is sparse (i.e., when there is a "hole"), it would use the designated array initializer form: `([0]=1 [1]=2 [4]=5 ...)`

Even though we currently do not support the designated initializer syntax (`([index]=value)`) for indexed arrays, I'm going to work on *the initializer list* [[ref1](https://github.com/oils-for-unix/oils/issues/661#issuecomment-599334537), [ref2](https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/2024-11-22.20Meeting/near/485170417)] after the transition to `SparseArray` is complete. Then, everything should become fine finally.

### DIFF
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -464,12 +464,15 @@ def SparseArray_Equals(lhs, rhs):
 def SparseArray_ToStrForShellPrint(sparse_val):
     # type: (value.SparseArray) -> str
     body = []  # type: List[str]
+
+    is_sparse = not mops.Equal(mops.IntWiden(SparseArray_Count(sparse_val)),
+                               SparseArray_Length(sparse_val))
+
     for index in SparseArray_GetKeys(sparse_val):
         if len(body) > 0:
             body.append(" ")
-        body.extend([
-            "[",
-            mops.ToStr(index), "]=",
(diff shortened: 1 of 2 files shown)


andychu:
Great thank you!  It's a good idea to make the transition easier

I haven't thought about the full details of the transition, but we can chat about it on Zulip 

I guess the main thing is feel free to use the CI to find breakages!  If it passes tests, it should be fine ... but I trust the tests to be thorough now

---

( Yes I think initializer list is in scope too - thanks for thinking about this!)

PR #2258, opened 2025-02-20
[refactor] Rename `{Sparse => Bash}Array`

akinomyoga:


andychu:
Oh it looks like `vendor/` had something unrelated called SparseArray -- please revert that directory

Otherwise looks good!

akinomyoga:
> Oh it looks like `vendor/` had something unrelated called SparseArray -- please revert that directory

Ah, thanks for pointing out this! I did know there were conflicting names in that file (because I ran `git grep` many times in refactoring), but I completely forgot that. I was distracted by `testdata/mycpp/oils_for_unix.mycpp.cc` and forgot about other places that I shouldn't modify.

Now I reverted it and amended the second commit.

andychu:
Thanks, this is great

Glad to see the migration went smoothly !!

akinomyoga:
Thanks. Yeah, it was much smoother than I expected. I hope problems won't happen later.

andychu:
I tested it out on Zulip, seems to work great, and it fixed the performance problems!

I think you wrote great tests !!  I don't expect any problems, but it's OK if there are

PR #2259, opened 2025-02-22
[cleanup] Remove `_a2sp` and `_opsp` and clean up others

akinomyoga:


test/spec.sh
```diff
@@ -986,6 +986,11 @@ ble-idioms() {
   run-file ble-idioms "$@"
 }
 
+ble-sparse() {
+  sh-spec spec/ble-sparse.test.sh \
```
andychu, line 990: this should use `run-file ble-sparse "$@"` 

and then the top of the file should have

`## compare_shells: bash mksh`

That is the new style  -- and hopefully I can get rid of all these tiny functions at some point, once they are all consistent

andychu (commented):
Thanks!  Just one small issue, which we can address later

test/spec.sh
akinomyoga: OK. I thought I just copied the configuration for `ble-idioms`, but `ble-idioms` seems to have been updated just yesterday.

PR #2262, opened 2025-02-23
[test/ble.sh] Add "--lib" in loading "out/ble.osh"

akinomyoga:
I rebased the `osh` branch in the upstream `akinomyoga/ble.sh` on the latest version of `ble.sh`.  As a consequence, sourcing `out/ble.osh` in the CI has started to fail (see Run 9147's [`ble-test-osh-py.txt`](https://op.oilshell.org/uuu/github-jobs/9147/app-tests.wwz/_tmp/soil/logs/ble-test-osh-py.txt) and [`ble-test-osh-cpp.txt`](https://op.oilshell.org/uuu/github-jobs/9147/app-tests.wwz/_tmp/soil/logs/ble-test-osh-cpp.txt)).

The latest version of `ble.sh` checks whether the current session has an associated TTY when loaded as an interactive setting.  When `ble.sh` cannot detect the TTY, it cancels its loading.  To load `ble.sh` in a non-interactive session, one needs to source `ble.sh` using "the library mode" by specifying the option `--lib`.

### DIFF
--- a/test/ble.sh
+++ b/test/ble.sh
@@ -73,7 +73,7 @@ HISTFILE=$HOME/.osh_history
 #	lib/test-canvas.sh \
 
 for script in \
-  out/ble.osh \
+  "out/ble.osh --lib" \
 	lib/test-main.sh \
 	lib/test-util.sh \
 	lib/test-decode.sh


andychu:
OK thanks!

Hm actually we run the `app-tests` image with `docker -t` ... So it is possible to use a TTY in the tests?

But then  we redirect the  stdout of the tests to a file in the `soil` framework?   

Oh maybe we also redirect stdin, let me see ....

andychu:
Yes, here we run the `app-tests` task with `docker -t`

https://github.com/oils-for-unix/oils/blob/master/soil/host-shim.sh#L258

And here we special case a script `test/nohup.sh` to have a TTY

https://github.com/oils-for-unix/oils/blob/master/soil/worker.sh#L494


So we can do the same thing with ble.sh if it's helpful

(This code could also be cleaned up a  bit, the special cases are hidden)

akinomyoga:
Thanks!

> And here we special case a script `test/nohup.sh` to have a TTY
> 
> https://github.com/oils-for-unix/oils/blob/master/soil/worker.sh#L494
> 
> So we can do the same thing with ble.sh if it's helpful

I think that for now, a TTY is not necessary for `ble.sh` because we only run unit tests of `ble.sh`.

PR #2265, opened 2025-02-24
[builtin/assign_osh] Fix `readonly` failing to create arrays with -a/-A

akinomyoga:
There is a dead code:

https://github.com/oils-for-unix/oils/blob/24146f4b0a223354c5f2e42e59cad00358ff82a5/builtin/assign_osh.py#L322-L334

The value of `rval` calculated in the above lines is not used anywhere. The next line only uses the original value `pair.rval`:

https://github.com/oils-for-unix/oils/blob/24146f4b0a223354c5f2e42e59cad00358ff82a5/builtin/assign_osh.py#L339

This is a regression introduced in [this hunk in commit `fd47f77`](https://github.com/oils-for-unix/oils/commit/fd47f772e49b2217c603a834e0521227e6ec2759#diff-81b973a51b80c984cbe64b92eb65bc97b31b7ce89fcd54920b7ed952410e3e85L291).

In fact, we can confirm that the `readonly` builtin doesn't behave in the expected way:

```console
$ bin/osh -c 'declare -a a; declare -p a'
declare -a a=()
$ bin/osh -c 'declare -A a; declare -p a'
declare -A a=()
$ bin/osh -c 'readonly -a a; declare -p a'
$ bin/osh -c 'readonly -A a; declare -p a'
$
```

This PR adds a spec test case and fixes an issue.

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -192,8 +192,8 @@ def _PrintVariables(mem, cmd_val, attrs, print_flags, builtin=_OTHER):
         return 1
 
 
-def _ExportReadonly(mem, pair, flags):
-    # type: (Mem, AssignArg, int) -> None
+def _ExportReadonly(mem, rval, pair, flags):
+    # type: (Mem, value_t, AssignArg, int) -> None
     """For 'export' and 'readonly' to respect += and flags.
 
     Like 'setvar' (scope_e.LocalOnly), unless dynamic scope is on.  That is, it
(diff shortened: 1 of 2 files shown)


andychu:
Oh wow, thanks for finding this, that is a bit embarrassing

I tend to rely on the tests a lot ... and then the downside is if we're missing test coverage, then that can lead to bugs ...

andychu:
We probably need to run more bash scripts end-to-end, some of them probably have `readonly -A -a`

Hence why I am glad we are expanding the ble.sh tests -- historically it has caught bugs that other tests haven't!

---

At some point I want to make some kind of "Github Actions Template" to run arbitrary bash with OSH ... so many people can help us test in parallel ... (debugging is still hard though)

akinomyoga:
Yeah, making a perfect test coverage is hard work in general.

PR #2266, opened 2025-02-24
[refactor osh/cmd_eval] Assert RHS of `ShAssignment` is not `None`

akinomyoga:
Before implementing the initializer-list approach to the assignment of the form `name=(...)`, I'd like to confirm the assumptions in the relevant places in the codebase.

In the function `cmd_eval._DoShAssignment`, there is a code path to handle the case where the RHS of ShAssignment is None, but is there a real case for it? The shell assignments should always have the form `var=rhs`. Even if `rhs` is empty, it should still generate an empty string as a RHS.

As far as I quickly checked the code, the RHS seems ensured to be not None. In this PR, I attempt to assert it never happens and see if the assertion passes the CI.

### DIFF
--- a/osh/cmd_eval.py
+++ b/osh/cmd_eval.py
@@ -973,8 +973,13 @@ class CommandEvaluator(object):
         which_scopes = self.mem.ScopesForWriting()
 
         for pair in node.pairs:
+            # The shell assignments should always have RHS.  An AssignPair
+            # stored in command_e.ShAssignment is constructed in
+            # cmd_parse._MakeAssignPair, where rhs is explicitly constructed to
+            # be CompoundWord or rhs_word.Empty.
+            assert pair.rhs, pair.rhs
+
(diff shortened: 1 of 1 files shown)


andychu:
OK yes thanks

This is the schema


```
   AssignPair = (Token left, sh_lhs lhs, assign_op op, rhs_word rhs)
```

- And yes `rhs_word` is intended to be set, because it doesn't have `?`
- Although  we **don't** statically check it with MyPy -- that requires turning on "strict optional" checking, which is hard, because it generates many errors

 
and

- I think that `declare x=`  is an `AssignPair` with `rhs_word.Empty`
- and `declare x` (no `=`)` doesn't have an AssignPair, I think it's just a word argument

---


Thanks for cleaning up the code!

akinomyoga:
Yeah, the builtin assignments (such as `declare a` and `readonly a`) are processed in a different place.

PR #2267, opened 2025-02-24
[refactor builtin/assign_osh] Remove duplicate codes in `NewVar`

akinomyoga:
The first commit is the main change. There is a duplicate code in `NewVar`. For `Readonly` and `Export`, a code to perform an assignment to a single variable is factorized into a function `_ExportReadonly`, but `NewVar` also has the same code. This PR modifies `NewVar` to use `_ExportReadonly` (with a small adjustment to switch scope).

The second commit is an extra commit renaming `_ExportReadonly` to `_AssignVarForBuiltin`. The name `_ExportReadonly` seems to have been named only after the fact that it is used by `export` and `readonly`, which doesn't explain what this function does. If we need to continue this naming, we need to update it to be `_ExportReadonlyNewVar` this time, which seems clumsy.

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -193,7 +193,7 @@ def _PrintVariables(mem, cmd_val, attrs, print_flags, builtin=_OTHER):
         return 1
 
 
-def _ExportReadonlyNewVar(mem, rval, pair, which_scopes, flags):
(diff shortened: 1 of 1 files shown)


andychu:
Looks good, thanks

PR #2268, opened 2025-02-24
[core/state] Do not initialize a sparse array with a long `List[str]`

### LINKED ISSUE #2263: Running ble.sh takes too much memory - C++ using more memory than Python
> Thank you.
> 
> ### Memory footprint 1.2 GB
> By the way, I noticed that the memory footprint becomes extremely large when `eval_unsafe_arith` is turned on in the C++ version. This is the result for sourcing `ble.osh`.
> 
> Python version	C++ version	Bash 5.3-beta (for ref)
> Without eval_unsafe_arith	69.23 MB (41.64s)	157.9 MB (6.364s)	14.77 MB (0.178s)
> With eval_unsafe_arith	69.17 MB (41.45s)	**1.207 GB** (6.421s)	N/A
> (Detailed commands and results)
> The same behavior is observed with `oils-for-unix` compiled from the tarball on the release page.
> 
> Why does this happen?
> 
> **edit**: By the way, the C++ version is faster than the Python version, but it is still 36x slower compared to Bash. The memory size is 10x as large as Bash even without `eval_unsafe_arith`. Hmm, but maybe this is simply because Bash doesn't do the full parsing at the sourcing stage. Bash performs some parsing at the execution step, so the performance should be different for the actual processing.

> Thank you.
> 
> ### Memory footprint 1.2 GB
> By the way, I noticed that the memory footprint becomes extremely large when `eval_unsafe_arith` is turned on in the C++ version. This is the result for sourcing `ble.osh`.
> 
> Python version	C++ version	Bash 5.3-beta (for ref)
> Without eval_unsafe_arith	69.23 MB (41.64s)	157.9 MB (6.364s)	14.77 MB (0.178s)
> With eval_unsafe_arith	69.17 MB (41.45s)	**1.207 GB** (6.421s)	N/A
> (Detailed commands and results)
> **With eval_unsafe_arith**
> 
> $ cat oshrc.source
> #!/usr/bin/env osh
> HISTFILE=$HOME/.osh_history
> shopt -s eval_unsafe_arith
> source out/ble.osh --norc
> echo "RSS: $(ps -o rss -p $$ | tail -n 1)" >&2
> exit
> $ ~/.mwg/git/oilshell/oil/bin/osh --rcfile oshrc.source
>   shopt -s checkwinsize
>   ^~~~~
> out/ble.osh:2815: 'shopt' got invalid option 'checkwinsize'
> osh warning: The 'RETURN' hook isn't implemented
> RSS: 69168
> [ble: elapsed 41.452s, CPU 99.8% (self)usr2ms/sys2ms+(child)usr41.109/sys264ms]
> $ ~/.mwg/git/oilshell/oil/_bin/cxx-asan/osh --rcfile oshrc.source
>   shopt -s checkwinsize
>   ^~~~~
> out/ble.osh:2815: 'shopt' got invalid option 'checkwinsize'
> osh warning: The 'RETURN' hook isn't implemented
> RSS: 1207040
> [ble: elapsed 6.421s, CPU 99.5% (self)usr1ms/sys3ms+(child)usr5.824/sys570ms]
> **Without eval_unsafe_arith**
> 
> $ cat oshrc.source
> #!/usr/bin/env osh
> HISTFILE=$HOME/.osh_history
> #shopt -s eval_unsafe_arith
> source out/ble.osh --norc
> echo "RSS: $(ps -o rss -p $$ | tail -n 1)" >&2
> exit
> $ ~/.mwg/git/oilshell/oil/bin/osh --rcfile oshrc.source
>   shopt -s checkwinsize
>   ^~~~~
> out/ble.osh:2815: 'shopt' got invalid option 'checkwinsize'
> osh warning: The 'RETURN' hook isn't implemented
> RSS: 69236
> [ble: elapsed 41.640s, CPU 99.8% (self)usr3ms/sys1ms+(child)usr41.288/sys273ms]
> $ ~/.mwg/git/oilshell/oil/_bin/cxx-asan/osh --rcfile oshrc.source
>   shopt -s checkwinsize
>   ^~~~~
> out/ble.osh:2815: 'shopt' got invalid option 'checkwinsize'
> osh warning: The 'RETURN' hook isn't implemented
> RSS: 157896
> [ble: elapsed 6.364s, CPU 99.6% (self)usr1ms/sys3ms+(child)usr5.845/sys494ms]
> **Bash 5.3-beta**
> 
> $ cat bashrc.source
> #!/usr/bin/env bash
> 
> HISTFILE=$HOME/.mwg/git/oilshell/oil/tmp/bash_history
> source out/ble.sh --norc --bash-debug-version=ignore
> echo "RSS: $(ps -o rss -p $$ | tail -n 1)" >&2
> exit
> $ bash --rcfile bashrc.source
> RSS: 14796
> [ble: elapsed 178ms, CPU 100.5% (self)usr2ms/sys2ms+(child)usr149ms/sys28ms]
> The same behavior is observed with `oils-for-unix` compiled from the tarball on the release page.
> 
> Why does this happen?
> 
> **edit**: By the way, the C++ version is faster than the Python version, but it is still 36x slower compared to Bash. The memory size is 10x as large as Bash even without `eval_unsafe_arith`. Hmm, but maybe this is simply because Bash doesn't do the full parsing at the sourcing stage. Bash performs some parsing at the execution step, so the performance should be different for the actual processing.

akinomyoga:
Fixes #2263.

This fixes an issue of #2263. This also fixes another issue https://github.com/oils-for-unix/oils/issues/2263#issuecomment-2677382000. In the previous versions, `unset[-1]=1` has created `unset=(1)` in the Python version and has core-dumped in the C++ version.

```console
$ bash -c 'unset[-1]=1; declare -p unset'
bash: line 1: unset[-1]: bad array subscript
$ bin/osh -c 'unset[-1]=1; declare -p unset'
declare -a unset=(1)
$ osh-0.27.0 -c 'unset[-1]=1; declare -p unset'  # 0.27.0 C++ version
Segmentation fault         (core dumped) osh-0.27.0 -c 'unset[-1]=1; declare -p unset'
```

The last commit 069162fd6 adds a spec test for the above bug.

This PR contains another commit a271b53c61b69413d53e136511131f13ccc4a413 to refactor the creation of a new sparse array. We've been passed an empty `List[str]` to `BashArray_FromList(list)`, but it is unnecessary to generate any list for an empty new sparse array. This commit adds a new function `BashArray_New()`.

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -288,7 +288,7 @@ def _ReconcileTypes(rval, flag_a, flag_A, blame_word):
                 # mycpp limitation: NewDict() needs to be typed
                 tmp = NewDict()  # type: Dict[str, str]
                 return value.BashAssoc(tmp)
-                #return bash_impl.BashArray_FromList([])
+                #return bash_impl.BashArray_New()
 
         if rval.tag() != value_e.BashAssoc:
             e_usage("Got -A but RHS isn't an associative array",
@@ -321,7 +321,7 @@ class Readonly(vm._AssignBuiltin):
         for pair in cmd_val.pairs:
             if pair.rval is None:
                 if arg.a:
-                    rval = bash_impl.BashArray_FromList([])  # type: value_t
+                    rval = bash_impl.BashArray_New()  # type: value_t
                 elif arg.A:
                     # mycpp limitation: NewDict() needs to be typed
                     tmp = NewDict()  # type: Dict[str, str]
@@ -456,7 +456,7 @@ class NewVar(vm._AssignBuiltin):
                 if arg.a:
                     if old_val.tag() not in (value_e.InternalStringArray,
                                              value_e.BashArray):
-                        rval = bash_impl.BashArray_FromList([])
+                        rval = bash_impl.BashArray_New()
                 elif arg.A:
                     if old_val.tag() != value_e.BashAssoc:
                         # mycpp limitation: NewDict() needs to be typed
--- a/core/bash_impl.py
+++ b/core/bash_impl.py
@@ -347,9 +347,15 @@ def BashAssoc_ToStrForShellPrint(assoc_val):
 # BashArray come here.
 
 
+def BashArray_New():
+    # type: () -> value.BashArray
+    d = {}  # type: Dict[mops.BigInt, str]
+    max_index = mops.MINUS_ONE  # max index for empty array
+    return value.BashArray(d, max_index)
+
+
 def BashArray_FromList(strs):
     # type: (List[str]) -> value.BashArray
-
     d = {}  # type: Dict[mops.BigInt, str]
     max_index = mops.MINUS_ONE  # max index for empty array
     for s in strs:
--- a/core/state.py
+++ b/core/state.py
@@ -2004,7 +2004,8 @@ class Mem(object):
                 cell, var_frame, _ = self._ResolveNameOrRef(
                     lval.name, which_scopes)
                 if not cell:
-                    self._BindNewArrayWithEntry(var_frame, lval, rval, flags)
+                    self._BindNewArrayWithEntry(var_frame, lval, rval, flags,
+                                                left_loc)
                     return
 
                 if cell.readonly:
@@ -2015,7 +2016,7 @@ class Mem(object):
                 with tagswitch(UP_cell_val) as case2:
                     if case2(value_e.Undef):
                         self._BindNewArrayWithEntry(var_frame, lval, rval,
-                                                    flags)
+                                                    flags, left_loc)
                         return
 
                     elif case2(value_e.Str):
@@ -2076,13 +2077,16 @@ class Mem(object):
             else:
                 raise AssertionError(lval.tag())
 
-    def _BindNewArrayWithEntry(self, var_frame, lval, val, flags):
-        # type: (Dict[str, Cell], sh_lvalue.Indexed, value.Str, int) -> None
+    def _BindNewArrayWithEntry(self, var_frame, lval, val, flags, blame_loc):
+        # type: (Dict[str, Cell], sh_lvalue.Indexed, value.Str, int, loc_t) -> None
         """Fill 'var_frame' with a new indexed array entry."""
-        no_str = None  # type: Optional[str]
-        items = [no_str] * lval.index
-        items.append(val.s)
-        new_value = bash_impl.BashArray_FromList(items)
+
+        new_value = bash_impl.BashArray_New()
+        error_code = bash_impl.BashArray_SetElement(new_value,
+                                                    mops.IntWiden(lval.index),
+                                                    val.s)
+        if error_code == error_code_e.IndexOutOfRange:
+            e_die("Index %d is out of bounds for array of length 0", blame_loc)
 
         # arrays can't be exported; can't have BashAssoc flag
         readonly = bool(flags & SetReadOnly)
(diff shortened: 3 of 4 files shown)


akinomyoga:
| | Python version | C++ version | Bash 5.2 | Bash 5.3-beta |
|---|---|---|---|---|
| Without eval_unsafe_arith | 69.30 MB (9.17s) | 38.23 MB (0.313s) | 31.97 MB (0.364s) | 14.77 MB (0.178s) |
| With eval_unsafe_arith | 69.29 MB (9.28s) | 38.57 MB (0.311s) | N/A | N/A |

Now, both the speed and the memory size are comparable to Bash. The C++ version is faster than Bash 5.2 and twice as slow as Bash 5.3. The memory size is similar to Bash 5.2 and about 2.5x as Bash 5.3.

andychu:
Oh awesome!

Do we know why bash 5.3 got faster, and uses less memory?  Is it because of some bash malloc changes?

andychu:
Excellent, thank you!!

Good to see the internal APIs making it easy to handle the error cases

akinomyoga:
> Do we know why bash 5.3 got faster, and uses less memory? Is it because of some bash malloc changes?

There has been a ridiculous bug in Bash 4.3..5.2, where two copies of a function were generated for every function (and one was not used at all). Actually, I submitted a patch to Bash last June:

https://lists.gnu.org/archive/html/bug-bash/2024-06/msg00000.html

andychu:
Wow, that's funny!  Great find

(OSH does have benchmarks in the CI, so we would find anything that doubles memory usage that way)

----

Oh well now I am a bit disappointed we're slower than bash 5.3 :-/    There is a GC-optimized build at _bin/cxx-opt+souffle/osh, but I tried it, and it's not much faster (less GC rooting)

YSH can be a lot faster, but I guess Python/mycpp/OSH does a lot of pointer chasing compared to C code, even though there is less copying

(I wonder if "small string optimization" will help ble.sh in particular, since it uses many integer-like strings?   That would save strings of 6 bytes inside the pointer, rather than allocating a GC object)

andychu:
(Also it's weird that bash doesn't have a good commit history - https://github.com/bminor/bash/commit/ba4ab055 )

andychu:
> Yeah, but I think what's measured by source ble.sh is just the function generation speed. I believe the execution speed is a different story.

Yeah that's true, for `source` of a file with  many functions,  then the interpreter is doing one thing

1. parsing the function bodies
2. entering it in a dict

And we are ~1x-3x slower on parsing (the 3x is the result of bash 5.2 optimizations I think)

https://oils.pub/release/0.27.0/benchmarks.wwz/osh-parser/

----

So yeah I agree the runtime speed is more important, since we are already at a "reasonable" speed (faster than bash 5.2)

Small string optimization isn't turned on, but I prototyped it here - https://github.com/oils-for-unix/oils/blob/master/mycpp/small_str_test.cc#L35

Although in general I think bigger speed ups will come from the work that @melvinw has done

- GC rooting optimization - this has a big speedup on the "bubble sort" microbenchmark, although I tested `_bin/cxx-opt+souffle/osh` on the SparseArray benchmarks, and it was only a small difference
- some kind of value types optimization -- not sure if/when we will get here, but chasing many pointers is probably expensive
  - this interacts with a few other things, so it's not straightforward - I have a blog post draft that mentions some of them

But it's good that we've fixed this BashArray problem by moving to the sparse representation!

PR #2270, opened 2025-02-24
[refactor core/bash_impl] Add `BashAssoc_New`

akinomyoga:
This is another preparation for the initializer-list implementation.

This PR adds `bash_impl.BashAssoc_New` so that the implementor doesn't have to care about `NewDict` in creating an instance of `BashAssoc`. This is also make it consistent with the `BashArray` creation through `BashArray_New` introduced in #2268.


andychu:
Looks great!

PR #2271, opened 2025-02-24
[perf osh/brace] move early check from BraceDetectAll to BraceDetect

akinomyoga:
Currently, an early check for brace expansions is performed only in one specific call path of `BraceDetect`, but this optimization should be able to be applied to all the calls of `BraceDetect`. `BraceDetect` is currently called from two places (excluding the unit tests):

1. The first place is `BraceDetectAll` (`osh/braces.py`), where the number of grammatical parts is checked for faster processing.
2. The second place is `CommandEvaluator._EvalRedirect` (`osh/cmd_eval.py`) where we do not expect brace expansions.  In this second place, the number of grammatical parts is not checked.

https://github.com/oils-for-unix/oils/blob/2e7a7ab9289f0835bd02604f4cefd307363452bd/osh/braces.py#L318-L325
https://github.com/oils-for-unix/oils/blob/2e7a7ab9289f0835bd02604f4cefd307363452bd/osh/cmd_eval.py#L478-L483

In this patch, we perform the early check for the brace expansions based on the number of grammatical parts inside `BraceDetect` instead of the caller sides.


andychu:
Very nice find, thank you!

PR #2272, opened 2025-02-24
[osh/word_parse] Perform tilde expansion on `rhs` of `dict=([key]=rhs)`

akinomyoga:
The right-hand side of `[k]=rhs` in the list assignment `a=([k]=v)` is subject to the tilde expansion. However, the current `master` doesn't perform it:

```console
$ bash -c "declare -A a=([k]=~); echo \"\${a['k']}\""
/home/murase
$ bin/osh -c "declare -A a=([k]=~); echo \"\${a['k']}\""
~
```

### Note: Bash 5.2 has a bug

It should be noted that Bash 5.2 has a similar bug. It is fixed in Bash 5.3.

```console
$ bash-5.1 -c "declare -A a=([k]=~); echo \"\${a['k']}\""
/home/murase
$ bash-5.2 -c "declare -A a=([k]=~); echo \"\${a['k']}\""
~
$ bash-5.3 -c "declare -A a=([k]=~); echo \"\${a['k']}\""
/home/murase
```

### DIFF
--- a/osh/word_parse.py
+++ b/osh/word_parse.py
@@ -1666,6 +1666,7 @@ class WordParser(WordEmitter):
         # If the first one is a key/value pair, then the rest are assumed to be.
         pair = word_.DetectAssocPair(words[0])
         if pair:
+            word_.TildeDetectAssign(pair.value)  # pair.value is modified
             pairs.append(pair)
 
             n = len(words)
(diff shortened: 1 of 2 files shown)


andychu:
Ah OK, makes sense!  

Funny that bash had the bug too

akinomyoga:
Yeah, Bash 5.2 had refactored its parsing codes, which introduced many bugs. I don't remember them one by one, but I think there were several other bugs introduced in Bash 5.2.

PR #2273, opened 2025-02-25
[builtin/assign_osh] Do not remove existing arrays with `readonly -a/-A`

akinomyoga:
In the current `master`, `readonly -a a` and `readonly -A a` remove existing indexed and associative arrays, respectively. See the following example:

```console
$ bash -c 'a=(1 2 3); readonly -a a; declare -p a'
declare -ar a=([0]="1" [1]="2" [2]="3")
$ osh -c 'a=(1 2 3); readonly -a a; declare -a a'
declare -ra a=()
```

This PR fixes the above behavior by initializing an empty array only when an existing value has a different type from the one expected by `-a` or `-A`. This behavioral change is similar to the one for the `declare` builtin introduced in #731.

Actually, this behavior of removing the existing arrays has been hidden under another bug fixed in #2265. Now #2265 was merged, and this problem has been revealed.

----

This patch moves the adjustment of the value into the function `_ReconcileTypes`, which is shared by `readonly` and `declare`.

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -10,7 +10,7 @@ from _devbuild.gen.runtime_asdl import (
     AssignArg,
 )
 from _devbuild.gen.value_asdl import (value, value_e, value_t, LeftName)
-from _devbuild.gen.syntax_asdl import loc, loc_t, word_t
+from _devbuild.gen.syntax_asdl import loc, loc_t
 
 from core import bash_impl
 from core import error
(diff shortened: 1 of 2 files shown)


builtin/assign_osh.py
```diff
@@ -267,31 +267,50 @@ def Run(self, cmd_val):
...
-
-        if rval.tag() != value_e.BashAssoc:
-            e_usage("Got -A but RHS isn't an associative array",
-                    loc.Word(blame_word))
+
+    if rval is None:
+        # declare -a foo=(a b); declare -a foo; should not reset to empty array
+        if flag_a:
+            old_val = mem.GetValue(pair.var_name)
+            if old_val.tag() not in (value_e.InternalStringArray,
```
andychu, line 288: It feels like InternalStringArray  should not be possible here -- i.e. it should not be stored in state.Mem.var_stack

But I guess that is a separate change, since it was already there

akinomyoga: It's possible with temporarily generatedd variables like `PIPESTATUS`. The following command hits this path with `old_val` being a temporary InternalStringArray instance which is dynamically created by `mem.GetValue`.

```console
$ bin/osh -c 'declare -a PIPESTATUS'
```

PR #2144, opened 2024-11-29
[osh] change BashArray `a+=()` to modify existing instance

akinomyoga:
https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/2024-11-22.20Meeting/near/485165500

PR #2274, opened 2025-02-26
[osh/sh_expr_eval] Check `readonly` before modifying the object

akinomyoga:
In the current `osh`, readonly arrays can be modified using `a+=(...)`:

```console
$ bash -c 'readonly -a a=(1 2 3); declare -p a; eval "a+=(4)" 2>/dev/null; declare -p a'
declare -ar a=([0]="1" [1]="2" [2]="3")
declare -ar a=([0]="1" [1]="2" [2]="3")
$ osh-0.27.0 -c 'readonly -a a=(1 2 3); declare -p a; eval "a+=(4)" 2>/dev/null; declare -p a'
declare -ra a=(1 2 3)
declare -ra a=(1 2 3 4)
```

----

- The first commit 808a628c8 makes `sh_expr_eval.OldValue` receive a new argument `blame_loc` and resolves the following TODO:

https://github.com/oils-for-unix/oils/blob/a3b3a609bec59b159ab4f90c432613fdb7d1bc27/osh/sh_expr_eval.py#L132

- The second commit 6bd51d50c fixes the issue that readonly arrays can be modified by `a+=(...)`.

### DIFF
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -127,7 +127,7 @@ def OldValue(lval, mem, exec_opts, blame_loc):
         else:
             raise AssertionError()
 
-    val = mem.GetValue(var_name)
+    val = mem.GetValueForRewrite(var_name, blame_loc)
     if exec_opts and exec_opts.nounset() and val.tag() == value_e.Undef:
         e_die('Undefined variable %r' % var_name, blame_loc)
 
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -92,8 +92,8 @@ _ = log
 #
 
 
-def OldValue(lval, mem, exec_opts):
-    # type: (sh_lvalue_t, state.Mem, Optional[optview.Exec]) -> value_t
+def OldValue(lval, mem, exec_opts, blame_loc):
+    # type: (sh_lvalue_t, state.Mem, Optional[optview.Exec], loc_t) -> value_t
     """Look up for augmented assignment.
 
     For s+=val and (( i += 1 ))
@@ -129,7 +129,7 @@ def OldValue(lval, mem, exec_opts):
 
     val = mem.GetValue(var_name)
     if exec_opts and exec_opts.nounset() and val.tag() == value_e.Undef:
-        e_die('Undefined variable %r' % var_name)  # TODO: location info
+        e_die('Undefined variable %r' % var_name, blame_loc)
 
     UP_val = val
     with tagswitch(lval) as case:
@@ -512,7 +512,8 @@ class ArithEvaluator(object):
         """ For x = y  and   x += y  and  ++x """
 
         lval = self.EvalArithLhs(node)
-        val = OldValue(lval, self.mem, self.exec_opts)
+        val = OldValue(lval, self.mem, self.exec_opts,
+                       location.TokenForArith(node))
 
         # BASH_LINENO, arr (array name without strict_array), etc.
         if (val.tag() in (value_e.InternalStringArray, value_e.BashAssoc,
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -127,7 +127,14 @@ def OldValue(lval, mem, exec_opts, blame_loc):
         else:
             raise AssertionError()
 
-    val = mem.GetValueForRewrite(var_name, blame_loc)
+    cell = mem.GetCellDeref(var_name)
+    if cell is not None:
+        if cell.readonly:
+            e_die("Can't assign to readonly variable %r" % var_name, blame_loc)
+        val = cell.val
+    else:
+        val = value.Undef
+
     if exec_opts and exec_opts.nounset() and val.tag() == value_e.Undef:
         e_die('Undefined variable %r' % var_name, blame_loc)
 
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -204,7 +204,11 @@ def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags):
     """
     lval = LeftName(pair.var_name, pair.blame_word)
     if pair.plus_eq:
-        old_val = sh_expr_eval.OldValue(lval, mem, None)  # ignore set -u
+        old_val = sh_expr_eval.OldValue(
+            lval,
+            mem,
+            None,  # ignore set -u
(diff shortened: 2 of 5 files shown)


akinomyoga:
This is a regression introduced by #2144, which was performed after [the Zulip discussion](https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/2024-11-22.20Meeting/near/485165500):

> @akinomyoga 
> ### Reason for making a new instance of BashArray on `a+=()`?
> I'm going to submit PRs again, but before that I have several questions about the current implementation. The first one is related to this line:
>
> https://github.com/oils-for-unix/oils/blob/ac916c7ec19b1c5ddf167121c75203523ea400ee/osh/cmd_eval.py#L224
>
> Is there a specific reason that it was written in the current way? Specifically, do I need to care about other places when I rewrite this part to add elements in the existing `a` instead of creating a new instance of `a`. For example, is there any other place, with which the reference to the BashArray instance is shared, and where the referenced instance of BashArray is assumed to be unchaneged?
>
> Or can I simply rewrite this part so that it rewrites the content of the existing instance (without changing other places)?
>
> I'm asking this because this affects the required interface of `bash_impl.SparseArray_*` (though I already prepared a commit with the assumption that we can simply modify the existing instance).
>
> ----
>
> @andychu 
> ### Mutation in place
> I guess this is fine? In OSH there is no way to tell the difference?
>
> In YSH there will be, but that is not a breaking change. You are not really supposed to mix them, but we can do whatever we want with
>
> ```
> declare -a a=()
>
> # there is no way to take a reference in bash/OSH, but there is in YSH
> var b = a  # this takes a REFERENCE, not a copy
> ```

osh/sh_expr_eval.py
```diff
@@ -974,6 +974,19 @@ def _VarNameOrWord(self, anode):
         no_str = None  # type: Optional[str]
         return (no_str, loc.Missing)
 
+    def _GetArithLoc(self, anode):
```
andychu, line 977: Hm actually I noticed we have `frontend/location.py TokenForArith(arith_expr_t)`

And that could return a `loc_t` rather than `Optional[Token]`, i.e. it should return `loc.Missing` instead of None

Because I think all the consumers can accept a `loc_t`, not just a Token

We can do that separately, but I think that function is more complete, because it works with Unary,Binary,TernaryOp

andychu (commented):
OK thanks, I see why you did it this way

We have OldValue(), but it gives you the VALUE, not the CELL. And the CELL is the thing that has the readonly flag

---

I am wondering if there is a nicer structure for state.Mem, but I don't have a strong conclusion now

https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/Brainstorming.20-.20structure.20of.20state.2EMem.20.2F.20builtin_assign


I guess this is a pretty small change ...

andychu:
Oh is it possible to use `mem.GetCell()`   in `OldValue()`?

And then check the `cell.readonly` flag?  That might be a bit cleaner?

akinomyoga:
> Oh is it possible to use `mem.GetCell()` in `OldValue()`?
> 
> And then check the `cell.readonly` flag? That might be a bit cleaner?

That would be a bit cleaner, but `mem.GetCell()` doesn't resolve the values of special variables which is processed in `GetValue`:

https://github.com/oils-for-unix/oils/blob/89daeb9171776c33aa15dfafd336b75c81593081/core/state.py#L2120-L2139

However, the special variables of osh are already broken, so maybe we do not have to care about them?

```console
$ osh -c 'p() { echo "[${PIPESTATUS[*]}][$(declare -p PIPESTATUS)]"; }; (exit 1) | (exit 2); p; PIPESTATUS=(999); p'
[1 2][]
[1 2][declare -a PIPESTATUS=(999)]
```

### Call both GetCell and GetValue?

Or do you mean we call both `mem.GetValue()` and `mem.GetCell()` to get the value by `GetValue` and to get the readonly flag by `GetCell`? For normal variables, it would cause two calls of `_ResolveNameOnly` which seems redundant.

In addition, when a name reference contains side effects (such as `declare -n ref='a[i++]'`), I'm afraid that it would cause an extra evaluation. However, the name reference to array elements does not seem to be supported by `osh` currently, and if it would be supported, we anyway need to combine the `OldValue` and `SetNamed` phases to share a single call of `_ResolveNameOrRef`, so refactoring is needed in any way. This is probably not an issue for the current version.

### GetCell doesn't resolve nameref

I also notice a difference between `GetCell` and `GetValue`: GetCell directly uses `_ResolveNameOnly`, while `GetValue` uses `_ResolveNameOrRef`. Doesn't this become a problem? If `GetCell` is used to get the readonly attribute, I guess we can still modify the readonly array through a name reference.

With this change,

```diff
diff --git a/osh/sh_expr_eval.py b/osh/sh_expr_eval.py
index 47046bc0c..cc5e6f61a 100644
--- a/osh/sh_expr_eval.py
+++ b/osh/sh_expr_eval.py
@@ -127,7 +127,11 @@ def OldValue(lval, mem, exec_opts, blame_loc):
         else:
             raise AssertionError()

-    val = mem.GetValueForRewrite(var_name, blame_loc)
+    cell = mem.GetCell(var_name)
+    if cell is not None and cell.readonly:
+        e_die("Can't assign to readonly variable %r" % var_name, blame_loc)
+
+    val = mem.GetValue(var_name)
     if exec_opts and exec_opts.nounset() and val.tag() == value_e.Undef:
         e_die('Undefined variable %r' % var_name, blame_loc)
```

We can still modify the array

```bash
$ bin/osh -c 'readonly -a a=(1 2 3); eval "declare -n r=a; r+=(4)" 2>/dev/null; declare -p a'
declare -ra a=(1 2 3 4)
```

We anyway need a version of `GetCell` that additionally resolves the name references.

andychu:
I think the reason GetCell() calls ResolveNameOnly() is for `declare -p`, so it doesn't dereference the variable

```
$ bash -c 'declare -n foo=a; declare -p foo'
declare -n foo="a"

$ bin/osh -c 'declare -n foo=a; declare -p foo'
declare -n foo=a
```

akinomyoga:
I added `mem.GetCellDeref()`. Now we use the result of `mem.GetCellDeref()` also for getting the old value (which means that we ignore the special variables in obtaining the original value for `+=`).

andychu:
Great, thank you!

PR #2275, opened 2025-02-26
[osh/word_] Disallow `BashAssocLiteral` in temporary environment

akinomyoga:
I think this should anyway be fixed when InitializerList is introduced, but I'd like to confirm the intended behavior. In the current osh, arrays of the form `(1 2 3)` (`ShArrayLiteral`) cannot be specified in tmpenv, but the form `([k]=v)` (`BashAssocLiteral`) can be specified:

```console
$ osh-0.27.0 -c 'A=(a a) declare -p A'
  A=(a a) declare -p A
  ^~
[ -c flag ]:1: Environment bindings can't contain array literals
$ osh-0.27.0 -c 'A=([a]=a) declare -p A'
declare -A A=(['a']=a)
```

I think `BashAssocLiteral` should also be rejected for consistency.

### DIFF
--- a/spec/array.test.sh
+++ b/spec/array.test.sh
@@ -281,6 +281,14 @@ A=a B=(b b) printenv.py A B
 ## OK bash status: 0
 ## OK mksh status: 1
 
+#### Associative arrays can't be used as env bindings either
+A=a B=([k]=v) printenv.py A B
+## status: 2
+## stdout-json: ""
+## OK bash stdout-json: "a\n([k]=v)\n"
+## OK bash status: 0
+## OK mksh status: 1
+
 #### Set element
 a=(1 '2 3')
 a[0]=9
--- a/osh/word_.py
+++ b/osh/word_.py
@@ -398,7 +398,8 @@ def HasArrayPart(w):
     # type: (CompoundWord) -> bool
     """Used in cmd_parse."""
     for part in w.parts:
-        if part.tag() == word_part_e.ShArrayLiteral:
+        if part.tag() in (word_part_e.ShArrayLiteral,
+                          word_part_e.BashAssocLiteral):
             return True
     return False
 


andychu:
Great thank you!

andychu:
bash does allow this, but I haven't encountered it in any programs, so I think for now it's OK to disallow:

```
$ bash -c 'ZZ=(z zz) env | grep ZZ'
ZZ=(z zz)
$ bash -c 'ZZ=(z=[a] zz) env | grep ZZ'
ZZ=(z=[a] zz)
```

Side note: I wonder if we even need the distinction between ShArrayLiteral and BashAssocLiteral now?  Does bash treat them as the same kind of thing?

andychu:
Oh bash does something very confusing I've never seen

```
$ bash -c 'declare -A ZZ=(x y z  [a]=b); declare -p ZZ'
declare -A ZZ=([z]="[a]=b" [x]="y" )

```

that doesn't seem right ...

spec/array.test.sh
```diff
@@ -281,6 +281,14 @@ A=a B=(b b) printenv.py A B
 ## OK bash status: 0
 ## OK mksh status: 1
 
+#### Associative arrays can't be used as env bindings either
+A=a B=([k]=v) printenv.py A B
+## status: 2
+## stdout-json: ""
+## OK bash stdout-json: "a\n([k]=v)\n"
```
andychu, line 288: Oh I forgot to say that I no longer use `stdout-json: ""` as much, since it's not as readable

I haven't converted all the old files, but I sometimes convert them when I change them

I think this style is more readable:

```
## STDOUT:
a
([k])=v
## END
```

akinomyoga: OK. I used this style because the previous test case used it on line 280. I'll fix it later.

akinomyoga: #2276

PR #2279, opened 2025-02-27
[osh] Implement InitializerList for `a=(v [k]=v [k]+=v)`

akinomyoga:
See commit messages.

### Note: Now `([k1]=v1 [k2]=v2)` is not a literal for BashAssoc

The construct `(...)` in `a=(...)` and `a+=(...)` always produces the new type InitializerList regardless of the forms of its items. The LHS object receives the initializer list and reads items one by one to modify its content. When the variable `a` is Undef, `a` is initialized to be an indexed array and processes the initializer list, so `a=([k1]=v1 [k2]=v2)` produces a sparse array (but not an associative array).

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -14,7 +14,7 @@ from _devbuild.gen.syntax_asdl import loc, loc_t
 
 from core import bash_impl
 from core import error
-from core.error import e_usage
+from core.error import e_usage, e_die
 from core import state
 from core import vm
 from display import ui
@@ -193,8 +193,9 @@ def _PrintVariables(mem, cmd_val, attrs, print_flags, builtin=_OTHER):
         return 1
 
 
-def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags):
-    # type: (Mem, value_t, AssignArg, scope_t, int) -> None
+def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags, arith_ev,
+                         flag_a, flag_A):
+    # type: (Mem, value_t, AssignArg, scope_t, int, sh_expr_eval.ArithEvaluator, bool, bool) -> None
     """For 'export', 'readonly', and NewVar to respect += and flags.
 
     Like 'setvar' (scope_e.LocalOnly), unless dynamic scope is on.  That is, it
@@ -203,28 +204,76 @@ def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags):
     Used for assignment builtins, (( a = b )), {fd}>out, ${x=}, etc.
     """
     lval = LeftName(pair.var_name, pair.blame_word)
-    if pair.plus_eq:
+
+    initializer = None  # type: value.InitializerList
+    if rval is None:
+        # When 'export e+=', then rval is value.Str('')
+        # When 'export foo', the pair.plus_eq flag is false.
+        # Thus, when rval is None, plus_eq cannot be True.
+        assert not pair.plus_eq, pair.plus_eq
+        # NOTE: when rval is None, only flags are changed
+        val = None  # type: value_t
+    elif rval.tag() == value_e.InitializerList:
         old_val = sh_expr_eval.OldValue(
             lval,
             mem,
             None,  # ignore set -u
             pair.blame_word)
-        # When 'export e+=', then rval is value.Str('')
-        # When 'export foo', the pair.plus_eq flag is false.
-        assert rval is not None
+        initializer = cast(value.InitializerList, rval)
+
+        val = old_val
+        if flag_a:
+            if old_val.tag() == value_e.BashAssoc:
+                # Note: OSH allows overwriting an existing BashArray with an
+                #   empty BashArray.  Bash does not allow this.
+                val = bash_impl.BashArray_New()
+            elif old_val.tag() in (value_e.Undef, value_e.Str,
+                                   value_e.BashArray):
+                # We do not need adjustemnts for -a.  These types are
+                # consistently handled within ListInitialize
+                pass
+            else:
+                e_die(
+                    "Can't convert type %s into BashArray" %
+                    ui.ValType(old_val), pair.blame_word)
+        elif flag_A:
+            if old_val.tag() in (value_e.Undef, value_e.Str,
+                                 value_e.BashArray):
+                # Note: We explicitly initialize BashAssoc for Undef.
+                # Note: OSH allows overwriting an existing BashArray with an
+                #   empty BashAssoc.  Bash does not allow this.
+                val = bash_impl.BashAssoc_New()
+            elif old_val.tag() == value_e.BashAssoc:
+                # We do not need adjustments for -A.
+                pass
+            else:
+                e_die(
+                    "Can't convert type %s into BashAssoc" %
+                    ui.ValType(old_val), pair.blame_word)
+
+        val = cmd_eval.ListInitializeTarget(val, pair.plus_eq, pair.blame_word)
+    elif pair.plus_eq:
+        old_val = sh_expr_eval.OldValue(
+            lval,
+            mem,
+            None,  # ignore set -u
+            pair.blame_word)
         val = cmd_eval.PlusEquals(old_val, rval)
     else:
-        # NOTE: when rval is None, only flags are changed
         val = rval
 
     mem.SetNamed(lval, val, which_scopes, flags=flags)
+    if initializer is not None:
+        cmd_eval.ListInitialize(val, initializer, pair.plus_eq,
+                                pair.blame_word, arith_ev)
 
 
 class Export(vm._AssignBuiltin):
 
-    def __init__(self, mem, errfmt):
-        # type: (Mem, ui.ErrorFormatter) -> None
+    def __init__(self, mem, arith_ev, errfmt):
+        # type: (Mem, sh_expr_eval.ArithEvaluator, ui.ErrorFormatter) -> None
         self.mem = mem
+        self.arith_ev = arith_ev
         self.errfmt = errfmt
 
     def Run(self, cmd_val):
@@ -266,7 +315,8 @@ class Export(vm._AssignBuiltin):
             which_scopes = self.mem.ScopesForWriting()
             for pair in cmd_val.pairs:
                 _AssignVarForBuiltin(self.mem, pair.rval, pair, which_scopes,
-                                     state.SetExport)
+                                     state.SetExport, self.arith_ev, False,
+                                     False)
 
         return 0
 
@@ -298,22 +348,12 @@ def _ReconcileTypes(rval, flag_a, flag_A, pair, mem):
                 rval = bash_impl.BashAssoc_New()
     else:
         if flag_a:
-            if rval.tag() not in (value_e.InternalStringArray,
-                                  value_e.BashArray):
-                e_usage("Got -a but RHS isn't an array",
+            if rval.tag() != value_e.InitializerList:
+                e_usage("Got -a but RHS isn't an initializer list",
                         loc.Word(pair.blame_word))
-
         elif flag_A:
-            # Special case: declare -A A=() is OK.  The () is changed to mean
-            # an empty associative array.
-            if rval.tag() == value_e.BashArray:
-                sparse_val = cast(value.BashArray, rval)
-                if bash_impl.BashArray_IsEmpty(sparse_val):
-                    return bash_impl.BashAssoc_New()
-                    #return bash_impl.BashArray_New()
-
-            if rval.tag() != value_e.BashAssoc:
-                e_usage("Got -A but RHS isn't an associative array",
+            if rval.tag() != value_e.InitializerList:
+                e_usage("Got -A but RHS isn't an initializer list",
                         loc.Word(pair.blame_word))
 
     return rval
@@ -321,9 +361,10 @@ def _ReconcileTypes(rval, flag_a, flag_A, pair, mem):
 
 class Readonly(vm._AssignBuiltin):
 
-    def __init__(self, mem, errfmt):
-        # type: (Mem, ui.ErrorFormatter) -> None
+    def __init__(self, mem, arith_ev, errfmt):
+        # type: (Mem, sh_expr_eval.ArithEvaluator, ui.ErrorFormatter) -> None
         self.mem = mem
+        self.arith_ev = arith_ev
         self.errfmt = errfmt
 
     def Run(self, cmd_val):
@@ -348,7 +389,8 @@ class Readonly(vm._AssignBuiltin):
             # - when rval is None, only flags are changed
             # - dynamic scope because flags on locals can be changed, etc.
             _AssignVarForBuiltin(self.mem, rval, pair, which_scopes,
-                                 state.SetReadOnly)
+                                 state.SetReadOnly, self.arith_ev, arg.a,
+                                 arg.A)
 
         return 0
 
@@ -356,11 +398,12 @@ class Readonly(vm._AssignBuiltin):
 class NewVar(vm._AssignBuiltin):
     """declare/typeset/local."""
 
-    def __init__(self, mem, procs, exec_opts, errfmt):
-        # type: (Mem, state.Procs, optview.Exec, ui.ErrorFormatter) -> None
+    def __init__(self, mem, procs, exec_opts, arith_ev, errfmt):
+        # type: (Mem, state.Procs, optview.Exec, sh_expr_eval.ArithEvaluator, ui.ErrorFormatter) -> None
         self.mem = mem
         self.procs = procs
         self.exec_opts = exec_opts
+        self.arith_ev = arith_ev
         self.errfmt = errfmt
 
     def _PrintFuncs(self, names):
@@ -463,7 +506,8 @@ class NewVar(vm._AssignBuiltin):
         for pair in cmd_val.pairs:
             rval = _ReconcileTypes(pair.rval, arg.a, arg.A, pair, self.mem)
 
-            _AssignVarForBuiltin(self.mem, rval, pair, which_scopes, flags)
+            _AssignVarForBuiltin(self.mem, rval, pair, which_scopes, flags,
+                                 self.arith_ev, arg.a, arg.A)
 
         return status
 
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -221,6 +221,14 @@ def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags, arith_ev,
             pair.blame_word)
         initializer = cast(value.InitializerList, rval)
 
+        # OSH compatibility: "declare -A assoc=(1 2)" is a usage
+        # error.  This code will soon be removed when the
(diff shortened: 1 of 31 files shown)


doc/ref/chap-osh-assign.md
```diff
@@ -26,11 +26,70 @@ This chapter describes OSH assignment, which looks like `x=value`.
...
+  a normal word.  In this case, this assigns `<value>` to the *next* element,
+  where the next element is determined by the LHS.  Unlike the previous two
+  forms, the `<value>` is subject to word splitting and pathname expansions as
+  if it is a normal argument to a command.
+
+The above three forms can be mixed within one initializer list, though there
+may be additional limitations depending on the type of the LHS of the
+assignment.
+
+When the assignment is performed with `=`, the content of the LHS value is
```
andychu, line 63: I would move this paragraph above after "a set of instructions to modify the existing value of the LHS"

doc/ref/chap-osh-assign.md
```diff
@@ -26,11 +26,70 @@ This chapter describes OSH assignment, which looks like `x=value`.
...
+- When the LHS variable is unset, the assignment creates an empty indexed array
+  and apply the initializer list to the created array (see
+  [sh-array](#sh-array)).
+- When the LHS is a scalar string, if the assignment operator is `=`, the
+  assignment overwrites the string with an empty array and applies the
+  initializer list to the array (see [sh-array](#sh-array)).  If the assignment
+  operator is `+=`, it causes an error.
+- When the LHS is an indexed or associative arrays, the initializer list is
+  applied to the original array following [sh-array](#sh-array) (for an indexed
+  array) or [sh-assoc](#sh-assoc) (for an assoative array).
```
andychu, line 79: typo associative

doc/ref/chap-type-method.md
```diff
@@ -29,7 +29,7 @@ These two types are for OSH code only.
 A bash array holds a sequence of strings.  Some entries may be unset, i.e.
 *not* an empty string.
 
-See [sh-array][] for details.  In YSH, prefer to use [List](#List) instances.
+See [sh-array][] for the initialization.  In YSH, prefer to use [List](#List) instances.
```
andychu, line 32: How about:

> See `sh-init-list` for initialization/mutation, and `sh-array` for other operations on BashArray.

or maybe we should called it sh-indexed-array and sh-assoc-array,  but I guess that is too long and doesn't fit

doc/ref/chap-type-method.md
```diff
@@ -38,7 +38,7 @@ See [sh-array][] for details.  In YSH, prefer to use [List](#List) instances.
 
 A bash associative array is a mapping from strings to strings.
 
-See [sh-assoc][] for details.  In YSH, prefer to use [Dict](#Dict) instances.
+See [sh-assoc][] for the initialization.  In YSH, prefer to use [Dict](#Dict) instances.
```
andychu, line 41: How about:

> See `sh-init-list` for initialization/mutation, and `sh-assoc` for other operations on BashArray.

doc/ref/toc-osh.md
```diff
@@ -121,7 +121,8 @@ X [Unsupported]   enable
 </h2>
 
 ```chapter-links-osh-assign
-  [Literals]      sh-array      array=(a b c)   array[1]=B   "${a[@]}"
+  [Literals]      sh-init-list  a=([0]=1 [0]+=2 3)
+                  sh-array      array=(a b c)   array[1]=B   "${a[@]}"
```
andychu, line 125: Maybe sh-init-list should also have the "value" like 

`a=(v [0]=1 ...)`

and maybe add a second example `a+=(v2 [c]=d [e]+=f)` or something

andychu (commented):
Wow thank you!  I read over the docs ONLY

That helps and clarifies what it does ... I don't think I fully understood that

I will have more comments later once I look at the code

spec/ble-sparse.test.sh
```diff
@@ -1200,5 +1200,49 @@ a[-1]=1
...
+argv.py "${a[@]}"
+## STDOUT:
+status=0
+['3']
+## END
+## N-I mksh status: 99
+## N-I mksh stdout-json: ""
+
+
+#### bash mangles indexed array #1 (keys undergoes arithmetic evaluation)
```
andychu, line 1220: Oh I see, why put this test in ble-sparse?   I think it would be better if the 2 tests are next to each other -- indexed array vs. associative array

builtin/assign_osh.py
```diff
@@ -203,28 +204,85 @@ def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags):
...
+            lval,
+            mem,
+            None,  # ignore set -u
+            pair.blame_word)
+        initializer = cast(value.InitializerList, rval)
+
+        # OSH compatibility: "declare -A assoc=(1 2)" is a usage
+        # error.  This code will soon be removed when the
+        # initialization of the form "assoc=(key value)" is supported.
+        if flag_A and len(initializer.assigns
```
andychu, line 227: I would add a () around the entire `if` condition and then reformat with yapf -- I do that because that way yapf can give a more natural wrapping

frontend/syntax.asdl
```diff
@@ -195,11 +195,15 @@ module syntax
      Token right
   )
 
-  AssocPair = (CompoundWord key, CompoundWord value)
+  AssocPair = (CompoundWord key, CompoundWord value, bool has_plus)
```
andychu, line 198: I noticed in `frontend/syntax.asdl` we have `assign_op = Equal | PlusEqual`, and `AssignPair` uses it

In core/runtime.asdl, `AssignArg` has `bool has_plus`

(One of them is static and one is dynamic)

It might be better to be consistent

osh/cmd_eval.py
```diff
@@ -190,6 +190,67 @@ def _HasManyStatuses(node):
...
+                if has_plus:
+                    old_val = bash_impl.BashArray_Copy(old_val)
+                else:
+                    old_val = bash_impl.BashArray_New()
+            return old_val
+        elif case(value_e.BashAssoc):
+            # OSH compatibility: assoc=(1 2) will create a new array.  This
+            # code will soon be removed when the initialization of the form
+            # "assoc=(key value)" is supported.
+            if not has_plus and len(
```
andychu, line 220: ditto, this could be wrapped better with `()` around the entire condition

osh/cmd_eval.py
```diff
@@ -979,30 +1024,45 @@ def _DoShAssignment(self, node, cmd_st):
             # be CompoundWord or rhs_word.Empty.
             assert pair.rhs, pair.rhs
 
-            if pair.op == assign_op_e.PlusEqual:
-                rhs = self.word_ev.EvalRhsWord(pair.rhs)
+            # RHS can be a string or initialier list.
```
andychu, line 1027: typo initializer

osh/word_.py
```diff
@@ -531,7 +530,11 @@ def DetectAssocPair(w):
             value = CompoundWord(parts[i + 1:])  # $a$b from
 
             # Type-annotated intermediate value for mycpp translation
-            return AssocPair(key, value)
+
+            # XXX---Isn't there a better way to detect "]+="?
+            has_plus = cast(Token, parts[i]).length == 3
+
+            return AssocPair(key, value, has_plus)
```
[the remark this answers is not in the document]
andychu: You might be able to use `lexer.IsPlusEquals(tok)`, depending on the token

Or write a similar function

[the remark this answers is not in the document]
andychu: Yeah in `osh/cmd_parse.py _MakeAssignPair`, it has this logic

```
     if left_token.id == Id.Lit_VarLike:  # s=1
         if lexer.IsPlusEquals(left_token):
             var_name = lexer.TokenSliceRight(left_token, -2)
             op = assign_op_e.PlusEqual
         else:
             var_name = lexer.TokenSliceRight(left_token, -1)
             op = assign_op_e.Equal

```

osh/word_eval.py
```diff
@@ -2162,23 +2166,26 @@ def EvalRhsWord(self, UP_w):
...
+                part0 = cast(word_part.InitializerLiteral, UP_part0)
+
+                assigns = []  # type: List[InitializerValue]
                 for pair in part0.pairs:
-                    k = self.EvalWordToString(pair.key)
-                    v = self.EvalWordToString(pair.value)
-                    d[k.s] = v.s
-                return value.BashAssoc(d)
+                    UP_pair = pair
+                    if pair.tag() == InitializerWord_e.ArrayWord:
```
andychu, line 2175: It might be nicer to use the `with  tagswitch` idiom here , even though it causes another level of indentation

The old code could have used it too, although it did an early return

andychu (commented):
This generally looks good --  I mostly looked at the  tests and the ASDL schema

I think we should rename ShArrayLiteral to YshArrayLiteral -- that is a good idea, since it's only used for YSH now

core/bash_impl.py
```diff
@@ -30,6 +32,20 @@ def BigInt_LessEq(a, b):
     return not mops.Greater(a, b)
 
 
+class ArrayIndexEvaluator(object):
+    """Abstract base class for things runnable in another process."""
```
andychu, line 36: This comment should be changed, it looks like copy and paste

core/value.asdl
```diff
@@ -27,6 +27,8 @@ module value
 
   IntBox = (int i)
 
+  InitializerValue = (str? key, str rval, bool plus_eq)
```
andychu, line 30: I think it might be cleaner to use frontend/syntax.asdl assign_op everywhere, although we can also refactor that separately probably

doc/ref/chap-osh-assign.md
```diff
@@ -26,11 +26,70 @@ This chapter describes OSH assignment, which looks like `x=value`.
...
+- Otherwise, it is an error.
 
 ### sh-array
 
-Array literals in shell accept any sequence of words, just like a command does:
+When an initializer list is assigned to an indexed array, the values will be
```
andychu, line 84: I think this text can use some more examples, like

    declare -a a  
    declare -a a=()

    declare -a a=(1 2)
    declare -a a=([k]=v 2)

    declare -a a+=(1 2)
    declare -a a+=([k]=v 2)

and so forth

It might be useful to use  a table format too ... I added a way to make tables, called `ul-table`, which you can see here:

```
$ fgrep -A 5 '<table>' doc/ref/*.md
doc/ref/chap-word-lang.md:<table>
doc/ref/chap-word-lang.md-
doc/ref/chap-word-lang.md-- thead
doc/ref/chap-word-lang.md-  - Reference
doc/ref/chap-word-lang.md-  - Expression
doc/ref/chap-word-lang.md-  - H-value
--
doc/ref/chap-word-lang.md:<table>
doc/ref/chap-word-lang.md-
doc/ref/chap-word-lang.md-- thead
doc/ref/chap-word-lang.md-  - Character
doc/ref/chap-word-lang.md-  - Where `${x}` would be obtained
doc/ref/chap-word-lang.md-- tr

````

I guess the table is like

(-A assoc or -a indexed, += or =, value or [key]=value or [key]+=value, type of original LHS)

(This reminds me of how I was saying IFS=x and $@ and '' has 3 dimensions -- now it appears we have 4 dimensions)

File: doc/ref/chap-osh-assign.md [hunk 16]
```diff
@@ -49,7 +108,12 @@ In YSH, use a [list-literal][] to create a [List][] instance.
 
 ### sh-assoc
 
-Associative arrays map strings to strings:
+When an initializer list is assigned to an associative array (which maps
```
andychu, line 111: ditto, I think this could use some more examples of the cases

osh/word_.py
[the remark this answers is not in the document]
akinomyoga: > You might be able to use `lexer.IsPlusEquals(tok)`, depending on the token

OK, `lexer.IsPlusEquals` seems to look into the string, so it can be applied to the present case as well.

doc/ref/chap-type-method.md
akinomyoga: Currently, `sh-array` only explains the special rule for the list initialization. The section doesn't explain anything about other operations. We might add the description about the operation `a[i]=v` in `sh-array`, but I'm not sure if we should describe the other operations as the filename is `chap-osh-assign`. In fact, the expansion of arrays of the form `${a[...]}` is explained in `op-bracket` in `chap-word-lang.md`.

osh/word_.py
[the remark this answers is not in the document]
akinomyoga: 4e18d258d

doc/ref/chap-osh-assign.md
akinomyoga: I mover the paragraph in 4c4f59333. Or do you think we should combine the two paragraphs into one?

doc/ref/chap-osh-assign.md
akinomyoga: 4c4f59333

doc/ref/chap-type-method.md
akinomyoga: In commit 4c4f59333, I updated it as "See [sh-init-list][] and [sh-array][] for initialization/mutation of BashArray." for now.

doc/ref/chap-type-method.md
akinomyoga: This is essentially the same as another conversation https://github.com/oils-for-unix/oils/pull/2279#discussion_r1973990831, so I'll update them together. Let's discuss it in the other conversation.

doc/ref/toc-osh.md
akinomyoga: > Maybe sh-init-list should also have the "value" like
> 
> `a=(v [0]=1 ...)`

I already included the "value" form as the third item.

In 4c4f59333, I have moved the third item to the first and adjusted the *value* of the values, indices, and keys.

> and maybe add a second example `a+=(v2 [c]=d [e]+=f)` or something - to make it clear that all the combinations do something

Yeah, that's a good idea. I added it in 4c4f59333.

spec/assoc.test.sh
```diff
@@ -439,28 +425,23 @@ v
 -{a,b}-
 ## END
 
-#### bash mangles array #1
+#### bash mangles indexed array #1 (associative array is OK)
+declare -A a
```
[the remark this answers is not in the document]
akinomyoga: As you noticed below, I moved the case without `declare -A` to `spec/ble-sparse.test.sh`.

spec/ble-sparse.test.sh
akinomyoga: Yeah, I think it would be better to have them in the same place. Where should we put these?

The original location was `spec/assoc.test.sh`, which implies that the test case is intended for  BashAssoc. Then, I moved the test case for BashArray (with the specified index `[i]=v`) to the file for the sparse array, i.e., `spec/sparse.sh`.

Actually, `spec/ble-sparse.test.sh` may be renamed to just `spec/sparse.test.sh`. What do you think?

core/bash_impl.py
akinomyoga: Ah, yes. I'll fix it.

spec/assoc.test.sh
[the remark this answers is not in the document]
akinomyoga: Let's discuss this in the next conversation https://github.com/oils-for-unix/oils/pull/2279#discussion_r1974623753.

builtin/assign_osh.py
akinomyoga: I fixed it in 28bf152247.

osh/cmd_eval.py
akinomyoga: 28bf152247

osh/cmd_eval.py
akinomyoga: Yeah, I fixed it in 850af58fe.

osh/word_eval.py
akinomyoga: 850af58fe

core/bash_impl.py
akinomyoga: I fixed it in 850af58fe.

doc/ref/chap-osh-assign.md
akinomyoga: > I think this text can use some more examples, like

I added an explanation in commit 1d9bab24f.

> It might be useful to use a table format too ... I added a way to make tables, called `ul-table`, which you can see here:

I added a table in commit 173d0d6ff.

> I guess the table is like
> 
> (-A assoc or -a indexed, += or =, value or [key]=value or [key]+=value, type of original LHS)
> 
> (This reminds me of how I was saying IFS=x and $@ and '' has 3 dimensions -- now it appears we have 4 dimensions)

I'm not sure whether I can create a table for four dimensions.

By the way, the dimension of `+=` vs `=` is introduced by OSH. The difference doesn't exist in Bash, while OSH has an additional check for the type of LHS and RHS for `+=`.

akinomyoga:
> I think we should rename ShArrayLiteral to YshArrayLiteral -- that is a good idea, since it's only used for YSH now

475676355

builtin/assign_osh.py
```diff
@@ -203,28 +204,85 @@ def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags):
...
+                pass
+            else:
+                e_die(
+                    "Can't convert type %s into BashArray" %
+                    ui.ValType(old_val), pair.blame_word)
+        elif flag_A:
+            if old_val.tag() in (value_e.Undef, value_e.Str,
+                                 value_e.BashArray):
+                # Note: We explicitly initialize BashAssoc for Undef.
+                # Note: OSH allows overwriting an existing BashArray with an
```
andychu, line 251: Arguably, we should not allow this?    Because then someone could write a working OSH script, and then it doesn't work in bash

That is already true in many places probably, but arguably BashArray and BashAssoc should try to be identical


```
$ bash -c 'declare -a a=(a b); declare -A a'
bash: line 1: declare: a: cannot convert indexed to associative array
```

spec/ble-sparse.test.sh
andychu: Oh I see, maybe we can have

```
spec/indexed-array  # aka sparse array, maybe we need indexed-array-1 and indexed-array-2, for the ble-sparse
spec/assoc-array

spec/bash-array-literal  # common stuff ?
```

This actually relates to a renaming I want to do

I think we should rename:

- BashArray / BashAssoc types ->  IndexedArray / AssocArray
- internal details: YshArrayLiteral and BashArrayLiteral
  - the BashArrayLiteral is for both IndexedArray / AssocArray

I think this is more consistent with bash, and also solves some of our naming issues ?

I will merge this PR, and then we can talk about it separately

andychu (commented):
Looks great, thank you!  I fixed a couple typos in the docs

I think we should consider the renaming mentioned

And I may make some more tweaks to the docs, but this is a big step forward!

spec/ble-sparse.test.sh
akinomyoga: #2282

builtin/assign_osh.py
akinomyoga: #2285

PR #2281, opened 2025-03-01
[osh] Implement Bash 5.1 `assoc=(key value ...)`

akinomyoga:


spec/assoc.test.sh
```diff
@@ -33,7 +33,7 @@ declare -A assoc=([k1]=foo [k2]='spam eggs')
 for v in "${assoc[@]}"; do echo $v; done | sort
 for v in "${!assoc[@]}"; do echo $v; done | sort
 
-# disallow this in OSH?  Changing type?
+# Bash 5.1 assoc=(key value). Bash 4.4 does not implement this.
 
 assoc=(foo 'spam eggs')
 argv.py "${assoc[@]}"
```
andychu, line 39: It might be clearer to test the output of `declare -p assoc` after this?

spec/assoc.test.sh
```diff
@@ -56,16 +56,22 @@ k2
 []
 ## END
 
-#### Can't initialize assoc array with indexed array
+#### Can initialize assoc array with the "(key value ...)" sequence
 declare -A A=(1 2 3)
 echo status=$?
+argv.py "${!A[@]}"
+argv.py "${A[@]}"
```
andychu, line 63: ditto, could also add declare -p output

Although if it doesn't match bash, it's OK to leave it out, if this test has full coverage

andychu (commented):
OK I see this behavior is documented in bash

https://www.gnu.org/software/bash/manual/html_node/Arrays.html

> When assigning to an associative array, the words in a compound assignment may be either assignment statements, for which the subscript is required, or a list of words that is interpreted as a sequence of alternating keys and values: name=(key1 value1 key2 value2 … ). These are treated identically to name=( [key1]=value1 [key2]=value2 … ). 


But I think it is confusing and doesn't improve any programs.  You can always use the clearer style, right?

What do you think about hiding this behavior behind `strict_array`, which is supposed to help with "type confusion" ?


```
$ bin/osh -o strict_array -c 'a=(x y); echo $a'
  a=(x y); echo $a
                ^~
[ -c flag ]:1: fatal: Array 'a' can't be referred to as a scalar (without @ or *)


$ bin/osh -o strict_array -c 's=foo; echo ${s[@]}'
  s=foo; echo ${s[@]}
              ^~
[ -c flag ]:1: fatal: Can't index string with @
```


I asked the same question in the previous PR ... it would require a bit more test coverage, but sometimes I put those in `test/runtime-errors.sh` (e.g. search for strict_array there)

---

It is not essential to add it to strict_array, if there are too many cases we have to handle, and it makes the code more complex

But I think it could be nice

The idea is that when you set `shopt --set strict:all`, you can get a less error prone subset of bash

spec/assoc.test.sh
akinomyoga: OK, I fixed it in 6c2014b4d.

spec/assoc.test.sh
akinomyoga: I fixed this also in 6c2014b4d.

akinomyoga:
> But I think it is confusing and doesn't improve any programs. You can always use the clearer style, right?

This was requested by users of Bash. The form `[key]=value` isn't so friendly when a user wants to save and load the data in a simple text format. For example, even if we have `[key]=value ...` in a text file. We can't do `assoc=($data)` because the detection of `[` and `]=` is performed at the grammar level and not applied to the result of expansions. Zsh implements the same.

Also, Bash 5.2 has `echo "${a[@]@k}"` to generate the sequence `key value ...` from an existing associative array.

> What do you think about hiding this behavior behind `strict_array`, which is supposed to help with "type confusion" ?

I added it in commit 7053b8b56.

> I asked the same question in the previous PR ... it would require a bit more test coverage, but sometimes I put those in `test/runtime-errors.sh` (e.g. search for strict_array there)

I added the test in 8a045d1ac.

andychu:
Thanks!  I noticed a build failure in an experimental "yaks" binary and I think I fixed it

The issue if we have a `yaks -> j8 -> bash_impl` dependency, and mycpp doesn't handle dependencies 100% automatically

I think it should pass now

akinomyoga:
> Thanks! I noticed a build failure in an experimental "yaks" binary and I think I fixed it

Thank you for fixing it!

akinomyoga:
I want to update a documentation.

akinomyoga:
I rebased.

> I want to update the documentation for `strict_array`.

I added 1a7a47996 for `assoc=(k v ...)` with `strict_array`.

akinomyoga:
I resolved conflicts.

andychu:
Great thank you!

PR #2282, opened 2025-03-01
[spec] Move similar tests for `Bash{Array,Assoc}` init into `array-literal.test.sh`

akinomyoga:
https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/Bash.20array.20name.20brainstorming/near/502780012

I moved `ble-sparse` to `array-sparse`. The original `array` is kept for BashArray for the non-sparse usage. `array-sparse` is roughly intended for uses with sparsity. I also moved `assoc` to `array-assoc`. Now everything related to arrays are under the prefix `array`. I also moved `assoc-zsh` to `zsh-assoc` because `zsh-assoc` doesn't test osh, and it seems to be intended for testing Zsh's behavior.

https://github.com/oils-for-unix/oils/pull/2279#discussion_r1976379111

I moved the tests `bash mangles indexed arrays #{1,2}` from `array-assoc.sh` and `array-sparse.sh` into a new file `array-literal.test.sh`


andychu:
Great thanks!

akinomyoga:
Thanks!

PR #2283, opened 2025-03-01
[osh] Promote a string to an array with `str+=(...)`

akinomyoga:
In Bash, string variables can always be used as if it is an array. This is also the case with `str+=(1 2 3)`, which converts the string variable to an array and modifies the content.

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -261,7 +261,7 @@ def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags, arith_ev,
                     ui.ValType(old_val), pair.blame_word)
 
         val = cmd_eval.ListInitializeTarget(val, initializer, pair.plus_eq,
-                                            pair.blame_word)
+                                            mem.exec_opts, pair.blame_word)
     elif pair.plus_eq:
         old_val = sh_expr_eval.OldValue(
             lval,
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -20,7 +20,7 @@ from core import vm
 from display import ui
 from frontend import flag_util
 from frontend import args
-from mycpp.mylib import log
+from mycpp.mylib import log, tagswitch
 from osh import cmd_eval
 from osh import sh_expr_eval
 from data_lang import j8_lite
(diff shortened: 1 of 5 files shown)


spec/append.test.sh
```diff
@@ -49,10 +49,8 @@ argv.py "${y[@]}"
 s='abc'
 s+=(d e f)
 echo $s
```
andychu, line 51: Hm s is an array  now, so I think we should also test `argv.py "${s[@]}"`

builtin/assign_osh.py
```diff
@@ -247,10 +247,16 @@ def _AssignVarForBuiltin(mem, rval, pair, which_scopes, flags, arith_ev,
...
-                # Note: We explicitly initialize BashAssoc for Undef.
+                # Note: We explicitly initialize BashAssoc for Undef and Str.
+                #   When applying +=() to Str, we associate an old value to the
+                #   key '0'.
                 # Note: OSH allows overwriting an existing BashArray with an
                 #   empty BashAssoc.  Bash does not allow this.
-                val = bash_impl.BashAssoc_New()
+                assoc_val = bash_impl.BashAssoc_New()
+                if pair.plus_eq and old_val.tag() == value_e.Str:
+                    bash_impl.BashAssoc_SetElement(assoc_val, '0',
```
andychu, line 257: What do you think about hiding this behavior behind `shopt -s strict_array` ?

It is to prevent confusion over types.  It currently does things like this:



```
$ bin/osh -o strict_array -c 'a=(x y); echo $a'
  a=(x y); echo $a
                ^~
[ -c flag ]:1: fatal: Array 'a' can't be referred to as a scalar (without @ or *)


$ bin/osh -o strict_array -c 's=foo; echo ${s[@]}'
  s=foo; echo ${s[@]}
              ^~
[ -c flag ]:1: fatal: Can't index string with @

```

akinomyoga: OK, I added it in commit 59ad69491.

spec/append.test.sh
akinomyoga: Thanks. I updated the test in commit c5fd118c7.

andychu:
Great thanks!

akinomyoga:
Thank you!

PR #2284, opened 2025-03-01
[doc/ref/chap-osh-assign] Fix descriptions of initializer list

akinomyoga:
Sorry, these are additional fixes to the descriptions added in #2279.


doc/ref/chap-osh-assign.md
```diff
@@ -239,20 +238,21 @@ item in order.  An item in the initializer list must be in the forms
...
-                                # (['k']=1).  Unlike BashArray, "k" is not
-                                # processed by arithmetic expansion.
-    declare -a a+=([a]=3 [b]=4) # This adds two elements to the existing array.
-                                # The result is ([a]=3 [b]=4 [k]=v)
-    declare -a a+=([k]=5)       # This overwrites an element in the existing
-                                # array. The result is ([a]=3 [b]=4 [k]=5).
+    declare -A a=([k]=v)         # This creates a BashAssoc with one element,
+                                 # (['k']=1).  Unlike BashArray, "k" is not
+                                 # processed by arithmetic expansion.
+    declare -a a+=([a]=3 [b]=4)  # This adds two elements to the original
```
andychu, line 251: should this just be `a+=()` without `declare -a`?

If we are going to include the flag, it should be -A right?

doc/ref/chap-osh-assign.md
```diff
@@ -239,20 +238,21 @@ item in order.  An item in the initializer list must be in the forms
...
+                                 # array.
+                                 # The result is ([a]=3 [b]=4 [k]=v)
+    declare -a a+=([k]=5)        # This overwrites an element in the original
```
andychu, line 254: same here, I think just `a+=` makes more sense

the `declare` is redundant

doc/ref/chap-osh-assign.md
```diff
@@ -73,20 +73,19 @@ applied.
...
-  original value is stored at key `"0"` instead of a BashArray.  If the
+  original value is stored at key `"0"`, instead of a BashArray.  If the
   assignment operator is `+=`, OSH issues an error "Can't append an array to
   string", while Bash is permissive.
-- When the LHS is an indexed or associative arrays, the initializer list is
-  applied to the original array following [sh-array](#sh-array) (for an indexed
-  array) or [sh-assoc](#sh-assoc) (for an associative array).  If the
+- When the LHS is an indexed or associative arrays, the original array is
+  directly used for the modification target.  If the
   assignemnt is performed through an assignment builtin and mismatching flag
```
andychu, line 89: typo assignment

doc/ref/chap-osh-assign.md
```diff
@@ -73,20 +73,19 @@ applied.
 In the first phase, the type adjustment is performed in the following way:
 
 - When the LHS variable is unset, the assignment creates an empty indexed array
```
andychu, line 75: We can talk about it later, but it is perhaps confusing to say indexed array (BashArray) -- i.e. to have two different terms

For now it's OK

doc/ref/chap-osh-assign.md
```diff
@@ -45,7 +45,7 @@ cleared before starting the modifications.  When the assignment is performed
 with `+=`, the modifications are applied to the existing content of the LHS
 value.
 
-An initializer list has the following form: `(<words>...)`, where each word has
+An initializer list has the following form: `(<items>...)`, where each item has
```
andychu, line 48: I prefer not using `<>` for placeholders,  like `<key>` and `<items>`

That's because bash has so much extra syntax, it's hard to tell if `<>` is part of it or not

Also I use `*` for repetition instead of `...`

So I think it is nicer to write this like

`'(' ITEMS* ')'` like a grammar

And then `[KEY]=VALUE`

and `VALUE`

etc.

doc/ref/chap-osh-assign.md
```diff
@@ -168,6 +162,11 @@ These rules are summarized in the following table.
   - `-A`
   - the original BashAssoc
   - <!-- empty -->
+- tr
+  - (others)
+  - <!-- empty -->
+  - N/A
+  - Error
 
 </table>
```
andychu, line 172: > When the result is BashAssoc, see [sh-array](#sh-array).

Should be `sh-assoc`

doc/ref/chap-osh-assign.md
akinomyoga: Ah, you are right. I just created these examples by rewriting those for BashArray, and I forgot to rewrite the flags.

I decided to remove `declare -a` / `declare -A`. I fixed it in commit 19d86b06e.

doc/ref/chap-osh-assign.md
akinomyoga: I also removed these `declare -a` before `a+=` in commit 19d86b06e.

doc/ref/chap-osh-assign.md
akinomyoga: Thanks. I fixed it in c383a4d54. I also found another typo by a spell checker.

doc/ref/chap-osh-assign.md
akinomyoga: Which one should we prefer in the documentation?

doc/ref/chap-osh-assign.md
akinomyoga: Thank you. I updated the notation in commit 5c7c4e4c7.

doc/ref/chap-osh-assign.md
akinomyoga: Right, thanks. I fixed it in commit 28359e9fc.

doc/ref/chap-osh-assign.md
andychu: I think we should try to make all of these match

- the type names (user visible types - it would be nice if internal types match too though)
- the help topic names
- the terms used in the text

We can talk about it on Zulip -- but it is not essential right now

It is probably more important to see if this can run any new tests in ble.sh and so forth, and then we can circle back to the names if necessary

The names now are decent, but I think could be improved a bit.

At least I may want to add a `bash-` or `sh-` prefix consistently , with `sh-` meaning "POSIX shell"  ...  Although some things like `+=` are not limited to bash

PR #2292, opened 2025-03-21
io.captureAll() Method (return stdout, stderr and status)

Melkor333:


core/executor.py
```diff
@@ -595,41 +596,59 @@ def RunSubshell(self, node):
...
         #log('Command sub started %d', pid)
 
         chunks = []  # type: List[str]
+        chunks2 = []  # type: List[str]
         posix.close(w)  # not going to write
+        posix.close(w2)  # not going to write
+        eofcount = 0
         while True:
-            n, err_num = pyos.Read(r, 4096, chunks)
+            fds, w, exc = select.select([r,r2], [], [])
```
Melkor333, line 620: @andychu I get the following error and don't know why 😬 

```
core/executor.py:621: error: Incompatible types in assignment (expression has type "List[Any]", variable has type "int")
```

Do you have an idea why this doesn't work here? I don't really get where it gets an `int` when I pass 3 lists...

(And don't worry, I'll make it a deparate function from `captureStdout`. I just wanted to test if it works that way on the phone and wanna clean it up on my laptop once it works)

andychu: Oh sorry I didn't respond to this earlier ...

I think MyPy has problems inferring the type of the empty list `[]`.  So it just makes it  `List[Any]`

Sometimes I have to do this:

```
tmp = []  # type: List[int]
tmp2 = []  # type: List[int]
```

Melkor333:
@andychu it compiles now but gives an indexError I don't really understand...
```
ysh-0.28$ var x = io.captureStdout(^(echo hi; echo hi>&2;))                         
libc++abi: terminating due to uncaught exception of type IndexError*
Aborted
```
Is it "easy" to compile the cpp tarball to throw nice errors with more info?
Also, do you think this `WaitForInputs` function (also the cpp version) makes roughly sense? Or would you do it completely differently?

cpp/core.cc
```diff
@@ -254,6 +254,26 @@ bool InputAvailable(int fd) {
...
+  fd_set fds;
+  FD_ZERO(&fds);
+  int n = len(fds_in);
+  for (int i = 0; i < n; ++i) {
+    FD_SET(fds_in->at(n), &fds);
+  }
+  int out = select(FD_SETSIZE, &fds, NULL, NULL, NULL);
+  for (int i = 0; i < n; ++i) {
+    if (InputAvailable(fds_in->at(n))) {
+      ret->append(fds_in->at(n));
```
andychu, line 268: Hm I wonder if this is causing the IndexError

I don't think you should call `InputAvailable()` here ... the `select()` is supposed to return the list of descriptors

I think you make a new list from what select "returns", unconditionally

core/executor.py
```diff
@@ -595,41 +595,62 @@ def RunSubshell(self, node):
...
-                    e_die_status(
-                        2,
-                        'Oils I/O error (read): %s' % posix.strerror(err_num))
+            # zero outputs mean something went wrong
+            if len(fds) == 0:
+                break
 
-            elif n == 0:  # EOF
+            for fd in fds:
+                if fd == r:
```
andychu, line 627: what is this condition for?

Melkor333:
I've done a first implementation of `io.captureOutputs` 🥳 
BUT the error handling needs a bit of "fleshing out". It does add a `status` but fails when `fail=false` is not passed as an extra argument. I wonder if the `status` should NOT be added to the dict when `fail=true` (default), as it will implicitly always be `0`.

Also. should this `fail` option be added to `captureStdout` as well? Though we return a string on that function and I don't think we want to change that!

Anyway one issue I have (and why the tests are failing) is that when the command is executed and FAILs, the "Oils error" `X command.Simple failed with status Y` is printed to stderr and thus added to the `stderr` variable:
```
test/spec.sh ysh-method-io -v
        our_shell           : ysh
        oils_failures_allowed: 0

ysh-method-io.test.sh
case    line    ysh
  0       4     pass    captureStdout() is like $()
  1      20     pass    captureStdout() failure
  2      40     pass    io->eval() is like eval builtin
  3      54     pass    captureOutputs() is like $() but captures stderr as well
  4      70     pass    captureOutputs() failure
  5      90     FAIL    captureOutputs() with fail=false
case: 5

[ysh stdout]
Expected '(Dict)   {"code":0}\n(Dict)   {"stdout":"one","stderr":"","status":1}\n'
Got      '(Dict)   {"code":0}\n(Dict)   {"stdout":"one","stderr":"  var c = ^(echo one; false; echo two)\\n                      ^~~~~\\n[ stdin ]:1: errexit PID 2185909: command.Simple failed with status 1","status":1}\n'

ysh stdout:
(Dict)   {"code":0}
(Dict)   {"stdout":"one","stderr":"  var c = ^(echo one; false; echo two)\n                      ^~~~~\n[ stdin ]:1: errexit PID 2185909: command.Simple failed with status 1","status":1}

ysh stderr:


  6     110     pass    io->eval() with failing command - caller must handle
  7     128     pass    io->eval() with exit

                ysh
        pass    7
        FAIL    1
        total   8

ysh-method-io: FATAL: Got 1 Oils failures, but 0 are allowed
ysh-method-io: FATAL: Got 1 total failures, but 0 are allowed
```

Is there a way to not have this error message printed? I suspect it has something to do with `p.StartProcess(trace.CommandSub)` that might need to be changed. 🤔 From taking a look at `core/process.py` it seems like there is currently no way to disable the "oils error", but I could be wrong.

core/executor.py
Melkor333: Made it more clear now. It checks if it's the `stdout_fd` and writes to the `stdout_chunks`, otherwise to the `stderr_chunks`.

andychu:
Is it correct to ignore that text on stderr?

Generally the stderr of the shell interpreter itself is  fair game, e.g. in shell:

```
$ x=$(declare bad-name 2>&1 ); echo "x=[$x]"
x=[-bash: declare: `bad-name': not a valid identifier]
```

Melkor333:
> Is it correct to ignore that text on stderr?
> 
> Generally the stderr of the shell interpreter itself is  fair game, e.g. in shell:
> 
> ```
> $ x=$(declare bad-name 2>&1 ); echo "x=[$x]"
> x=[-bash: declare: `bad-name': not a valid identifier]
> ```
> 
> 
> 
> We asked for stderr there, so we got stderr
> 
> there is a `shopt --unset verbose_errexit` option I believe
> 

Ah, I thought this message is also printed when an executable fails, but it's not (I should do less believing and more testing!). I'll just make the test case consistent then.
Do you think there's something else to consider or  do you think it's good when the spec test succeeds?

Melkor333:
@andychu everything done now. :-)

builtin/method_io.py
```diff
@@ -232,6 +232,41 @@ def Call(self, rd):
...
+        # type: (state.Mem, vm._Executor) -> None
+        self.mem = mem
+        self.shell_ex = shell_ex
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+        unused = rd.PosValue()
+        cmd = rd.PosCommand()
+        # Allow us to not fail on error
+        fail = rd.NamedBool("fail", True)
```
andychu, line 246: I think it might be OK to NEVER fail, because returning the status sort of implies that

core/executor.py
```diff
@@ -632,6 +639,64 @@ def CaptureStdout(self, node):
...
+                    if err_num == EINTR:
+                        pass  # retry
+                    else:
+                        # Like the top level IOError handler
+                        e_die_status(
+                            2,
+                            'Oils I/O error (read): %s' % posix.strerror(err_num))
+                elif n == 0:  # EOF
+                    eofcount += 1
+                if eofcount > 1:
```
andychu, line 686: I would say `eofcount == 2`, and then `assert 0 <= eofcount < 2`

core/pyos.py
```diff
@@ -282,6 +282,14 @@ def InputAvailable(fd):
     return len(r) != 0
 
 
+def WaitForInputs(fd):
+    # type: (List[int]) -> List[int]
+    # Get a list of fds and return fds which have available input
+    # read, write, except
+    r, w, exc = select.select(fd, [], fd)
```
andychu, line 289: Hm I want to see what we should do about `exc`

Well at least we can see what other programs / shells do

from `man select`, I guess none of thse apply since it is a pipe ...  well that could be in a comment here



```
       POLLPRI
              There is some exceptional condition on the file descriptor.  Possibilities include:

              •  There is out-of-band data on a TCP socket (see tcp(7)).

              •  A pseudoterminal master in packet mode has seen a state change on the slave (see ioctl_tty(2)).

              •  A cgroup.events file has been modified (see cgroups(7)).
```

andychu: If we're not going to use `w` and `exc`, then we can call them `_, _`

Also, if we ignore `exc`, then I guess we should not pass the third param to select

core/pyos.py
```diff
@@ -282,6 +282,14 @@ def InputAvailable(fd):
     return len(r) != 0
 
 
+def WaitForInputs(fd):
```
andychu, line 285: fd implie a single descriptor -- I would call this `fd_list`

Some other names could be `WaitToRead(fd_list)` or `WaitForReading(fd_list)`

cpp/core.cc
```diff
@@ -254,6 +254,27 @@ bool InputAvailable(int fd) {
...
 
+List<int>* WaitForInputs(List<int>* fds_in) {
+  auto* ret = NewList<int>();
+  fd_set fds;
+  FD_ZERO(&fds);
+  int n = len(fds_in);
+  for (int i = 0; i < n; ++i) {
+    FD_SET(fds_in->at(i), &fds);
+  }
+  int out = select(FD_SETSIZE, &fds, NULL, NULL, NULL);
```
andychu, line 265: yeah here we don't pass the list of "exc"?   So we probably shouldn't pass it in Python, if there are no errors to handle

I'm not sure if that affect "close" or not ...

andychu, line 265: `out` should be something like `num_ready`

cpp/core.cc
```diff
@@ -254,6 +254,27 @@ bool InputAvailable(int fd) {
   return select(FD_SETSIZE, &fds, NULL, NULL, &timeout) > 0;
 }
 
+List<int>* WaitForInputs(List<int>* fds_in) {
```
andychu, line 257: The param name should match Python, so it should be something like  `fd_list`

cpp/core.cc
```diff
@@ -254,6 +254,27 @@ bool InputAvailable(int fd) {
...
+  auto* ret = NewList<int>();
+  fd_set fds;
+  FD_ZERO(&fds);
```
andychu, line 260: `fds` could be something like `fd_bits`, to distinguish from `fd_list`

It is a bit mask passed to the kernel

doc/ref/chap-type-method.md
```diff
@@ -963,9 +963,30 @@ If the command fails, `captureStdout()` raises an error, which can be caught
 with `try`.
 
     try {
-      var s = _io->captureStdout(c)
+      var s = io.captureStdout(c)
     }
 
+### captureOutputs()
+
+Capture stdout and stderr of a command a string.
```
andychu, line 971: - typo "a string"
- should mentino status

doc/ref/chap-type-method.md
```diff
@@ -963,9 +963,30 @@ If the command fails, `captureStdout()` raises an error, which can be caught
...
+
+    var c = ^(echo hi; echo hello >&2;)
+    var outputs = _io.captureOutputs(c)  # => { stdout: "hi", stderr: "hello", status: 0 }
+
+It's like `io.captureStdout` but captures both stdout and stderr.  Trailing newlines `\n` are
```
andychu, line 976: We may want to keep trailing newlines?

The only reason captureStdout() does it is because that's how shell command sub work

I also wonder if there are better names for `captureOutputs()` ...  I was thinking `captureStdoutStderr()`, but that leaves out `status`

or `capture3()` or something

spec/ysh-blocks.test.sh
```diff
@@ -92,6 +92,9 @@ proc my-cd (; b) {
   # Yup, this is a problem
   var s = io.captureStdout(b)
   echo "stdout $s"
+
+  setvar s = io.captureOutputs(b)
+  echo "stdout $[s.stdout]"
```
andychu, line 97: this is a bit confusing because it doesn't change any of the assertions ?  what happened?

andychu (commented):
OK thanks, this looks good!  Just some naming issues mainly

andychu:
and yeah I think we can probably leave out `fail=true`?  If we don't have a use case for that, we can leave it out

And maybe don't strip newlines

doc/ref/chap-type-method.md
Melkor333: I thought especially of `capture3`, too. But it's IMO a bit too "bound to the output". E.g. if we ever want to add something like `additional_fds=[3, 4]`, or execution time, etc. Therefore `Outputs` seemed the cleanest.

andychu: I also thought it could be `captureStd()` or `captureAll()`

I feel like captureStdout() + captureOutputs() is a tiny bit awkward, because is "stderr" and "status" an "output"?  It is hard to find a good name though

I guess All() can also refer to the fact that it doesn't strip newlines

Melkor333: hm `captureAll` sounds good! Let me change it to that and fix the remaining points.

core/executor.py
Melkor333: LOL I ran right into that assert. If a FD returns EOF I shouldn't check for it anymore 😄

spec/ysh-blocks.test.sh
Melkor333: Soo turns out this test was marked as "allowed failure" because it didn't work in the past. But it does work now, so I added the proper `STDOUT`. I think it's behaviour is correct now.

Melkor333:
@andychu I don't get why the CI fails. Seems like something unrelated?

andychu:
Hmm yeah we started getting this:

```
mkdir: created directory '_tmp/code-blocks/doc'
mkdir: created directory '_tmp/code-blocks/doc/ref'
  File "doctools/cmark.py", line 65, in <module>
    raise AssertionError("Couldn't find libcmark.so")
AssertionError: Couldn't find libcmark.so
```

which is unrelated to your change

That is odd, I didn't change anything ... looking into it

andychu:
OK I resolved the CI error, let me take another look

core/executor.py
```diff
@@ -632,6 +639,62 @@ def CaptureStdout(self, node):
...
+                        pass  # retry
+                    else:
+                        # Like the top level IOError handler
+                        e_die_status(
+                            2,
+                            'Oils I/O error (read): %s' % posix.strerror(err_num))
+                elif n == 0:  # EOF
+                    open_fds.remove(fd)
+
+            if len(open_fds) == 0:
```
andychu, line 686: I am not sure we need both `break` conditions ... but I guess it avoids calling select() with an empty list, so maybe that's useful

doc/ref/chap-type-method.md
```diff
@@ -963,7 +963,27 @@ If the command fails, `captureStdout()` raises an error, which can be caught
...
+    }
+
+### captureAll()
+
+Capture stdout and stderr and status of a command.
+
+    var c = ^(echo hi; echo hello >&2;)
+    var outputs = io.captureAll(c)  # => { stdout: "hi", stderr: "hello", status: 0 }
+
+It's vaguely similar to `io.captureStdout` but captures both stdout, stderr into a `dict`.
```
andychu, line 976: I would remove the word "vaguely"

And capitalize `Dict`, which is a type

andychu (commented):
OK looks great!  Just some minor cosmetic things

Thank you

core/executor.py
Melkor333: According to [stackoverflow](https://stackoverflow.com/questions/15109527/return-value-of-select-with-empty-sets):

>     Some code calls select() with all three sets empty, nfds zero, and a non-NULL timeout as a fairly portable way to sleep with subsecond precision.

I have the feeling we might get an indefinite sleep without the second break...

andychu: Ah OK good point, thanks for looking that up

Melkor333:
@andychu sorry for the delay. finally finished the cosmetics.

btw. I didn't intend to fix the `shell.nix` but I had this minor change still configured and it just worked when I briefly tested it today. Reverting my change didn't work, despite that it should work with a more recent libreadline. It might be an inconsistency in the nix build working half of the time (which would be super odd).

andychu:
Merged, thank you!

PR #2293, opened 2025-04-06
[builtin/bind] Fix bind -X output after bind -r removal

KingMob:
## Bug

After calling `bind -r` to remove a `bind -x` binding, the binding itself no longer works, but `bind -X` still listed it sometimes.

The reason is, bash 4.x's unbinding code does not properly clean up the command map (the one for `bind -x`) in all circumstances, and by basing my code on bash 4.x (for older readline compatibility), I replicated the same bug. 

## Fix

Switch to bash 5.x's unbinding method, where we first retrieve the fn used by a binding before removing it, and if it's the -x command callback fn, we also remove it from the associated command map. 

(This code existed, but was commented out, because the first attempt used some functions specific to bash-5.x readline.)

## Extra, extra!

- Updated binding error mesg from last PR
- Removed `unbind_shell_cmd` from public API
- Added a quick spec test of bind -m and bind -x.


andychu:
Great, thanks for the clear description!

Issue #834, opened 2020-10-02
Implement kill builtin


andychu:
A user ran into this.  The external one isn't the same.  The internal one also takes job IDs like `%1`

https://oilshell.zulipchat.com/#narrow/stream/121540-oil-discuss/topic/Bash.20Builtin.20compatibility.3A.20.60kill.20-.C2.ABsigspec.C2.BB.60

jcob-sikorski:
In what files builtin commands generally reside?

andychu:
I think this one can go here: https://github.com/oilshell/oil/blob/master/osh/builtin_process.py

simonLeary42:
should this issue be closed? https://github.com/oils-for-unix/oils/pull/2478

andychu:
Yes this can be closed now, thanks!

PR #2298, opened 2025-04-27
Add two failing tests for kill builtin

AtomsForPeace:
Adds two failing tests for #834. mksh is the only one with slightly unusual behavior in that it returns a different code.

### DIFF
--- a/spec/builtin-kill.test.sh
+++ b/spec/builtin-kill.test.sh
@@ -1,0 +1,28 @@
+## oils_failures_allowed: 2
+## compare_shells: dash bash-4.4 mksh zsh
+
+# Tests for builtins having to do with killing a process
+
+#### Kills the process with SIGTERM
+# Test 1: Basic SIGTERM
+sleep 0.1 &
+pid=$!
+kill -15 $pid
+wait $pid
+echo $?  # Should be 143 (128 + SIGTERM)
+## STDOUT:
+143
+## END
+# For some reason mksh doesn't return the same as the others.
+## OK mksh stdout: 0
+
+#### Kills the process with SIGKILL
+# Test 2: Basic SIGKILL
+sleep 0.1 & 
+pid=$!
+kill -9 $pid 
+wait $pid
+echo $?  # Must be 137 (128 + SIGKILL) 
+## STDOUT:
+137
+## END
--- a/test/spec.sh
+++ b/test/spec.sh
@@ -174,6 +174,10 @@ builtin-process() {
   test/spec-py.sh run-file builtin-process "$@"
 }
 
+builtin-kill() {
+  test/spec-py.sh run-file builtin-kill "$@"
+}
+
 builtin-read() {
   test/spec-py.sh run-file builtin-read "$@"
 }


spec/builtin-kill.test.sh
```diff
@@ -0,0 +1,28 @@
+## oils_failures_allowed: 2
+## compare_shells: dash bash-4.4 mksh zsh
```
andychu, line 2: Please change this to 'bash' (to specify bash 5), since bash-4.4 is only for old tests

andychu (commented):
Thanks!

So the funny thing is that OSH succeeds because these tests work with the external `/usr/bin/kill`

https://op.oilshell.org/uuu/github-jobs/9447/

https://op.oilshell.org/uuu/github-jobs/9447/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-kill.html

I guess one way we can test it is by a new dir  in `PATH` ?

Like `PATH=tmp_dir_with_executable_kill:$PATH`

then this will be found before /usr/bin/kill, but **after** a builtin `kill` I believe

That's one way, not sure if there is a better way

andychu:
BTW the reason bash has `kill` is so that `kill %1` and so forth work.  The `%1` is a shell-specific syntax that looks up a job in bash memory -- it doesn't exist outside the shell, so `/bin/kill` doesn't support that syntax

Although honestly I think there should be a syntax for the process group ID, like `$[pgid(1)]` could return the process group ID of job ID 1 ....

AtomsForPeace:
> Thanks!
> 
> So the funny thing is that OSH succeeds because these tests work with the external `/usr/bin/kill`
> 
> https://op.oilshell.org/uuu/github-jobs/9447/
> 
> https://op.oilshell.org/uuu/github-jobs/9447/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-kill.html
> 
> I guess one way we can test it is by a new dir in `PATH` ?
> 
> Like `PATH=tmp_dir_with_executable_kill:$PATH`
> 
> then this will be found before /usr/bin/kill, but **after** a builtin `kill` I believe
> 
> That's one way, not sure if there is a better way

Ah cool, I didn't even notice that it passed. It fails on my local machine. I wonder why that is.

andychu:
Hm do you have an externally `kill` binary?

On my machine it's:

```
$ which kill
/usr/bin/kill
```

AtomsForPeace:
> Hm do you have an externally `kill` binary?
> 
> On my machine it's:
> 
> ```
> $ which kill
> /usr/bin/kill
> ```

Same on my machine, yeah. Here's my results. 

|→ test/spec.sh builtin-kill
builtin-kill.test.sh
case    line    dash    bash-4  mksh    zsh     osh
  0       6     pass    pass    ok      pass    FAIL    Kills the process with SIGTERM
  1      19     pass    pass    pass    pass    FAIL    Kills the process with SIGKILL

                dash    bash-4  mksh    zsh     osh
        pass    2       2       1       2       0
        ok      0       0       1       0       0
        FAIL    0       0       0       0       2
        total   2       2       2       2       2

builtin-kill: note: Got 2 allowed Oils failures

andychu:
Can you add  `-v` to see why they fail?

It should explain the errors

We want them to fail on the CI as well, so then we can know if it's indeed fixed

AtomsForPeace:
Here it is with -v:

|→ test/spec.sh builtin-kill -v
        compare_shells      : dash bash-4.4 mksh zsh
        oils_failures_allowed: 2

builtin-kill.test.sh
case    line    dash    bash-4  mksh    zsh     osh
  0       6     pass    pass    ok      pass    FAIL    Kills the process with SIGTERM
case: 0
mksh stdout:
0

mksh stderr:


case: 0
[osh stdout] Expected '143\n', got '127\n'
osh stdout:
127

osh stderr:
[%1] PID 40496 Done
  wait $pid
       ^~~~
[ stdin ]:4: 40496 isn't a child of this shell


  1      19     pass    pass    pass    pass    FAIL    Kills the process with SIGKILL
case: 1
[osh stdout] Expected '137\n', got '127\n'
osh stdout:
127

osh stderr:
[%1] PID 40509 Done
  wait $pid
       ^~~~
[ stdin ]:4: 40509 isn't a child of this shell



                dash    bash-4  mksh    zsh     osh
        pass    2       2       1       2       0
        ok      0       0       1       0       0
        FAIL    0       0       0       0       2
        total   2       2       2       2       2

builtin-kill: note: Got 2 allowed Oils failures

andychu:
Oh actually a simple way to make it fail is probably `builtin kill`

i.e. add the `builtin`  prefix on the front of each one

andychu:
Cool, now it fails

https://op.oilshell.org/uuu/github-jobs/9502/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-kill.html#details-0-osh

which means the CI passes :)

---

I am reminded that we have 134 OSH spec test failures now ... we do want to make a renewed focus on getting this number down

Thanks for adding the tests!  And feel free to post any questions on Zulip, or I also opened up Github discussions recently

https://github.com/oils-for-unix/oils/discussions

PR #2303, opened 2025-05-09
docs: fix captureAll output strings

Melkor333:
The strings were missing `\n`! I also formatted it as if it was `= r` removing the spaces.


andychu:
Thanks, sorry for delay

PR #2304, opened 2025-05-11
Updated Zephyr report link

KingMob:


andychu:
This is our  copy  of the Python 2.7 source code ... maybe a better place to put the link is `asdl/README.md` ?

That file links to a blog post, which also has a link

OK now I see it's broken the same way

http://www.oilshell.org/blog/2016/12/11.html

I'll fix it

It's annoying when webmasters don't respect link structure ...

KingMob:
> It's annoying when webmasters don't respect link structure

For all that we talk about link rot, I'm unaware of any website tool explicitly designed to preserve links by default. My librarian archivist friends hate that aspect of modern tech.

Closing this unmerged, iiuc.

PR #2306, opened 2025-05-18
Add readline environment variable support for bind -x commands

KingMob:
This adds support for setting and checking READLINE_LINE and READLINE_POINT env vars for arbitrary shell commands in bind -x.

The shell code can read them to learn what's currently on the line and where the cursor is. If either variable is changed, readline will update the command line/cursor.

This is compatible with bash 4, but not (fully) bash 5. Newer readline offers READLINE_ARGUMENT and READLINE_MARK as well.

----

Since this is the final piece of bind support (in Python/C anyway), I also cleaned up a lot of comments and debug printing.


andychu:
OK awesome, looks like it's passing all the tests now!

https://op.oilshell.org/uuu/github-jobs/9572/interactive.wwz/_tmp/spec/stateful/index.html

I will take a look at the code in more detail

I wouldn't be surprised if say running Warp finds some issues, or Koichi  finds some issues (since `ble.sh` is very thorough)

But this is great progress!

And we can also talk about C++ ... I think if we have some part of  Warp or ble.sh working, that will  be motivating

KingMob:
So, in an irony of ironies for my testing, it seems Warp is interfering with `bind -x` when a shell is "Warpified". 

I expected some of this, because part of how Warp offers nice GUI popups is by overriding standard keyseqs like Ctrl-r. And I'm not surprised that READLINE_LINE and READLINE_POINT `bind -x` commands fail, because I doubt they're using readline under the hood when they talk to the underlying bash process, and it looks like they haven't implemented support for a relatively-niche bash-specific feature.

What I *didn't* expect was that *all* bind -x bindings are blocked in Warpified mode, even custom ones that they shouldn't care about. I'm going to file it as a bug.

Screenshot 2025-05-19 at 12 41 41 PM

KingMob:
Also, I know Warp still uses a couple flags not yet supported in osh: `history -a` and `shopt -s histappend`

pyext/line_input.c
```diff
@@ -1154,32 +1066,21 @@ on_bind_shell_command_hook(int count /* unused */, int key /* unused */) {
...
-    // debug_print("Command return code: %d", cmd_return_code);
-    // debug_print("New line buffer: '%s'", line_buffer);
-    // debug_print("New point: '%s'", point);
-
-    // if (save_readline_state() != 1 || restore_readline_state() != 1) {
-    //     PyErr_SetString(PyExc_RuntimeError, "Failed to update readline state");
-    //     result = 1;
-    //     goto cleanup;
-    // }
+    update_line_if_needed(post_line_buffer);
```
andychu, line 1082: What do these functions do?  Some comments would be useful

builtin/readline_osh.py
```diff
@@ -49,46 +48,67 @@ def __exit__(self, type, value, traceback):
...
+        self.mem.Unset(location.LName('READLINE_POINT'), scope_e.GlobalOnly)
+
+        return (status, post_line_buffer, post_point)
+
+    def _get_rl_env_var(self, envvar_name):
+        # type: (str) -> Optional[str]
+        """Retrieve the value of an env var, return None if undefined"""
+
+        envvar_val = self.mem.GetValue(envvar_name, scope_e.GlobalOnly)
+        if envvar_val.tag() == value_e.Str:
```
andychu, line 93: To test if this value is either `Str` or `Undef` or else, please use this `with tagswitch` idiom

https://oils.pub/release/latest/doc/oils-repo/mycpp/README.html#with-tagstr_switch-switch-statement

You can also grep the code for `tagswitch` -- we use it all over

It translates to a switch statement in C++!

builtin/readline_osh.py
```diff
@@ -49,46 +48,67 @@ def __exit__(self, type, value, traceback):
...
 
-        return (status, line_buffer, str(point))
+        # Retrieve READLINE_* env vars and compare for changes
+        readline_line = self._get_rl_env_var('READLINE_LINE')
+        readline_point = self._get_rl_env_var('READLINE_POINT')
+        post_line_buffer = readline_line if readline_line is not None else line_buffer
+        post_point = int(
+            readline_point) if readline_point is not None else point
+
+        self.mem.Unset(location.LName('READLINE_LINE'), scope_e.GlobalOnly)
```
andychu, line 83: I think we should use a context manager for unsetting to be "exception safe".    e.g. `eval.Run` may  raise an exception, and then this line won't be run

We have an idiom for this - `with ctx_Foo(...)`

https://oils.pub/release/latest/doc/oils-repo/mycpp/README.html#python-context-manager-c-constructor-and-destructor-raii

You can see many examples in say `core/state.py` -- the `__init__` pushes the state, and the `__exit__` pops it.  (And for our purposes, the `__enter__` must be there, but is ignored)

builtin/readline_osh.py
```diff
@@ -49,46 +48,67 @@ def __exit__(self, type, value, traceback):
...
+        """Retrieve the value of an env var, return None if undefined"""
+
+        envvar_val = self.mem.GetValue(envvar_name, scope_e.GlobalOnly)
+        if envvar_val.tag() == value_e.Str:
+            return cast(value.Str, envvar_val).s
+        elif envvar_val.tag() == value_e.Undef:
+            return None
+        else:
+            # bash has silent weird failures if you set the readline env vars
+            # to something besides a string, but I think an exception is better
```
andychu, line 99: Yeah this is a common issue -- bash doesn't check errors very well

Oh this is a good reason to be exception safe, which I commented about ...

Is the `TypeErr` caught somewhere?  Does it end up printing a message to `stderr`?

---

Sometimes GNU readline callbacks swallow errors, which is not great

andychu (commented):
Thank you, looks great!  Mostly style comments ...

builtin/readline_osh.py
KingMob: I'm adding it as `ctx_EnvVars`. It's in `readline_osh.py`, but it's general enough that it could be used elsewhere, if there's other places osh needs to temporarily set/unset env vars.

builtin/readline_osh.py
KingMob: As near as I can tell, the equivalent bash callback, `bash_execute_unix_command`, doesn't return a meaningful status code, so I doubt we should, either.

`bash_execute_unix_command` _does_ return 1 if it can't locate a keymap (rare, I suspect), but otherwise, it always returns 0, ignoring the return code of the bound shell commands. (The only thing the command return code is used for is to change the redrawing behavior when the return code is 124...whatever that's about.)

Just to make sure, I checked with bash 4 and 5. Running `bind -x` commands has no effect on `$?`, so it's not user-visible afaict. (Maybe ble.sh has more insight into this, though.)

----

Given this, I think there's no point in raising an exception, so I'm going to keep printing out the error, but remove the `raise`.

KingMob:
OK, I addressed the feedback. See my comments in the 2 unresolved conversations.

----

Also, I fixed one minor missing piece that I just noticed at the very end. When deciding how to redraw after a bind -x command, we now check to ensure that the command return code isn't 124, to mimic bash.

andychu:
> Also, I know Warp still uses a couple flags not yet supported in osh: history -a and shopt -s histappend

I believe both of these have come up in the past, so are in scope ... thanks for mentioning it

builtin/readline_osh.py
andychu: OK that makes sense, generally I think printing more errors than bash makes sense, but failing hard sometimes leads to incompatibilities

PR #2308, opened 2025-05-26
hide "but oils didn't start it" behind xtrace

Melkor333:
I hope the `not None` to safe a nested `if` is OK. 😇


andychu:
Thanks, I think we should do this

But I'm also wondering what the problem is ... let me check it out for a minute

andychu:
I guess I thought this problem was RARE, but it seems very common, based on the logs ...

PR #2317, opened 2025-06-19
error-catalog.md: add parse_backslash info and hints

bar-g:


bar-g:
bump

andychu:
I looked over this - I think the existing wording is clearer

This seems to "assault" the reader with a lot of information that doesn't necessarily help solve the problem

bar-g:
Your sparse stuff/style is unfortunately very much incomplete, and not systematic.
A larger problem is that as such, it doesn't scale to bring more comprehensive help to the user.

Instead, it actually even brings more puzzles after facing an error.

> Did you mean `\\z`?

what?

Therefore I rather started listing involved options, facts, and then real example alternatives for the user.

Maybe, the facts could be shortened, or their inlined examples be moved to their ends, but I don't think your response is appropriate.
(Besides that the current infos hadn't helped me to understand things I added.)

bar-g:
from the old PR:
> https://github.com/oils-for-unix/oils/pull/1824#issuecomment-1964834619
> this belongs in a new file doc/error-catalog.md

bar-g:
done fixing typos, small improvements and missing bits

PR #2318, opened 2025-06-21
no-quotes.sh: set state back to oldstate

bar-g:
In osh we can't assume errexit is always enabled.

So we set oldstate afterwards again (and disable the verbose trace of this operation in case it was enabled).

### DIFF
--- a/stdlib/osh/no-quotes.sh
+++ b/stdlib/osh/no-quotes.sh
@@ -43,10 +43,12 @@ nq-run() {
   local __status
 
   # Tricky: turn errexit off so we can capture it, but turn it on against
+  local oldstate
+  oldstate="$(set +o)"
   set +o errexit
   ( set -o errexit; "$@" )
   __status=$?
-  set -o errexit
+  set +vx; eval "$oldstate"
 
   out_status=$__status
 }
@@ -62,10 +64,12 @@ nq-capture() {
   local __stdout
 
   # Tricky: turn errexit off so we can capture it, but turn it on against
+  local oldstate
+  oldstate="$(set +o)"
   set +o errexit
   __stdout=$(set -o errexit; "$@")
   __status=$?
-  set -o errexit
+  set +vx; eval "$oldstate"
 
   out_status=$__status
   out_stdout=$__stdout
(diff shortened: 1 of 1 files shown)


andychu:
Hm thanks for looking at this !

Two comments

- I don't understand why `set +vx` is ndeed
- I wonder if we can do it without the command sub, without forking?  There are some things like `shopt -o -p errexit` and `shopt -q errexit`






(wow github's UI has gotten really slow ... hard to type now!)

andychu:
Yeah I think it would be better to query just that one option, and then set a variable that's either `-o` or `+o` for restoration


```
$ bash -c 'shopt -o -q errexit; echo $?; set -e; shopt -o -q errexit; echo $?'
1
0
```

bar-g:
I was assuming that the capture library (or any `.sh` stdlib) is to be posix compatible, because osh serves to run (and test) posix scripts, and assumed posix scripts are to be able to use this also when running in other posix shells.

Thought `shopt` isn't posix.
```
$ dash -c 'shopt'
dash: 1: shopt: not found
```

I used what I think is the [generic] posix shell idom for "save, modify, restore basic shell settings".  The `set +vx` disables printing all settings as they are being restored, in case it was enabled (temporarily or before).

Using "$-" itself would not need a command sub, but I did not find a posix solution to check if it contains an 'e', any idea?.

bar-g:
Uh, checked back on the $(set +o) method I was using, but sure that posix solution depends on bash's posix mode or `shopt -s inherit_errexit` explicitly enabled, so would require to ensure that...

So maybe this: `case "$-" in (*e*) restore_ee='set -e';; (*) restore_ee=':';; esac`

andychu:
Oh sorry for the delayed response -- the goal of the OSH stdlib is to work on BASH and OSH, not dash

Because dash doesn't have some essential things like `set -o pipefail`

andychu:
Hm thanks for taking the time to add this

I guess it is reasonable ... I have a feeling it will break for some reason because I never run with `set +e` -- i.e. it would be better to have tests

But if you're going to USE this, then I guess you can tell us if it breaks ?

bar-g:
Tests could be:

OSH-Test:
```
$ osh -c 'echo $-; source $LIB_OSH/no-quotes.sh; nq-run _ true; echo $-'

e
```
Expect the same output twice (currently the empty string!).

bar-g:
Other ideas might be to instead give this lib a name that makes it clearer which is the limited use of it, and generally return with error on loading if errexit is off.


Short name + short description maybe:
`errcap.capture-cmd-outputs-into-vars.sh`

Short loading:
`source $LIB_OSH/errcap.*.sh`

PR #2322, opened 2025-06-24
chap-stdlib.md: correct all() section to denote all() examples

bar-g:


andychu:
Oh thanks!  This is why we need a scheme to **test** all the examples !

I did that with `doc/ysh-tour.md`, but it's a bit hard with the Oils reference

PR #2323, opened 2025-06-25
quote.test.sh: copy-modify-write attempt to test ctrl-Z escape

bar-g:
No idea really.

### DIFF
--- a/spec/quote.test.sh
+++ b/spec/quote.test.sh
@@ -198,6 +198,11 @@ echo $'\1 \11 \11 \111' | od -A n -c | sed 's/ \+/ /g'
  $ 001 \t \t I \n
 ## END
 
+#### $'' ctrl sequence escapes
+echo $'\cZ' | od -A n -c | sed 's/ \+/ /g'
+## STDOUT:
+  032 \n
+## END
 
 #### OSH allows invalid backslashes
 case $SH in (dash|mksh) exit ;; esac


andychu:
- What does the output look like locally?   with `test/spec.sh quote`?
  - the idea is to make it PASS under bash, and FAIL under OSH
- Can you send this from a branch of the `oils` repo?
  - that way we can see the pass/fail on the WEB too
  - mentioned here - https://github.com/oils-for-unix/oils/wiki/Contributing

andychu:
I sent you a github invite, so you can push to your own branch in `oils-for-unix/oils`

And run the CI

This will help with the docs too!  The docs are automatically rendered by the CI

andychu:
On the surface, this looks right though!

Probably can be expanded a bit more, but it looks like the right idea

Lately I use a hex format, which I find nicer

```
spec/nul-bytes.test.sh:  echo -n "$1" | od -A n -t x1
```

but I see the tests are not consistent



```
$ grep -w od spec/*
spec/assign-extended.test.sh:bash -c 'source tmp.bash; echo "$s1 $s2"; echo -n "$s3" "$s4" | od -A n -t x1'
grep: spec/bin: Is a directory
spec/builtin-echo.test.sh:echo -en '\03777' | od -A n -t x1 | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo -en '\04000' | od -A n -t x1 | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo $flags '\0777' | od -A n -t x1 | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo -en 'abcd\x6' | od -A n -c | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo -e '\x' '\xg' | od -A n -c | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo $flags 'abcd\04' | od -A n -c | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo -en 'abcd\u006' | od -A n -c | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo $flags '\u6' | od -A n -c | sed 's/ \+/ /g'
spec/builtin-echo.test.sh:echo $flags '\0' '\1' '\8' | od -A n -c | sed 's/ \+/ /g'
spec/case_.test.sh:  #echo -n $c | od -A n -t x1
spec/nul-bytes.test.sh:  echo -n "$1" | od -A n -t x1
spec/nul-bytes.test.sh:  echo -n "$s" | od -A n -t x1
spec/nul-bytes.test.sh:  echo -n "$s" | od -A n -t x1

```

bar-g:
I found the invite now, only looked at wrong place first (notifications).

I put local building endeavor on the todo, my daily 5 Watts "machine" is probably not that well suited.

bar-g:
Don't wait for me to add this, if it's a missing test.

andychu:
OK thanks, I commited this with more test cases here - https://github.com/oils-for-unix/oils/commit/d90601eac9416924fb4d15db7d015848c12d1f41

Apparently `\ca` and `\cA` are  the same thing as Ctrl-A, which means to mask the ASCII value by `0x1f` == 31

PR #2327, opened 2025-06-26
chap-builtin-cmd.md: add info about slow, unbuffered 'read'

bar-g:


doc/ref/chap-builtin-cmd.md
```diff
@@ -454,6 +456,7 @@ You may want to use `fromJson8()` or `fromJson()` after reading a line.
 
 (Unlike OSH [read](#read), none of these features remove NUL bytes.)
 
+NOTE: For fast, non-interctive, buffered reading use [io.stdin]: chap-type-method.html#stdin
```
andychu: I think this won't render correctly

You can build it locally with `build/doc.sh all-ref`

And also send it from a branch on `oils-for-unix/oils`, so the CI can run, and then we  can view the docs rendered by the CI

andychu (commented):
OK, I think it's good to mention some of these things

doc/ref/chap-builtin-cmd.md
bar-g: ah, changed syntax

andychu:
OK is this ready to look at?

See the note here:

> If you only push commits, I won't look at the PR again, because I'll assume it's in progress

https://github.com/oils-for-unix/oils/wiki/Contributing#pull-requests

bar-g:
Ok, gave it another pass, but maybe mark any "fast" --long-options (if there are some? I don't know maybe --all ??)

(Good to know commits put it on hold, can nicely save some back and forth.)

bar-g:
All right, please take a look.

Structured it according to the 2 of 3 ways of doing IO.

What shows is that the current long flags aren't really clear about verbatim "raw" (seemingly only in char-based-slow) vs. block-based-fast. (They give no idea to tell without reading documentation, in contrast to having something like `read --chars` vs. `read --block`.)

I'm not that sure about dabbing the legacy `read` algorithm as "text processing", to maintain the connection between the parts, but couldn't think of something.

bar-g:
If you don't like "ysh adds reduced options", maybe call them "streamlined"?

bar-g:
https://github.com/oils-for-unix/oils/commit/fdfd13702066433b2135976ccd2341404e111f4a

PR #2328, opened 2025-06-27
[builtin/fc] Initial implementation of `fc -l`

ghost:
#### Summary

This is the first pull request for `fc`. This just implements most of the display features. On thing that's missing is with `[first]` and `[last]`, you can index with a string that does a "starts with" type of match to find the index. Seems like it would be pretty complex, so I thought of tabling that for now.

I wasn't sure if I needed to update documentation or anything. Just let me know, and I'll get it done 🙂 👍

### DIFF
--- a/builtin/readline_osh.py
+++ b/builtin/readline_osh.py
@@ -465,7 +465,9 @@ class Fc(vm._Builtin):
 
             return 0
 
-        # TODO:
+        # TODO: -e
         if arg.e is not None:
             pass
-        e_usage('edit not implemented', loc.Missing)
+
+        # We currently require -l
+        e_usage("command editing isn't yet implemented", loc.Missing)
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1634,21 +1634,28 @@ Flags:
 
     fc FLAG* FIRST? LAST?
 
-Display or execute commands from the shell's history entries.
+"Fix a command" from the shell's history.
 
-FIRST and LAST are numbers. A positive number refers to the index of the
-history entry. A negative number refers to an offset from the current command.
-When displaying history entries, if FIRST is not specified `-16` is the
-default value.
+`fc -l` displays commands.  FIRST and LAST specify a range of command numbers,
+where:
+
+- A positive number is an index into the history list.
+- A negative number is an offset from the current command.
+- If FIRST is omitted, the value `-16` is used.
+- If LAST is omitted, the current command is used.
 
 Flags:
 
-    -l  Displays the history entries.
-    -n  Excludes the line numbers when displaying history entries.
-    -r  Reverses the order of the displayed history entries.
-<!--    -e EDITOR
-    -s -->
+    -l  List commands (rather than editing)
+    -n  Omit line numbers
+    -r  Use reverse order (newest first)
 
+<!-- 
+Not implemented
+
+-e EDITOR
+-s
+-->
 
 ## Unsupported
 
--- a/doc/ref/toc-osh.md
+++ b/doc/ref/toc-osh.md
@@ -66,10 +66,10 @@ preserve_anchor_case: yes
   [Child Process] jobs        wait
                   fg        X bg       X kill        X disown 
   [External]      test [      getopts
-  [Conditional]   cmd/true    cmd/false               colon :
+  [Conditional]   cmd/true    cmd/false                colon :
   [Introspection] help        hash       cmd/type    X caller
   [Word Lookup]   command     builtin
-  [Interactive]   alias       unalias    history     fc     X bind
+  [Interactive]   alias       unalias    history       fc     X bind
 X [Unsupported]   enable
 ```
 


test/spec.sh
```diff
@@ -546,6 +550,10 @@ xtrace() {
   test/spec-py.sh run-file xtrace "$@"
 }
 
+ysh-TODO-deprecate() {
```
andychu, line 553: Oh interesting, I think you have a different glob order, so this file has a bunch of ordering diffs

Actually I noticed an issue where OSH has a different glob order, so I wonder if it's related to that!

ghost: I think I'm going to need help with this one, @andychu. I'm pretty sure I'm running it with Bash v5.2, but I can't get it to output any differently. I don't think it's an OSH issue, but I could be wrong. 😅

andychu: Ah I think it's a locale issue  ... try this maybe?

```
andy@hoover:~/git/oils-for-unix/oils$ unset LC_ALL

andy@hoover:~/git/oils-for-unix/oils$ unset LC_COLLATE

andy@hoover:~/git/oils-for-unix/oils$ test/spec-runner.sh gen-task-file 

andy@hoover:~/git/oils-for-unix/oils$ git diff

```

Then I think the diff will be smaller

And if that works, then we should put those settings inside the `gen-task-file` function

We want to make the sorting  insensitive  to the machine it's running on!

ghost: Okay, I think I understand what's going on. The current order that my commit does _should_ be with collation unset. If `ysh-TODO-deprecate` was placed next to `ysh-tuple` and not `ysh-assign`, for example, then that would be collation.

It looks like Alpine (musl libc), might not have collation support at all ([source](https://wiki.musl-libc.org/open-issues#Locale-limitations)), and even trying to add support via a package like `musl-locales` doesn't help. Now, if you want it to be sorted with collation, I could run this script on another machine, or if you want no collation, I could add some form of `unset LC_COLLATE` to either the top of the script or to just `gen-task-file`. Very curious what your thoughts are after this comment 😄

ghost: Yeah, when I run this on a non-Alpine machine, `LANG=en_US.UTF-8` gives the result that's currently in `master` at the time of this message. `LANG=C` or `unset LANG` gives what's in this pull request at the time of this message.

I say `LANG` because that was what was set by default, but if I unset that, and use `LC_COLLATE=en_US.UTF-8` instead, same sort as master. So I believe collation is being done in what's in `master` at the time of this message, essentially.

andychu: Ah OK, you can leave it then!  This new order should be the canonical order -- it can be sorted by bytes, without locales

Thanks!

ghost:
@andychu, I’m not sure what the current CI errors are. Is there anything I can run locally to work through them? If it’s nothing introduced or if nothing needs to be changed, I think this initial implementation of `builtin/fc` is ready for a final review 🙂

andychu:
Ah so this code type checks now, which is  great!

But now we run into some mycpp limitations

I just pushed a couple fixes - https://github.com/oils-for-unix/oils/pull/2328/commits/b6552f18f9b259ab82681761ffabdb48691fe233

The error is kind of hidden at the bottom of the logs: https://op.oilshell.org/uuu/github-jobs/9807/cpp-small.wwz/_tmp/soil/logs/oils-cpp-smoke.txt

```
builtin/readline_osh.py:436: KeyError: <mypy.nodes.TupleExpr object at 0x7f18335881d0>
```

That is not a good error, but it at least it gives you the line number

----

Now it doesn't like `reversed(xrange(...`

....

andychu:
There is a big doc about mycpp here - https://oils.pub/release/latest/doc/oils-repo/mycpp/README.html

But I'd say the bottom line is that we may have to fit it in one of these two idioms

- explicit for loop - we do support `for i in xrange()` as a special case
- MAYBE list comprehension


```
mylist = [x for x in xrange(0, 5)]   # not sure if this works in mycpp?

f([x for x in ])   # note: this doesn't work, because list comprehensions must be on the RHS of an assignment
```

We are sort of writing "garbage-collected C++ in Python syntax" :-)  :-/

andychu:
And then test locally with

```
./NINJA-config.sh
ninja   # this ends up running mycpp
```

If you get bad error messages, let me know!

ghost:
Wow, this `mycpp` is a bit painful if I'm being honest 😅 I haven't been able to get asan to work in Alpine, so I modified `build/ninja_main.py` and `test/spec-cpp.sh` locally to to use the `opt` variant. There's probably a better way, but I couldn't find it.

I see that the test failures are specifically for the loop, so I guess I need to just fiddle with this loop until C++ is happy.

ghost:
I think I got it. I think there's no support for negative steps in `xrange`.

ghost:
@andychu, I think this should be good for a review now 🙂 Thanks for all your guidance on this one!

builtin/readline_osh.py
```diff
@@ -376,3 +376,94 @@ def Run(self, cmd_val):
...
+
+        attrs, arg_r = flag_util.ParseCmdVal('fc', cmd_val)
+        arg = arg_types.fc(attrs.attrs)
+
+        # Returns 0 items in non-interactive mode?
+        num_items = readline.get_current_history_length()
+
+        first_arg, first_arg_loc = arg_r.Peek2()
+        if first_arg is None:
+            if num_items > 16:
```
andychu, line 409: Hm, where does the number 16 come from?  Is that what bash does?

May be worth a comment

builtin/readline_osh.py
```diff
@@ -376,3 +376,94 @@ def Run(self, cmd_val):
...
+            else:
+                first_index = 1
+        else:
+            try:
+                # TODO: Support string arg
+                first_index = int(first_arg)
+            except ValueError:
+                e_usage('got invalid argument %r' % first_arg, first_arg_loc)
+        if first_index < 0:
+            first_index = num_items + first_index
```
andychu: You should be able to use `first_index += num_items` here -- mycpp should allow it

builtin/readline_osh.py
```diff
@@ -376,3 +376,94 @@ def Run(self, cmd_val):
...
+        if last_arg is None:
+            last_index = num_items - 1
+        else:
+            try:
+                # TODO: Support string arg
+                last_index = int(last_arg)
+            except ValueError:
+                e_usage('got invalid argument %r' % last_arg, last_arg_loc)
+        if last_index < 0:
+            last_index = num_items + last_index
```
andychu: ditto, can use +=

spec/builtin-fc.test.sh
```diff
@@ -0,0 +1,258 @@
+## tags: interactive
+## compare_shells: bash
+## oils_failures_allowed: 2
+
+#### fc -l
```
andychu: Can you describe the tests in more detail?    `-l` is supposed to list the commands

Although actually I'm a bit confused in reading the test

I think it would be better not to use `seq 3` in the hist file?

Can we make it `echo 1; echo 2; echo 3` so it looks like a real history file?

spec/builtin-fc.test.sh
```diff
@@ -0,0 +1,258 @@
...
+
+## STDOUT:
+1	 history -r
+2	 1
+3	 2
+4	 3
+^D
+## END
+
+#### fc -ln
```
andychu: ditto, I would write 'fc -ln lists history commands without numbers'

spec/builtin-fc.test.sh
```diff
@@ -0,0 +1,258 @@
...
+
+## STDOUT:
+	 3
+	 2
+	 1
+	 history -r
+^D
+## END
+
+#### fc -l with default offset
```
andychu: "fc  -l without args uses the default command history offset" (if that is accurate)

spec/builtin-fc.test.sh
```diff
@@ -0,0 +1,258 @@
...
+case $SH in bash) echo '^D' ;; esac
+
+## STDOUT:
+2	 1
+3	 2
+4	 3
+^D
+## END
+
+#### fc -l with first as offset
```
andychu: is this the "fc -l passed negative number as offset"  case ?

andychu:
Oh also you can auto-format the code with

```
devtools/format.sh install-yapf  # install a specific version that we need to agree on


devtools/format.sh yapf-changed  # better: make sure everything is committed on your branch,
                                 # it only auto formats what's changed

```

ghost:
Thanks for the comments. I just want to clarify that two big things I'm pulling from are the Bash manual pages on `fc` (for offset, index, `-16` default offset, other interesting future corner cases like `-0`, etc.), and the current implementation of `history` in Oils (the less-descriptive tests). I'm also unsure where to place documentation to note things like offsets, indexes, etc. I'll try my best to address the feedback 👍

spec/builtin-fc.test.sh
ghost: > I think it would be better not to use seq 3 in the hist file?
>
> Can we make it echo 1; echo 2; echo 3 so it looks like a real history file?

I'm torn on this, because writing all the echos for "X" number of tests seems like a lot. I do agree that the output doesn't look very history-like. Maybe a reasonable middleground could be something like:

```sh
for i in $(seq 3); do
    printf 'echo %s\n' "$i"
done > tmp
```

Or using an xargs to make it a bit more concise. The extra echoes isn't too big for something with a history of 3 or 4, but my biggest size concern would be for listing out all the 16+ echoes twice for the default page size (offset) test.

andychu: How about this:

```
$ printf 'echo %s\n' a b c
echo a
echo b
echo c
```


This takes advantage of `printf` "vectorization", which I didn't know about before starting OSH!


In bash, you can also do

```
$ printf 'echo %s\n' {a..z}
echo a
echo b
echo c
echo d
echo e
echo f
echo g
echo h
```

andychu:
Ah yes for docs, you can start with

- `doc/ref/toc-osh.md` - now I see `X fc` - we can remove the `X` to say it's implemented !
- `doc/ref/chap-builtin-cmd.md` - search for `#history`, and then you can add notes about what we support under `### fc`

Oh yes it annoys me that bash has a different man page and a `help` page

It also has an  `info` page from what I remember!

andychu:
And to build docs, it's `build/doc.sh all-ref`

Let me know if you have other questions ... thank you for working on this!

andychu:
> Wow, this `mycpp` is a bit painful if I'm being honest 😅 I haven't been able to get asan to work in Alpine, so I modified `build/ninja_main.py` and `test/spec-cpp.sh` locally to to use the `opt` variant. There's probably a better way, but I couldn't find it.
> 
> I see that the test failures are specifically for the loop, so I guess I need to just fiddle with this loop until C++ is happy.

Ah sorry I missed this comment

Oh interesting that ASAN didn't work

```
ninja _bin/cxx-opt/osh
```

will build the optimized binary ... but yeah I think `cxx-asan` is hard-coded for `test/spec-cpp.sh`

Hm ... if this causes a problem we should address it, not sure how at the moment

I thought that ASAN was built into GCC, and available everywhere.  I wonder what the error messages were?

andychu:
And actually code that passes tests in Python is mergeable ... I often view C++ as a "second step"

But I thought this PR was very close!   and you did figure out the issues very quickly, so it seems OK now?

mycpp is definitely the worst part of the codebase (it shouldn't crash with long stack traces), but it was a challenge to even get working!

ghost:
Re: ASAN stuff on Alpine, perhaps we can pull this into a Zulip topic to follow-up on it. The issue took probably a full hour of changing the loop & re-testing until I finally found out that the culprit was the `xrange` loop step argument.

This PR should be quite close now, since tests are passing, it's basically just some nits to work through. I'll see how quickly I can wrap it up. Again, I appreciate your guidance on this request. Hopefully future requests will be a bit smoother now that I've gotten acquainted with many parts of this code base 😅

spec/builtin-fc.test.sh
ghost: That is such a beautiful TIL 🤩

ghost:
@andychu, I think this should be good for another review 👍 Thank you!

andychu:
Thank you, looks great!

I edited the docs a bit based on the bash man page

Hopefully it is accurate, because I find it is extremely confusing!

> When listing, a first or last of 0 is equivalent to -1 and -0 is equivalent to the current command (usually the fc command); otherwise 0 is equivalent to -1 and -0 is invalid.  If last is not specified, it is set to the current command for listing (so that ``fc -l -10'' prints the last 10 commands) and to first otherwise.  If first is not specified, it is set to the previous command for editing and -16 for listing.

Feel free to edit my edits if you decide to implement more of `fc`

PR #2331, opened 2025-06-29
[deps] Add more required wedge dependencies for Alpine

jameschensmith:
Running the one-time setup does not currently work on a fresh Alpine install.

Currently, "coreutils" is needed for some advanced usage of "xargs". "findutils" is needed for some advanced usage of "date".


andychu:
Excellent, thank you!

PR #2332, opened 2025-06-30
args.ysh: Add support for flag escaping via --

quexxon:
Ref: [#oil-dev > A missing feature in args.ysh?](https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/A.20missing.20feature.20in.20args.2Eysh.3F/with/526243441)

### DIFF
--- a/stdlib/ysh/args-test.ysh
+++ b/stdlib/ysh/args-test.ysh
@@ -380,6 +380,9 @@ proc test-escaped-args {
 
   try { call parseArgs(spec, :|-f 1.0 pos_arg -- -i foo bar|) }
   assert [2 === _error.code]
+  # still an error when no args follow `--`
+  try { call parseArgs(spec, :|-i --|) }
+  assert [2 === _error.code]
 }
 
 if is-main {
--- a/stdlib/ysh/args-test.ysh
+++ b/stdlib/ysh/args-test.ysh
@@ -357,15 +357,15 @@ proc test-escaped-args {
 
   var args = parseArgs(spec, :|-f 1.0 pos_arg -- -i foo bar|)
 
-  assert [type(args.float) === 'Float']
+  assert ['Float' === type(args.float)]
   assert [floatsEqual(args.float, 1.0)]
-  assert [args.int === null]
-  assert [args.pos === 'pos_arg']
-  assert [type(args.rest) === 'List']
-  assert [len(args.rest) === 3]
-  assert [args.rest[0] === '-i']
-  assert [args.rest[1] === 'foo']
-  assert [args.rest[2] === 'bar']
+  assert [null === args.int]
+  assert ['pos_arg' === args.pos]
+  assert ['List' === type(args.rest)]
+  assert [3 === len(args.rest)]
+  assert ['-i' === args.rest[0]]
+  assert ['foo' === args.rest[1]]
+  assert ['bar' === args.rest[2]]
 
   # positional args are still required
   try { call parseArgs(spec, :|-f 1.0 -- -i foo bar|) }
--- a/stdlib/ysh/args-test.ysh
+++ b/stdlib/ysh/args-test.ysh
@@ -347,6 +347,41 @@ proc test-multi-value-flags {
   assert [2 === _error.code]
 }
 
+proc test-escaped-args {
+  parser (&spec) {
+    flag -f --float (Float)
+    flag -i --int (Int)
+    arg pos
+    rest rest
+  }
+
+  var args = parseArgs(spec, :|-f 1.0 pos_arg -- -i foo bar|)
+
+  assert [type(args.float) === 'Float']
+  assert [floatsEqual(args.float, 1.0)]
+  assert [args.int === null]
+  assert [args.pos === 'pos_arg']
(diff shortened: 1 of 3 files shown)


stdlib/ysh/args-test.ysh
```diff
@@ -347,6 +347,41 @@ proc test-multi-value-flags {
...
+  parser (&spec) {
+    flag -f --float (Float)
+    flag -i --int (Int)
+    arg pos
+    rest rest
+  }
+
+  var args = parseArgs(spec, :|-f 1.0 pos_arg -- -i foo bar|)
+
+  assert [type(args.float) === 'Float']
```
andychu: I know this file is inconsistent, but we should have `assert ['Float' === `, with expected on the left

Similar to `assert [2 === `

Because the error message actually does say "expected" and "got".  (This order is conventional in Python and Java I believe)

stdlib/ysh/args-test.ysh
```diff
@@ -347,6 +347,41 @@ proc test-multi-value-flags {
...
+
+  # it's an error to provide `--` without `rest`
+  parser (&spec) {
+    flag -f --float (Float)
+    flag -i --int (Int)
+    arg pos
+  }
+
+  try { call parseArgs(spec, :|-f 1.0 pos_arg -- -i foo bar|) }
+  assert [2 === _error.code]
```
andychu, line 382: How about adding a test case for `:| -i -- |` with no `rest` -- since `--` is not an arg I think that should be accepted?

stdlib/ysh/args.ysh
```diff
@@ -165,9 +165,19 @@ func parseArgs(spec, argv) {
...
   while (i < argc) {
     var arg = argv[i]
-    if (arg.startsWith('-')) {
+
+    if (escape_remaining) {
+      if (not spec.rest) {
+        error "Too many arguments, unexpected '$arg'" (code=2)
+      }
+
+      call rest->append(arg)
```
andychu: Ah OK I see this error message and append logic is from below

I might want to put this in one place, but I can see that might be difficult with the current structure

andychu (commented):
Thanks for adding this!

Just some minor comments about the tests

I am not sure about the loop structure, but if it passes the tests, then we could do that later, if desired

stdlib/ysh/args-test.ysh
quexxon: I'll add that test case. That was actually a question. I left it a little loose so there's no error if `--` is the last argument and `rest` is not defined - because there's no harm in that case. Alternatively, we could be more strict with a dedicated error anytime `--` is provided and `rest` is not defined.

andychu: Yeah if I am unsure about something, I often survey Python and JavaScript, e.g. what `argparse` does

I am not 100% sure in this case, but I think they will allow the user to specify `--` even if the program doesn't want 
args


Well I guess the other case is 

```
ls -- *.py
```

The glob can be empty in YSH, unlike in bash !   I think that is `shopt -s nullglob`

In that case the program DOES want args , though

quexxon: Node.js's [util.parseArgs](https://nodejs.org/docs/latest/api/util.html#utilparseargsconfig) allows a final `--` when positional arguments are disallowed (i.e., `strict === true` or `allowPositionals === false`). Passing an argument after `--` raises a parse error. That's not documented, but observed behavior on v20.17.

quexxon: Python's `argparse` is more strict. A final `--` is disallowed unless the final positional arg has something like `nargs='*'`.

andychu: Oh, thank you for surveying !!!  That's kinda annoying, we don't have consensus!

I guess I don't have an opinion then?   we just decide on way or the other

I was thinking about `ls -- *.py`, but that is the case where you allow args

I guess you can argue there is no reason to pass `--` if there will never be an args

Say you have `cp`.  It accepts a source and destination, and there is no `rest`

```
cp src dest --          # this should probably be disallowed?
```

Yeah that looks wrong to me, so if I am thinking about this right, I guess disallow it?

andychu: and if you consider something like


```

$ whoami
andy

$ whoami --
andy

$ whoami a
whoami: extra operand ‘a’
Try 'whoami --help' for more information.

$ whoami -- a
whoami: extra operand ‘a’
Try 'whoami --help' for more information.
```

Well that gives us the opposite conclusion -- it does allow `--`  !   Because they use a common flag parser to process `--help --version`


---


To me, `cp src dest --` looks wrong, and `whoami --` is questionable ...

stdlib/ysh/args-test.ysh
quexxon: 👍🏻 done

stdlib/ysh/args-test.ysh
quexxon: Added the requested test case, and made the change to strictly prohibit `--` without `rest`.

andychu:
Excellent, thank you!

PR #2343, opened 2025-07-04
[builtin/assign_osh] Do not print variable name for unknown types

akinomyoga:
From private communication:

> While checking one of my WIP branches, I noticed that var d = {}; declare -p d produces declare -- d, which appears to be an unset value of the variable d. This is an intersection area of OSH and YSH, so this isn't a copatibility issue, but I'm not sure if the current design makes sense. What do you think?
>
> Andy Chu: Oh yes, I think I noticed that too
> I think we could "pretend d doesn't exist"?  Just not print itAndy Chu: Because there is no way to express it with the declare -p syntax ...

### DIFF
--- a/builtin/assign_osh.py
+++ b/builtin/assign_osh.py
@@ -196,7 +196,7 @@ def _PrintVariables(
                 ["=", bash_impl.BashArray_ToStrForShellPrint(sparse_val)])
 
         else:
-            pass  # note: other types silently ignored
+            continue  # note: other types silently ignored
 
         print(''.join(decl))
         count += 1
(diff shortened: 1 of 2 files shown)


andychu:
Looks great, thank you!

---

Hm I see this error

https://github.com/oils-for-unix/oils/actions/runs/16065846876/job/45340148438


```
Run soil/github-actions.sh publish-cpp-tarball
$OILS_GITHUB_KEY not set
```

Oh yes this is the issue where you have to work in the `oils-for-unix/oils` repo unfortunately, rather than in a fork


That is one thing I wanted to look at for the 4th grant.  The issue is safetly publishing tarballs / authorization ....

I will merge this anyway, since I think the tests will pass

akinomyoga:
> Oh yes this is the issue where you have to work in the `oils-for-unix/oils` repo unfortunately, rather than in a fork

Ah, OK. I had completely forgotten that restriction.

akinomyoga:
Thanks!

PR #2344, opened 2025-07-04
[builtin/pure_osh] Print all variables with `set`

akinomyoga:
Currently, `set` doesn't print arrays. This PR makes `set` to print arrays.

Actually, the original code for `set` was a duplicate of the other builtins in `assign_op.py`, so I switched the `set` implementation to use `assign_op.PrintVariables` with refactoring. In the refactoring, I switched to use `builtin_i` instead of local constants `_OTHER`, `_READONLY`, and `_EXPORT` (while assuming that `0` is not assigned to mean any builtins (currently [builtin_id starts with 1](https://github.com/oils-for-unix/oils/blob/4c4d91d3c4cb0ac43d7ce6e900cafc996766f367/frontend/builtin_def.py#L93)).


akinomyoga:
OK, there was an existing test case. I updated the existing test case and dropped the new test case.

builtin/assign_osh.py
```diff
@@ -25,35 +25,33 @@
...
         mem,  # type: state.Mem
-        errfmt,  # type: ui.ErrorFormatter
-        cmd_val,  # type: cmd_value.Assign
-        attrs,  # type: _Attributes
+        errfmt,  # type: Optional[ui.ErrorFormatter]
+        cmd_val,  # type: Optional[cmd_value.Assign]
+        flag,  # type: Dict[str, value_t]
         print_flags,  # type: bool
-        builtin=_OTHER,  # type: int
+        builtin=0,  # type: int
```
andychu: I usually use `consts.NO_INDEX` for the `builtin_i` variable, which is -1

Technically I think  it is not necessary, since builtins start at `1`, so 0 is fine

But I think it is useful to keep is consistent, in case we refactor things

akinomyoga: Ah, OK. If there is an existing practice, I should follow it. I'll update it.

akinomyoga: [`9bd3658..908e1e8`](https://github.com/oils-for-unix/oils/compare/9bd3658e41a75f484016faebaac70a7e94ba6cb8..908e1e8cbbbe0e16bf09f75151444ba466b33336)

builtin/assign_osh.py
```diff
@@ -19,41 +19,34 @@
 from core import vm
 from data_lang import j8_lite
 from display import ui
-from frontend import flag_util
-from frontend import args
+from frontend import args, consts, flag_util
```
akinomyoga, line 22: Whlie importing `consts` for `consts.NO_INDEX`, I merged those two lines, but was there a reason that those were separated?

andychu: I usually put each FILE on a seprate line, but if there are names within a file, those go on the same line

I don't actually remember why I do that -- I think before auto-formatting it was a bit easier to maintain.

This other style is shorter, which may be nice since we count lines :-)

andychu:
Great thank you!

builtin/assign_osh.py
akinomyoga: OK!

akinomyoga:
Thank you!

PR #2348, opened 2025-07-05
[osh/word_eval] Fix word-splitting bugs

akinomyoga:
Bug 1: word splitting is applied to `${!prefix*}` inside quoted string `"..."`

```console
$ bash -c 'IFS=x axb=1; echo "${!axb*}"x'
axbx
$ bin/osh -c 'IFS=x axb=1; echo "${!axb*}"x'
a bx
```

Bug 2: word splitting is not applied to `${#v}`

```console
$ bash -c 'IFS=5 v=hello; echo [${#v}]'
[ ]
$ bin/osh -c 'IFS=5 v=hello; echo [${#v}]'
[5]
```

Bug 3: word splitting is applied to `${v:-val}` inside quoted string `"..."`

```console
$ bash -c 'IFS=x v=; echo "${v:-axb}"x'
axbx
$ bin/osh -c 'IFS=x v=; echo "${v:-axb}"x'
a bx
```

### DIFF
--- a/spec/word-split.test.sh
+++ b/spec/word-split.test.sh
@@ -861,3 +861,52 @@ for i in ""$A""; do echo =$i=; done
 =def=
 ==
 ## END
+
+
+#### Regression: "${!v*}"x should not be split
+case $SH in dash|mksh|ash|yash) exit 99;; esac
+IFS=x
+axb=1
+echo "${!axb*}"
+echo "${!axb*}"x
+## STDOUT:
+axb
+axbx
+## END
+## N-I dash/mksh/ash/yash status: 99
+## N-I dash/mksh/ash/yash STDOUT:
+## END
+
+
+#### Regression: ${!v} should be split
+v=hello
+IFS=5
+echo ${#v}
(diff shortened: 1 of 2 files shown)


spec/word-split.test.sh
```diff
@@ -861,3 +861,52 @@ for i in ""$A""; do echo =$i=; done
...
+axbx
+## END
+## N-I dash/mksh/ash/yash status: 99
+## N-I dash/mksh/ash/yash STDOUT:
+## END
+
+
+#### Regression: ${!v} should be split
+v=hello
+IFS=5
```
andychu, line 883: oof I didn't think of this case :-)

andychu (commented):
Thank you for finding and fixing these bugs!

andychu:
I like that the code was fairly simple, which means the data structures are probably right

I didn't finish looking at the glob bug -- but the code's control flow looks a bit complex, which makes me think the data structures are wrong

But sometimes it is OK to accept some complexity to move forward, let me look

PR #2349, opened 2025-07-05
[osh/word_eval] Fix behaviors of `\` in unquoted substitutions

akinomyoga:
This PR fixes glob quoting issues.

Bug 1: `'a\b'` shouldn't become `'a\\b'`.

```console
$ bash -c 'v="a\\b"; set -f; echo *$v'
*a\b
$ bin/osh -c 'v="a\\b"; set -f; echo *$v'
*a\\b
```

Bash behavior: `\` should escape the next glob chars in unquoted substitutions, and the single `\` should not match the literal `\`.

```bash
$ mkdir x; touch 'x/a\b' 'x/ab*'
$ bash -c 'v="*\*"; echo x/$v'
x/ab*
$ bin/osh -c 'v="*\*"; echo x/$v'
x/a\b
```

### DIFF
--- a/osh/glob_.py
+++ b/osh/glob_.py
@@ -3,15 +3,10 @@
 import libc
 
 from _devbuild.gen.id_kind_asdl import Id, Id_t
-from _devbuild.gen.syntax_asdl import (
-    CompoundWord,
-    Token,
-    word_part_e,
-    glob_part,
-    glob_part_e,
-    glob_part_t,
-)
-from core import pyutil
+from _devbuild.gen.syntax_asdl import (CompoundWord, Token, word_part_e,
+                                       glob_part, glob_part_e, glob_part_t,
+                                       loc_t)
+from core import pyutil, error
 from frontend import match
 from mycpp import mylib
 from mycpp.mylib import log, print_stderr
@@ -96,6 +91,23 @@ def GlobEscape(s):
     return pyutil.BackslashEscape(s, GLOB_META_CHARS)
 
 
+def GlobEscapeUnquotedSubstitution(s):
+    # type: (str) -> str
+    """For the substitution results such as $v with v='a\*b.txt'."""
+
+    # When glob expansion is not performed, unquoted substitutions need to be
+    # literally preserved in the final result.  A backslash inside unquoted
+    # substitutions should be treated as escaping in glob expansion, yet it
+    # needs to be preserved in the final result without glob expansion.  This
+    # means that a backslash inside unquoted substitutions is treated
+    # differently from quoted backslash, so we need a distinct representaton.
+    # We choose \@ here.
+    #
+    # XXX--This representation is affected by the known IFS='\' bug, but the
+    # bug will be fixed in the coming PR.
+    return s.replace('\\', r'\@')
+
+
 # Bug fix: add [] so [[:space:]] is not special, etc.
 ERE_META_CHARS = r'\?*+{}^$.()|[]'
 
@@ -124,6 +136,7 @@ def GlobUnescape(s):
     unescaped = []  # type: List[int]
     i = 0
     n = len(s)
+    c_backslash = mylib.ByteAt('\\', 0)
     while i < n:
         c = mylib.ByteAt(s, i)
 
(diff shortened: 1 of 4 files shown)


osh/glob_.py
```diff
@@ -96,6 +91,23 @@ def GlobEscape(s):
...
+    # type: (str) -> str
+    """For the substitution results such as $v with v='a\*b.txt'."""
+
+    # When glob expansion is not performed, unquoted substitutions need to be
+    # literally preserved in the final result.  A backslash inside unquoted
+    # substitutions should be treated as escaping in glob expansion, yet it
+    # needs to be preserved in the final result without glob expansion.  This
+    # means that a backslash inside unquoted substitutions is treated
+    # differently from quoted backslash, so we need a distinct representaton.
+    # We choose \@ here.
```
akinomyoga: > the data structures are wrong

Yeah, that was exactly the case. The issue is explained in this code comment. In short, the two backslashes in `'a\b'$v` (with `v='c\d'`) have different behaviors, but they are both escaped as `\\` in the old representation.

akinomyoga (commented):
> *From https://github.com/oils-for-unix/oils/pull/2348#issuecomment-3040339758*
>
> I didn't finish looking at the glob bug -- but the code's control flow looks a bit complex, which makes me think the data structures are wrong
> 
> But sometimes it is OK to accept some complexity to move forward, let me look

osh/glob_.py
```diff
@@ -455,33 +501,37 @@ def _Glob(self, arg, out):
...
+          Number of items appended, or -1 when glob expansion did not happen.
+        Raises:
+          error.FailGlob for fatal failglob error.
         """
+
         if self.exec_opts.noglob():
-            # we didn't glob escape it in osh/word_eval.py
-            out.append(arg)
-            return 1
+            # We cannot reconstruct the original string from the glob pattern,
```
andychu: OK this part is interesting, let me think about whether there can be a different data structure

osh/glob_.py
```diff
@@ -96,6 +91,23 @@ def GlobEscape(s):
...
+
+    # When glob expansion is not performed, unquoted substitutions need to be
+    # literally preserved in the final result.  A backslash inside unquoted
+    # substitutions should be treated as escaping in glob expansion, yet it
+    # needs to be preserved in the final result without glob expansion.  This
+    # means that a backslash inside unquoted substitutions is treated
+    # differently from quoted backslash, so we need a distinct representaton.
+    # We choose \@ here.
+    #
+    # XXX--This representation is affected by the known IFS='\' bug, but the
```
andychu, line 113: OK I see the problem here ... But it seems like we probably should not use `\@` in the long term ??

I guess that's what this comment is indicating

Do we know what kind of representation we will want, or maybe we need to look at IFS=\ more  before figuring it out?

osh/glob_.py
```diff
@@ -455,33 +501,37 @@ def _Glob(self, arg, out):
...
-        # type: (str, List[str]) -> int
+    def Expand(self, arg, out, blame_loc):
+        # type: (str, List[str], loc_t) -> int
         """Given a string that could be a glob, append a list of strings to
-        'out'.
+        'out' when glob expansion happens.
 
         Returns:
-          Number of items appended, or -1 for fatal failglob error.
+          Number of items appended, or -1 when glob expansion did not happen.
```
akinomyoga, line 519: It should also be noted that I changed the meaning of `return -1` of this function (`globber.Expand`).

andychu (commented):
OK thank you, this looks like great progress, and I think it points to one of the **fundamental** problems with the code

I think I would call it the "double backslash escaping" problem

1. `\*` is escaping a glob character
2. and then `\ ` can also escape an IFS character

And then what happens when IFS is `\` ??   It gets VERY confusing

I wonder if we should look at IFS=\ a little  more  before committing  to the `\@` representation ?  

In general,  I wanted to avoid this style of "multi-pass string processing" in Oils, but I think I did this to try to avoid allocations

e.g. processing a single string causes fewer allocations than making a tiny data structure for `\\` `\\\\` `\@` etc.

But it also makes the code more confusing!

osh/glob_.py
akinomyoga: I already fixed the `IFS=\` bug in my local branch based on this glob fix.

Also, I considered different representations and different approaches, but I think this is the most efficient and simple way.

akinomyoga:
> I think I would call it the "double backslash escaping" problem
> 
> 1. `\*` is escaping a glob character
> 2. and then `\ ` can also escape an IFS character
> 
> And then what happens when IFS is `\` ?? It gets VERY confusing

The present fix is unrelated to the double-escaping issue. Even with the single-level escaping for globbing, the current glob escaping has conflicts between the backslash in `echo a'b\c'` and `v='b\c'; echo a$v`.

osh/glob_.py
andychu: OK well I guess I'm saying I would like to change the data structures even more, so the code is simpler, but we might not know what that will be yet

This area is hard, because many things depend on each other!

osh/glob_.py
andychu: Oh cool, is it possible to send a PR for that before this is merged?  I guess you are saying they are not strictly related, but I am curious about it

akinomyoga: I submitted Draft PR #2353 (but this includes also all the changes from #2349 and #2350). The substantial change of #2353 can be checked from [`c45877d..bs-IFS`](https://github.com/oils-for-unix/oils/compare/c45877d6865e65535d25e5499dbfd4634912663c..bs-IFS).

This is actually a re-implementation of the splitter.

andychu:
BTW the tests look good here, and that is the most important part

And the follow-up also fixes IFS=\, which is very important!

I am wondering about the `-1` return value, and the `\@` representation

I would like to use more static **types** rather than string rewriting, but maybe we should wait until later to do that.  Since the tests are very good, I think it can be refactored

---

I am working on producing a list of the failures with OSH in Alpine Linux.  So then we can see how many more bugs there are ... and if there are any more bugs in this same area!

andychu:
OK I looked at this more carefully, it makes sense.  The `@` char is distinct from GLOB_META_CHARS

Actually we may want to add an assert to verify that

----



I noticed GlobUnescapeUnquotedSubstitution only differs from GlobUnescape by one line:

```
             if mylib.ByteInSet(c2, GLOB_META_CHARS):
                 unescaped.append(c_backslash)  # backslash preserved
                 unescaped.append(c2)
```

I will probably merge this as-is, but then refactor it

That may cause some conflicts with the `IFS=\` fix, but it already has some conflicts, so we'll have to address that

---

Thanks for fixing this!

andychu:
OK I fixed the dash/ash assertions - the problem was that the test is using `touch {a,b}.txt`, but dash/ash don't support brace expansion

So I expanded it to `touch a.txt b.txt`, and now dash/ash match bash.    (And also I discovered that ksh and yash both have the divergent mksh behavior you noted!)

Also I simplified the quoting.  I believe it is nicer to write `v='\*'` rather than `v="\*"`, even though they are the same thing, because it is more obvious that the backslash is literal.  

i.e. single quotes never have \ escapes, but double quotes sometimes do

---

I think we can merge this; I may do some refactorings in a separate commit

PR #2350, opened 2025-07-05
[builtin/read_osh] read -n respects backslashes when not -r, and other fixes

akinomyoga:
Bug: `read -n 4` should not repeat `_ReadPortion` even if the previous read ended with incomplete backslash

```console
$ bash -c 'echo "abc\def\ghijxxxx" | (read -n 4; declare -p REPLY)'
declare -- REPLY="abcd"
$ bin/osh -c 'echo "abc\def\ghijxxxx" | (read -n 4; declare -p REPLY)'
declare -- REPLY=abcdefghij
```

Bash behavior: Counting of `read -n 5` should exclude escaping backslash

```console
$ bash -c 'echo "a\b\c\d\exxxx" | (read -n 5; declare -p REPLY)'
declare -- REPLY="abcde"
$ bin/osh -c 'echo "a\b\c\d\exxxx" | (read -n 5; declare -p REPLY)'
declare -- REPLY=abc
```

### DIFF
--- a/builtin/read_osh.py
+++ b/builtin/read_osh.py
@@ -145,8 +145,8 @@ def _ReadN(num_bytes, cmd_ev):
     return ''.join(chunks)
 
 
-def _ReadPortion(delim_byte, max_chars, cmd_ev):
-    # type: (int, int, CommandEvaluator) -> Tuple[str, bool]
+def _ReadPortion(delim_byte, max_chars, allow_escape, cmd_ev):
+    # type: (int, int, bool, CommandEvaluator) -> Tuple[str, bool]
     """Read a portion of stdin.
 
     Reads until delimiter or max_chars, which ever comes first. Will ignore
@@ -157,9 +157,10 @@ def _ReadPortion(delim_byte, max_chars, cmd_ev):
     ch_array = []  # type: List[int]
     eof = False
 
-    bytes_read = 0
+    chars_read = 0
+    backslash = False
     while True:
-        if max_chars >= 0 and bytes_read >= max_chars:
+        if max_chars >= 0 and chars_read >= max_chars:
             break
 
         ch, err_num = pyos.ReadByte(0)
@@ -174,6 +175,16 @@ def _ReadPortion(delim_byte, max_chars, cmd_ev):
             eof = True
             break
 
+        elif backslash:
+            backslash = False
+            if ch == pyos.NEWLINE_CH:
+                continue
+            ch_array.append(pyos.BACKSLASH_CH)
+            ch_array.append(ch)
+        elif allow_escape and ch == pyos.BACKSLASH_CH:
+            backslash = True
+            continue
+
         elif ch == delim_byte:
             break
 
@@ -184,7 +195,7 @@ def _ReadPortion(delim_byte, max_chars, cmd_ev):
         else:
             ch_array.append(ch)
 
-        bytes_read += 1
+        chars_read += 1
 
     return pyutil.ChArrayToString(ch_array), eof
 
@@ -496,28 +507,22 @@ class Read(vm._Builtin):
 
         # Read MORE THAN ONE line for \ line continuation (and not read -r)
         parts = []  # type: List[mylib.BufWriter]
-        join_next = False
-        status = 0
-        while True:
-            chunk, eof = _ReadPortion(delim_byte, mops.BigTruncate(arg.n),
+        chunk, eof = _ReadPortion(delim_byte, mops.BigTruncate(arg.n), not raw,
                                       self.cmd_ev)
 
-            if eof:
-                # status 1 to terminate loop.  (This is true even though we set
-                # variables).
-                status = 1
+        # status 1 to terminate loop.  (This is true even though we set
+        # variables).
+        status = 1 if eof else 0
 
-            #log('LINE %r', chunk)
-            if len(chunk) == 0:
-                break
-
+        #log('LINE %r', chunk)
+        if len(chunk) > 0:
+            join_next = False
             spans = self.splitter.SplitForRead(chunk, not raw, do_split)
             done, join_next = _AppendParts(chunk, spans, max_results,
                                            join_next, parts)
 
             #log('PARTS %s continued %s', parts, continued)
-            if done:
-                break
+            assert done
 
         entries = [buf.getvalue() for buf in parts]
         num_parts = len(entries)
--- a/spec/builtin-read.test.sh
+++ b/spec/builtin-read.test.sh
@@ -773,3 +773,104 @@ status=1
 status=0
 ## END
 ## N-I dash/ash/mksh/zsh stdout-json: ""
+
+#### read -n 0
+case $SH in zsh) exit 99;; esac  # read -n not implemented
+
+echo 'a\b\c\d\e\f' | (read -n 0; argv.py "$REPLY")
+
+## STDOUT:
+['']
+## END
+# ash appears to treat 0 as unspecified
+## OK ash STDOUT:
+['abcdef']
+## END
+## N-I zsh status: 99
+## N-I zsh STDOUT:
+## END
+
+#### read -n and backslash escape
+case $SH in zsh) exit 99;; esac  # read -n not implemented
+
+echo 'a\b\c\d\e\f' | (read -n 5; argv.py "$REPLY")
+echo 'a\ \ \ \ \ ' | (read -n 5; argv.py "$REPLY")
+
+## STDOUT:
+['abcde']
+['a    ']
+## END
+## OK mksh STDOUT:
+['a\x08d\x1b']
+['a ']
+## END
+## OK ash STDOUT:
+['abc']
+['a  ']
+## END
+## N-I zsh status: 99
+## N-I zsh STDOUT:
+## END
+
+#### read -n 4 with incomplete backslash
+case $SH in zsh) exit 99;; esac  # read -n not implemented
+
+echo 'abc\def\ghijklmn' | (read -n 4; argv.py "$REPLY")
+echo '   \xxx\xxxxxxxx' | (read -n 4; argv.py "$REPLY")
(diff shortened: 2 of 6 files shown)


andychu:
OK interesting .... I think the underlying issue might be that there is a matrix of behaviors again:

```
(-r to suppress escapes, not raw) x ( delimiter, -n chars)
```


And this is saying the `-n "not raw"` case does not match bash ...

- in the first case, bash gives you 4 chars, but it reads 5 bytes
- in the second case, bash gives you 5 chars, but reads 9 because of the backslashes

OK I see bash makes more sense than OSH does ...

I will look at the code, thank you!

andychu:
OK maybe a simpler way to say this is that `read -n` did not support `\` escaping at all!

builtin/read_osh.py
```diff
@@ -174,6 +176,16 @@ def _ReadPortion(delim_byte, max_chars, cmd_ev):
             eof = True
             break
 
+        elif backslash:
+            backslash = False
+            if ch == delim_byte:
+                continue
```
akinomyoga: Sorry, I now checked the behavior of Bash again and realized that Bash's behavior is slightly different from this.

akinomyoga: Bash seems to swallow "backslash + newline" (instead of "backslash + delimiter") regardless of `delim_byte`. I pushed a fix 54332fb76.

builtin/read_osh.py
```diff
@@ -157,9 +157,12 @@ def _ReadPortion(delim_byte, max_chars, cmd_ev):
     ch_array = []  # type: List[int]
     eof = False
 
-    bytes_read = 0
+    chars_read = 0
+    ch_backslash = mylib.ByteAt('\\', 0)
+    ch_newline = mylib.ByteAt('\n', 0)
```
akinomyoga: Another thing is the recommended way to get the code of a character. `pyos.NEWLINE_CH` seems to be hardcoded to be 10, but do you assume ASCII (for 7-bit characters)?

andychu: Yes I think we should create `BACKSLASH_CH = 92  # \ ` to be consistent (unfortunately `ord('\\')` will not work as a constant at the top level, but it's OK if we comment it

---

Oils is assuming UTF-8, so `\n` is always 10.  I hope that assumption will not bite us as far as building distros ...

I believe the OpenBSD shell did this - they only support UTF-8

I know this conflicts with libc, which supports multiple encodings (and we still have to fix inconsistency there)

But I think it reduces the number of bugs, and so far no user has complained.  I hope that will continue to be the case!

andychu: It also sorta relates to my philosophy about a "distributed shell"

I hope the same shell script will mean the same thing on different machines, irrespective of global variables like LANG

But this can also cause incompatibility

We might have to support LANG=C and LANG=utf-8, but that shouldn't affect NEWLINE=10 or BACKSLASH=92

akinomyoga: Thank you. I pushed 592aaf460.

spec/builtin-read.test.sh
```diff
@@ -773,3 +773,99 @@ status=1
...
+case $SH in (zsh) exit 99;; esac  # read -n not implemented
+
+echo 'a\b\c\d\e\f' | (read -n 5; argv.py "$REPLY")
+echo 'a\ \ \ \ \ ' | (read -n 5; argv.py "$REPLY")
+
+## STDOUT:
+['abcde']
+['a    ']
+## END
+## OK mksh STDOUT:
```
andychu, line 803: If we are choosing the bash behavior for OSH (and it seems reasonable to), then I think we should have

- `BUG mksh`
- `BUG-2 ash`  - I added BUG-2 and BUG-3 awhile ago, so we can see how many different behaviors there are, at a glance

spec/builtin-read.test.sh
```diff
@@ -773,3 +773,99 @@ status=1
...
+printf '%s\n' 'a b\,c d'   | (read -d ,; argv.py "$REPLY")
+
+## STDOUT:
+['a bc d']
+['a b,c d']
+['a bc d\n']
+['a b,c d\n']
+## END
+# mksh/zsh swallows "backslash + delim" instead.
+## OK mksh/zsh STDOUT:
```
andychu: Likewise I think this is BUG rather than OK

"OK" is usually used if there is some harmless difference, like `\$` vs `'$'`  -- although there are some inconsistencies, I'm sure

spec/builtin-read.test.sh
```diff
@@ -773,3 +773,99 @@ status=1
 status=0
 ## END
 ## N-I dash/ash/mksh/zsh stdout-json: ""
+
+#### read -n 0
+case $SH in (zsh) exit 99;; esac  # read -n not implemented
```
andychu: minor style preference: I usually write `zsh)` instead of `(zsh)` now

For awhile I used `(zsh)`, but now YSH uses the symmetrical style for expressions, like

`case (x) { (42) { echo number } }`

So I try to keep it distinct

builtin/read_osh.py
```diff
@@ -174,6 +175,16 @@ def _ReadPortion(delim_byte, max_chars, cmd_ev):
...
             break
 
+        elif backslash:
+            backslash = False
+            if ch == pyos.NEWLINE_CH:
+                continue
+            ch_array.append(pyos.BACKSLASH_CH)
+            ch_array.append(ch)
+        elif allow_escape and ch == pyos.BACKSLASH_CH:
+            backslash = True
```
andychu, line 185: OK one thing I notice is that in _AppendParts(), we handle the `\` for the `read -d delim` case, with `span_e.Backslash`

Let me see if we can make the cases more consistent, i.e.  `(delim, -n) x (raw, not raw)`

andychu:
Also, since the other shells differ, I think we should document this in `chap-builtin-cmd.md` under `read`

I guess the behavior is that in `read -n`,  `\c` is always counts a single character, including `\\` ?

And for `read -d delim`, `\` is meant to escape the delimiter

builtin/read_osh.py
andychu: Oh now I see there is a `assert done`, so I think the `while True` loop can be removed?  And both return values `done, join_next` are unnecessary?

If so, this change seems like it can simplify the code, which is nice

akinomyoga: Yes, and the `while` loop is removed in the next PR #2353. I just tried to minimize the change.

akinomyoga: Should I remove the `while` loop in this PR?

andychu: If you did it in the next change, it's OK, we can fix it with that change


I made some other comments about the spec test assertions, which I think will not conflict with the next change

Thanks

spec/builtin-read.test.sh
akinomyoga: We can't say this is a bug of mksh because mksh manual says `-n NUM`  is number of bytes (while Bash manual says number of characters). They implement a similar but different feature.

spec/builtin-read.test.sh
akinomyoga: > "OK" is usually used if there is some harmless difference, like `\$` vs `'$'` -- although there are some inconsistencies, I'm sure

OK, so different designs that make sense are also categorized as `BUG`. Then, I think the name `BUG` is confusing. Shouldn't it be renamed to a different name? But then, there is no way to distinguish real bugs and different designs, so I probably want another category. Or, this should probably be `N-I` because it doesn't implement the feature that Bash implements?

spec/builtin-read.test.sh
akinomyoga: I see also `ok-2` and `ok-3` in the tests. What are they? If insignificant differences are `OK`, and significant differences are all BUG, what is the use case for `OK-2`? Maybe, `OK-2` is used to represent a distinct design.

spec/builtin-read.test.sh
akinomyoga: I just copied the existing ones in the same file. I fixed it ([`592aaf4..77d499c`](https://github.com/oils-for-unix/oils/compare/592aaf46067f89840acc640a7faff703e83336fd..77d499caef4680c3abe17100de6678562e245589)). I'll later submit a PR to fix existing ones (**edit**: #2358).

spec/builtin-read.test.sh
akinomyoga: [`77d499c..c36072b`](https://github.com/oils-for-unix/oils/compare/77d499caef4680c3abe17100de6678562e245589..c36072b7cbf60a55e793750ccd41b6aab3ddb690) I decided to use `OK-2` and `OK-3` because those shells behave as they document it. Just designs are different.

spec/builtin-read.test.sh
akinomyoga: [`77d499c..c36072b`](https://github.com/oils-for-unix/oils/compare/77d499caef4680c3abe17100de6678562e245589..c36072b7cbf60a55e793750ccd41b6aab3ddb690)

builtin/read_osh.py
akinomyoga: [`c36072b..7a6fdc3`](https://github.com/oils-for-unix/oils/compare/c36072b7cbf60a55e793750ccd41b6aab3ddb690..7a6fdc3ced21847900a2640939469cb05e7e7f3b)

akinomyoga:
> Also, since the other shells differ, I think we should document this in `chap-builtin-cmd.md` under `read`

007230d47

> I guess the behavior is that in `read -n`, `\c` is always counts a single character, including `\\` ?

Yes.

> And for `read -d delim`, `\` is meant to escape the delimiter

Yes.

andychu:
Thank you!

I think I will look at the Alpine package that failed because of `IFS=\`, I want to confirm with the end-to-end test

The change looks very promising

Issue #2338, opened 2025-07-02
mpfr package build failure in Nix / Alpine - IFS=\ bug


andychu:
[#nix > mpfr package build failure - DLT_OBJDIR](https://oilshell.zulipchat.com/#narrow/channel/307442-nix/topic/mpfr.20package.20build.20failure.20-.20DLT_OBJDIR/with/477522248) 

```
gcc: warning: \\\-DLT_OBJDIR=".libs/: linker input file unused because linking not done
gcc: error: \\\-DLT_OBJDIR=".libs/: linker input file not found: No such file or directory
gcc: warning: \\\-DLT_OBJDIR=".libs/: linker input file unused because linking not done
gcc: error: \\\-DLT_OBJDIR=".libs/: linker input file not found: No such file or directory
gcc: warning: \\\-DLT_OBJDIR=".libs/: linker input file unused because linking not done
gcc: error: \\\-DLT_OBJDIR=".libs/: linker input file not found: No such file or directory
```

This one is HARD, may require rewrite of IFS word splitting - `test/spec.sh word-split`

This is the one Samuel and Aidan worked on last year

PR #2353, opened 2025-07-05
[osh/split] Fix the `IFS='\'` bug of word splitting

akinomyoga:
The first commit "waiting" contains the changes in the other PRs, #2349 and #2350, so it's not the changes of this PR.

The actual changes can be checked in [`c45877d..bs-IFS`](https://github.com/oils-for-unix/oils/compare/c45877d6865e65535d25e5499dbfd4634912663c..bs-IFS).


akinomyoga:
I think the fix is done, but this is still WIP:

- [x] I need to clean up the unused code of the old splitter implementation.
  - I finished it in commit 037b7160a6. I think I'll submit it in a separate PR because it is another big change.
- [x] I want to add tests for the splitting in `read`.

andychu:
OK wow, this looks very promising based on the tests it passes!  

https://op.oilshell.org/uuu/github-jobs/9851/ovm-tarball.wwz/_tmp/spec/osh-py/word-split.html

And the code isn't too long either, that is nice

I may want to manually test it against the original bug, as additional confirmation

https://github.com/oils-for-unix/oils/issues/2338

akinomyoga:
I added tests for splitting in `read` ([`ebdee55..55ab2b2`](https://github.com/oils-for-unix/oils/compare/ebdee55b3cad30decd6f8d9feef6f6885a3a367a..55ab2b20ed5125e7659457ccaa65a3ead435f05b)).

akinomyoga:
The behavior of the word splitting in command arguments is perfect, but there still seems to be a behavior of the `read` builtin different from Bash's. I'll try to fix it.

andychu:
I verified that this fixes the build of the `mpfr4` package!   which used something like `IFS=\`

[osh](https://oilshell.zulipchat.com/#narrow/channel/502349-osh)
>
[Reproducing mpfr4 build failure](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/Reproducing.20mpfr4.20build.20failure/with/527505399)


Thank you!

I will look in more detail at this PR, at the glob fix PR

It might take awhile, since I am also kicking off these Alpine Linux builds for the 4th  grant planning -- I want to see if we can make it easier to **connect** the spec tests and the actual build failures in the wild, to **prioritize**

This one was important because it's both a POSIX incompatibility and something that prevented us from running real shell scripts!

andychu:
BTW I published the results of the Alpine Linux builds, and this `IFS=\` bug is by far the most common

https://pages.oils.pub/

e.g. if you look at `error grep` here

https://op.oils.pub/aports-build/2025-07-29-shard5.wwz/_tmp/aports-report/2025-07-29-shard5/diff-baseline.html

then you will often see these `\\\-D` bugs with OSH

https://op.oils.pub/aports-build/2025-07-29-shard5.wwz/_tmp/aports-report/2025-07-29-shard5/error/libidn2.txt

----

I think you said this depends on #2349, so I need to look at that more !

akinomyoga:
The C++ version is failing with

```
osh: mycpp/gc_str.cc:394: BigStr* BigStr::rstrip(BigStr*): Assertion `len(chars) == 1' failed.
```

Is this an intentional design of mycpp?

andychu:
Thanks for revising this!  I will look more carefully

Ah yes I think we only support one character for `rstrip()` right now, because we don't use this capability elsewhere.  I will see if we should add it to mycpp, or if we adjust the Python code

> self.args[-1].rstrip(self.ifs_space)

osh/split.py
```diff
@@ -317,3 +321,128 @@ def Split(self, s, allow_escape):
...
+            if self.state == state_i.DE_White1:
+                self._GenerateWord()
+            self.char_buff.append(byte)
+            self.state = state_i.Black
+
+    def PushTerminator(self):
+        # type: () -> List[str]
+        if self.state in (state_i.DE_White1, state_i.Black):
+            self._GenerateWord()
+            if self.max_split >= 0 and len(self.args) == self.max_split + 2:
```
andychu, line 444: Hm why do we need to "fix this up" here?  Is it possible to avoid calling _GenerateWord() in some cases?  That is what appends to self.args

But then it looks like we MAY have TOO MANY args after calling _GenerateWord() ?  So we have to do some joining

If it's not possible to avoid, we should have a comment explaining why

andychu (commented):
This looks great, thank you

Folding in max_split does look like cleaner code

But then I wonder about the part at the end where we may fix up `args[-2]` and `args[-1]`

andychu:
(and by the way, I merged with master, so you should  pull again if you plan to make edits)

osh/split.py
```diff
@@ -317,3 +321,128 @@ def Split(self, s, allow_escape):
             i += 1
 
         return spans
+
+
+class IfsSplitterState(object):
+
```
andychu, line 327: I know you said you would like to write a blog post about this algorithm, which I appreciate

I think it would be nice to summarize in the docstring `"""` here, perhaps something like:

- This algorithm is used for 3 purposes
  - word splitting of unquoted $x
  - the `read` builtin, which also requires allow_escape (read -r) and max_split (read a b c)
  - the YSH shSplit() function
- How to use it?
  - call PushLiteral for constant parts like `--flag=` in `--flag=$x`, and PushFragment for `$x` I think
  - then call PushTerminator
- Internal state
  - the state variable: `state_i.{Start,DE_White,Backslash,Black}`
  - self.args - the words that will eventually be returned by PushTerminator()
  - self.frags - used by PushLiteral?  and _FlushCharBuff()
  - self.char_buff 
  - self.white_buff - I am not clear on the difference between char_buff and white_buff, maybe they could be renamed?  It says white_buff is for max_split?

andychu (commented):
I looked more closely at this

The general structure looks nice and clean

I think some brief comments about the details would help (and then later a blog post would help people who have less background with shell, e.g. some people don't  know you can do `--flag=$x` and `--flag="$x"` and so forth)

andychu:
I filled in the C++ part, and all the tests pass.  (The C++ API is now a bit inconsistent, but that's not the main issue and it's something we can improve later, if necessary)

I am going to test this out on Alpine Linux with `regtest/aports`

I would still like to read an article about the algorithm and how you developed it, and link it from the source code, but it's important to test it too

Thank you!

akinomyoga:
Sorry for the delay. This PR has been based on #2349, but after #2349 is merged, you modified the code of #2349 in commit 0f31a51ada694e4928f6054da6214b3c96317cb4. As far as I remember, there was a reason for how I implemented #2349, and I think the simplification applied in 0f31a51ada694e4928f6054da6214b3c96317cb4 should have broken it, although it was not covered by the spec test cases. Then, I didn't have time to re-examine #2349 before working on this PR. I'm now busy and also am going to be busy for the next few months, so I don't have time to do the investigation again.

osh/split.py
akinomyoga: This is because it is the Bash behavior. When `read` requests $n$ words, and there are $n+m$ words, the extraction rule for the first $n$ words and the rest $m$ words are different. We need to read the $`n`$th word and the extra words under different rules, where we need to keep them separately. I'll explain it in a blog post.

akinomyoga:
More specifically, I need to re-examine the impact and consistency on 0f31a51ada694e4928f6054da6214b3c96317cb4 and 9041a6f8d835b7b1a3501889e6ec045a5e65b0c5.

akinomyoga:
> I would still like to read an article about the algorithm and how you developed it, and link it from the source code, but it's important to test it too

Where am I supposed to submit the blog post? Is there a repository for it?

andychu:
Hm regarding 

https://github.com/oils-for-unix/oils/commit/0f31a51ada694e4928f6054da6214b3c96317cb4

We have two functions GlobUnescape() and GlobUnescapeBackslash() now

https://github.com/oils-for-unix/oils/blob/master/osh/glob_.py#L169

I was thinking that GlobUnescapeBackslash() is equivalent to `s.replace('\\@', '\\')` now, i.e. the whole function can be one line

But I don't understand why exactly that works.  So I agree there is some potential for more problems

---

But these 2 changes are BIG progress, so  we can go back to it when we have time.  And also I am doing a bigger round of testing right now, with `regtest/aports`.  We'll see how many bugs it fixes

I had thought about trying to "rewrite the whole thing" since I think it's a bit confusing, but I think what you did helped us make progress, and fixed many many bugs

---

If I were to try to rewrite it, I think that using more static types rather than `\@` could help, but there are possibly some  challenges with using `mycpp` that way.   

i.e. it would be nice if we had different static types for strings in different stages of escaping / unescaping, but I'm not sure if that's possible

But it also might be better just to fix more bugs as we encounter them, with the existing style

andychu:
Regarding the blog post, you could put the post on your own site and I can link to it, and circulate it

Or I can also send you an invite to the `oils.pub`  repo.  It is mostly a plain markdown file, though there might be quirks in the toolchain

Whichever you prefer is fine

---

There is also the possibility of putting something in `doc/`, which is published with each release.  But I think it helps to have a "first person" account too, more like a story than a doc

Thank you again!

Issue #2307, opened 2025-05-21
simple command argv should be evaluated BEFORE redirects


andychu:
Failing spec test cases here:

https://github.com/oils-for-unix/oils/commit/4ee4eb3e01b560fdf32f85f30dce475c810bbc57

andychu:
Repro:

`test/spec.sh redir-order`

This is sort of a "meaty" fun bug I think ... it will test your knowledge of the core Oils interpreter

Requiring changes in `frontend/syntax.asdl` I believe

andychu:
I'm closing bugs marked "Bug Fixed in Code", but they are still on our Kanban board, so we can write about them.  The writing is arguably more important than the fix!

Issue #20, opened 2017-07-30
errors with make SHELL=osh


andychu:
Reported here: 

https://www.reddit.com/r/oilshell/comments/6p2tp0/please_try_the_first_osh_release/dktf71s/?utm_content=permalink&utm_medium=front&utm_source=reddit&utm_name=oilshell

Steps:

    make SHELL=osh
    osh install

This is actually two bugs: typo in install, should be `>&` not `<&`.  And we should implement `<&`.

PR #2363, opened 2025-07-09
Evaluate command.Simple argv before redirects

PossiblyAShrub:
Fixes: https://github.com/oils-for-unix/oils/issues/2307

Need to clean-up a bit more, but this should fix the issue. Want to check its impact on other tests first.

---

### TODO List

- [x] Make CI Green
- [x] Clean-up the fix
  - [x] Address/test TODO comments
  - [x] Can we reduce return value count in `_DoRedirects`? (also could have a better name)
  - [ ] Cleaner way to report the `io_errors` in `_DoSimple` and `_DoRedirect`

### DIFF
--- a/osh/cmd_eval.py
+++ b/osh/cmd_eval.py
@@ -1090,39 +1090,53 @@ class CommandEvaluator(object):
 
         run_flags = executor.IS_LAST_CMD if node.is_last_cmd else 0
 
-        # NOTE: RunSimpleCommand may never return
-        if len(node.more_env):  # I think this guard is necessary?
-            if self.exec_opts.env_obj():  # YSH
-                bindings = NewDict()  # type: Dict[str, value_t]
-                with state.ctx_EnclosedFrame(self.mem, self.mem.CurrentFrame(),
-                                             self.mem.GlobalFrame(), bindings):
-                    self._EvalTempEnv(node.more_env, 0)
+        #print('n', node.redirects)
+        status, redirects, io_errors = self._DoRedirects(node.redirects or [])
+        #print(status, redirects, io_errors)
+        if status != 0:
+            return status
 
-                # Push this on the prototype chain
-                with state.ctx_EnvObj(self.mem, bindings):
-                    status = self._RunSimpleCommand(cmd_val, cmd_st, run_flags)
+        #print(redirects)
+        with vm.ctx_Redirect(self.shell_ex, len(redirects), io_errors):
+            # NOTE: RunSimpleCommand may never return
+            if len(node.more_env):  # I think this guard is necessary?
+                if self.exec_opts.env_obj():  # YSH
+                    bindings = NewDict()  # type: Dict[str, value_t]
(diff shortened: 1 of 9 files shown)


andychu:
Awesome, thanks for doing this, and for sending the post mortem

This looks right - the tests pretty much show that!

spec/toysh-posix.test.sh
```diff
@@ -10,7 +10,7 @@ abc=${a?bc} echo hello; echo blah
 ## stdout-json: ""
 
 #### setting readonly var (bash is only one where it's non-fatal)
-# http://landley.net/notes.html#20-06-2020
+# https://landley.net/notes-2020.html#12-06-2020
```
andychu: Hm  adding `-2020.html` is right, but the anchor still starts with `#20-06`

https://landley.net/notes-2020.html#20-06-2020

osh/cmd_parse.py
```diff
@@ -385,9 +386,12 @@ def _MakeSimpleCommand(
     more_env = []  # type: List[EnvPair]
     _AppendMoreEnv(preparsed_list, more_env)
 
+    if not len(redirects):
```
andychu: style nitpick: I think this  reads better as `if len(redirects) == 0` , although it's the same thing

osh/cmd_eval.py
```diff
@@ -1724,18 +1740,20 @@ def _DoTimeBlock(self, node):
 
         return status
 
-    def _DoRedirect(self, node, cmd_st):
-        # type: (command.Redirect, CommandStatus) -> int
+    def _DoRedirects(self, redirects):
```
andychu: I think this function could use a different name ... now it is a helper function for

- `_DoRedirect()` - which is part of the big `command_t` switch, for the `command.Redirect`
- `_DoSimple()` - the `command.Simple` case

There is _EvalRedirect

I guess this is NOT just _EvalRedirects

Maybe we can call it `_EvalAndApply(redirects)` or something

I think the `_Do*` names are better for the big _Dispatch() switch

andychu: Oh now I see it could be `_EvalAndPushRedirects`  - because it only does the "first half", not the undoing of the redirects



(what annoys me a bit is that the API is not a symmetric Push/Pop because of error handling, but that is separate from this change!)

andychu (commented):
This looks excellent, thank you!  Just cosmetic comments

andychu:
And yeah I think the `io_errors` is OK for now ... the API was always a bit asymmetrical, but it is correct

PossiblyAShrub:

andychu:
Great thank you!

I changed it from `List[Redir]` to `List[Redir]?`, which should translate to `Optional[List[Redir]]` in MyPy

That is NOT checked right now, but I guess it's good for documentation.  (and eventually we could check it, though it probably isn't high priority)


---

I think  an interesting bug, and it is funny that shell scripts depend on it -- the `imap` package with the OSCFLAGS code

Issue #2376, opened 2025-07-24
Nested io.stdin loops leak lines


PossiblyAShrub:
Example case:

```
proc inner {
  printf 'a\nb\nc\n' | for x in (io.stdin) {
    echo x=$x
  }
}

printf 'x\ny\n' | for _ in (io.stdin) {
  inner
}
```

I would expect the output to be:

```
x=a
x=b
x=c
x=a
x=b
x=c
```

However, it is:

```
x=y
x=a
x=b
x=c
```

It seems like the "y" leaks into the inner `io.stdin`.

This may be related to https://github.com/oils-for-unix/oils/issues/2356

andychu:
Hm yeah I think that is very related, something probably needs to change

`io.stdin` is basically modelled after `sys.stdin` in Python, but Python doesn't have pipelines ... so that complicates things

---

Or another way to put it is that buffered I/O isn't available AT ALL in shell, so adding it has some unexpected consequences ... it doesn't "compose" with other shell features like pipelines

bar-g:
Yes, looks very related. Somehow, `io.stdin` doesn't seem to be bound to the stdin of its specific pipeline or redirection block (frame?).

Could look like it would set up only one global stdin buffer which then feeds into any reading command. [EDIT: regardless of the commands specific pipeline or any redirection context.]

Issue #2356, opened 2025-07-06
iterating over `io.stdin` twice has unexpected  results


bar-g:
[#oil-help > io.stdin scope issue? or rationale+workaround?](https://oilshell.zulipchat.com/#narrow/channel/264891-oil-help/topic/io.2Estdin.20scope.20issue.3F.20or.20rationale.2Bworkaround.3F/with/525932785)

When using buffered io.stdin to read from stdin within a pipe, any remaining pipeline-stdin-buffer should be closed together with the terminated pipeline.

Currently the buffer persists and seems to only get discarded after EOF, causing problems like unexpected results and data-loss on subsequent input at a later pipeline's stdin:

```
# two lines written, only the one header line read
ysh-0.31$ echo $'a \n b' | { for i, line in (io.stdin) { echo $i: $line; break }; }
0: a 

# adding two more lines,  and process only one line
ysh-0.31$ echo $'c \n d' | { for i, line in (io.stdin) { echo $i: $line; break }; };
0:  b

# previous data lost when adding two more, and now going through all
ysh-0.31$ echo $'e \n f' | { for i, line in (io.stdin) { echo $i: $line; } }
0: e 
1:  f
```

related:
manual close for `io.stdin` #2355
fopen/ redir blocks to auto-close io.stdin buffers #2357

bar-g:
Of course only close the io.stdin buffers reading within the pipe, not any other buffer, e.g. a buffer with paused reading within an enclosing, outside `redir { ... }` block.

bar-g:
Hm, might this even be a security question? Considering the buffer overflow into the second pipeline in the example, and that it could possibly evaluate based on the input.

bar-g:
So you want to discuss here after closing the other issues?

I intentionally posted all three issues for discussion ahead of filing, but with no response on zulip:
[#oil-help > io.stdin scope issue? or rationale+workaround? @ 💬](https://oilshell.zulipchat.com/#narrow/channel/264891-oil-help/topic/io.2Estdin.20scope.20issue.3F.20or.20rationale.2Bworkaround.3F/near/526777129)

### fopen/ redir blocks to auto-close io.stdin buffers #2357

the code from zulip that shows the buffer is not closed after running the block:
`echo $'a \n b' | fopen { for line in (io.stdin) { echo $line; break } }; for line in (io.stdin) { echo $line; break }`


### manual close for io.stdin buffer #2355
[...]

PR #2377, opened 2025-07-24
Add spec test for next io.stdin bug (#2376)

PossiblyAShrub:


andychu:
Thanks, how about adding a case for https://github.com/oils-for-unix/oils/issues/2356   as well ?

The metadata for allowed failures should also be updated

I'm not sure what we will do about this, but it is useful to have these handy

PossiblyAShrub:
Done. This is ready for another review.

andychu:
Great thanks!

(btw having some kind of blog post for spec tests would be useful and on topic ... I wrote about it a long time ago, but it's probably useful to new contributors to have an updated version, and another viewpoint.  Or it can also go in the contributor guide)

PR #2381, opened 2025-07-30
eggex.md add collected basic matching info

bar-g:


bar-g:
hope this branching thing works correct now

bar-g:
ok, done.
Didn't see at first that the second zulip thread touched the same document

bar-g:
should be fine now

bar-g:
Realized "line breaks" is a common term, e.g. https://www.regular-expressions.info, added that with having a unifyed character listing.

doc/eggex.md
```diff
@@ -118,13 +118,27 @@ Constructs like `. ^ $ \< \>` are deprecated because they break these rules.
...
-depends on the underlying regex library.
+The `dot` (wildcard) primitive matches "any" single character in whatever way
+the underlying regex library interprets this.
+Differences usually only concern special characters:
 
-- YSH uses `libc`, which accepts POSIX ERE syntax.  So `dot` aka `.` matches
-  any character, unless the `reg_newline` flag is true.
-- If Eggex were compiled to Python, `dot` aka `.` matches any character
-  *except* a newline, unless the `re.DOTALL` flag is true.
+**NUL-character:** `\0` (c-string), `\x00` (c-string/eggex) or `\y00` (j8)
```
andychu, line 125: I don't think we need to mention that `\0`  is the spelling for NUL

Although it is a good point that it's `\x00` in eggex, and `\y00` in strings ...  I guess we are also following the rule that we match the underlying regex library

```
ysh-0.34$ = / [ \x00 ] /
<Eggex 0xb44>
```

Oh yeah this is sort of interesting, I wonder if it is a bug

```
var p = / [ \x00 ] /

ysh-0.34$ = ${p}
(Str)   b'[\u{0}]'
```

It is translating x\00 to  the `\u{0}` syntax which seems wrong ?

doc/eggex.md
```diff
@@ -508,6 +522,8 @@ However, cases like this are a fatal runtime error:
   - `(?:noncapturing)`
 
 With egg expressions, each construct has a **distinct syntax**.
+Even if not all constructs are supported by every library.
```
andychu, line 525: I think it might be worth having a separate section about "compiling to libc ERE" or something

andychu:
I'm looking into the bug here

[#oil-dev > Eggex translation of \x00](https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/Eggex.20translation.20of.20.5Cx00/with/532332135)

doc/eggex.md
bar-g: > It is translating x\00 to the \u{0} syntax which seems wrong ?

Could it instead be translated to something universal, like `\0`, as accepted for passing args to awk/grep/sed? And also accept that same escape in places where ysh accepts eggexes internally?

(The approach to translate eggexes to verbatim newline and NUL chars doesn't work for passing regexes to external commands anyway.)

> [#language-design > eggex: matching newline @ 💬](https://oilshell.zulipchat.com/#narrow/channel/384942-language-design/topic/eggex.3A.20matching.20newline/near/528590705)
> this looks really like a problem with current eggex translations to me.

andychu:
BTW I looked into this area more, and there are some bugs with OSH that need to be addressed first

- Oils should call setlocale() to respect ENV, which is usually UTF-8, and NOT rely on GNU readline to do so
- I believe that NUL has to be disallowed altogether, because libc ERE doesn't have  char escapes
- likewise, bytes above 127 are a bit problematic, as they would assume utf-8
  - the safest thing would be to allow only bytes 1 to 127
- eggex should not use `\x01` syntax; it should use unambiguous `\y01` (byte) or `\u{1}` (code point) syntax

andychu:
This is subsumed by https://github.com/oils-for-unix/oils/commit/8a44dc82e87c338d9153bbe381e38a502d48b525

PR #2382, opened 2025-07-30
chap-type-method: refine "notes on whitespace"

bar-g:


doc/ref/chap-type-method.md
```diff
@@ -246,9 +246,9 @@ that pattern is removed if it matches the start or end.
 
 #### A note on whitespace
 
-When stripping whitespace, Oils decodes the bytes in string as utf-8
-characters. Only the following Unicode codepoints are considered to be
-whitespace.
+When ***trimming*** whitespace (the default multi-line `/space/` or `blank`),
```
andychu: I think stripping => trimming is a good change

The `/space/` and `blank` phrasing is confusing

andychu:
I think it might be more helpful to add a separate scription of what the default behavior is exactly

Right now it says: "With no arguments, whitespace is removed."

---

Also I am not sure why it's surprising that search/replace don't treat whitespace specially, but trim does?

doc/ref/chap-type-method.md
bar-g: Hm, the behavior seems a bit unclear, and in need of better doc.

I understood you in the docs an zulip as if `/space/` would be identical to the default, and the default does trimm multiple newlines. Whereas `/blank/` is known not to match newlines.

However, trying it, `/space/` actually works different then the default, only trimming a single newline:

```
ysh-0.34$ = u' a b \n\n'.trim()

(Str)   'a b'
ysh-0.34$ = u' a b \n\n'.trim(/space/)

(Str)   b'a b \n'

```

Trimming `/blank/` works as expected, not matching newlines.

```
ysh-0.34$ = u' a b \n\n'.trim(/blank/)

(Str)   b'a b \n\n'

```

bar-g: ah, it's `/space*/` ?

bar-g:
Is the default `trim(/space*/)` ?

What is the difference in whitespace treating between trim and search/replace ?

bar-g:
Are threre even more differences than non-breaking whitespace consideration?

---
Only considering basic unicode definitions seems common for all stripping (trimming and search/replace).
But considering non-breaking whitespace is only valid for trimming, not for searching whitespace.

Whith the changed note only talking about "trimming" instead of "stripping" (which would includes search/replace) it's already a bit clearer implicitly, but mentioning the difference is what would clearly document it.

doc/ref/chap-type-method.md
bar-g: Ok, done consolidating the text to current understanding.

bar-g:
OK, I think I also added the correct utf-8 encodings, thank to Aidan's hint.

A maybe separate idea: Isn't  `matchLeft()` basically `search()` with some implicit prefix concatenation of  `/ %start / `? Could be a good explaining example, if you'd have one?

bar-g:
Ok, added info on fatal utf-8 errors from zulip thread.

bar-g:
ok, tested and showing: differently needed byte-code escape examples

bar-g:
Oh, please note that listing the different escape syntax may seem a bit convoluted, but I found that really necessary. 

It's listing the ways it works in Oils, which is very confusing if encountering errors when some escape that had just worked somewhere now doesn't work in another context.

andychu:
I updated it and made it more concise ... now I see your note about the UTF-8 encoding of say `\u{a0}`, which I removed, because  it didn't belong in this section

I think we could clarify that in the `OILS-ERR-*` for the decoding errors

But I am not sure where exactly it comes up, and what the error looks like

andychu:
I see

```
ysh-0.34$ = b'\yff'.trim()
  = b'\yff'.trim()
  ^
[ interactive ]:4: fatal: UTF-8 decode: Bad encoding at offset 0 in string of 1 bytes
```

But I don't see the need to list all the encodings like


```
`u'\u{a0}'`,`b'\yC2\yA0'`,`$'\xC2\xA0'`,`/ \xC2\xA0 /`
```

I think we can put create a separate section about "syntax synonyms" or something

It could be in a "strings" doc or something

PR #2383, opened 2025-07-30
chap-expr-lang: note that repeated concatenation with ++ is slow

bar-g:


doc/ref/chap-expr-lang.md
```diff
@@ -344,6 +344,8 @@ and `List` objects:
     ysh$ = M
     (List)   ["one", "two", "three", "4"]
 
+String concatenation is slow, list appending is faster for looping.
```
andychu: I think it would be clearer to say something like:

"Note: Avoid repeated string concatenation in a loop.  It's algorithmically faster to append to a list, and then join() at the end."

Actually I think we couild put that in `doc/ysh-faq`, and then **link** to it from the reference.

bar-g: Ok, added a note like that, and made following examples locally understandable (no backreference).

Also noticed some `ysh$ ` prompts at some examples, and removed them to unify all examples in the doc.

Seams you would have to wheigh in some actual question, is it really a basic FAQ?
> https://github.com/oils-for-unix/oils/pull/1829#issuecomment-1969806225
> andychu commented on Feb 28, 2024
> I prefer putting most information in doc/ref. Then if you think it's hard to find, write a short FAQ entry that > describes the actual question you had, and link directly to doc/ref

andychu:
Hm this is a bit hard to review because it also removes a lot of `ysh$` stuff ... I think we are a bit inconsistent about examples, but that is a separate discussion from `++` in a loop

bar-g:
Yeah, a bit, but still possible to check through the `ysh$` removals.
For the output I now changed it to the single/multi-line style of the ysh-tour throughout.

Search for "Notes:" to get near to the adapted notes.

andychu:
I added a FAQ here : https://github.com/oils-for-unix/oils/commit/a07ef4585ec1bf0ec8fa01bf6a4c90172b59e39f

PR #2386, opened 2025-08-10
Use `mktemp -d` when possible in configure script

quexxon:
Because `mktemp -d` is near universally supported by Unix-likes, use it to create a random per-execution directory under ${TMPDIR:-/tmp}. This closes the vector for a well-known attack where pre-known filepaths in /tmp are modified. Before this change, the caller would need to set TMPDIR manually to avoid the risk.

Ref: [#oil-dev > temp dir provisioning in configure script](https://oilshell.zulipchat.com/#narrow/channel/121539-oil-dev/topic/temp.20dir.20provisioning.20in.20configure.20script)

### DIFF
--- a/configure
+++ b/configure
@@ -15,7 +15,7 @@
 # - Other settings: LTO, PGO?  Consider moving prefix, LTO, PGO to build and
 #   install steps.
 
-TMP=${TMPDIR:-/tmp}  # Assume that any system has $TMPDIR set or /tmp exists
+TMP="$(mktemp -d || echo "${TMPDIR:-/tmp}")"
 readonly TMP  # POSIX sh supports 'readonly'
 
 log() {


configure
```diff
@@ -15,7 +15,7 @@
 # - Other settings: LTO, PGO?  Consider moving prefix, LTO, PGO to build and
 #   install steps.
 
-TMP=${TMPDIR:-/tmp}  # Assume that any system has $TMPDIR set or /tmp exists
+TMP="$(mktemp -d || echo "${TMPDIR:-/tmp}")"
```
andychu: I think TMPDIR should come first, so something like

    TMP=${TMPDIR:-"$(mktemp -d || echo /tmp)"}

TMPDIR is supported by many applications, e.g. in theory you could redirect all temp files (not just Oils) to `/mnt/otherdisk/tmp` if you were doing some kind of huge build

Probably 99.9% of people don't set that, so the common case will still be `mktemp -d`

quexxon: `mktemp -d` is one such application that supports `TMPDIR`. The temporary directory will be created under a user-specified base directory (not applicable here) or `$TMPDIR` or `/tmp` in that order.

andychu: Ah OK I didn't realize that!  Looks like it is pretty standard (busybox and GNU)

So let's remove our own TMPDIR, and add a comment to remind ourselves, in case anyone wants to override it:

```
# note: mktemp -d respects $TMPDIR
TMP=$(mktemp -d || echo /tmp)
```

quexxon: If we remove our `$TMPDIR`, won't we then fail to honor TMPDIR in the rare case that `mktemp -d` is not available?

PR #2387, opened 2025-08-15
[chap-type-method.md] fix leftMatch intro sentence

bar-g:


doc/ref/chap-type-method.md
```diff
@@ -307,9 +307,9 @@ The `%start` or `^` metacharacter will only match when `pos` is zero.
 
 ### leftMatch()
 
-`leftMatch()` is like `search()`, but it checks
+`leftMatch()` is like `search()` with an implicit `%start` constrain
```
andychu: `constrain` should be `constraint`

But how about saying 

`leftMatch()` is like `search()`, but the pattern must match at the beginning of the string.  (This is similar to `%start`)

doc/ref/chap-type-method.md
```diff
@@ -307,9 +307,9 @@ The `%start` or `^` metacharacter will only match when `pos` is zero.
...
 
-    var m = 'hi world' => leftMatch(/[aeiou]/)  # search for vowels
+    var m = 'hi world' => leftMatch(/[aeiou]/)  # search first-character vowels
```
andychu: how about `# is the first char a vowel?`

It is awkward to use the word "search", when we're saying that leftMatch() is different than search()

andychu (commented):
Thanks, good catch!

doc/ref/chap-type-method.md
bar-g: yeah, true. Think it's just `# match any first char vowel`.

doc/ref/chap-type-method.md
bar-g: Hm, combining both: ``, but the pattern must match at the beginning of the string. (Implicit `%start` constraint.)``

doc/ref/chap-type-method.md
andychu: hm "match any first char vowel" does not seem like grammatical english

"match if the first char is a vowel"  is better

PR #2388, opened 2025-08-15
document some results from zulip thread

bar-g:
https://oilshell.zulipchat.com/#narrow/channel/447377-api-design/topic/A.20note.20on.20whitespace

### DIFF
--- a/doc/unicode.md
+++ b/doc/unicode.md
@@ -11,11 +11,18 @@ Notes on Unicode in Shell
 
 ## Philosophy
 
-Oils is UTF-8 centric, unlike `bash` and other shells.
+Oils is UTF-8 and JSON centric, unlike `bash` or other C-string and UNIX centric shells.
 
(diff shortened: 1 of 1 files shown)


doc/unicode.md
```diff
@@ -11,12 +11,19 @@ Notes on Unicode in Shell
 
 ## Philosophy
 
-Oils is UTF-8 centric, unlike `bash` and other shells.
+Oils is UTF-8 and JSON centric, unlike `bash` or other C-string and UNIX centric shells.
```
andychu, line 14: the JSON part doesn't really make sense; it is potentially confusing because JSON has elements of both UTF-8 and UTF-16

andychu (commented):
I don't think this is an improvement

I think the main things we need to do with unicode are:

- support LANG= ... (already mentioned)
- disallow \x01 in Eggex, which is ambiguous.    It should be `\y01` or `\u{1}` like in the rest of YSH, which is unambiguous

PR #2390, opened 2025-08-19
[doc/ref] Clarify regex search() vs. leftMatch()

andychu:
There is a quirk because ^ is used to implement leftMatch().  This is different than Python, but unavoidable due to the underlying regex engine, and harmless.


doc/ref/chap-type-method.md
```diff
@@ -311,49 +311,58 @@ Respects unicode.
...
 
 ### leftMatch()
 
-`leftMatch()` is like `search()`, but the pattern must match at the beginning
-of the string.
-
-    var m = 'hi world'.leftMatch(/[a e i o u]/)
-    = m           # => Null because 'h' is not a vowel
+Test whether the regex matches the string at the given `pos`, which is `0` by
+default.  Returns a [Match](#Match) value, or `null` if it doesn't match.
```
bar-g: I think I would also mention the relation to search here, instead of in the notes below.
So altogether something like this:

Check whether a regex matches the given string starting at `pos`, which is `0` by
default.  Returns a [Match](#Match) value, or `null` if it doesn't match.

`leftMatch` does regex matching like `search`, but *only* at the specified position. It does not search any further.

doc/ref/chap-type-method.md
```diff
@@ -311,49 +311,58 @@ Respects unicode.
...
 
-    var lexer = / <capture digit+> | <capture space+> /
+    var m = 'hi world'.leftMatch(/[a e i o u]/, pos=3)
+    = m           # null, because 'w' is not a vowel
 
-See [YSH Regex API](../ysh-regex-api.html).
+`leftMatch()` and `pos` can be used to implement iterative lexers.  See [YSH
+Regex API](../ysh-regex-api.html).
+
+Notes:
```
bar-g, line 362: Moving all of the clarification of search() vs. leftMatch() up-front, will allow to focus on python vs. ysh things here:

Notes:
YSH's `leftMatch()` vs. `search()` is mostly inspired by Python's `re.match` vs. `re.search`, however:
- `leftMatch` is implemented adding `^` to the regex under the hood.
- `leftMatch` does not override the `reg_newline` option of eggexes.

This means:
- With `reg_newline` set, `leftMatch()` actually *also matches on additioanal fixed positions*.
- It's possbible to use an eggex with the `reg_newline` option set, to check if it matches a string at the specified position on the first line, *or* at the beginning of any following line. (With just a sigle `leftMatch()`  call.)
- Eggexes with `reg_newline` are a speciality of `leftMatch()` to be used with care.

andychu:
OK thanks, I added a note about reg_newline

I also found a bug with leftMatch() using ^ and alternation |

which I will fix

bar-g:
Great, I think you did improve the leftMatch intro even further.

But I just can't quite empathize with "reference docs" that produce vagueness (the "notes"), instead of pointing out the facts and what that means (docs allowing for a more competent understanding).

doc/ref/chap-type-method.md
bar-g: I find this style allows for better understandig.

Issue #324, opened 2019-05-31
history should not behave like a "two-level store"


andychu:
The thread over in #320 made me realize what is wrong with bash history.  It has two parts:

- the history in the current shell process
- the persisted history in the `~/.bash_history` file

When you exit a process, it appends to the history file `shopt -s histappend`

This is really confusing because most people use multiple shells at once!

--------

It's not clear what the solution is, but ideas are:

- History could be stored in a unique file keyed by (PID, timestamp).  Then all of them would have to be searched for recent commands?)
- use an sqlite database (don't want this to be a required dependency)
- try doing file system tricks for atomic writes (note that network file systems don't always obey POSIX)

KingMob:
I think keying history by pwd is not bad. Realistically, storing PID and pwd would allow us to do nice things like create a more contextual history. We can prioritize recent and pwd-relevant commands first. We could do history completions based on that, etc. Just spitballing.

The whole idea of a linear list of commands in a text file feels really old. I understand not wanting to add sqlite as a dep, but could a db allow us to store ALL (or at least a lot more) history efficiently?  E.g., just 500 commands consumes 18 kb in my .bash_history; what if they were stored in a db, compressed, deduped, store invocation times separately from the command itself, etc.

----

OTOH, that's a lot of complication, I admit.

Here's a simpler suggestion: the bash docs say that if you set HISTTIMEFORMAT, every command gets a timestamp added as a history file comment. We could probably add other metadata as comments if you want to keep using a similar file format.

PR #2396, opened 2025-08-24
Add more spec tests for history builtin and associated envvars

KingMob:
I've been pretty busy with my real job, but I was able to get some time to put together the necessary history tests for my next steps with histappend, and some extras, too.

## Stupid bash behavior

So, shell option `histappend` is tricky to test, because bash usually appends _anyway_. At some point shortly after people had the resources to run multiple shells on the same machine, they realized overwriting the history file on exit was a bad idea.

The actual bash logic when histappend is unset is to compare the number of new commands in the session with the size of the in-mem history. If the new commands # exceeds the history size, it overwrites the history file entirely, otherwise it appends. This is friendlier to multiple bash sessions, since even with file truncation, it'll keep recent entries written by other bash sessions.

The only thing setting `shopt -s histappend` does is force bash to _always_ append... which honestly, should be the default.

## `history -r` bug

I also added tests for other history behaviors. One in particular stood out.

The existing spec test for `history -r` was not quite correct. `history -r` actually _appends_, rather than replaces, the file's contents to the current in-mem history. If run twice, you should expect to see all the new commands doubled up.

The old version only tested running `history -r` once, which can't distinguish between replacing and appending, so I fixed that.

BUT, for whatever reason, it's only failing on the Python side. I'm not sure why, but the test passes on the C++ side. Not sure what that's about yet.

## Spec test list
1. history -w
2. history -n
3. Fix incorrect history -r test
4. HISTSIZE test
5. HISTFILESIZE test
6. set +/-o history
7. shopt histappend

### DIFF
--- a/spec/builtin-history.test.sh
+++ b/spec/builtin-history.test.sh
@@ -1,5 +1,7 @@
 ## tags: interactive
 ## compare_shells: bash
+## oils_failures_allowed: 8
+## oils_cpp_failures_allowed: 7
 
 #### history -a
 rm -f tmp
@@ -32,28 +34,64 @@ history -a
 ^D
 ## END
 
-#### history -r
-rm -f tmp
-echo 'foo' > tmp
 
-echo '
+#### history -w
+
+cd $TMP
+echo 'cmd orig' > tmp
+HISTFILE=tmp
+history -c
+echo 'cmd new' > /dev/null
+history -w
+grep 'orig' tmp > /dev/null
+echo "found=$?"
+## STDOUT:
+found=1
+## END
+
+
+#### history -r 
+# Reads from the history file, and appends it to the current history
+
+cd $TMP
+printf "cmd orig%s\n" {1..10} > tmp
+HISTFILE=tmp
+
 history -c
 
-HISTFILE=tmp
 history -r
-history
-' | $SH -i
+history -r
 
-# match osh's behaviour of echoing ^D for EOF
-case $SH in bash) echo '^D' ;; esac
+history | grep orig | wc -l
 
 ## STDOUT:
-    1  HISTFILE=tmp
-    2  history -r
-    3  foo
-    4  history
-^D
+20
 ## END
+
+
+#### history -n
+# Reads new commands from the history file, and appends them to the current history
+# Based on line ranges, not contents
+
+cd $TMP
+
+printf "cmd orig%s\n" {1..10} > tmp1
+cp tmp1 tmp2
+printf "cmd new%s\n" {1..10} >> tmp2
+
+history -c
+HISTFILE=tmp1 history -r
+HISTFILE=tmp2 history -n
+
+history | grep orig | wc -l
+history | grep new | wc -l
+
+## STDOUT:
+10
+10
+## END
+
+
 
 #### HISTFILE is defined initially
 echo '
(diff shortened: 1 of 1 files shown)


andychu:
> The only thing setting shopt -s histappend does is force bash to always append... which honestly, should be the default.

Hm interesting ... I am in favor of simplifying OSH like this

And I **have** noticed `shopt -s histappend` missing in real scripts - well I guess that was in Warp bash code maybe?  Or mabye Starship bash

I forget, but yeah it might be reasonable to implement the **desired** behavior, and then `shopt -s histappend` is a no-op



---

Thanks for looking into `history -r` too!    These details are useful ...

Actually can you link to your comments in this PR in the spec tests themselves, or maybe copy the comments in?  I'm not sure when someone will work on this, but it is good info to have

spec/builtin-history.test.sh
```diff
@@ -32,29 +34,65 @@ history -a
...
+echo 'cmd new' > /dev/null
+history -w
+grep 'orig' tmp > /dev/null
+echo "found=$?"
+## STDOUT:
+found=1
+## END
 
+
+#### history -r
```
andychu: I think it will help to move the comment up to the description line, so that we can see it in the console / HTML reports

e.g.

"history -r reads from history file and appends to history in memory"

andychu:
This also reminds me of

- #324

Having history both on disk and in-memory causes many silly issues and broken mental models, like the ones you mention


Logically, I might imagine  history is a global append-only table of:

```
(shell pid, time, command)
```

or even

```
(machine ID, shell pid, time, command)
```

for a distributed shell

spec/builtin-history.test.sh
```diff
@@ -32,29 +34,65 @@ history -a
...
-    1  HISTFILE=tmp
-    2  history -r
-    3  foo
-    4  history
-^D
+20
 ## END
 
+
+#### history -n
```
andychu: likewise "history -n reads new commands from history file, and appends to history in memory"

spec/builtin-history.test.sh
```diff
@@ -32,29 +34,65 @@ history -a
 ^D
 ## END
 
-#### history -r
-rm -f tmp
-echo 'foo' > tmp
 
-echo '
+#### history -w
```
andychu: I guess this one would be "history -w writes history from memory to disk" ? 

or the spec test seems to be doing something slightly different

It could use some comments in the body too

spec/builtin-history.test.sh
```diff
@@ -195,4 +233,131 @@ status=1
 ## END
 
 
+#### HISTSIZE shrinks the history when changed
```
andychu: Oh interesting, I didn't know this

This is similar to `LANG=C` - it actually makes a setlocale() syscall when the var is changed !!

I don't like that, because the syntax doesn't tell you that!  It's just a magic function call that looks like a variable!

spec/builtin-history.test.sh
```diff
@@ -195,4 +233,131 @@ status=1
...
+HISTFILE=tmp
+HISTFILESIZE=5
+cat tmp | wc -l
+
+## STDOUT:
+5
+## END
+
+
+#### set -o history / set +o history
```
andychu: maybe "recording history can be toggled with set -o history"

spec/builtin-history.test.sh
```diff
@@ -195,4 +233,131 @@ status=1
...
+## STDOUT:
+status=0
+shopt -s histappend
+status=0
+shopt -u histappend
+^D
+## END
+
+
+#### shopt histappend
```
andychu: I think this is the one where we might not want to match bash behavior, for simplicity

If so then let's make a note here

I think it makes sense to implement the better behavior ... It's possible people will notice the difference, but it's OK if they push back and we have to "concede"

andychu (commented):
Excellent, thanks for doing this, and for the informative comments!

history is something I use very little of myself ... I basically have an alias `hi` for "history 30", and I type `!123` - that's about it

spec/builtin-history.test.sh
KingMob: LANG is a rabbit hole I don't want to go down. All I know is it royally messed with my attempts to run the Sonarr/Radarr projects.

spec/builtin-history.test.sh
KingMob: Fine with me. I know Warp is trying to set it, so that works for me.

KingMob:
Alright, I expanded the test names with useful descs, and I also added a simpler `history -d` test than the pre-existing one. The pre-existing one tested negative indices, which aren't supported in bash 4.

----

Weirdly, the C++ ALSO passes the new simple history -d test, though the Python one doesn't.

----

For the shopt histappend test, I used `## OK osh STDOUT:` but should I say `## pass osh STDOUT:` instead? https://github.com/oils-for-unix/oils/wiki/Spec-Tests doesn't say much about the qualifiers.

KingMob:
> history is something I use very little of myself ... I basically have an alias `hi` for "history 30", and I type `!123` - that's about it

My workflow is I tend to use `^R` to search history a lot. I don't know how standard that binding is, but it seems pretty widespread. Or I just push "Up" until I find what I'm looking for 😄

andychu:
Great thank you!

Issue #2395, opened 2025-08-23
backtick interpolation errors could be disabled in ysh (not in ysh:upgrade)


bar-g:
```
ysh-0.34$ var tst = "that `backticks` fail"
  var tst = "that `backticks` fail"
                  ^
[ interactive ]:10: Invalid backtick: use $(cmd) or \` in YSH strings
```

bar-g:
Hm, well I gues you say use '\\`', is that also the way in python?

bar-g:
```
>>> print("that is `no`")
that is `no`
```

bar-g:
* python prints ' \\` ' verbatim in double-quotes, so that would only be compatible in osh/ysh

Hm, seems I should rather have raised a question... it's now:

If it's really correct to not free and disable shell`s backticks-syntax, how do we handle interpolatable text output?

Would that be that quote module, to insert backslashes on import and remove them on export?

andychu:
```

$ osh -o ysh:upgrade -c 'echo "`echo hi`"'
hi


$ osh -o ysh:all -c 'echo "`echo hi`"'
  echo "`echo hi`"
        ^
[ -c flag ]:1: Use $(cmd) instead of backticks (parse_backticks)


```

This is working as intended

We don't change the meaning of existing code; instead we disallow old code

If we allowed backticks in YSH strings, then you have to know if you're looking at OSH or YSH to figure out what it means.

bar-g:
Thanks, understood now, it's the shell compatibilty that's not to break.

(Made additions to those error messages.)

PR #2397, opened 2025-08-25
[word_parse.py] mention compat reason in backtick errors

bar-g:
explanation related to #2395

### DIFF
--- a/osh/word_parse.py
+++ b/osh/word_parse.py
@@ -1095,7 +1095,7 @@ class WordParser(WordEmitter):
 
             elif self.token_kind == Kind.Left:
                 if self.token_type == Id.Left_Backtick and is_ysh_expr:
-                    p_die("Invalid backtick: use $(cmd) or \\` in YSH strings",
+                    p_die("In YSH use $(cmd) instead of backticks, or compatibly denote one with \\` (parse_backticks)",
                           self.cur_token)
 
                 part = self._ReadDoubleQuotedLeftParts()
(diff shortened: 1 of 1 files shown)


osh/word_parse.py
```diff
@@ -1095,7 +1095,7 @@ def _ReadLikeDQ(self, left_token, is_ysh_expr, out_parts):
 
             elif self.token_kind == Kind.Left:
                 if self.token_type == Id.Left_Backtick and is_ysh_expr:
-                    p_die("Invalid backtick: use $(cmd) or \\` in YSH strings",
+                    p_die("In YSH use $(cmd) instead of backticks, or compatibly denote one with \\` (parse_backticks)",
```
andychu: I think the new string is longer, and isn't  any clearer

If you want it to be more clear, adding OILS-ERR-X and then an entry to `doc/error-catalog.md` is the way we do it

bar-g: Ok, shortened.
But mentioning the shopt at both places.
An OILS-ERR would basicly be empty (nothing to add there).

osh/word_parse.py
```diff
@@ -1095,7 +1095,7 @@ def _ReadLikeDQ(self, left_token, is_ysh_expr, out_parts):
 
             elif self.token_kind == Kind.Left:
                 if self.token_type == Id.Left_Backtick and is_ysh_expr:
-                    p_die("Invalid backtick: use $(cmd) or \\` in YSH strings",
+                    p_die("Invalid backtick: use $(cmd) or compatible \\` (parse_backticks)",
```
andychu: this is not correct - the error message is not affected by parse_backticks

YSH expressions like

```
var x = "`echo hi`"
```

don't need any such option, because it's not valid OSH code

on the other hand, 

```
echo "`echo hi`"
```

is valid OSH code, so it requires an option

andychu:
I don't think this change is substantial enough; if there were an OILS-ERR-X then it would probably be

bar-g:
added AFAIU

Issue #2398, opened 2025-08-26
Nested Subshells Don't Parse


melvinw:
gzip has a utility script called zdiff. It gets used at some point in the check step when building gzip from source. It also makes use of nested subshells in a few places. The osh parser is tripping over these.

Here's the relevant output from the failed build.
```
make[3]: Entering directory '/home/udu/aports/main/gzip/src/gzip-1.14'
  GEN      gzip.doc.gz
  GEN      check-local
                                ((gzip -cdfq -- "$file2" 4>&-
                                                ^
./zdiff:137: Unexpected token after arithmetic expression (Id.Word_Compound != Id.Arith_RParen)
make[3]: *** [Makefile:2631: check-local] Error 2
```

I'll send a patch shortly to add a spec test, and eventually a fix, for this and I want to be able to reference an issue from those commits.

melvinw:
Turns out this is a dupe of #2337

Issue #2337, opened 2025-07-02
libffi, gzip, freeswitch - (((grep  not parsed like bash/dash


andychu:
[#osh > ((( not parsed like bash](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/.28.28.28.20not.20parsed.20like.20bash/with/477011885)

andychu:
`test/spec.sh divergence`

andychu:
from #2398 

> gzip has a utility script called zdiff. It gets used at some point in the check step when building gzip from source. It also makes use of nested subshells in a few places. The osh parser is tripping over these.
> 
> Here's the relevant output from the failed build.
> 
> ```
> make[3]: Entering directory '/home/udu/aports/main/gzip/src/gzip-1.14'
>   GEN      gzip.doc.gz
>   GEN      check-local
>                                 ((gzip -cdfq -- "$file2" 4>&-
>                                                 ^
> ./zdiff:137: Unexpected token after arithmetic expression (Id.Word_Compound != Id.Arith_RParen)
> make[3]: *** [Makefile:2631: check-local] Error 2
> ```
> 
> I'll send a patch shortly to add a spec test, and eventually a fix, for this and I want to be able to reference an issue from those commits.

andychu:
Let me remind myself of the **failing cases**

## Latest Results

https://op.oils.pub/aports-build/2025-10-15-main.wwz/_tmp/aports-report/2025-10-15-main/diff_merged.html#t:notable_disagree=6a

## marked `##2337`

`freeswitch`

```
       cputype=`(((grep cpu /proc/cpuinfo | cut -d: -f2) ; ($PRTDIAG -v |grep -i sparc) ; grep -i cpu /var/run/dmesg.boot ) | head -n 1) 2> /dev/null`
                        ^
[ backticks in ./configure ]:21650: Parser expected Id.Arith_RParen, got Id.Word_Compound
```

`isl25 isl26`

```
       cputype=`(((grep cpu /proc/cpuinfo | cut -d: -f2) ; ($PRTDIAG -v |grep -i sparc) ; grep -i cpu /var/run/dmesg.boot ) | head -n 1) 2> /dev/null`
                        ^
[ backticks in ./configure ]:6981: Parser expected Id.Arith_RParen, got Id.Word_Compound
```


### marked `##2337`

`gzip`

```
                                ((gzip -cdfq -- "$file2" 4>&-
                                                ^
[...]

PR #2399, opened 2025-08-26
add spec test for nested subshell parse error

melvinw:
gzip has a utility script called zdiff. It gets used at some point in the check step when building gzip from source. It also makes use of nested subshells in a few places. The osh parser is tripping over these. Here's an example from a simple script.

    ((echo foo) | tr o 0)
         ^~~
    _tmp/foo.sh:1: Unexpected token after arithmetic expression (Id.Word_Compound != Id.Arith_RParen)

Other shells seem to handle this case, so in preparation for an inbound fix, this commit adds a new spec test to cover it.

Related to #2398

### DIFF
--- a/spec/divergence.test.sh
+++ b/spec/divergence.test.sh
@@ -1,5 +1,5 @@
 ## compare_shells: bash dash mksh zsh ash
-## oils_failures_allowed: 2
+## oils_failures_allowed: 3
 
 # This file relates to:
 #
@@ -54,4 +54,12 @@ good
 bad
 
 ## STDOUT:
+## END
+
+#### Nested subshell (#2398)
+
+# found on line 137 of the zdiff util from gzip
+((echo foo) | tr o 0)
+## STDOUT:
+f00
 ## END


andychu:
Oh yes I saw this in the logs ... I wonder if this is the same as   https://github.com/oils-for-unix/oils/issues/2337

Awhile ago I added a couple tests in `test/spec.sh divergence`

----

I think this one might require some design; there is the issue of

- `((` vs `( (` with space
- `$((` vs `$( (` with space - similar bug I think I may have seen



In theory I think `regtest/aports/cause.awk` can help us de-dupe these bugs, though probably we don't have a smooth process yet

melvinw:
Oh great! We can close this then

andychu:
I was thinking about this issue, and I realized it does help to have all the real examples encoded in the spec tests

The one I added was this

```
# spaces help
good() {
  cputype=`( ( (grep cpu /proc/cpuinfo | cut -d: -f2) ; ($PRTDIAG -v |grep -i sparc) ; grep -i cpu /var/run/dmesg.boot ) | head -n 1) 2> /dev/null`
}

bad() {
  cputype=`(((grep cpu /proc/cpuinfo | cut -d: -f2) ; ($PRTDIAG -v |grep -i sparc) ; grep -i cpu /var/run/dmesg.boot ) | head -n 1) 2> /dev/null`
  #echo cputype=$cputype
}

```


The `good` one has spaces for `( ( (`

---

But this one is slightly different, so  I  think it's useful to have it too.   And to note that it came from gzip.  (hm I didn't note which package the other error is from)


Because we don't know what the exact solution is yet.

But they should be in the same file, I guess `spec/divergence` is OK

andychu:
So yeah I think having the exact line helps, and then we can reduce to something more minimal as we fix it

The reducing process is often how I figure out what the bug actually is

melvinw:
Ah ok. I'll reopen this and plop it in `spec/divergence`

andychu:
great thanks!

Issue #2401, opened 2025-08-27
dtc package - printf %b doesn't parse octal forms correctly


PossiblyAShrub:
```
osh$ printf '%b\n' "\012\013\014\015\336\352\255\276\357" | xxd
00000000: 0a0b 0c0d 5c33 3336 5c33 3532 5c32 3535  ....\336\352\255
00000010: 5c32 3736 5c33 3537 0a                   \276\357.

bash$ printf '%b\n' "\012\013\014\015\336\352\255\276\357" | xxd
00000000: 0a0b 0c0d deea adbe ef0a                 ..........
```

The issue here is that `printf '%b'` shouldn't use `echo -e` formatting but rather `$''` quoting (I think).

This is causing at least one test in the [`dtc` package to fail](https://op.oils.pub/aports-build/2025-08-14-fix.wwz/_tmp/aports-report/2025-08-14-fix/shard2/error/dtc.txt).

SeerLite:
This should be closed, right? Since it seems to have been fixed via #2403.
(Came here from the [blog post explaining this bug](https://aolsen.ca/writings/reproducing-dtc-failure). Very useful and easy to follow!)

andychu:
Yes we can close it now

It still appears on https://github.com/orgs/oils-for-unix/projects/1/views/1 (in the last column)

PR #2403, opened 2025-08-27
Parse \123 octal in printf %b

PossiblyAShrub:
Fixes: #2401


PossiblyAShrub:
With this change, I can successfully build `dtc` with `regtest/aports-run.sh build-packages dtc osh-as-sh`.

(When I'm using the CI build of oils on this branch per [#osh > regtest/aports testing on custom CI version](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/regtest.2Faports.20testing.20on.20custom.20CI.20version))

PossiblyAShrub:
@andychu This is ready for a review.

andychu:
Hm it's funny that `printf %b` is different than `echo -e`

from `help printf`

```
    In addition to the standard format specifications described in printf(1),
    printf interprets:
    
      %b        expand backslash escape sequences in the corresponding argument

```


and then `help echo`

```
      -e        enable interpretation of the following backslash escapes
    
    `echo' interprets the following backslash-escaped characters:
      \a        alert (bell)
      \b        backspace
...
      \0nnn     the character whose ASCII code is NNN (octal).  NNN can be
                0 to 3 octal digits

```

andychu:
Thank you!  

bash can't decide whether octal is 3 or 4 digits ...   It has 3 different  choices, which I made a note of in the spec tests:

```
# - echo -e: 
#   - NO  3 digit octal  - echo -e '\141' does not work
#   - YES 4 digit octal
# - printf %b
#   - YES 3 digit octal
#   - YES 4 digit octal
# - printf string (outer)
#   - YES 3 digit octal
#   - NO  4 digit octal
# - $'' and $PS1
#   - YES 3 digit octal
#   - NO  4 digit octal
```


IMO the underlying cause for `#bug-post-mortem` is that bash has a lot of copy-paste code, and they didn't think about or document these things

PR #2406, opened 2025-09-03
spell out the "special default" of trim() and whitespace contexts

bar-g:
* emphasize the special case of the trim() default
* refer to "blank space" for "normal space" to show its connection to /blank/
  (wording as in wikipedia: "U+0020 represents blank space" https://en.wikipedia.org/wiki/Whitespace_character)
* up-front "unicode" in referring sentence


bar-g:
two-dimensional /space/ vs. /blank/ line-elements is a useful connotation to remember their difference

bar-g:
shortened to minimal descriptions in parens

PR #2407, opened 2025-09-03
[unicode.md] default split() "by space" like trim() except non-breaking

bar-g:


bar-g:
You wrote:
> `s.split()` doesn't have a default "split by space", which should probably respect unicode space, like `trim()` does

However, it shouldn't split on no-break space.

So, would it still need to match any special unicode char?

The ascii vertical space `\y0B` (unicode `\u{b}`) is covered by `/space/`.
```
ysh-0.35$ = b'\y0b\u{b}'.split(/space/)
(List)  ['', '', '']
```

andychu:
This is confusing

bar-g:
I do think this one may actually be ok to close.

Because the answer to my question "would it still need to match any special unicode char?" seems to be "no".

I think `/space/` seems to be the correct default, and this would then not require any further unicode splitting. Because the non-breaking chars are out-of-scope for splitting (i.e. for breaking) and all remaining chars from the `trim()` "whitespace" list are included in `/space/`.

This would mean that the todo idea that `split()` "should probably respect unicode space" can be negated. With the current list of `trim()` chars, it doesn't need to split on any unicode chars.

Assigning `/space/` as the default value seems enough.

PR #2408, opened 2025-09-05
[chap-type-method.md] list evalToDict (as header for ToC)

bar-g:


bar-g:
I added a new separate header (and also moved the old text to it):
### io/evalToDict()
putting that header _after_ the io/eval() note about "pure eval()" (so that note practically moved before the new header)

PR #2410, opened 2025-09-09
feat: new build option 'build-packages-overlayfs'

Melkor333:
As discussed in Zulip, a way to build all packages completely independently. This doesn't use `xargs ...` yet, so definitely needs more improvements. But my local test with a bunch of packages seemed to work. The only questionable thing: I can't tell how much overhead it creates.

A quick test showed that `aaudit` builds in about 2.5 seconds with `overlayfs` and about 2 seconds without. But that could just be the preparation, in which case it would mean that there is about half a second added per package, which is about 15mins additional time required. If it's overhead at build time, well, then it will probably take longer :)

### DIFF
--- a/regtest/aports-setup.sh
+++ b/regtest/aports-setup.sh
@@ -12,6 +12,8 @@
 #
 # Other commands:
 #   $0 remove-chroot
+#
+set -e
 
 : ${LIB_OSH=stdlib/osh}
 source $LIB_OSH/bash-strict.sh
--- a/regtest/aports-setup.sh
+++ b/regtest/aports-setup.sh
@@ -12,8 +12,6 @@
 #
 # Other commands:
 #   $0 remove-chroot
-#
-set -e
 
 : ${LIB_OSH=stdlib/osh}
 source $LIB_OSH/bash-strict.sh
--- a/regtest/aports-common.sh
+++ b/regtest/aports-common.sh
@@ -2,6 +2,7 @@
 
 readonly CHROOT_DIR=_chroot/aports-build
 readonly CHROOT_HOME_DIR=$CHROOT_DIR/home/udu
+readonly OVERLAY_BASE_DIR=$(pwd)/_chroot/overlay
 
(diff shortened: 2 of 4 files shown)


andychu:
Cool!  I tried this and it worked

```
_chroot/overlay/
├── lzip
│   ├── env.sh
│   ├── etc
│   │   └── apk
│   │       └── world
│   ├── home
│   │   └── udu
│   │       ├── aports
│   │       │   └── main
│   │       │       └── lzip
│   │       ├── oils
│   │       │   └── _tmp
│   │       │       └── aports-guest
│   │       │           └── baseline
│   │       │               ├── lzip.log.txt
│   │       │               └── lzip.task.tsv
│   │       └── packages
│   │           └── main
│   │               └── x86_64
│   │                   ├── APKINDEX.tar.gz
│   │                   ├── lzip-1.25-r0.apk
│   │                   └── lzip-doc-1.25-r0.apk
│   ├── lib
│   │   └── apk
│   │       └── db
│   │           ├── installed
│   │           ├── lock
│   │           ├── scripts.tar
│   │           └── triggers
│   ├── tmp
│   └── var
│       └── cache
│           ├── apk
│           │   ├── APKINDEX.654befff.tar.gz
│           │   └── APKINDEX.6f88eeb6.tar.gz
│           └── distfiles
├── lzip_merged
└── lzip_work
    └── work  [error opening dir]

```


I think this is a promising approach

andychu:
And yeah now I understand a little more about how `abuild` works!  It apparently keeps its state in `/lib/apk/db` ?

regtest/aports-setup.sh
```diff
@@ -12,6 +12,8 @@
 #
 # Other commands:
 #   $0 remove-chroot
+#
+set -e
```
andychu: btw `bash-strict.sh` does `set -e`

andychu:
OK I am merging this with some slight edits.  Great idea, thank you !!

PR #2419, opened 2025-09-14
aports: add interactive mode

melvinw:
This only works with the overlayfs build for now. If the INTERACTIVE environment variable is set, we will drop into a shell after attempting to build a package with abuild. This is the last thing that happens before unmounting the overlay we used for the build.


andychu:
Great thanks!

Issue #2411, opened 2025-09-09
sfic - C++ IndexError* crash


andychu:
```
checking for tdb_store in -ltdb... yes
terminate called after throwing an instance of 'IndexError*'
Aborted
>>> ERROR: sfic: build failed
```

https://op.oils.pub/aports-build/2025-09-08.wwz/_tmp/aports-report/2025-09-08/shard14/osh-as-sh/log/sfic.log.txt

melvinw:
Was able to reproduce and get some more info about this by doing the following. Haven't had a chance to dig too deep into the trace yet. Hoping to do that later tonight or tomorrow

1. Reproduce the failure from the top with `regtest/aports-run.sh build-packages 'sfic' osh-as-sh`
2. Drop into the chroot as `udu` and add `set -x` to `~udu/aports/main/sfic/src/sfic-0.1.7/configure` right before the crash: line 3417
3. Re-run the build by hand with `abuild -f -r -C ~/aports/main/sfic builddeps build`
```
>>> sfic: Analyzing dependencies...
>>> sfic: Installing for build: build-base autoconf automake tdb-dev
fetch http://dl-cdn.alpinelinux.org/alpine/latest-stable/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/latest-stable/community/x86_64/APKINDEX.tar.gz
(1/10) Installing m4 (1.4.19-r4)
(2/10) Installing perl (5.40.3-r0)
(3/10) Installing autoconf (2.72-r1)
(4/10) Installing automake (1.17-r1)
(5/10) Installing tdb-libs (1.4.12-r0)
(6/10) Installing tdb-dev (1.4.12-r0)
(7/10) Installing .makedepends-sfic (20250911.211149)
(8/10) Installing perl-error (0.17030-r0)
(9/10) Installing perl-git (2.49.1-r0)
(10/10) Installing git-perl (2.49.1-r0)
Executing busybox-1.37.0-r19.trigger
OK: 336 MiB in 93 packages
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking whether make sets ${MAKE}... yes
checking for working aclocal... found
checking for working autoconf... found
checking for working automake... found
checking for working autoheader... found
checking for working makeinfo... missing
checking for x86_64-alpine-linux-musl-gcc... cc
checking for C compiler default output... a.out
checking whether the C compiler works... yes
[...]

PR #2420, opened 2025-09-14
add spec test for #2411

melvinw:
This commit adds a failing spec test to tickle a crash that happens when evaluating a simple command that is prefixed with variable bindings.

### DIFF
--- a/spec/temp-binding.test.sh
+++ b/spec/temp-binding.test.sh
@@ -1,5 +1,5 @@
 ## compare_shells: dash bash zsh mksh ash yash
-## oils_failures_allowed: 0
+## oils_failures_allowed: 1
 
 # forked from spec/ble-idioms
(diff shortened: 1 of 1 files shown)


spec/temp-binding.test.sh
```diff
@@ -152,3 +152,10 @@ x=
 x=
 x=
 ## END
+
+#### argv from unset var (#2411)
```
andychu: I would make this a bit more descriptive, like "Temp binding, then empty argv from unquoted unset var"

to distinguish from the other cases we looked at

Or actually now I just put the code in the description, like

```
FOO=bar $unset - temp binding, then empty argv from unquoted unset var
```

That can make it easier to find spec tests, which is not always easy now

melvinw: Ah sure. Done!

PR #2434, opened 2025-09-25
[chap-builtin-cmd.md] --raw-line slow but no EOF error

bar-g:
also mention @(...)

### DIFF
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -444,19 +444,20 @@ recommended:
 
 ---
 
-This flag replaces shell's `IFS= read -r` idiom, reading one byte a time in an
-unbuffered fashion:
+This flag still reads one byte a time in an unbuffered fashion, like shell's
+`IFS= read -r` idiom, but does not error upon input- or file-ending (EOF):
 
     read --raw-line             # unbuffered read of line, omitting trailing \n
     read --raw-line (&x)        # fills $x
 
-    read --raw-line --with-eol  # include the trailing \n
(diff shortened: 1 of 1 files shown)


andychu:
My recommendation for all these PRs is to make the MINIMUM change to the docs

Often your edits make the docs better in some ways, and worse in others

I think you are not a native English speaker, so you should defer to the existing text, at least at first

---

Or send 2 changes - one that makes the minimum change, and one that has suggested edits

Most likely I would accept the first

bar-g:
your feedback...

what are you refering to with "the first"? you would accept...

Though, the change already is rather minimal, concisely:

* start sentence with performance topic, to continue the theme started with the first two "flags" (moving in the wind? cloudy language, haha)
* mention difference to legacy options
* mention related command splice

If you think there is something like an article (like "the") or anything else missing in my sentence but not in yours two lines above, or anything better, why don't you just mention it?

Unlike your above message, mentioning things specifically could actually help in solving what fits and what doesn't.

bar-g:
It's what the title says: Add info "--raw-line slow but no EOF error" + the useful `@(...)` command splice (mentioned in the commit message).

bar-g:
On zulip you commented that "Right now it's mixed in with changes that make it worse".

I reorderd your sentence to first mention the performance topic (to keep the reference/listing/topic of the earlier text going), and added the EOF difference.

And I added the very useful command splice to your hinting paragraph.

What do you think is making what worse? What could be better? Specific suggestions/examples.

You've not given any specific feedback, that could be useful enough to learn what you wanted or to edit this PR.

Issue #2435, opened 2025-09-27
Command subs have different truthiness in OSH


melvinw:
I noticed this while debugging some test failures while building `ifupdown-ng`.

`eval` behaves the same in OSH and other shells
```
$ eval true; echo $?
0
$ eval false; echo $?
1
```

However, OSH seems to swallow the exit code for both cases here when using backticks and `$()`.
```
$ `true`; echo $?
0
$ `false`; echo $?
0
$ $(true); echo $?
0
$ $(false); echo $?
0
```

Other shells, e.g. bash, behave differently.
```
$ `true`; echo $?
0
$ `false`; echo $?
1
$ $(true); echo $?
0
$ $(false); echo $?
1
```

andychu:
Hm is this the whole root cause for

- #2416 

?  Or is it a separate bug?

----


If there is going to be one fix, we should probably have one bug

We can change the title of the bug

Right now I think it's


- **Symptom** `abuild` etc. "cannot provide test case names with -l"
- **Cause** "In OSH, unquoted command subs that evaluate to empty string / empty argv have different `$?` than other shells"



That's the Symptom -> Root Cause framing we discussed, e.g. with the title of    https://aolsen.ca/writings/reproducing-dtc-failure

melvinw:
> Hm is this the whole root cause for
> 
>     * [abuild, ifupdown-ng lddtree, tiny-cloud -  "cannot provide test case names with -l" #2416](https://github.com/oils-for-unix/oils/issues/2416)
> 
> 
> ? Or is it a separate bug?

Oh, sure. We can repurpose that other issue.

PR #2436, opened 2025-09-27
spec: add test for command sub exit code behavior

melvinw:
This commit adds a failing spec test to capture the difference between how OSH and other shells handle the exit code for command subs that don't print anything.

See #2435

### DIFF
--- a/spec/divergence.test.sh
+++ b/spec/divergence.test.sh
@@ -64,7 +64,7 @@ bad
 f00
 ## END
 
-#### Command substitution exit codes (#2435)
+#### Exit code when command sub evaluates to empty str, e.g. `false` (#2435)
 
 # OSH exits with 0 while others exit with 1
 `true`; echo $?


spec/divergence.test.sh
```diff
@@ -63,3 +63,27 @@ bad
 ## STDOUT:
 f00
 ## END
+
+#### Command substitution exit codes (#2435)
```
andychu: I would be more specific in the title, so it's easier to find later (I use `grep`

Like "Exit code of unquoted command substitution that evaluates to empty string / empty argv - $(false)

andychu (commented):
Thanks for finding this!  Great bug, and now we can see that all shells agree, except OSH

spec/divergence.test.sh
melvinw: Updated!

PR #2437, opened 2025-09-27
Add multiline herestring to idioms

Melkor333:
I hope the very simple example is fine :)

I also fixed the title of the sigil saying to use the deprecated `:` which has become `&`.


doc/idioms.md
```diff
@@ -297,6 +297,23 @@ Yes:
...
 
+### Use multiline here string instead of here doc
+
+Ysh has multiline here strings with `'''` and `"""` which are cleaner than heredocs, as they can be properly indented.
+
+No:
+
+    cat <<EOF
+    hello world
+    EOF
```
bar-g: to be fair `<<-EOF` and `<<-'EOF'` allow (tab) indention, but tabs get lost in the www

doc/idioms.md
```diff
@@ -297,6 +297,23 @@ Yes:
     read -0 (&myvar) < file.bin
     echo $record
 
+### Use multiline here string instead of here doc
```
andychu: I would say "Use Multi-line Strings Instead of Here Docs"

doc/idioms.md
```diff
@@ -297,6 +297,23 @@ Yes:
...
+
+Ysh has multiline here strings with `'''` and `"""` which are cleaner than heredocs, as they can be properly indented.
```
andychu: s/Ysh/YSH

s/heredocs/here docs/

Rather than "cleaner", I'd say they are "easier to remember"

I explained that in this post:

https://www.oilshell.org/blog/2021/09/multiline.html#an-orthogonal-design-is-easier-to-use-and-remember

e.g. you don't have to remember a new rule about `EOF` and `'EOF'` with quotes - you can use the rule you already know about single- and double-quotes

So then there are at least two yes-no examples:

---

No:

```
cat <<EOF
hello $name
EOF
```

Yes:


```
cat <<< """
hello $name
"""
```



and then


No:

```
cat <<'EOF'
Price is $4.99
EOF
```

Yes:


```
cat <<< '''
Price is $4.99
'''
```



That shows the quoting rules are more consistent


The indentation is also an improvement, since you don't have to remember `<<-`, and you can use spaces, not just tabs

doc/idioms.md
```diff
@@ -283,7 +283,7 @@ Idiomatic:
 (As always, `set` can be used when you care about compatibility with other
 shells.)
 
-### Use `:` When Mentioning Variable Names
+### Use `&` When Mentioning Variable Names
```
andychu, line 362: Oof, thanks for fixing this!

andychu (commented):
Thanks for adding this!

doc/idioms.md
andychu: It's also probably more accurate to say that "Rather than here docs, YSH multi-line strings can be used with the `<<<` operator"

Because multi-line strings can be used **with or without** `<<<` 

We decoupled the "multi-line" aspect from the "stdin" aspect

Melkor333: Should I add an additional example, or what do you think?

andychu: I think we need more examples

Do you think someone would understand it from the existing text?

I mean we can always link to the blog post, but it's kinda lame because that's over 4 years old now

andychu:
OK sorry for the big delay on reviewing this!  I think we need to find a better UI for code review

---

I simplified the text a bit, and  moved it to its own section

It shouldn't have been under the "YSH builtins" section

I added another topic with the new section "Strings and Quoting"

PR #2442, opened 2025-09-29
check argv length before indexing (fixes #2411)

### LINKED ISSUE #2411: sfic - C++ IndexError* crash

melvinw:
We have a helper to check if prefix bindings should pe preserved when running a simple command. This check requires checking the first argument of the command line, but doesn't check the length before indexing. This commit patches the helper to return early if argv is empty. From there, the usual command execution logic will take care of the rest: either a noop, or erroring out if strict_argv is enabled.

### DIFF
--- a/osh/cmd_eval.py
+++ b/osh/cmd_eval.py
@@ -304,6 +304,9 @@ def _PrefixBindingsPersist(cmd_val):
             return True
         elif case(cmd_value_e.Argv):
             cmd_val = cast(cmd_value.Argv, UP_cmd_val)
+            if len(cmd_val.argv) == 0:
+                return True
+
             arg0 = cmd_val.argv[0]
 
             # exec is an EXCEPTION to the SPECIAL builtin rule.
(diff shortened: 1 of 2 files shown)


osh/cmd_eval.py
```diff
@@ -304,6 +304,9 @@ def _PrefixBindingsPersist(cmd_val):
             return True
         elif case(cmd_value_e.Argv):
             cmd_val = cast(cmd_value.Argv, UP_cmd_val)
+            if len(cmd_val.argv) == 0:
+                return True
```
andychu, line 311: Hm it's somewhat surprising to me that the binding persists - I was thinking it should be `return False` !!!

```
$ bash -c 'FOO=bar $undef; echo FOO=$FOO'
FOO=bar


$ bash -c 'FOO=bar echo; echo FOO=$FOO'  # does NOT persist
FOO=
```

I would have thought FOO=bar goes out of scope in the first case, like the second

It looks like the spec test covers it, but how about adding a comment about this behavior?

andychu (commented):
Thanks, I'm glad this was a localized fix!

osh/cmd_eval.py
melvinw: Done!

andychu:
Great thanks!

Issue #2429, opened 2025-09-21
[osh as sh] gdbm - test failures


PossiblyAShrub:
Error: https://op.oils.pub/aports-build/2025-09-15-order.wwz/_tmp/aports-report/2025-09-15-order/shard3/error/gdbm.txt

Full Logs: https://op.oils.pub/aports-build/2025-09-15-order.wwz/_tmp/aports-report/2025-09-15-order/shard3/osh-as-sh/log/gdbm.log.txt

PossiblyAShrub:
This one ended up being fairly easy to reproduce:

The test logs (which are kinda hard to read...) show this:

```
## ---------------------- ##
## Detailed failed tests. ##
## ---------------------- ##

#                             -*- compilation -*-
30. gdbmtool00.at:17: testing Input from file ...
./gdbmtool00.at:20:

cat >input <<'_ATEOF'

status
version
quit
_ATEOF


gdbmtool -f input | sed 's/^GDBM version.*/GDBM version/'

--- /dev/null
+++ /home/udu/aports/main/gdbm/src/gdbm-1.24/tests/testsuite.dir/at-groups/30/stderr
@@ -0,0 +1,4 @@
+  gdbmtool -f input | sed 's/^GDBM version.*/GDBM version/'
+  ^~~~~~~~
+/home/udu/aports/main/gdbm/src/gdbm-1.24/tests/testsuite.dir/at-groups/30/test-source:36: Can't execute '/home/udu/aports/main/gdbm/src/gdbm-1.24/tests/gdbmtool': Permission denied
--- -
+++ /home/udu/aports/main/gdbm/src/gdbm-1.24/tests/testsuite.dir/at-groups/30/stdout
@@ -1,6 +1 @@
-Database file: junk.gdbm
-Database is not open
-define key string
-define content string
-GDBM version

```

But `/home/udu/aports/main/gdbm/src/gdbm-1.24/tests/gdbmtool` is a directory!

I reran the test with `echo $PATH` and found that `PATH=/home/udu/aports/main/gdbm/src/gdbm-1.24/tests:/home/udu/aports/main/gdbm/src/gdbm-1.24/tools/:/usr/local/bin/:/usr/sbin/:/usr/bin/:/sbin/:/bin/`.

From this we can get a minimal reproduction:

```
$ mkdir -p bin test/hello
$ printf '#!/usr/bin/env sh\necho hi\n' >bin/hello
$ chmod +x bin/hello
$ bin/hello
hi

[...]

Issue #689, opened 2020-04-06
Don't try to execute directories, or give a better error message


andychu:
> (24) I tested zsh and mksh, and they return 126 here.

OK, I tried `zsh`, `mksh` and `dash`, and they all try to execute the directory. `yash` and `bash` detects the empty string. This is interesting. I tried more shells: `posh`, `ksh` and `busybox sh` try to execute the directory, and `csh`, `tcsh` and `fish` detects the empty string. Anyway, I think it's better to detect the empty string rather than trying to execute the directory.

> I guess it's a special case.

The following cases may also be handled specially. I think one can just skip directory names in searching executable files from `PATH`. Directories are not executable even if the permission contains an executable flag.

```bash
osh$ ..
  ..
  ^~
[ interactive ]:1: Can't execute '/home/murase/bin/..': Permission denied
osh$ mkdir -p foo/bar
osh$ PATH=foo:$PATH
osh$ bar
  bar
  ^~~
[ interactive ]:9: Can't execute 'foo/bar': Permission denied
```

andychu:
Shells disagree on this: many appear to add a special case, some just execute the directory and fail with `126`.

iacore:
YSH does this though... isn't it supposed to be a clean state implementation?

andychu:
Oh wow, this bug got lost ... it turned out be an important bug in #2444 !

andychu:
This will be fixed with the next release

Issue #527, opened 2019-11-08
Allow registering internal hooks when variables are changed


andychu:
* We need this for `LANG=C` support.  See #522 
*  This would likely make some other things more efficient

andychu:
bash has a ton of these hooks in `variables.c`.  We might only need a few

```
static struct name_and_function special_vars[] = {                                                                                                                                                                                            
  { "BASH_COMPAT", sv_shcompat },                                                                                                                                                                                                             
  { "BASH_XTRACEFD", sv_xtracefd },                                                                                                                                                                                                           
                                                                                                                                                                                                                                              
#if defined (JOB_CONTROL)                                                                                                                                                                                                                     
  { "CHILD_MAX", sv_childmax },                                                                                                                                                                                                               
#endif                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                              
#if defined (READLINE)                                                                                                                                                                                                                        
#  if defined (STRICT_POSIX)                                                                                                                                                                                                                  
  { "COLUMNS", sv_winsize },                                                                                                                                                                                                                  
#  endif                                                                                                                                                                                                                                      
  { "COMP_WORDBREAKS", sv_comp_wordbreaks },                                                                                                                                                                                                  
#endif                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                              
  { "EXECIGNORE", sv_execignore },                                                                                                                                                                                                            
                                                                                                                                                                                                                                              
  { "FUNCNEST", sv_funcnest },                                                                                                                                                                                                                
                                                                                                                                                                                                                                              
  { "GLOBIGNORE", sv_globignore },                                                                                                                                                                                                            
                                                                                                                                                                                                                                              
#if defined (HISTORY)                                                                                                                                                                                                                         
  { "HISTCONTROL", sv_history_control },                                                                                                                                                                                                      
  { "HISTFILESIZE", sv_histsize },                                                                                                                                                                                                            
  { "HISTIGNORE", sv_histignore },                                                                                                                                                                                                            
  { "HISTSIZE", sv_histsize },                                                                                                                                                                                                                
  { "HISTTIMEFORMAT", sv_histtimefmt },                                                                                                                                                                                                       
#endif                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                              
#if defined (__CYGWIN__)                                                                                                                                                                                                                      
  { "HOME", sv_home },                                                                                                                                                                                                                        
#endif                                                                                                                                                                                                                                        
[...]

PR #2444, opened 2025-10-03
Fix: OSH tries to execute directories on PATH

PossiblyAShrub:
Changes:
- I discovered and added a failing test for https://github.com/oils-for-unix/oils/issues/2445
- Added a failing test for https://github.com/oils-for-unix/oils/issues/2429 and it's extension to `source`
- Fixed https://github.com/oils-for-unix/oils/issues/2429 and the related issue with `source`

### DIFF
--- a/cpp/pylib.cc
+++ b/cpp/pylib.cc
@@ -60,7 +60,7 @@ bool isfile(BigStr* path) {
   if (::stat(path->data_, &st) < 0) {
     return false;
   }
-  return !S_ISDIR(st.st_mode);
+  return S_ISREG(st.st_mode);
 }
 
 }  // namespace path_stat
--- a/cpp/pylib.cc
+++ b/cpp/pylib.cc
@@ -55,4 +55,12 @@ bool isdir(BigStr* path) {
   return S_ISDIR(st.st_mode);
 }
 
+bool isfile(BigStr* path) {
+  struct stat st;
+  if (::stat(path->data_, &st) < 0) {
+    return false;
+  }
+  return !S_ISDIR(st.st_mode);
+}
+
 }  // namespace path_stat
--- a/spec/command_.test.sh
+++ b/spec/command_.test.sh
@@ -193,3 +193,54 @@ echo status=$?
 ## stdout: status=1
 ## OK osh stdout: status=2
 ## BUG dash/bash stdout: status=0
+
+#### Executing command with same name as directory in PATH (#2429)
+
+# Make the following directory structure. File type and permission bits are
+# given on the left.
+# [drwxr-xr-x]  _tmp
+# +-- [drwxr-xr-x]  bin
+# |   \-- [-rwxr-xr-x]  hello
+# +-- [drwxr-xr-x]  notbin
+# |   \-- [-rw-r--r--]  hello
+# \-- [drwxr-xr-x]  dir
+#     \-- [drwxr-xr-x]  hello
+mkdir -p _tmp/bin
+mkdir -p _tmp/bin2
+mkdir -p _tmp/notbin
+mkdir -p _tmp/dir/hello
+printf '#!/usr/bin/env sh\necho hi\n' >_tmp/notbin/hello
+printf '#!/usr/bin/env sh\necho hi\n' >_tmp/bin/hello
+chmod +x _tmp/bin/hello
+
+DIR=$PWD/_tmp/dir
+BIN=$PWD/_tmp/bin
+NOTBIN=$PWD/_tmp/notbin
+
+# The command resolution will search the path for matching *files* (not
+# directories) WITH the execute bit set.
+
+# Should find executable hello right away and run it
+PATH="$BIN:$PATH" hello
+echo status=$?
+
+hash -r  # Needed to clear the PATH cache
+
+# Will see hello dir, skip it and then find&run the hello exe
+PATH="$DIR:$BIN:$PATH" hello
+echo status=$?
+
+hash -r  # Needed to clear the PATH cache
+
+# Will see hello (non-executable) file, skip it and then find&run the hello exe
+PATH="$NOTBIN:$BIN:$PATH" hello
+echo status=$?
+
+## STDOUT:
+hi
+status=0
+hi
+status=0
+hi
+status=0
+## END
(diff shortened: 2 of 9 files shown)


PossiblyAShrub:
This is ready for a review.

cpp/pylib.cc
```diff
@@ -55,4 +55,12 @@ bool isdir(BigStr* path) {
   return S_ISDIR(st.st_mode);
 }
 
+bool isfile(BigStr* path) {
+  struct stat st;
+  if (::stat(path->data_, &st) < 0) {
+    return false;
+  }
+  return !S_ISDIR(st.st_mode);
```
andychu: There are more options than directories and files, so this should not test !S_ISDIR()

Looks like it should be `S_ISREG()` for a regular file

The Python stdlib bottoms out here:

https://github.com/oils-for-unix/oils/blob/master/Python-2.7.13/Lib/genericpath.py#L34

And I never remember where to look these things up.  Claude got it on the second try

https://claude.ai/share/4922349a-4f4e-42e5-a575-afb68683a748

First try `man 2 stat` was wrong, but them `man 7 inode` shows all the functions like S_ISDIR() and so forth

```
           S_ISREG(m)  is it a regular file?

           S_ISDIR(m)  directory?

           S_ISCHR(m)  character device?

           S_ISBLK(m)  block device?

           S_ISFIFO(m) FIFO (named pipe)?

           S_ISLNK(m)  symbolic link?  (Not in POSIX.1-1996.)

           S_ISSOCK(m) socket?  (Not in POSIX.1-1996.)

```

BUT I also want to check what other shells do

spec/command_.test.sh
```diff
@@ -193,3 +193,54 @@ echo status=$?
 ## stdout: status=1
 ## OK osh stdout: status=2
 ## BUG dash/bash stdout: status=0
+
+#### Executing command with same name as directory in PATH (#2429)
+
+# Make the following directory structure. File type and permission bits are
```
andychu, line 199: nice comment!

andychu (commented):
Thanks for finding the other related bugs !  

As mentioned,  isfile() should be different

But then I also want to check what other shells do -- how exactly they exclude directories


---

And a weird coincidence is that someone just dug up this bug from 2020!

https://github.com/oils-for-unix/oils/issues/689#issuecomment-3367333428

( I may not have tested the other shells correctly then )

andychu:
OK I found this file - https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/exec.c

and then

```
/*
 * Resolve a command name.  If you change this routine, you may have to
 * change the shellexec routine as well.
 */

void
find_command(char *name, struct cmdentry *entry, int act, const char *path)
{
```

and then it checks if the file is executable

```
		if (!test_exec(fullname, &statb))
			continue;
```


and then it does use S_ISREG

```
static int test_exec(const char *fullname, struct stat64 *statb)
{
	if (!S_ISREG(statb->st_mode))
		return 0;

	if ((statb->st_mode & 0111) != 0111 &&
#ifdef HAVE_FACCESSAT
	    !test_file_access(fullname, X_OK)
#else
	    !test_access(statb, X_OK)
#endif
	   )
		return 0;

	return 1;
}
```


So using `isfile()` seems like it's OK, with the `S_ISREG()` correction



---

I am not sure why it checks mode `0111`, but we can start with this

Or maybe consult bash

andychu:
Oh I also noticed it does the single stat() call, and then reuses it for the  `S_ISREG()` and then the `X_OK` check


```

	if (!S_ISREG(statb->st_mode))
		return 0;

	if ((statb->st_mode & 0111) != 0111 &&
#ifdef HAVE_FACCESSAT
	    !test_file_access(fullname, X_OK)
#else
	    !test_access(statb, X_OK)
```


I am not sure if `posix.access(path)` and then `stat() S_ISREG()` does more sys calls, but it seems like it may


I am not sure if this matters though ... I guess I think that passing a path NAME twice could be a little odd.  And what if the file is CHANGED between the two calls

I think a single stat() call could make the check atomic

But the dash code has an `#ifdef`, hmmm

---

I guess  we should check if this fixes the bug before diving into that

andychu:
Looking at `findcmd.c` in bash, it does call stat() first with the name, and then eaccess X_OK with the name too

```
   if (stat (name, &finfo) < 0)
     return (0);

...

   /* Use eaccess(2) if we have it to take things like ACLs and other
      file access mechanisms into account.  eaccess uses the effective
      user and group IDs, not the real ones.  We could use sh_eaccess,
      but we don't want any special treatment for /dev/fd. */
   if (exec_name_should_ignore (name) == 0 && eaccess (name, X_OK) == 0)
     r |= FS_EXECABLE;

```

So probably the two checks we are doing are fine

andychu:
And thanks for doing the changes across `cpp/` and `pyext/` 

I think that is an annoying artifact of our metaprogramming style, and some people can stuck there

It's probably something that can go in the dev guide, and we can even link to commits like this

PossiblyAShrub:
Okay, this is ready for another review.

I am going to draft a write up of what all the APIs are here and put it into a bug post-mortem. I did some research myself and it seems like all the components are:

1. Call `stat` and check if the path resolves to an existing, regular file
```
	if (!S_ISREG(statb->st_mode))
		return 0;
```

2. (Optimization) if all the exec bits are set, then no need to call access, we can assume the file is executable
```
	if ((statb->st_mode & 0111) != 0111 &&
```

3. Call access (or, really, faccessat) if it's available. Bash uses an `#ifdef` here because it wants to support systems that don't have the faccessat syscall
```
#ifdef HAVE_FACCESSAT
	    !test_file_access(fullname, X_OK)
#else
	    !test_access(statb, X_OK)
```

Now the code does parts (1) and (3, without the `#ifdef`).

I am not sure if the optimization in (2) is actually fully correct. (3) will definitely check it, so it's safe to leave (2) out for now. I'll have to do some more research into (2). I think it is possible for no execute bits to be set in the traditional owner/group/others but still have execute permissions with [ACLs](https://www.redhat.com/en/blog/linux-access-control-lists). But I am not sure!

On that last question [this stack exchange entry](https://unix.stackexchange.com/questions/358494/setting-x-executable-bit-using-acl) seems to think that "it is possible for no execute bits to be set in the traditional owner/group/others but still have execute permissions with ACLs" is *false*. But I want to *test* that first.

andychu:
Great thank you!  

And thanks for finding the associated bugs ... the PATH= invalidation reminds me of the `LANG=` hook too

andychu:
The related hooks issue is

- https://github.com/oils-for-unix/oils/issues/527

PR #2449, opened 2025-10-05
mycpp: fix seg fault in pretty printer

melvinw:
The pretty printer found a gap in the stack roots optimizer. It was
falling over on code with the following form:

    x = alloc_something()...
    do_something(x, function_that_collects())

We werent' generating a root for `x` because its use isn't captured by
the live variable analysis. This commit adds a rule to the rooting
policy to capture this case and adds an example that we can use to check
for regressions.

### DIFF
--- a/mycpp/NINJA_subgraph.py
+++ b/mycpp/NINJA_subgraph.py
@@ -443,6 +443,8 @@ def MycppExamples(ru, ph):
                 impl = 'C++'
                 if translator == 'mycpp-nosouffle':
                     impl = 'C++-NoSouffle'
+                elif translator == 'mycpp-souffle':
+                    impl = 'C++-Souffle'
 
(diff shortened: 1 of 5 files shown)


andychu:
FWIW I computed the diff here:

https://github.com/oils-for-unix/mycpp-output/commit/4e0fd4db26701a9a2cddac71d86610e1fdc8eb0a#diff-96d60a33d4c764fae7fdb12525611c636ccdc1d76dec3b53b3904e68db1ccb6c

Our benchmarks are not informative now,  because it compares  nosouffle vs default, and nosouffle is the default

https://op.oilshell.org/uuu/github-jobs/10229/benchmarks2.wwz/_tmp/gc-cachegrind/index.html


Let me see about that

PR #2450, opened 2025-10-06
Deps build errors

btertoolen:
Fixed some small issues with building dependencies, which seem to no longer build when using gcc >= 14 (in the case of bash and zsh) or when using linux kernel >= 6.8 (busybox).

For busybox the recommended workaround was just disabling the module.

For bash and zsh the fix was treating certain compile errors as warnings (which was the default behavior for those errors before gcc 14). So I think these changes should also not really have an impact on existing working environments with an older gcc.

For reference here were the compilation issues:

zsh: `termcap.c:45:14: error: conflicting types for ‘boolcodes’; have ‘char *[]’`
busybox: `tc.c:296:24: error: invalid use of undefined type ‘struct tc_cbq_lssopt’`
bash: `parse.y:5741:21: error: implicit declaration of function ‘count_all_jobs’ [-Wimplicit-function-declaration]`

### DIFF
--- a/deps/source.medo/bash/WEDGE
+++ b/deps/source.medo/bash/WEDGE
@@ -20,7 +20,7 @@ wedge-make() {
 
   # --without-bash-malloc for Alpine only -- musl libc incompatibility?
   # https://patchwork.ozlabs.org/project/buildroot/patch/20170523171931.18744-1-dsabogalcc@gmail.com/
-  time $src_dir/configure --without-bash-malloc --prefix=$install_dir CFLAGS="-std=gnu99"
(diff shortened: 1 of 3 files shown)


andychu:
Perfect, thank you!

I pushed this change to sourcehut, and Fedora now passes !

https://op.oilshell.org/uuu/sourcehut-jobs/git-7eba547cb39389815327aeb4c58cf200723bddb4/

https://op.oilshell.org/uuu/sourcehut-jobs/1583292/dev-setup-fedora.wwz/_build/wedge/logs/index.html

Issue #2198, opened 2024-12-19
trap without args should print traps


andychu:
I ran into this in some real-world code of mine; I have an "atexit" library I use, and as a sanity check, it greps for EXIT in the output of `trap`; this is not perfect, in that it can have false-positives, but I like that better than stomping on another EXIT handler.

_Originally posted by @jasom in https://github.com/oils-for-unix/oils/issues/1881#issuecomment-2305495947_

andychu:
Sort of related: #2339

andychu:
I would count this as the post-mortem - >
[Lessons From Improving the Trap Builtin](https://oilshell.zulipchat.com/#narrow/channel/266575-blog-ideas/topic/Lessons.20From.20Improving.20the.20Trap.20Builtin/with/544343545)

Issue #2370, opened 2025-07-18
The trap builtin only registers the first given signal


quexxon:
In a difference from the POSIX shell [trap builtin](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_29), Oils doesn't support registering multiple signals with the trap handler. There is no error if multiple signals are provided, but only the first will be registered.

## OSH
```
osh-0.34$ trap ps TERM INT EXIT
osh-0.34$ trap -p
15 TrapState
osh-0.34$ exit
# no ps invocation
```
```
osh-0.34$ trap ps EXIT
osh-0.34$ trap -p
EXIT TrapState
osh-0.34$ exit
    PID TTY          TIME CMD
1616736 pts/1    00:00:09 fish
2128462 pts/1    00:00:00 osh
2128866 pts/1    00:00:00 ps
```

## Bash
```
bash$ trap ps TERM INT EXIT
bash$ trap -p
trap -- 'ps' EXIT
trap -- 'ps' SIGINT
trap -- 'ps' SIGTERM
bash$ exit
exit
    PID TTY          TIME CMD
1616736 pts/1    00:00:09 fish
2129031 pts/1    00:00:00 bash
2131675 pts/1    00:00:00 ps
```

Issue #1082, opened 2022-01-27
Shell constructs that print source code


andychu:
- `trap -p` lists traps
- `jobs -l` prints a pipeline in a very specific format
- `set` and `declare` can  print functions (ble.sh apparently uses some of this)
  - I think they are later `eval`'d

andychu:
This might be easier after new token representation

- #812

andychu:
The test framework needs this

```
assert [42 === x]
```

and then pretty printing also needs it

```
$ pp [42 + x]

   42 + x === 43

```

andychu:
- #1992

andychu:
Idea for this: maybe `ctx_SaveSourceLines` makes it a doubly linked list?

That can matter for multi-line expressions?

We already have it for BraceGroup

But would we want it for entire `func` and `proc` signatures?  I guess so

Hm

I guess the only reason we keep the comment and blank lines in memory is because you MIGHT re-print them?

---

But bash loses the formattingt oo

```
$ bash -c '
> f() {
>   # comment
>   echo hi
> }
> declare -F f
> '
bash: line 6: declare: f: not found
f () 
{ 
    echo hi
}
f
```

So maybe we don't need to save those intermediate lines

We just need to associate the subset of lines with function nodes?  Hmm hm

PR #2453, opened 2025-10-08
Fix various issues with `trap` builtin

### LINKED ISSUE #2339: Implement 'trap 0' - mawk package on Alpine
[#osh > alpine build failures - mawk - trap 0 requires arguments](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/alpine.20build.20failures.20-.20mawk.20-.20trap.200.20requires.20arguments/with/518913398)

last-genius:
Makes the builtin `trap` much closer to bash - fixing some of the disagreements that triggered build issues/broke tools.

1. `trap` without args should act like `trap -p`, printing traps
2. `trap -p` should be more informative and helpful
3. `trap SIGNAL_NAME` should reset the handler for that signal
4. `trap SIGNAL_NUM` should reset the handler for that signal
5. `trap SIGNAL_NUM (SIGNAL_NUM|SIGNAL_NAME)*` should reset handlers for all the signals mentioned

All of these are fixed now - with spec tests making sure we don't regress them.

Fixes #2339 and #2198.

---

There is one more unresolved issue (=1 failing spec test):

6. `trap` does not write into a pipe, see e.g. `trap -p | cat -` vs. `trap -p`. This means osh's `trap` output can't be filtered with `grep` and much more.

I'm still looking into this one...

### DIFF
--- a/spec/builtin-process.test.sh
+++ b/spec/builtin-process.test.sh
@@ -1,5 +1,8 @@
-## oils_failures_allowed: 2
 ## compare_shells: dash bash mksh zsh
+## oils_failures_allowed: 2
+## oils_cpp_failures_allowed: 3
+# case #24 with ulimit -f 1 is different under C++ for some reason - could be due to the python2
+# intepreter and SIGXFSZ
 
 #### exec builtin 
 exec echo hi
@@ -413,6 +416,7 @@ after 1
 
 
 #### ulimit -f 1 prevents files larger 512 bytes
+trap - XFSZ  # don't handle this
 
 rm -f err.txt
 touch err.txt
@@ -448,8 +452,8 @@ cat err.txt
 512 status=0
 ## END
 
-## OK osh status: 0
-## OK osh STDOUT:
+## OK disabledosh status: 0
+## OK disabledosh STDOUT:
 512 status=0
 513 status=0
 
@@ -488,6 +492,8 @@ echo outer=$?
 ## STDOUT:
 outer=153
 ## END
+
+# not sure why this is different
 ## OK osh STDOUT:
 inner=1
 outer=0


andychu:
Wow thanks for making so much progress on this, quickly!  I am taking a look

I tried `trap -p | cat`, and am surprised by that issue !!!   Hmmm ...

andychu:
Wow this is super weird: `trap -l | cat` works, but `trap -p | cat` doesn't ?   But they are both right next to each other in the code, using `print()`

```
$ osh -c 'trap "echo hi" INT; trap -l | cat'
   EXIT
   ERR
   RETURN
   DEBUG
 1 SIGHUP

```

andychu:
FWIW I think one of the bugs here just came up in a new package failure!

https://op.oils.pub/aports-build/2025-10-08-comm.wwz/_tmp/aports-report/2025-10-08-comm/shard3/osh-as-sh/test-suite/bogofilter/test-suite.log.txt


```
  oldtraps=`trap`
            ^
[ backticks in ./t.lock3 ]:18: 'trap' requires a code string
```

So thanks for working on this!  I will take a closer look

spec/builtin-trap.test.sh
```diff
@@ -33,7 +33,135 @@ echo $?
 0
 ## END
 
+#### trap -p should print the handlers and full signal names
+case $SH in dash) exit ;; esac
+trap "echo INT" INT
+trap "echo EXIT" EXIT
+# osh's trap mishandles pipes, a separate test verifies them
```
andychu: Thanks for finding this!  I started a thread on Zulip about it, and am surprised by this bug, but yes let's address it separately

Sorry you had to work around it everywhere !!

https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/odd.20bug.3A.20trap.20-p.20.7C.20cat/with/543832915

spec/builtin-trap.test.sh
```diff
@@ -33,7 +33,135 @@ echo $?
...
+# bash installs several default handlers, so grep filters them out for easier testing
+case $SH in bash) trap -p | grep echo ;;
```
andychu: Hm actually I thought the `case $SH in bash)` was to work around the `trap -p | cat` bug, but now it doesn't appear so

I don't see the default handlers you're mentioning

Can you reproduce this on your machine?

```
$ . build/dev-shell.sh 

$ which bash
/home/andy/wedge/oils-for-unix.org/pkg/bash/5.2.21/bin/bash

$ bash -c 'trap -p'
<no output>
```

This requires the `build/deps.sh install-wedges-fast` step from https://github.com/oils-for-unix/oils/wiki/Contributing


i.e. it seems like bash starts with an empty set of traps too

Actually it doesn't seem to depend that much on the bash version, but we have an exact version, so we can get the same behavior

spec/builtin-trap.test.sh
```diff
@@ -33,7 +33,135 @@ echo $?
...
+trap EXIT
+case $SH in bash) trap -p | grep echo ;;
+               *) trap -p;;
+esac
+echo ---
+trap INT
+case $SH in bash) trap -p | grep echo ;;
+               *) trap -p;;
+esac
+# osh, as opposed to bash here, does not preserve the exit code
```
andychu: Hm it looks like `bash`  runs `grep` here, not `osh` ?    So this comment would be misleading

But as mentioned above, I *think* we should be able to get rid of the `case $SH` for bash

spec/builtin-trap.test.sh
```diff
@@ -33,7 +33,135 @@ echo $?
...
+---
+---
+## END
+## N-I mksh status: 1
+## N-I dash/ash status: 2
+## N-I dash/mksh/ash STDOUT:
+EXIT
+## END
+
 #### trap without args prints traps, like trap -p
```
andychu: Hm does this comment match the test case?  I don't see trap without args?

spec/builtin-trap.test.sh
```diff
@@ -94,16 +221,9 @@ status=0
...
-## STDOUT:
-status=2
-## END
-## OK dash/ash STDOUT:
-status=1
-## END
-## BUG mksh STDOUT:
-status=0
-## END
+## status: 2
```
andychu: The style here WAS a bit inconsistent ... but actually what I have settled only lately is the following

```
trap foo
if test $? -ne 0; then
  echo ok
fi

## STDOUT:
ok
## END
## BUG mksh STDOUT:
## END

```

The reason is that this makes it easier to compare against OTHER shells.  I started doing this a few months ago:

https://pages.oils.pub/spec-compat/2025-09-14/renamed-tmp/spec/compat/PASSING.html

In many cases, it makes sense to specify zero or non-zero -- the exact exit code can be a bit "overspecified"

frontend/signal_def.py
```diff
@@ -82,7 +84,12 @@ def GetNumber(sig_spec):
     return _SIGNAL_NAMES.get(sig_spec, NO_SIGNAL)
 
 
-_SIGNAL_NAMES = _MakeSignals()
+def GetName(sig_num):
+    # type: (int) -> str
+    return _SIGNAL_NUMBERS.get(sig_num)
+
+
+_SIGNAL_NAMES, _SIGNAL_NUMBERS = _MakeSignals()
```
andychu, line 92: Ah, I see we didn't support signal numbers at all.  Thank you for adding this!

builtin/trap_osh.py
```diff
@@ -229,19 +230,60 @@ def _ParseTrapCode(self, code_str):
 
         return node
 
+    def _GetCommandString(self, body):
+        # type: (command_t) -> str
+        doc = 'TrapState'  # type: str
```
andychu: hm we should probably have some other default?

builtin/trap_osh.py
```diff
@@ -229,19 +230,60 @@ def _ParseTrapCode(self, code_str):
...
+        if body.tag() == command_e.Simple:
+            bgroup = cast(command.Simple, body)
```
andychu: Ah I think this shouldn't be called `bgroup` - I thought it was BraceGroup

It could be `simple` or `simple_cmd` I guess, for command.Simple

builtin/trap_osh.py
```diff
@@ -229,19 +230,60 @@ def _ParseTrapCode(self, code_str):
...
+            if bgroup.blame_tok:
+                doc = bgroup.blame_tok.line.content
```
andychu: I think getting the first line is not what bash does -- although of course other shells DIFFER too

So we have to decide on a behavior

This makes sense for now, since the tests pass, but we might construct some more test cases later, and adjust this

Related issue: https://github.com/oils-for-unix/oils/issues/1082

Thanks for getting it working!

builtin/trap_osh.py
```diff
@@ -229,19 +230,60 @@ def _ParseTrapCode(self, code_str):
 
         return node
 
+    def _GetCommandString(self, body):
```
andychu: This function could use a docstring, or maybe renaming it to `_GetCommandSourceCode()` or `_GetSourceCode()` is enough

Or `_GetSourceSnippet()`

builtin/trap_osh.py
```diff
@@ -229,19 +230,60 @@ def _ParseTrapCode(self, code_str):
...
         attrs, arg_r = flag_util.ParseCmdVal('trap', cmd_val)
         arg = arg_types.trap(attrs.attrs)
 
-        if arg.p:  # Print registered handlers
-            # The unit tests rely on this being one line.
-            # bash prints a line that can be re-parsed.
-            for name, _ in iteritems(self.trap_state.hooks):
-                print('%s TrapState' % (name, ))
+        # 'trap' with no arguments is equivalent to 'trap -p'
+        if arg.p or arg_r.n == 1:  # Print registered handlers
```
andychu, line 276: Oh interesting you had to use `arg_r.n` here -- I don't think we do that elsewhere

I guess the `trap` builtin syntax is very  unusual


We usually use the public methods `Read(), ReadRequired(), AtEnd()` etc

builtin/trap_osh.py
```diff
@@ -229,19 +230,60 @@ def _ParseTrapCode(self, code_str):
...
+            for name, handler in iteritems(self.trap_state.hooks):
+                print("trap -- '%s' %s" %
```
andychu: The source snippet may contain single quotes, in which case `'%s'` will not be valid

Let's see what other shells do


OK bash does quoting

```
$ trap "echo 'hi'" INT; trap -p
trap -- 'echo '\''hi'\''' SIGINT

```

dash does different quoting, but same idea:

```
$ trap "echo 'hi'" INT; trap   
trap -- 'echo '"'"'hi'"'" INT
```

Let's use the function `j8_lite.ShellEncode(s)` - I think that will give similar results

builtin/trap_osh.py
```diff
@@ -229,19 +230,60 @@ def _ParseTrapCode(self, code_str):
...
+                      (self._GetCommandString(handler), name))
 
-            for sig_num, _ in iteritems(self.trap_state.traps):
-                print('%d TrapState' % (sig_num, ))
+            for sig_num, handler in iteritems(self.trap_state.traps):
+                sig_name = signal_def.GetName(sig_num)
+                print("trap -- '%s' %s" %
```
andychu: ditto here for the quoting -- it might be worth extracting a helper function, to have the format in one place

builtin/trap_osh.py
```diff
@@ -254,42 +296,48 @@ def Run(self, cmd_val):
...
 
         code_str, code_loc = arg_r.ReadRequired2('requires a code string')
-        sig_spec, sig_loc = arg_r.ReadRequired2(
-            'requires a signal or hook name')
+        # Per POSIX, if the first argument to trap is an unsigned integer
+        # then reset every condition
+        # https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/V3_chap02.html#tag_18_28
+        if _IsUnsignedInteger(code_str, code_loc):
+            code_num = int(code_str)
+            # 0 is the number for the EXIT pseudo-signal in bash
```
andychu: Hm I am thinking of how to consolidate the parsing a bit, since the `trap` syntax is so messy

There were some discussions on lobste.rs recently about "parse, don't validate" and "make illegal states representable"

But I think we can do that later; the more important thing is that the tests are thorough, so future refactorings don't break things

And they seem to look good -- although I don't remember if I saw one for resetting multiple signals with names

e.g. some cases I came up with

```
$ bash -c 'trap "echo int" INT; trap "echo e" EXIT; trap - int 0 3; trap '

$ bash -c 'trap "echo int" INT; trap "echo e" EXIT; trap - int 0 3 -99; trap '
bash: line 1: trap: -99: invalid signal specification
```

builtin/trap_osh.py
```diff
@@ -254,42 +296,48 @@ def Run(self, cmd_val):
...
-                self.trap_state.RemoveUserHook(sig_key)
-                return 0
-
-            if sig_num != signal_def.NO_SIGNAL:
-                self.trap_state.RemoveUserTrap(sig_num)
-                return 0
-
-            raise AssertionError('Signal or trap')
+        if code_str == '-':
+            self._ResetSignal(sig_key, sig_num)
```
andychu: Hm yeah we call _ResetSignal 3 times, because there are 3 syntaxes

I kinda want to consolidate that into a type like:

```
trap_mutation =
  Add(int trap_id, string code)
| Remove(int trap_id)
```

But that doesn't have to be in this PR

(We use algebraic data types with Zephyr ASDL, e.g. */*.asdl, which is from CPython.   But it may take a bit of getting used to.  Then again it seems like you dug into the  `signal_def.py` metaprogramming, which I appreciate !  )

spec/builtin-trap.test.sh
last-genius: `bash -c 'trap -p'` is empty for me as well, but it isn't in the spec tests. Without grep:
```
[bash stdout]
Expected "trap -- 'echo EXIT' EXIT\ntrap -- 'echo INT' SIGINT\nEXIT\n"
Got      "trap -- 'echo EXIT' EXIT\ntrap -- 'echo INT' SIGINT\ntrap -- '' SIGPIPE\ntrap -- '' SIGXFSZ\nEXIT\n
```

spec/builtin-trap.test.sh
last-genius: Fixed

spec/builtin-trap.test.sh
last-genius: > Hm it looks like bash runs grep here, not osh ? So this comment would be misleading

oh, right. I've changed the comment:
```
# since grep doesn't match anything, bash's exit code is 1 here, while
# osh's is 0. Add another successful echo command below to work around it
```

builtin/trap_osh.py
last-genius: This way it's similar to what `osh` does currently, but it's not particularly helpful, I agree

builtin/trap_osh.py
last-genius: thanks for the context. yeah, this is the simplest thing i could come up with that would work in these conditions. if we had a generic "print this AST as source code" function it'd be a much better fit here.

bar-g:
> 5\. `trap SIGNAL_NUM (SIGNAL_NUM|SIGNAL_NAME)*` should reset handlers for all the signals mentioned

Seems a general problem, also: The trap builtin only registers the first given signal #2370

builtin/trap_osh.py
last-genius: Fixed

builtin/trap_osh.py
last-genius: Renamed to `_GetCommandSourceCode()`

builtin/trap_osh.py
last-genius: Renamed.

spec/builtin-trap.test.sh
last-genius: Changed accordingly.

builtin/trap_osh.py
last-genius: Good catch! Fixed this now in a separate commit and added new spec tests

last-genius:
> Hm I am thinking of how to consolidate the parsing a bit, since the trap syntax is so messy

> I kinda want to consolidate that into a type like:
> 
> ```
> trap_mutation =
>   Add(int trap_id, string code)
> | Remove(int trap_id)
> ```

The logic *is* getting a bit convoluted, especially given this also needs fixing (thanks for pointing that out!):

> Seems a general problem, also: The trap builtin only registers the first given signal #2370

I'll look into using algebraic types here, thanks for the tip!

last-genius:
> Seems a general problem, also: The trap builtin only registers the first given signal https://github.com/oils-for-unix/oils/issues/2370

This should be fixed! I think the logic is looking a bit more sensible now - imo we could add a strict mode (discussed on Zulip) and algebraic types at a later point, and leave this PR as a collection of fixes.

last-genius:
argh, the failing cpp spec test is because the CPP and Python code sort traps differently somewhere...

CPP:
```
trap -- 'echo test' EXIT
trap -- 'echo test' SIGTERM
trap -- 'echo test' SIGINT
```
Python:
```
trap -- 'echo test' EXIT
trap -- 'echo test' SIGINT
trap -- 'echo test' SIGTERM
```

andychu:
expanding bug links for reference

- #2339 
- #2198 
- #2370

spec/builtin-trap.test.sh
andychu: Oh I was scratching my head, and figured  out why

```
>>> import subprocess
>>> subprocess.call(['bash', '-c', 'trap -p; echo hi'])
trap -- '' SIGPIPE
trap -- '' SIGXFSZ
hi
0

```


for some reason python2 calls subprocesses with a different signal state

Maybe we can do `trap - PIPE XFSZ` at the beginning of each test, which is a bit cleaner

spec/builtin-trap.test.sh
andychu: Hm I don't see a bare `trap` invocation?  unless github UI is misleading me

the title is "trap without args prints traps, like trap -p"

So I would expect to see `trap` somewhere

andychu:
> argh, the failing cpp spec test is because the CPP and Python code sort traps differently somewhere...
> 
> CPP:
> 
> ```
> trap -- 'echo test' EXIT
> trap -- 'echo test' SIGTERM
> trap -- 'echo test' SIGINT
> ```
> 
> Python:
> 
> ```
> trap -- 'echo test' EXIT
> trap -- 'echo test' SIGINT
> trap -- 'echo test' SIGTERM
> ```

Oh that is unfortunate ... that is probably because the dict sort order is slightly different in the runtimes !!

Gah

Well one fix is to pipe to `| sort`, but then the other very odd  bug might appear ...

---

I think it is reasonable then to change it in the metadata, and put a TODO on that spec test 

```
$ grep oils_cpp_failures spec/*
spec/array-sparse.test.sh:## oils_cpp_failures_allowed: 2
```

I will put the `trap|cat` bug up for grabs too -- I did look at it, but didn't see a reason

andychu:
And yes I think it is OK to leave this PR as a collection of fixes

and then we can refactor when doing OSH/YSH design

spec/builtin-trap.test.sh
last-genius: it's just below, L50:
```
case $SH in bash) trap | grep true ;;
               *) trap ;;
```

last-genius:
> Well one fix is to pipe to `| sort`, but then the other very odd bug might appear ...
> 
> I think it is reasonable then to change it in the metadata, and put a TODO on that spec test

I've decided to sort the output by signal IDs, that's what bash is doing.

But I'm getting this linker error:
```
undefined reference to `List<int>::sort()'
```

rolfan-ish:
- In this comment I document the behavior of gnu bash.
- It's purpose is purely informative.
- I post it in this issue is because it may prove useful.
- It hasn't been thoroughly tested, but it should be OK.
- Bash 5.3 was used for the source code inspection.
- Got a headache, but it was *fun*.

## Some unusual terms/syntax used
- `VAL`: Value, not empty, not an option
- `SIGNAL`: Value that represents a valid signal
- `SIGNUM`: Number that represents a valid signal
- `NO_<NAME>`: Isn't `<NAME>`
- `<PRE> & <PRO>`: Is `<PRE>` and `<PRO>`
- `<PRE> ^ <PRO>`: Is `<PRE>` xor `<PRO>`

## Source
The implementation of the builtin is found in `bultins/trap.def`, a file fed to a code generator, producing `bultins/trap.c`.

## Behavior (exhaustive when taking into account the next section)
### print usage `$?=0`
- `trap --help [-p] [-P] [-l] [-] [--] [VAL*]`
### print usage and error `$?=258=2`
- `trap NO_SIGNUM` *if* POSIX
- `trap NO_SIGNAL`
- `trap -VAL & NO_SIGNAL`
### print error about misuse `$?=258=2`
- `trap -P -p [-l] [-] [--] [VAL*]`
- `trap -P [-l]`
### list signal names
- `trap -l [[-P VAL*] ^ -p] [VAL*]`
### display all signals and their handlers
- `trap -p [SIGNAL*]` *if* POSIX
- `trap` *if* POSIX
Note: `trap -p` and `trap` are equal
### display all handlers and their signals
- `trap -p [SIGNAL*]` *if not* POSIX
- `trap` *if not* POSIX
Note: `trap -p` and `trap` are equal
### display handlers
- `trap -P SIGNAL*`
### set handler of signals
- `trap NO_SIGNUM SIGNAL*`
### revert handlers of signals
- `trap SIGNAL` *if not* POSIX
- `trap SIGNUM SIGNAL*`
- `trap - SIGNAL*`
### make signals ignored
- `trap '' SIGNAL*`

## Notes on the signals
- an execution error is emitted if a SIGNAL is not a signal `$?=1`
- when reverting some signals have special treatment, the following snippet is said special treatment (fed to clang-format):
```c
  switch (sig) {
  case SIGINT:
    /* XXX - should we do this if original disposition
       was SIG_IGN? Yes, because setup_async_signals can
       set the original disposition to SIG_IGN to keep
       restore_signal from restoring the SIGINT
       disposition from when the shell was invoked, and
       we want to allow an asynchronous subshell to
       use `trap' to restore the default disposition or
       set a trap command. */
    if (interactive)
      set_signal_handler(SIGINT, sigint_sighandler);
    /* special cases for interactive == 0 */
    else if (interactive_shell &&
             (sourcelevel || running_trap || parse_and_execute_level))
      set_signal_handler(SIGINT, sigint_sighandler);
    else
      set_signal_handler(SIGINT, termsig_sighandler);
    break;

  case SIGQUIT:
    /* Always ignore SIGQUIT, but allow a posix-mode shell
       that is running asynchronously and has ignored
       SIGQUIT to reset it to the default. POSIX interp 751. */
    if (posixly_correct && signal_is_async_ignored(SIGQUIT))
      set_signal_handler(SIGQUIT, termsig_sighandler);
    else
      set_signal_handler(SIGQUIT, SIG_IGN);
    break;
  case SIGTERM:
#if defined(JOB_CONTROL)
  case SIGTTIN:
  case SIGTTOU:
  case SIGTSTP:
#endif /* JOB_CONTROL */
    if (interactive)
      set_signal_handler(sig, SIG_IGN);
    break;
  }
```

andychu:
Oh this explains the stray signal handlers in Python 2 subprocess, but not Python 3 subprocess

That is annoying!  I might add a preexec_fn


https://docs.python.org/3/library/subprocess.html

> If restore_signals is true (the default) all signals that Python has set to SIG_IGN are restored to SIG_DFL in the child process before the exec. Currently this includes the SIGPIPE, SIGXFZ and SIGXFSZ signals. (POSIX only)

> Changed in version 3.2: restore_signals was added.

andychu:
OK yes you're right, we should order by signal number -- that's what bash does (and probably other shells)

I noticed a bit of a smell in the code -- `signal_def::PrintSignals()` is generated code, whereas really we want a way to iterate over signals and get the max one

I want to reduce the amount of code gen

----


So I just allowed the sorting failure for now, with

```
## oils_cpp_failures_allowed: 3
```

and then I'll do some `signal_def.py` refactoring as a followup

---

Thank you for getting this all working!

andychu:
FWIW I fixed the sorting issue: https://github.com/oils-for-unix/oils/commit/c52b099ddad71bbea88864ec5d015d8778d87f35

I wanted to do a bunch of cleanup on the metaprogramming too -- we no longer need to generate a PrintSignals() for `trap -l` - that's just the same iteration  over the signal numbers, in order !

Thanks again

andychu:
Ah one thing I missed is that when the loop was added, the error location wasn't adjusted.  It always pointed to the first one

```
$ bin/osh -c 'trap - INT Z'                                                                        
  trap - INT Z                                                             
         ^~~                                                               
[ -c flag ]:1: Invalid signal or hook 'Z'                                  
```

For reference, I fixed it here: https://github.com/oils-for-unix/oils/commit/c5d4d1079e91a4b8c603b316eab775152db05e0b

```
$ bin/osh -c 'trap - INT Z'                                                                        
  trap - INT Z                                                             
             ^                                                             
[ -c flag ]:1: Invalid signal or hook 'Z'                   
```

One of our claims on the home page is to have precise error messages - https://oils.pub/osh.html  :-)

last-genius:
Thanks for fixing up the code nicely and fixing the sorting quirk!

PR #2454, opened 2025-10-08
implement read -u

### LINKED ISSUE #2340: Implement read -u
[#nix > read -u FD](https://oilshell.zulipchat.com/#narrow/channel/307442-nix/topic/read.20-u.20FD/with/499074800)

btertoolen:
Implement the -u flag for the read command, which is used to specify a specific file descriptor to use as input. 

Fixes #2340

### DIFF
--- a/spec/builtin-read.test.sh
+++ b/spec/builtin-read.test.sh
@@ -681,6 +681,8 @@ reply=1234
 ## N-I dash/mksh stdout-json: ""
 
 #### read -u 3 -d b -N 6
+case $SH in ash|zsh) exit ;; esac
+
 # file descriptor
 read -u 3 -d b -N 4 3<<EOF
 ababababa
@@ -695,7 +697,10 @@ echo reply=$REPLY
 reply=abab
 reply=ab
 ## END
-## N-I dash/mksh stdout-json: ""
+## N-I ash/zsh stdout-json: ""
+## BUG mksh stdout-json: ""
+## BUG mksh status: 2
+
 
 #### read -N doesn't respect delimiter, while read -n does
 case $SH in dash|zsh|ash) exit ;; esac


builtin/read_osh.py
```diff
@@ -444,22 +442,22 @@ def _Run(self, cmd_val):
...
-        # type: (arg_types.read, List[str]) -> int
+    def _Read(self, arg, names, input_fd):
+        # type: (arg_types.read, List[str], int) -> int
 
         # read a certain number of bytes, NOT respecting delimiter (-1 means
         # unset)
         arg_N = mops.BigTruncate(arg.N)
         if arg_N >= 0:
-            s = _ReadN(arg_N, self.cmd_ev)
+            s = _ReadN(input_fd, arg_N, self.cmd_ev)
```
andychu, line 460: It looks like this is `read -u 9 -N 5` --  how about adding a spec test for this too?

btertoolen: Good point! I added it

andychu: Hm looks like you added -d and not -N ?

I guess both test cases are worthwhile

btertoolen: My bad, I somehow complete mixed the 2 up in my head. But now they really should be there :)

andychu:
Ah looks like a test is still failing.   When I get a CI  failure, I go to

then I click on the failure - https://op.oilshell.org/uuu/github-jobs/10284/#job-ovm-tarball

then RESULTS - https://op.oilshell.org/uuu/github-jobs/10284/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

then I scroll down and see the new test passes for OSH and bash, but it fails for other shells

https://op.oilshell.org/uuu/github-jobs/10284/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-read.html

So it looks like the issue is that the other shells don't implement `-N`, and the assertions have to be adjusted

---

Sometimes I use the `case $SH in dash|zsh) exit ;; esac` pattern to  skip running tests for a shell, if they just don't have a feature

---

Thanks for working on this!

btertoolen:
zsh ash and dash indeed do not support it. Mksh is supposed to support it, but still the test fails. I put it as a bug for now, I dont think its worth looking into exactly what is going wrong right? If theres a better way to deal with this let me know

andychu:
Excellent, thank you!

PR #2456, opened 2025-10-09
Update aports.md

btertoolen:
Add a clarifying note on the expected directory layout for regtest scripts


andychu:
Looks good!

And thanks for updating the shell function names ... Right now only I'm running those (often on a rented Hetzner box), but I'd like others to run it too!

Issue #2409, opened 2025-09-07
sqsh - (test) Unexpected trailing word '--'


andychu:
```
  while ( test $# -ne 0 -a "$1" != "--" ); do
                                   ^
./autoconf/install-man:25: (test) Unexpected trailing word '--'
```

andychu:
https://op.oils.pub/aports-build/2025-08-26-ifs.wwz/_tmp/aports-report/2025-08-26-ifs/shard15/osh-as-sh/log/sqsh.log.txt

andychu:
Hm I tried to reproduce this, but can't?

```
osh -c 'test $# -ne 0 -a "$1" != "--" '
```

I do get that message with


```
$ osh -c 'test $# -ne 0 -a "$1" != "--" foo '
  test $# -ne 0 -a "$1" != "--" foo
                                ^~~
[ -c flag ]:1: (test) Unexpected trailing word 'foo'

```


OSH seems to match bash:

```
$ bash -c 'test $# -ne 0 -a "$1" != "--" foo '
bash: line 1: test: too many arguments

```

andychu:
The bug is still present in the latest run, weird - https://op.oils.pub/aports-build/2025-09-18-bash.wwz/_tmp/aports-report/2025-09-18-bash/shard14/osh-as-sh/log/sqsh.log.txt

kvrigor:
Minimal repro:


```sh
#
# Generate test script (based on aports/main/sqsh/src/sqsh-2.5/autoconf/install-man)
#
cat << 'EOF' > repro.sh
#!/bin/sh
#

has_error=0

while ( test $# -ne 0 -a "$1" != "--" ); do
        case $1 in
                -d)
                        shift
                        echo "Detected option: -d $1"
                        ;;
                -t)
                        shift
                        echo "Detected option: -t $1"
                        ;;
                -s)
                        shift
                        echo "Detected option: -s $1"
                        ;;
                *)
                        has_error=1
                        ;;
        esac
        shift
done

shift       # get rid of the -- in the options line

if (test ${has_error} -eq 1); then
        echo "use: repro.sh [-d dummy_arg] [-t dummy_arg] [-s dummy_arg]"
        exit 1
fi
EOF

#
# bash test (PASS)
#
bash repro.sh -d test
# Detected option: -d test

#
# osh test (FAIL)
#
osh repro.sh -d test
#  while ( test $# -ne 0 -a "$1" != "--" ); do
#                                   ^
[...]

PR #2457, opened 2025-10-10
Added conditional expression tests with unary operator-like input

kvrigor:
See https://github.com/oils-for-unix/oils/issues/2409#issuecomment-3390181101

### DIFF
--- a/doc/known-differences.md
+++ b/doc/known-differences.md
@@ -293,27 +293,6 @@ In ysh, you can use:
       echo 'exists'
     }  # => exists
 
-### Tokens in the conditional part of `test <cond1> <op> <cond2>` can be misinterpreted as unary operators
-
-Consider a script that takes a [POSIX-style] argument:
-
-    set -- -o
-    test $# -ne 0 -a "$1" != "--" && echo "Option passed is $1"
-    # ERROR: Unexpected trailing word '--'
-
-Here OSH would misinterpret '-o' as the logical OR operator and would throw
-an error. This doesn't happen if the argument doesn't happen to be a [bash conditional expression]:
-
-    set -- -y
-    test $# -ne 0 -a "$1" != "--" && echo "Option passed is $1"
-    # OK!
-
-Oddly enough, such misparsing never occurs in a single-condition `test`:
-
-    set -- -o
-    test "$1" != "--" && echo "Option passed is $1"
-    # OK!
-
 ## Data Structures
 
 ### Arrays aren't split inside ${}
@@ -528,5 +507,5 @@ External:
 [spec/command-sub]: ../test/spec.wwz/command-sub.html
 [spec/loop]: ../test/spec.wwz/loop.html
 [spec/alias]: ../test/spec.wwz/alias.html
-[POSIX-style]: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html
-[bash conditional expression]: https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html#Bash-Conditional-Expressions-1
+
+


kvrigor:
Added tests 27-29 at `test/spec.sh parse-errors`

image

kvrigor:
@andychu maybe we should test all flags in this [list](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html#Bash-Conditional-Expressions-1)?

spec/parse-errors.test.sh
```diff
@@ -219,3 +219,16 @@ echo =word
...
 =word
 ## END
+
+
+#### Single conditional with unary expression-like input
+test "-o" != "--"
+## status: 0
+
+#### Two conditionals with reserved unary expression
+test 1 -ne 0 -a "-o" != "--"
```
andychu: I think it helps to keep the test case fairly close to the `sqsh` code


```

$ osh -c 'set -- -o; test $# -ne 0 -a "-o" != "--" '
  set -- -o; test $# -ne 0 -a "-o" != "--"
                                      ^
[ -c flag ]:1: (test) Unexpected trailing word '--'
```

and then let's substitute -o

```
$ osh -c 'test $# -ne 0 -a "-o" != "--" '
  test $# -ne 0 -a "-o" != "--"
                           ^
[ -c flag ]:1: (test) Unexpected trailing word '--'
```


and then let's try not quoting it

```
$ osh -c 'test $# -ne 0 -a -o != "--" '
  test $# -ne 0 -a -o != "--"
                         ^
[ -c flag ]:1: (test) Unexpected trailing word '--'

```

The way I usually do this is


```
test CODE1
echo status=$?

test CODE2
echo status=$?

test CODE3
echo status=$?

```

And then we can see the pattern of 0 and 1 with OSH and other shells

That is a way to "zero-in" on the bug -- the conditions where it happens, and where it doesn't happen

andychu: And then let's try a different flag -- it still fails:

```
$ osh -c 'set -- -z; test $# -ne 0 -a "$1" != "--" '
  set -- -z; test $# -ne 0 -a "$1" != "--"
                                      ^
[ -c flag ]:1: (test) Unexpected trailing word '--'
```

then let's try a non-flag, it works now (agrees with other shells):

```
osh -c 'set -- z; test $# -ne 0 -a "$1" != "--" '
```

andychu:
```
osh -c 'set -- -o; test "$1" != "--" '
```

and this is similar to your case -- it doesn't disagree if we remove the clause before the  `-a`

andychu:
After I push, I  visit  https://op.oilshell.org/uuu/github-jobs/

and then look at the results

---

In this case,  I see a couple failures, so I click:

https://op.oilshell.org/uuu/github-jobs/10281/#job-ovm-tarball

then I click on "results"

https://op.oilshell.org/uuu/github-jobs/10281/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

If you scroll down, you see that `parse-errors` is failing


----

So then we need to bump up `oils_allowed_failures: 1` at the top


However, please move this  test to `spec/divergence` -- that is kind of a "grab bag" of "failing test cases where OSH diverges from other shells"


(`parse-errors` is for EXPECTED parse errors , to check that we issue the right errors.   But this is a wrong error, that will go away when we fix the bug)

---

Thank you!  This is very helpful

spec/parse-errors.test.sh
kvrigor: Suggested test changes done and moved them to `spec/divergence.test.sh`

spec/divergence.test.sh
```diff
@@ -137,3 +137,34 @@ status=0
 hey
 status=0
 ## END
+
+#### Unexpected trailing error
```
andychu: this should say which builtin it is, and where the bug came from, like

test builtin - Unexpected trailing word '--' - #2409

doc/known-differences.md
```diff
@@ -293,6 +293,27 @@ In ysh, you can use:
       echo 'exists'
     }  # => exists
 
+### Tokens in the conditional part of `test <cond1> <op> <cond2>` can be misinterpreted as unary operators
```
andychu: ???

We are going to FIX this bug!  It is necessary to run many scripts


We don't need this doc change

andychu:
The idea is to produce a failing test FIRST, so you or somebody else can verify that it's fixed

doc/known-differences.md
kvrigor: Whoops I thought all divergence tests have to be documented:

https://github.com/oils-for-unix/oils/blob/cc7920f08ac1db789658dab94e460b88d203f7ba/spec/divergence.test.sh#L4-L8

Anyway I've reverted this

andychu:
It is related, but the idea is that we EITHER fix a bug OR decide that it's a known difference

It's NOT always clear when we encounter a bug which category it fall in

For this one , it is clear -- there is no reason not to fix it

PR #2459, opened 2025-10-11
Add strict_arg_parse option in cd builtin

### LINKED ISSUE #2425: [osh as ash] nginx - cd got too many arguments - upstream APKBUILD bug
This is because `osh` is like `bash` and 

    cd /tmp /tmp    # error, 2 args

But it's not an error in busybox ash

---

Could send a patch upstream?

Or also hide it behind `shopt --set strict_arg_parse` for all builtins?   We have an inconsitsency here

https://op.oils.pub/aports-build/2025-09-17-ash2.wwz/_tmp/aports-report/2025-09-17-ash2/shard7/osh-as-sh/log/nginx.log.txt

EarthenSky:
Close #2425

Adds the `strict_arg_parse` option under `strict:all`, changing the default behavour of cd to allow additional (ignored) arguments. A small change, but a lot of learning the codebase!

```sh
osh-0.36# cd / /
osh-0.36# shopt -p strict:all
shopt -u strict_arg_parse
shopt -u strict_argv
...
osh-0.36# shopt -s strict:all
osh-0.36# shopt -p strict:all
shopt -s strict_arg_parse
shopt -s strict_argv
...
osh-0.36# cd / /
  cd / /
       ^
[ interactive ]:5: 'cd' got too many arguments
osh-0.36# exit
```

I've also tested that the `nginx` package now builds properly (no other incosistencies were hiding in the package!)

TODO:
- [x] Check whether I should add a test?
- [x] Figure out how to run the formatter
- [x] Decide whether to include more builtins under the `strict_arg_parse` option (probably yes)

### DIFF
--- a/spec/builtin-cd.test.sh
+++ b/spec/builtin-cd.test.sh
@@ -23,7 +23,7 @@ status=1
 status=0
 ## END
 
-#### strict cd with 2 or more args
+#### cd with 2 or more args - with strict_arg_parse
 
 shopt -s strict_arg_parse
 
@@ -46,19 +46,17 @@ status=0
 failed with multiple args
 ## END
 
-## OK dash/ash STDOUT:
+## N-I dash/ash STDOUT:
 status=0
 status=0
 ## END
 
-#### cd with 2 or more args
-
-shopt -u strict_arg_parse
+#### cd with 2 or more args is allowed (strict_arg_parse disabled)
 
 mkdir -p foo
 cd foo bar
 
-## OK ash/dash/osh status: 0
+## status: 0
 ## OK bash/zsh status: 1
 ## OK mksh status: 2
 


andychu:
This looks good!  And thanks for verifying nginx builds

Now I need to explain the CI - let me copy what I wrote in another PR :-)

Basically I visit https://op.oilshell.org/ , and follow the links

andychu:
Basically I follow the links, until - https://op.oilshell.org/uuu/github-jobs/10292/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

then I can see `builtin-cd` fails, which makes sense

https://op.oilshell.org/uuu/github-jobs/10292/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-cd.html

the one with 2 or more fails

----

I might SPLIT that into 2 cases

- existing test with `shopt --set strict_arg_parse`
- new test without 

And then you may have to fiddle with the assertions for other shells, because they will differ, or they don't support `strict_arg_parse`


and

have details

andychu:
I guess the idea is that we match busybox ash now, UNLESS strict_arg_parse, in which case we match the other shells

So the 2 test cases can reflect that

----

As far as running the formatter, it's on the cheat sheet -- `devtools/format.sh yapf-changed` is usually what I do (on a file that's committed to a branch)

andychu:
And for docs, please add a line here -- I use `grep` to find the right place in the reference

```
$ grep strict_ doc/ref/*
doc/ref/chap-cmd-lang.md:persist.  This is enforced by `shopt --set strict_env_binding`.
doc/ref/chap-mini-lang.md:- `shopt --set strict_word_eval` exposes "syntax errors" in `name[index]`, and
doc/ref/chap-option.md:      strict_parse_equals     # Disallow '=x' to avoid confusion with '= x'
doc/ref/chap-option.md:      strict_parse_slice      # No implicit length for ${a[@]::}
doc/ref/chap-option.md:    X strict_parse_utf8       # Source code must be valid UTF-8
doc/ref/chap-option.md:      strict_argv             # No empty argv
doc/ref/chap-option.md:      strict_arith            # Fatal parse errors (on by default)
doc/ref/chap-option.md:      strict_array            # Arrays and strings aren't confused

```


Thank you for working on this!

spec/builtin-cd.test.sh
```diff
@@ -44,11 +46,22 @@ status=0
 failed with multiple args
 ## END
 
-## BUG dash/ash STDOUT:
+## OK dash/ash STDOUT:
```
EarthenSky: I changed this to "OK" (we got a value that was not ideal, but expected) because it seems a bit more correct than "BUG - we verified the value of a known bug". I use OK for the non-strict case below as well, for consistency.

EarthenSky:
@andychu I'm mostly happy with the PR now (just fixing a parsing test failure from `pushd` due to strict_arg_parse not being set for the test). However, there are a lot more builtins that could support `strict_arg_parse` (in terms both of accepting additional arguments and failing when provided additional arguments).

Do you think it's valuable for me to implement more now, or should I make an issue for it to be done later?

andychu:
Hm I peeked at the CI

```
===== CASE: -c pushd x y =====


  pushd x y
        ^
[ -c flag ]:1: pushd: '/home/uke/oil/x': No such file or directory

test/parse-errors.sh: fatal: Should FAIL under _bin/cxx-asan/osh: expected status 2, got 1
```

Not sure we should change `pushd` here ... I think our goal is just to run `nginx`


----


```
$ busybox ash -c 'pushd x y; echo hi'
ash: pushd: not found
hi
```

Plus `pushd` does not exist in busybox, so we do not need it to be compatible with busybox

builtin/dirs_osh.py
```diff
@@ -221,7 +221,7 @@ def Run(self, cmd_val):
 
         arg_r.Next()
         extra, extra_loc = arg_r.Peek2()
-        if extra is not None:
+        if extra is not None and self.mem.exec_opts.strict_arg_parse():
```
andychu: As mentioned, I don't see any motivation for changing anything besides `cd`

frontend/option_def.py
```diff
@@ -138,9 +139,9 @@ def DoneWithImplementedOptions(self):
     ('for_loop_frames', False),
 ]
 
-# TODO: Add strict_arg_parse?  For example, 'trap 1 2 3' shouldn't be
+# TODO: Add wider support for strict_arg_parse?  For example, 'trap 1 2 3' shouldn't be
```
andychu: This `trap` comment is now obsolete --  I just learned that trap DOES accept multiple args!

But I think there are a few other examples of builtins that only USE one arg,  and maybe accept MORE

We could look at those in a follow-up

spec/builtin-cd.test.sh
```diff
@@ -23,7 +23,9 @@ status=1
 status=0
 ## END
 
-#### cd with 2 or more args
+#### strict cd with 2 or more args
```
andychu: I would add strict_arg_parse in the title, like:

cd with 2 or more args - with strict_arg_parse


It makes it easier to find/read later

spec/builtin-cd.test.sh
andychu: I would make this one N-I for "not implemented", because busybox and dash don't implement strict_arg_parse

(neither does bash, but it gives the same result here)

spec/builtin-cd.test.sh
```diff
@@ -44,11 +46,22 @@ status=0
...
 ## END
 
+#### cd with 2 or more args
+
+shopt -u strict_arg_parse
+
+mkdir -p foo
+cd foo bar
+
+## OK ash/dash/osh status: 0
```
andychu: I  would change this to `status: 0` without the qualification

Because the tests are for the "OSH spec"

spec/builtin-cd.test.sh
```diff
@@ -44,11 +46,22 @@ status=0
 failed with multiple args
 ## END
 
-## BUG dash/ash STDOUT:
+## OK dash/ash STDOUT:
 status=0
 status=0
 ## END
 
+#### cd with 2 or more args
```
andychu: how about "cd with 2 or more args is accepted (without strict_arg_parse)"

spec/builtin-cd.test.sh
```diff
@@ -44,11 +46,22 @@ status=0
...
+
+shopt -u strict_arg_parse
```
andychu: We can delete this line, just so that we're sure it's the default.  This is how cd behaves normally

builtin/dirs_osh.py
EarthenSky: Got you! My thinking was to implement little bug fixes here and there, but I can understand trying to keep change minimal, with the specific goal of compatibility building alpine packages.

I've removed the commit to change pushd and popd.

frontend/option_def.py
EarthenSky: Sounds good. I'll make an issue (https://github.com/oils-for-unix/oils/issues/2473) regarding this & remove the comment (https://github.com/oils-for-unix/oils/pull/2459/commits/985eaca490d13b47e9e6cc74e56c42ee5d3ecc9e).

spec/builtin-cd.test.sh
EarthenSky: Good idea! https://github.com/oils-for-unix/oils/pull/2459/commits/6625f1b4feffedf77e761cb2191294d952b4aa84

spec/builtin-cd.test.sh
EarthenSky: https://github.com/oils-for-unix/oils/pull/2459/commits/6625f1b4feffedf77e761cb2191294d952b4aa84

spec/builtin-cd.test.sh
EarthenSky: https://github.com/oils-for-unix/oils/pull/2459/commits/6625f1b4feffedf77e761cb2191294d952b4aa84

spec/builtin-cd.test.sh
EarthenSky: https://github.com/oils-for-unix/oils/pull/2459/commits/6625f1b4feffedf77e761cb2191294d952b4aa84

doc/ref/chap-option.md
```diff
@@ -198,6 +198,7 @@ Parsing options:
 
 Runtime options:
 
+      strict_arg_parse        # Disallow additional arguments to some builtins
```
andychu, line 201: Now I'm thinking `strict_extra_args` could be a good name, but meh it's actually OK if we change this later

(before 1.0)

builtin/dirs_osh.py
```diff
@@ -120,7 +120,7 @@ def Run(self, cmd_val):
         # At most 1 arg is accepted
         arg_r.Next()
         extra, extra_loc = arg_r.Peek2()
```
andychu, line 122: I should check if `arg_r.Done()` encapsulates this whole check, but that can be done later with #2473

andychu:
Excellent, thank you!

Issue #2466, opened 2025-10-12
'trap -p | cat' does not work - doesn't write anything to pipe?


andychu:
I looked into it, and it's very weird

[odd bug: trap -p | cat](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/odd.20bug.3A.20trap.20-p.20.7C.20cat/with/543825704)

andychu:
This is simply about trap in subshells!  duh

PR #2468, opened 2025-10-12
[spec/builtin-trap] Add test case for ignored traps in subshells

rolfan-ish:
POSIX specifies that if the action of trap is null the shell shall ignore each specified condition. Currently oils doesn't adhere.

POSIX also says:
> When a subshell is entered, traps that are not being ignored shall
> be set to the default actions, except in the case of a command
> substitution containing only a single trap command, when the traps
> need not be altered.

However `dash` and `bash` don't fully follow this behavior, which I am pretty sure is a bug in both shells. I haven't tested other shells yet, including oils.

TODO: Add the spec tests for the exception of traps.

### DIFF
--- a/spec/builtin-trap.test.sh
+++ b/spec/builtin-trap.test.sh
@@ -47,6 +47,25 @@ trap | cat
 bye
 ## END
 
+#### ignored signals are active inside subshells
+trap '' USR1
+subshell_result=$(
+  trap '' USR2
+  trap - USR2
+  trap
+)
+
+# NOT a subshell
+trap > traps.txt
+shell_result=$(cat traps.txt)
+
+if test "$shell_result" = "$subshell_result"; then
+  echo ok
+fi
+
+## STDOUT:
+ok
+## END
 
 #### trap accepts/ignores --
 trap -- 'echo hi' EXIT


andychu:
I would think of it in terms of what `strace` says each shell is doing, not necessarily what POSIX says

"ignored" is an ambiguous term

- does it mean it executes nothing?
- does it mean `SIG_IGN` ?

rolfan-ish:
Is it good now?
Otherwise I would appreciate any pointers to resources/examples in order to catch on the guidelines.

Note: Detailed information on the commits can be found in the commit messages.

rolfan-ish:
Looking at the source code the fix seems simple.
TODO: decide how much to adhere to POSIX (specially on `case 0`). As much as bash? As much as ash? Not at all like `dash` and `mksh`?

andychu:
Ah let's look at the results - https://op.oilshell.org/uuu/github-jobs/10320/

https://op.oilshell.org/uuu/github-jobs/10320/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

https://op.oilshell.org/uuu/github-jobs/10320/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-trap.html

andychu:
OK the general idea does look good !  Other shells pass and OSH fails

We need assertions for the other shells, like

```
## BUG mk STDOUT:
## END
```

See https://github.com/oils-for-unix/oils/wiki/Spec-Tests

---

And we are inhibited by

- https://github.com/oils-for-unix/oils/issues/2466

this is a really bad bug ... seems like it may block this spec test, gah

andychu:
I think we have to figure out what OSH does here first:

[Most shells don't respect traps in subshells: pipelines etc.](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Most.20shells.20don't.20respect.20traps.20in.20subshells.3A.20pipelines.20etc.2E/with/543825704)

I think these tests are a bit confused by this issue, especially because bash behaves differently

Let's discuss on Zulip about what we're trying to accomplish

PR #2471, opened 2025-10-13
Emulate BASH_VERSINFO

kvrigor:
From the [Bash guide](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-BASH_005fVERSINFO):

---

**BASH_VERSINFO**
A readonly array variable whose members hold version information for this instance of Bash. The values assigned to the array members are as follows:

- `BASH_VERSINFO[0]`: The major version number (the release).
- `BASH_VERSINFO[1]`: The minor version number (the version).
- `BASH_VERSINFO[2]`: The patch level.
- `BASH_VERSINFO[3]`: The build version.
- `BASH_VERSINFO[4]`: The release status (e.g., beta).
- `BASH_VERSINFO[5]`: The value of MACHTYPE.

### DIFF
--- a/core/shell.py
+++ b/core/shell.py
@@ -332,7 +332,7 @@ def Main(
         # 2025-09: bash 5.3 is the latest version; can increase this with
         # future Oils releases
         state.SetGlobalString(mem, 'BASH_VERSION', '5.3')
-        state.SetGlobalArray(mem, 'BASH_VERSINFO', ['5', '3', '0', '0', 'release', 'x86_64-pc-linux-gnu']) # major minor patch build release-status machine-type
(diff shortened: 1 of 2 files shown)


core/shell.py
```diff
@@ -332,6 +332,7 @@ def Main(
         # 2025-09: bash 5.3 is the latest version; can increase this with
         # future Oils releases
         state.SetGlobalString(mem, 'BASH_VERSION', '5.3')
+        state.SetGlobalArray(mem, 'BASH_VERSINFO', ['5', '3', '0', '1', 'release', 'x86_64-pc-linux-gnu']) # major minor patch build release-status machine-type
```
andychu: Let's put "unknown" for the last entry -- because Oils will be built on non x86_64 machines

And how did you choose  0 1 ?  That is kinda arbitrary ... but maybe 0 0 is better?

kvrigor: That was copy-pasted from my machine ... now fixed

Issue #2191, opened 2024-12-10
Redirecting to `/dev/null` fails with `noclobber`


Janfel:
Hi, I’ve noticed that OSH raises an error when redirecting to `/dev/null` while `noclobber` is set.

~~~ console
$ set -C
$ >/dev/null
  >/dev/null
  ^
[ interactive ]:2: Can't open '/dev/null': File exists (noclobber)
[ interactive ]:2: I/O error applying redirect: File exists
~~~

It seems that OSH applies `noclobber` to non-regular files as well as regular files. Compare POSIX (emphasis mine):

> Output redirection using the '>' format shall fail if the noclobber option is set (see the description of [set](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#set) -C) and the file named by the expansion of word exists **and is a regular file**. Otherwise, redirection using the '>' or ">|" formats shall cause the file whose name results from the expansion of word to be created and opened for output on the designated file descriptor, or standard output if none is specified. If the file does not exist, it shall be created; otherwise, it shall be truncated to be an empty file after being opened.
> – [POSIX 2018, Section 2.7.2 Redirecting Output](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_07_02)

andychu:
Thanks for noticing this!   and finding the documentation

If you have any time or interest, a patch would be very welcome -- I usually use `grep` to find the tests and the code


```
$ grep noclobber */*.py
core/process.py:                # noclobber flag is OR'd with other flags when allowed
core/process.py:                noclobber_mode = O_EXCL if self.exec_opts.noclobber() else 0
core/process.py:                    # NOTE: This is different than >| because it respects noclobber, but
core/process.py:                    mode = O_CREAT | O_WRONLY | O_TRUNC | noclobber_mode
core/process.py:                    mode = O_CREAT | O_WRONLY | O_APPEND | noclobber_mode
core/process.py:                    if e.errno == EEXIST and self.exec_opts.noclobber():
core/process.py:                        extra = ' (noclobber)'
frontend/option_def.py:    ('C', 'noclobber'),
frontend/option_def.py:    # set -o noclobber, etc.
osh/word_eval.py:    if exec_opts.noclobber():
```

and


```
andy@hoover:~/git/oilshell/oil$ grep noclobber spec/*

spec/redirect.test.sh:set -o noclobber
spec/sh-options.test.sh:#### noclobber off
[...]

Issue #37, opened 2017-09-14
runtime issues with OSH, bash_profile, etc.


andychu:
- [x] `test -r` (and `-w`)
- [x] `$UID` and `$EUID`

@icarroll I just implemented -r and -w here:

https://github.com/oilshell/oil/commit/89a89be9fd1ed5109e1631b327223e206cfda228

Let's use this thread to keep track of all the issues you encounter.

I believe issue #36 is also related, since you said the SSH connection closes on Ctrl-C.

icarroll:
What shell variables are specified by posix? Bash sets a large number of them, but I don't know how much shell code out in the wild depends on them being present.

andychu:
My strategy has been to implement on demand (as long as they are sane).  On issue #26 I implemented a lot of things for Nix.

Do you have a link to RedHat scripts and WSO2IS scripts, like this?

https://github.com/NixOS/nixpkgs/blob/master/pkgs/stdenv/generic/setup.sh#L318

Then I can just 'source' them and see what happens.

icarroll:
I'm sure they're available somewhere. If nothing else I can make my own fork here on github.

andychu:
OK great, it would be nice to have a configuration that we can both run.

It's easier to improve the error messages that way too.

icarroll:
I haven't been able to find the right repos on github so I'll just create a new repo and commit the relevant files.

andychu:
Closing this since it's been dormant for awhile.  Feel free to file new issues based on any problems you encounter.

PR #2474, opened 2025-10-14
Add noclobber spec tests

PossiblyAShrub:
Add tests for:
- https://github.com/oils-for-unix/oils/issues/2191
- https://github.com/oils-for-unix/oils/issues/2475

### DIFF
--- a/spec/sh-options.test.sh
+++ b/spec/sh-options.test.sh
@@ -530,17 +530,17 @@ show-strict() {
 show-strict
 shopt -s strict:all
 show-strict
-shopt -u strict_arith
+shopt -u strict_argv
 show-strict
 ## STDOUT:
+shopt -u strict_arg_parse
 shopt -u strict_argv
-shopt -u strict_arith
 -
+shopt -s strict_arg_parse
 shopt -s strict_argv
-shopt -s strict_arith
 -
-shopt -s strict_argv
-shopt -u strict_arith
+shopt -s strict_arg_parse
+shopt -u strict_argv
 -
 ## END
 ## N-I dash status: 2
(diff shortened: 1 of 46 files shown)


andychu:
https://op.oilshell.org/uuu/github-jobs/10328/

https://op.oilshell.org/uuu/github-jobs/10328/ovm-tarball.wwz/_tmp/spec/osh-py/redirect.html


Hm looks like dash fails the new case `#37` for some reason?

andychu:
Thanks for finding this!  That is pretty obscure, but makes sense

PossiblyAShrub:
> Hm looks like dash fails the new case #37 for some reason?

Yeah, dash is weird because redirects like `echo a &> /dev/null` will still place output on `stdout`:

```
$ dash -c 'echo a &> /dev/null' 2>/dev/null
a
$ dash -c 'echo a &> /dev/null' 1>/dev/null
$ # No output
```

spec/sh-options.test.sh
```diff
@@ -320,6 +320,92 @@ f.o
 .oo
 ## END
 
+#### noclobber on >>
+rm -f $TMP/no-clobber
```
andychu, line 324: BTW the spec test framework now defaults to creating a new dir for each test case, and cd's to that dir

So you no longer have to use $TMP, it is OK to just use `touch no-clobber`

spec/sh-options.test.sh
```diff
@@ -320,6 +320,92 @@ f.o
...
+## END
+
+#### noclobber on &> >
+set -C
+
+rm -f $TMP/no-clobber
+echo foo > $TMP/no-clobber
+echo stdout=$?
+echo bar > $TMP/no-clobber
+echo again=$?
```
andychu, line 343: One thing I've started doing to avoid having to specify `again=1` for some shells and `again=2` for  others is

```
if test $? -ne 0; then 
  echo ok
fi
```

and then you can assert STDOUT of 'ok'

This also helps the compatibility reports at https://pages.oils.pub/

Because it means we can get more accurate assessements of other shells, without writing specific assertions for them.  If they happen to use `again=2` or `again=3` even, it is generally OK

Issue #2469, opened 2025-10-13
py3-pylibmc - kill builtin isn't implemented


andychu:
https://op.oils.pub/aports-build/2025-10-08-comm.wwz/_tmp/aports-report/2025-10-08-comm/shard51/osh-as-sh/log/py3-pylibmc.log.txt

```
kill: invalid number '%%'
>>> ERROR: py3-pylibmc: check failed
```

andychu:
Notes:

- the `kill` builtin is just like the external kill command, **except** it supports the `%%` job syntax
  - right now I think `fg` and `bg` support `%` syntax
  - this may require some refactoring, but the first step is to make it work
- I would recommend looking at dash / busybox ash source code to see what the minimum `kill` builtin is

Here is the bash version, but the first step is to have a minimum version that simply implement `kill %%`, with tests

That will make the package work!

```
$ help kill
kill: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
    Send a signal to a job.
    
    Send the processes identified by PID or JOBSPEC the signal named by
    SIGSPEC or SIGNUM.  If neither SIGSPEC nor SIGNUM is present, then
    SIGTERM is assumed.
    
    Options:
      -s sig    SIG is a signal name
      -n sig    SIG is a signal number
      -l        list the signal names; if arguments follow `-l' they are
                assumed to be signal numbers for which names should be listed
      -L        synonym for -l
    
    Kill is a shell builtin for two reasons: it allows job IDs to be used
    instead of process IDs, and allows processes to be killed if the limit
    on processes that you can create is reached.
    
    Exit Status:
    Returns success unless an invalid option is given or an error occurs.

```

andychu:
MAY be helpful

[Good example of failing test cases, then Python/C++ change](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Good.20example.20of.20failing.20test.20cases.2C.20then.20Python.2FC.2B.2B.20change/with/525374830)

btertoolen:
I'd like to have a look at this if no one is working on it?

andychu:
Yes thanks (and please assign yourself!  )

andychu:
And I might start with `test/spec.sh builtin-kill`

I haven't looked at them closely -- Adam added them a few months ago:

```
$ git log spec/builtin-kill.test.sh
commit 6a9e7e6914ee5196cfbbd0538c2af6ab1521f901
[...]

PR #2478, opened 2025-10-14
Kill builtin

btertoolen:
Implement the kill builtin.

Some points im not sure about:
* I took the list of available signals as defined here: https://man7.org/linux/man-pages/man7/signal.7.html, but it lists some niche architectures that have different signal numbers. I dont know if we should take those into consideration or if they are too "legacy" anyway
* For some spec tests I didnt care about the stdout output, but the test framework kept complaining if i tried to leave the expected stdout output out of the test. So for some of them i just ended up redirecting to stdout
* I put the kill command in the process_osh.py file, but maybe a different grouping is preferred

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,9 +72,8 @@ build](https://op.oilshell.org/) tests them at every commit.
 
 ---
 
-Remember that this is the **dev build** in the Oils repo.  [The Oils Repo Is
-Different From the Tarball
-Releases](https://github.com/oils-for-unix/oils/wiki/The-Oils-Repo-Is-Different-From-the-Tarball-Releases).
+Remember that this is the **dev build** in Python, not a released binary.  [The
+Oils Repo Is Different From the Tarball Releases](https://github.com/oils-for-unix/oils/wiki/The-Oils-Repo-Is-Different-From-the-Tarball-Releases).
 
 ### Important: We Accept Small Contributions!
 


andychu:
I just pushed a change to fix some errors - although there is a sorting error left

Funny thing that this is exactly what @last-genius ran into on the `trap` PR - lack of sorting

---

However I just realized that `trap -l` and `kill -l` are identical in bash !

So you can extract a function from `builtin/trap_osh.py`

The one that looks like this

```
    def _PrintNames(self):
        # type: () -> None
        for hook_name in _HOOK_NAMES:
            # EXIT is 0, but we hide that
            print('   %s' % hook_name)

        # Iterate over signals and print them
        n = signal_def.MaxSigNumber() + 1
        for sig_num in xrange(n):

            sig_name = signal_def.GetName(sig_num)
            if sig_name is None:
                continue
            print('%2d %s' % (sig_num, sig_name))

```


I think we just print the signals, NOT the hooks.  So in OSH, `trap -l` and `kill -l` are slightly different


```
$ bin/osh -c 'trap -l'
   EXIT
   ERR
   RETURN
   DEBUG
 1 SIGHUP
 2 SIGINT
 3 SIGQUIT
 4 SIGILL
 5 SIGTRAP

```

Thanks for working on this, and sorry about the bad mycpp error message!

andychu:
direct link to code: https://github.com/oils-for-unix/oils/blob/master/builtin/trap_osh.py#L313

builtin/process_osh.py
```diff
@@ -682,4 +683,167 @@ def Run(self, cmd_val):
...
+    """Send a signal to a process"""
+    def __init__(self, job_control, job_list, waiter):
+        # type: (process.JobControl, process.JobList, Waiter) -> None
+        self.job_control = job_control
+        self.job_list = job_list
+        self.waiter = waiter
+        self.exec_opts = waiter.exec_opts
+        # TODO: this may not be true for all systems,
+        # see https://man7.org/linux/man-pages/man7/signal.7.html
+        self.sigspec_to_signame = {
```
andychu: It looks like this can use the same code gen as `builtin/trap_osh.py` - it uses functions from `signal_def` to turn names into numbers, and vice versa

Let me know if not

btertoolen: Done!

builtin/process_osh.py
```diff
@@ -682,4 +683,87 @@ def Run(self, cmd_val):
...
+                        return 1
+
+            try:
+                target_pid = int(cmd_val.argv[-1])
+            except ValueError:
+                print_stderr("error: invalid process id provided")
+                return 1
+
+        if signal_to_send == signal_def.NO_SIGNAL:
+            signal_to_send = 15
```
andychu: How about adding a comment that this is SIGTERM?

Also I just learned that bash lets you convert numbers to names like this:

```
$ kill -l 15 1 0
TERM
HUP
EXIT
```

and then it also does the inverse

```
$ kill -L TERM HUP
15
1

```

builtin/process_osh.py
```diff
@@ -682,4 +683,87 @@ def Run(self, cmd_val):
...
+        return signal_def.GetNumber(signal_name)
+
+    
+    def Run(self, cmd_val):
+        # type: (cmd_value.Argv) -> int
+
+        signal_to_send = signal_def.NO_SIGNAL
+        target_pid = 0
+
+        if len(cmd_val.argv) == 3 and cmd_val.argv[1].startswith('-'):
```
andychu: I think it is nicer to use the arg_r.Peek2() API   and `arg_r.Next()` API here

It gives you the arg and a location. 

You can construct the args.Reader directly without calling ParseCmdVal() like this:

```
$ grep '= args.Reader' */*.py
bin/oils_for_unix.py:    arg_r = args.Reader(argv, locs=[missing] * len(argv))
bin/osh_eval.py:    arg_r = args.Reader(argv, [missing] * len(argv))
builtin/assign_osh.py:        arg_r = args.Reader(cmd_val.argv, locs=cmd_val.arg_locs)
builtin/assign_osh.py:        arg_r = args.Reader(cmd_val.argv, locs=cmd_val.arg_locs)
builtin/assign_osh.py:        arg_r = args.Reader(cmd_val.argv, locs=cmd_val.arg_locs)
builtin/completion_osh.py:        arg_r = args.Reader(cmd_val.argv, cmd_val.arg_locs)

```

builtin/process_osh.py
```diff
@@ -682,4 +683,87 @@ def Run(self, cmd_val):
...
+            elif len(first_arg[1:]) > 2:
+                signal_to_send = self._SignameToSignum(first_arg[1:])
+                if(signal_to_send < 0):
+                    return 1
+
+            if signal_to_send != signal_def.NO_SIGNAL:
+                try:
+                    target_pid = int(cmd_val.argv[2])
+                except ValueError:
+                    print_stderr("error: invalid process id provided")
```
andychu: the `errfmt` API wlil let you print an error message with the location

Example from this file:


```
$ bin/osh -c 'wait 123'
  wait 123
       ^~~
[ -c flag ]:1: Job 123 was't found

```


```
               self.errfmt.Print_("Job %s was't found" % job_id,
                                   blame_loc=location)
                return 127

```

builtin/process_osh.py
```diff
@@ -682,4 +683,87 @@ def Run(self, cmd_val):
...
+
+            try:
+                target_pid = int(cmd_val.argv[-1])
+            except ValueError:
+                print_stderr("error: invalid process id provided")
+                return 1
+
+        if signal_to_send == signal_def.NO_SIGNAL:
+            signal_to_send = 15
+        if(target_pid == 0):
```
andychu: Reading a required arg can use the  `arg_r.ReadRequired2` API

```
$ grep ReadRequired */*.py
builtin/hay_ysh.py:            var_name, _ = arg_r.ReadRequired2("expected variable name")
builtin/meta_oils.py:            code_str, eval_loc = arg_r.ReadRequired2('requires code string')
builtin/meta_oils.py:        path_arg, path_loc = arg_r.ReadRequired2('requires a file path')

```

builtin/process_osh.py
```diff
@@ -682,4 +683,87 @@ def Run(self, cmd_val):
...
+class Kill(vm._Builtin):
+    """Send a signal to a process"""
+    def __init__(self, job_control, job_list, waiter):
+        # type: (process.JobControl, process.JobList, Waiter) -> None
+        self.job_control = job_control
+        self.job_list = job_list
+        self.waiter = waiter
+        self.exec_opts = waiter.exec_opts
+
+    def _PrintSignals(self):
```
andychu: How about extracting a free function in `builtin_trap.py` and calling it from both places?  Then if we change the format, it will be changed in both `kill` and `trap`

builtin/process_osh.py
```diff
@@ -682,4 +683,87 @@ def Run(self, cmd_val):
...
 
 
+class Kill(vm._Builtin):
+    """Send a signal to a process"""
+    def __init__(self, job_control, job_list, waiter):
+        # type: (process.JobControl, process.JobList, Waiter) -> None
+        self.job_control = job_control
+        self.job_list = job_list
+        self.waiter = waiter
+        self.exec_opts = waiter.exec_opts
```
andychu: not using exec_opts?

actually it seems like we're not using most of these args?    Please remove them if not

andychu (commented):
Thanks for working on this!  It looks good

I would restructure the arg parsing a bit with arg.{Peek2,Next,ReadRequired2}

I think it will make the logic cleaner and less indented

Although we can also do that in a follow-up change if you like, since these tests pass

andychu:
Oh and also I remember the original failure in #2469 had `kill %%` 

I don't remember what our job parsing / lookup looks like now, but it possibly needs to be consolidated

builtin/process_osh.py
```diff
@@ -682,4 +684,119 @@ def Run(self, cmd_val):
...
+    def Run(self, cmd_val):
+        # type: (cmd_value.Argv) -> int
+
+        signal_to_send = signal_def.NO_SIGNAL
+        target_pid = 0
+
+        arg_r = args.Reader(cmd_val.argv, locs=cmd_val.arg_locs)
+        arg_r.Next()
+        sigspec_arg, sigspec_arg_loc = arg_r.Peek2()
+        if len(cmd_val.argv) == 3 and sigspec_arg.startswith('-'):
```
andychu: I think we can avoid checking the length of argv, and this whole condition


because below we check `if sigspec_arg.startswith('-')` anyway

builtin/process_osh.py
```diff
@@ -682,4 +684,119 @@ def Run(self, cmd_val):
...
+            # if we land in this if or elif statement,
+            # we are dealing with the -sigspec argument, which
+            # flag-util.ParseCmdVal cant deal with so we parse it manually
+            if(sigspec_arg.startswith('-')):
```
andychu: We don't need the  `()` around the if conditions (when they don't span multiple lines)

builtin/process_osh.py
```diff
@@ -682,4 +684,119 @@ def Run(self, cmd_val):
...
+                pid_arg, pid_arg_loc = arg_r.ReadRequired2("You must provide a process id")
+                target_pid = self._ParsePid(pid_arg, pid_arg_loc)
+
+            else:
+                first_positional, first_positional_loc = arg_r.ReadRequired2("You must provide a process id")
+                second_positional, second_positional_loc = arg_r.Peek2()
+                if second_positional is not None:
+                    # 1st positional is our signal to send, 2nd is the receiving pid
+                    signal_to_send = self._SignameToSignum(first_positional)
+                    if(signal_to_send < 0):
```
andychu: ditto, don't need `()`

builtin/process_osh.py
```diff
@@ -682,4 +684,119 @@ def Run(self, cmd_val):
...
+            attrs, arg_r = flag_util.ParseCmdVal('kill',
+                                                 cmd_val,
+                                                 accept_typed_args=False)
+            arg = arg_types.kill(attrs.attrs)
+            if arg.l or arg.L:
+                arg_l, arg_loc = arg_r.Peek2()
+                signums_to_list = [] # type: List[Tuple[int, loc_t]]
+                signames_to_list = [] # type: List[Tuple[str, loc_t]]
+                while arg_l is not None:
+                    if(arg_l.isdigit()):
```
andychu: ditto, don't need `()`

builtin/process_osh.py
```diff
@@ -682,4 +684,119 @@ def Run(self, cmd_val):
...
+                    if(signal_to_send < 0):
+                        e_usage('invalid signal specification %r' % sigspec_arg, sigspec_arg_loc)
+
+            if signal_to_send != signal_def.NO_SIGNAL:
+                arg_r.Next()
+                pid_arg, pid_arg_loc = arg_r.Peek2()
+                target_pid = self._ParsePid(pid_arg, pid_arg_loc)
+
+        # we didn't have to deal with a -sigspec argument, so we can parse as usual
+        if signal_to_send == signal_def.NO_SIGNAL:
```
andychu: I think we could return early for the `kill -TERM 123` case, so then we don't have to indent this `kill -l` logic under `if signal_to_send ...`

I like the early return style for the different usages (similar to the `trap` builtin)

builtin/trap_osh.py
```diff
@@ -309,14 +309,7 @@ def _PrintNames(self):
...
 
-        # Iterate over signals and print them
-        n = signal_def.MaxSigNumber() + 1
-        for sig_num in xrange(n):
-
-            sig_name = signal_def.GetName(sig_num)
-            if sig_name is None:
-                continue
-            print('%2d %s' % (sig_num, sig_name))
+        signal_def.PrintSignals()
```
andychu: As mentioned on Zulip, `signal_def.py` is not translated

So it could go in either `builtin/process_osh.py` I guess .... (there doesn't seem to a common library file in `builtin/`)

spec/builtin-kill.test.sh
```diff
@@ -28,3 +28,132 @@ echo $?  # Must be 137 (128 + SIGKILL)
...
+builtin kill -n 9 %-
+wait $pid
+echo $?
+## STDOUT:
+0
+## N-I dash/mksh STDOUT:
+## END
+
+
+#### List specific processes
```
andychu: s/ specific processes / specific signals/

andychu:
Oh also the Python code can be auto-formatted with:


```
devtools/format.sh install-yapf  # install a specific version that we need to agree on

...

devtools/format.sh yapf-changed  # better: make sure everything is committed on your branch,
                                 # it only auto formats what's changed
```


andychu:
I made a few changes to the code and tests and found some problems:

- the exit code `$?` of `kill` appears incorrect
  - we weren't testing this; we were only testing the exit code of  `wait`.  Now we test both
- The algorithm for `kill -l 10 TERM` doesn't match bash
  - this is mixed names and numbers
  - I enhanced the spec test to catch this; it now fails
- `kill -s NAME` didn't have a test?
- I believe `kill TERM 123` is misinterpreted
  - it's not the same as `kill -TERM 123` with a `-`
- I removed some dead code in the arg parsing
  - you can use the file `test/runtime-errors.sh` to test the error messages visually

```
$ test/runtime-errors.sh test-kill-builtin-usage 

===== CASE: -c kill --15 99 =====


  kill --15 99
       ^~~~
[ -c flag ]:1: 'kill' got invalid signal '--15'

===== CASE: -c kill -TE 99 =====


  kill -TE 99
       ^~~
[ -c flag ]:1: 'kill' got invalid signal '-TE'

===== CASE: -c kill -15 =====


  kill -15
  ^~~~
[ -c flag ]:1: 'kill' expected a PID

```

andychu:
Here  is an example of how `kill TERM` is not the same as `kill -TERM`

It's probably best not to use TERM since that's the default

maybe use HUP so then the differing exit code may catch the bug



```
andy@hoover:~/git/oils-for-unix/oils$ bash -c 'kill TERM b c; echo $?'
bash: line 1: kill: TERM: arguments must be process or job IDs
bash: line 1: kill: b: arguments must be process or job IDs
bash: line 1: kill: c: arguments must be process or job IDs
1
andy@hoover:~/git/oils-for-unix/oils$ mksh -c 'kill TERM b c; echo $?'
E: mksh: kill: TERM: arguments must be jobs or process IDs
E: mksh: kill: b: arguments must be jobs or process IDs
E: mksh: kill: c: arguments must be jobs or process IDs
1
andy@hoover:~/git/oils-for-unix/oils$ bin/osh -c 'kill TERM b c; echo $?'
  kill TERM b c; echo $?
            ^
[ -c flag ]:1: 'kill' invalid process id specification 'b'
2

```

andychu:
Also the tests were testing `%-` (previous job) but not `%%` (current job):

https://github.com/oils-for-unix/oils/pull/2478/commits/988699dc71dc5c65a9419628e0ed2d9caa326997#diff-8b0151f613369f1e8b91780e5007ec1d9ccf37871ca1be900ede452db1dbb5b3

I believe `kill %%` is the thing that regtest/aports required

btertoolen:
@andychu This is ready for review, I checked all your comments. I did notice in some spec tests the zsh behavior changed on the CI, did the version get updated?

builtin/process_osh.py
```diff
@@ -682,4 +694,109 @@ def Run(self, cmd_val):
...
+        return signal_def.GetNumber(signal_name)
+
+    def _ParsePid(self, pid_arg, pid_arg_loc):
+        # type: (str, loc_t) -> int
+        if pid_arg.startswith("%"):
+            job = self.job_list.JobFromSpec(pid_arg)
+            if job is None:
+                e_usage("invalid signal specification %r" % pid_arg,
+                        pid_arg_loc)
+            else:
```
andychu: I would remove the `else` here, since `e_usage()` is the error path, and the normal path is `job.ProcessGroupId()`

builtin/process_osh.py
```diff
@@ -682,4 +694,109 @@ def Run(self, cmd_val):
...
+                return job.ProcessGroupId()
+        else:
+            try:
+                target_pid = int(pid_arg)
+                return target_pid
```
andychu: I would return target_pid after the `except`, since it can't raise ValueError

builtin/process_osh.py
```diff
@@ -682,4 +694,109 @@ def Run(self, cmd_val):
...
+        if signal_name.startswith("SIG"):
+            signal_name = signal_name[3:]
+        return signal_def.GetNumber(signal_name)
+
+    def _ParsePid(self, pid_arg, pid_arg_loc):
+        # type: (str, loc_t) -> int
+        if pid_arg.startswith("%"):
+            job = self.job_list.JobFromSpec(pid_arg)
+            if job is None:
+                e_usage("invalid signal specification %r" % pid_arg,
```
andychu: The error message mentions "signal" but this is a PID/job

builtin/process_osh.py
```diff
@@ -682,4 +694,109 @@ def Run(self, cmd_val):
...
+            signal = self._SignameToSignum(sigspec)
+        return signal
+
+    def Run(self, cmd_val):
+        # type: (cmd_value.Argv) -> int
+        signal_to_send = 15  # sigterm, the default signal to send
+        arg_r = args.Reader(cmd_val.argv, locs=cmd_val.arg_locs)
+        arg_r.Next()  # skip command name
+        first_positional, first_positional_loc = arg_r.ReadRequired2(
+            "you must provide a process id")
```
andychu: This error looks like this:

```
$ bin/osh -c 'kill -9'
  kill -9
  ^~~~
[ -c flag ]:1: 'kill' you must provide a process id
```

It is meant to enforce a certain grammatical convention

If you write `ReadRequired('expected a process ID')` then it will read grammatically

builtin/process_osh.py
```diff
@@ -682,4 +694,109 @@ def Run(self, cmd_val):
...
+        # checking for -sigspec argument
+        if first_positional.startswith('-') and (
+                first_positional[1:].isdigit() or len(first_positional) > 2):
+            signal_to_send = self._SigspecToSignal(first_positional[1:])
+            if signal_to_send == signal_def.NO_SIGNAL:
+                e_usage("invalid signal specification %r" % first_positional,
```
andychu: same with all the error messages:

```
$ bin/osh -c 'kill -foo'
  kill -foo
       ^~~~
[ -c flag ]:1: 'kill' invalid signal specification '-foo'
```

It can be more like


```
'kill' expected a signal, got '-foo'
```

builtin/process_osh.py
```diff
@@ -682,4 +694,109 @@ def Run(self, cmd_val):
...
+                else:
+                    num = self._SignameToSignum(arg_l)
+                    if num < 0:
+                        e_usage("invalid signal specification %r" % arg_l,
+                                arg_loc)
+                    print(str(num))
+                done_listing = True
+                arg_r.Next()
+
+            if not done_listing:
```
andychu: I think this can be moved to the top, without the bool:

```
if arg_r.AtEnd():  # no arguments given
  PrintSignals()
```

builtin/process_osh.py
```diff
@@ -682,4 +694,109 @@ def Run(self, cmd_val):
...
+                first_positional[1:].isdigit() or len(first_positional) > 2):
+            signal_to_send = self._SigspecToSignal(first_positional[1:])
+            if signal_to_send == signal_def.NO_SIGNAL:
+                e_usage("invalid signal specification %r" % first_positional,
+                        first_positional_loc)
+            else:
+                arg_pid, arg_pid_loc = arg_r.ReadRequired2(
+                    "you must provide a process id")
+                pid = self._ParsePid(arg_pid, arg_pid_loc)
+                posix.kill(pid, signal_to_send)  # Send signal
```
andychu: It's a bit awkward there are 3 posix.kill() calls in the function

I think there can be just one

Remember I mentioned in the `trap` builtin, I refactored it to `_RemoveTheRest()`

It can be `_KillArgs(arg_r, signal_to_send)`

I can help with that if you like

andychu (commented):
This looks good -- mostly some stylistic things

I'm not sure what exactly is going on with zsh

It fails on my machine, but it succeeds in CI?  Which is not supposed to happen

I am building a new version of zsh, but it is NOT deployed yet, so shouldn't have affected anything

andychu:
As mentioned  I use this function to quickly test the error messages

```
$ test/runtime-errors.sh test-kill-builtin-usage 

===== CASE: -c kill --15 99 =====


  kill --15 99
       ^~~~
[ -c flag ]:1: 'kill' invalid signal specification '--15'

===== CASE: -c kill -TE 99 =====


  kill -TE 99
       ^~~
[ -c flag ]:1: 'kill' invalid signal specification '-TE'

===== CASE: -c kill -15 =====


  kill -15
  ^~~~
[ -c flag ]:1: 'kill' you must provide a process id

```

But now they look ungrammatical


This is how they used to look, which is grammatical:

```
$ test/runtime-errors.sh test-kill-builtin-usage 

===== CASE: -c kill --15 99 =====


  kill --15 99
       ^~~~
[ -c flag ]:1: 'kill' got invalid signal '--15'

===== CASE: -c kill -TE 99 =====


  kill -TE 99
       ^~~
[ -c flag ]:1: 'kill' got invalid signal '-TE'

===== CASE: -c kill -15 =====


  kill -15
  ^~~~
[ -c flag ]:1: 'kill' expected a PID
```

andychu:
Hm looks like not all the comments were addressed?

There are some comments attached to the LHS now for some reason

btertoolen:
@andychu my bad, missed those somehow. I pushed a commit to fix them

andychu:
Merged, thank you!

I did find one bug --  here the validation check apparently changed from `< 0` to `< signal_def.NO_SIGNAL`


https://github.com/oils-for-unix/oils/commit/371276f245920c6ad63f75441ebb57bd74636b08

But it should be `== signal_def.NO_SIGNAL` (which is -1)

So the error didn't trigger, and we were missing test coverage

---

I added test coverage.  And I did find one deviation from bash -- the `kill -L` validation is more aggressive than bash.  We exit early instead of printing an error and keep goigng.

I'm not sure we need to fix that, or  it could be a good first issue for someone who wants to dip their toes in

Issue #2413, opened 2025-09-11
e2fsprogs test failures - `x=1>`  is lexed/parsed as `x=` then `1>`


PossiblyAShrub:
Minimal reproduction:

```
osh$ echo x=1> tmp; cat tmp
x=
bash$ echo x=1> tmp; cat tmp
x=1
```

OSH incorrectly absorbs the `1` in the last word.

---

I debugged [the `e2fsprogs` test suite failure](https://op.oils.pub/aports-build/2025-09-10-overlayfs.wwz/_tmp/aports-report/2025-09-10-overlayfs/shard2/osh-as-sh/log/e2fsprogs.log.txt) and found that OSH failed to correctly execute this command:

```sh
dd if=$TEST_BITS of=$TEST_DATA bs=4k count=2 seek=1> /dev/null 2>&1
```

The issue was that the `>` redirection absorbed the `1` so the command run was `dd if=$TEST_BITS of=$TEST_DATA bs=4k count=2 seek=` which resulted in a test failure.

Here's the parse tree with and without a space

```
$ osh -nc 'dd if=$TEST_BITS of=$TEST_DATA bs=4k count=2 seek=1 > /dev/null 2>&1'  # Expected
(C
  (w <Lit_Chars dd>)
  (w <Lit_VarLike "if="> ($ TEST_BITS))
  (w <Lit_VarLike "of="> ($ TEST_DATA))
  (w <Lit_VarLike "bs="> <Lit_Chars 4k>)
  (w <Lit_VarLike "count="> <Lit_Chars 2>)
  (w <Lit_VarLike "seek="> <Lit_Chars 1>)
)

$ osh -nc 'dd if=$TEST_BITS of=$TEST_DATA bs=4k count=2 seek=1> /dev/null 2>&1'  # Actual
(C
  (w <Lit_Chars dd>)
  (w <Lit_VarLike "if="> ($ TEST_BITS))
  (w <Lit_VarLike "of="> ($ TEST_DATA))
  (w <Lit_VarLike "bs="> <Lit_Chars 4k>)
  (w <Lit_VarLike "count="> <Lit_Chars 2>)
  (w <Lit_VarLike "seek=">)
)
```

PossiblyAShrub:
Some fun trivia, the github bash highlighter also seems to be affected by this issue!

<img width="516" height="64" alt="Image" src="https://github.com/user-attachments/assets/f1108893-711d-4816-9f52-fa7801fac169" />

[...]

PR #2479, opened 2025-10-14
[osh] Make redirects less eager to consume their left-hand side args

last-genius:
Compare bash:

```
$ echo x=1> tmp; cat tmp
x=1
```

against osh before this fix:
```
$ echo x=1> tmp; cat tmp
x=
```

Here the redirect absorbs '1' where it shouldn't.

To fix this, "absorb" the preceding argument in `WordParser` instead of in the lexer.

Drop the regexes capturing `FD_NUM` and `FD_VAR_NAME` as a part of `Id.Redir_*` in `lexer_def.py`. Instead, introduce `RedirOp` - a new type of `word_t` which can be emitted by `WordParser` when either `Lit_Number` or `Lit_RedirVarName` are the only word in a `CompoundWord` group before a `Redirect` interrupts it.

This way, the words leading up to the redirection operator will only be absorbed when they stand apart from words preceding them (not in `x=1>`, but in `x= 1>`, for example), given that they follow the rules for `FD_NUM` and `FD_VAR_NAME`.

This also allows us to avoid awkwardly checking the redirection string for `{` or convertability to an integer, instead reusing the results of the lexer when matching on `Lit_Number|Lit_RedirVarName`.

Fixes: #2413

### DIFF
--- a/osh/word_parse.py
+++ b/osh/word_parse.py
@@ -557,8 +557,8 @@ class WordParser(WordEmitter):
         self._SetNext(lex_mode_e.VSub_Zsh)  # Move past ${(foo)
 
         # Can be empty
-        w = self._ReadCompoundWord3(lex_mode_e.VSub_Zsh,
-                                       Id.Right_DollarBrace, True)
+        w = self._ReadCompoundWord3(lex_mode_e.VSub_Zsh, Id.Right_DollarBrace,
+                                    True)
         self._GetToken()
         return word_part.ZshVarSub(left_token, w, self.cur_token)
 
@@ -1831,9 +1831,8 @@ class WordParser(WordEmitter):
 
     def _ReadCompoundWord(self, lex_mode):
         # type: (lex_mode_t) -> CompoundWord
-        """Returns either word.Compound or word.Redir"""
 
-        # This lexer mode can return redirects
+        # This is the ONLY lexer mode that can return word.Redir
         assert lex_mode != lex_mode_e.ShCommand, lex_mode
 
         w = self._ReadCompoundOrRedir(lex_mode)
@@ -1843,7 +1842,7 @@ class WordParser(WordEmitter):
     def _ReadCompoundWord3(self, lex_mode, eof_type, empty_ok):
         # type: (lex_mode_t, Id_t, bool) -> CompoundWord
 
-        # This lexer mode can return redirects
+        # This is the ONLY lexer mode that can return word.Redir
         assert lex_mode != lex_mode_e.ShCommand, lex_mode
 
         w = self._ReadCompoundOrRedir3(lex_mode, eof_type, empty_ok)


last-genius:
Haven't worked out the issues with mycpp, but the Python version should be passing the spec tests now.

last-genius:
osh fails `brace-expansion` spec test because it's behaving like bash now:
```
$ bash -c 'echo {x}_{a,b}'
{x}_a {x}_b

Before PR:
$ bin/osh -c 'echo {x}_{a,b}'
{x}_{a,b}

After PR:
$ bin/osh -c 'echo {x}_{a,b}'
{x}_a {x}_b
```

last-genius:
Before:
```
$ bin/osh --ast-format text -nc 'echo {x}_{a,b}'
(command.Simple
  blame_tok:(Token id:Lit_Chars length:4 col:0 line:(SourceLine line_num:1 content:"echo {x}_{a,b}" src:(source.CFlag)))
  more_env:[]
  words:[
    (CompoundWord parts:[...0x7fb970eb8b40])
    (CompoundWord
      parts:[
        (Token id:Lit_LBrace length:1 col:5 line:...0x7fb970e6d410)
        (Token id:Lit_Chars length:1 col:6 line:...0x7fb970e6d410)
        (Token id:Lit_RBrace length:1 col:7 line:...0x7fb970e6d410)
        (Token id:Lit_Chars length:1 col:8 line:...0x7fb970e6d410)
        (Token id:Lit_LBrace length:1 col:9 line:...0x7fb970e6d410)
        (Token id:Lit_Chars length:1 col:10 line:...0x7fb970e6d410)
        (Token id:Lit_Comma length:1 col:11 line:...0x7fb970e6d410)
        (Token id:Lit_Chars length:1 col:12 line:...0x7fb970e6d410)
        (Token id:Lit_RBrace length:1 col:13 line:...0x7fb970e6d410)
      ]
    )
  ]
  is_last_cmd:F
)
```

After:
```
$ bin/osh --ast-format text -nc 'echo {x}_{a,b}'
(command.Simple
  blame_tok:(Token id:Lit_Chars length:4 col:0 line:(SourceLine line_num:1 content:"echo {x}_{a,b}" src:(source.CFlag)))
  more_env:[]
  words:[
    (CompoundWord parts:[...0x7f5a9664cb40])
    (word.BracedTree
      parts:[
        (Token id:Lit_RedirVarName length:3 col:5 line:...0x7f5a96584110)
        (Token id:Lit_Chars length:1 col:8 line:...0x7f5a96584110)
        (word_part.BracedTuple words:[(CompoundWord parts:[(Token id:Lit_Chars length:1 col:10 line:...0x7f5a96584110)]) (CompoundWord parts:[(Token id:Lit_Chars length:1 col:12 line:...0x7f5a96584110)])])
      ]
    )
  ]
  is_last_cmd:F
)
```

andychu:
Oh interesting about brace expansion ... well if it's like bash, maybe that's OK!

andychu:
Hm in the Python spec tests, it looks like bash and zsh fail for some reason:

https://op.oilshell.org/uuu/github-jobs/10338/ovm-tarball.wwz/_tmp/spec/osh-py/divergence.html

It might be help analysis to split up some of the test cases to see the pattern of pass/fail for bash/zsh/OSH

---

This looks promising though, thank you for sending it out so quickly !!

andychu:
Oh  and I usually iterate on it in Python first, to get the algorithm right.  And then I make sure type checking passes.  

The C++ part can be separate, which I can help with 

Glad to see this!

last-genius:
> Hm in the Python spec tests, it looks like bash and zsh fail for some reason:
> 
> https://op.oilshell.org/uuu/github-jobs/10338/ovm-tarball.wwz/_tmp/spec/osh-py/divergence.html

Fixed now

last-genius:
two `redirect` spec tests are failing though, will take a deeper look tomorrow.

([one](https://op.oilshell.org/uuu/github-jobs/10343/ovm-tarball.wwz/_tmp/spec/osh-py/redirect.html#details-32-osh) is the reason i put a "TODO" in the code - the limit for fds has effectively been dropped here, need to think about where to implement it. [the other failure](https://op.oilshell.org/uuu/github-jobs/10343/ovm-tarball.wwz/_tmp/spec/osh-py/redirect.html#details-3-osh) is more curious)

last-genius:
> Oh interesting about brace expansion ... well if it's like bash, maybe that's OK!

I've changed this brace expansion test to expect `osh` to behave like `bash`.

All of the CI is green now! mycpp took some time to figure out, as usual...

last-genius:
I've ended up not needing any lookahead here. I just do a "look-behind", I guess, by checking if `Lit_Number/Lit_RedirVarName` are the only members of a `CompoundWord` group before a redirection interrupts it, and returning a `RedirOp` instead.

So I think I'll tackle #2337 separately, since it *does* require lookahead, at least from what I can see at this point after investigating it for a while.

andychu:
I took a quick look, this is very promising, thank you !!

https://op.oilshell.org/uuu/github-jobs/10352/ovm-tarball.wwz/_tmp/spec/osh-py/redirect.html

The diff looks pretty clean too, which is a good sign :-)

I may split the tests up a bit, because I like to see a horizontal pattern of green for some cases, and then the disagreements separately -- e.g. where exactly dash or mksh might  agree and disagree


----


Regarding #2337, I am actually confused to what *exactly* bash does.    I think we need quite a few test cases to figure that out.

Although, it's also true that the `regtest/aports` provides exactly what we need to make work!   That is one of the most common bugs

I guess keep us updated on Zulip about what you find exactly ... i.e. we have to "discover" the spec for what the parser does, but we have a good motivation in making all the packages work


----

  #2409 is another possible candidate -- I believe it relates to the fact that we share a parser for `[[` and `[`, which all other shells do too

But we have a bug for some reason.    There are good test cases for it in `spec/divergence`

*[ Is a Builtin, But [[ Is Part of the Language* - https://www.oilshell.org/blog/2016/10/12.html

frontend/syntax.asdl
```diff
@@ -247,6 +247,8 @@ module syntax
   # At runtime: RHS of 'declare x='.
   rhs_word = Empty | Compound %CompoundWord
 
+  RedirOp = (Token? loc, Token op)
```
andychu: I think we should name it `Token left_arg` or `Token where` to be clearer -- because we use `loc` or `blame_loc` a lot for **syntax** locations, which are also tokens

And I would add a comment like explaining the 3 cases: `>    3>   {myvar}>`

frontend/location.py
```diff
@@ -394,6 +395,14 @@ def LeftTokenForCompoundWord(w):
         return None
 
 
+def LeftTokenForRedirOp(w):
+    # type: (RedirOp) -> Optional[Token]
```
andychu: I think this should just be Token, not Optional[Token]

(unfortunately we don't have MyPy configured to strictly enforce this)

frontend/syntax.asdl
andychu: Or maybe it can just be `left_tok`

Now I see there is a type `redir_loc`, which is why you chose that name ... but yeah I think `left_tok` is accurate and consistent

osh/word_parse.py
```diff
@@ -303,8 +305,8 @@ def _ReadVarOpArg(self, arg_lex_mode):
         return w
 
     def _ReadVarOpArg2(self, arg_lex_mode, eof_type, empty_ok):
-        # type: (lex_mode_t, Id_t, bool) -> CompoundWord
-        """Return a CompoundWord.
+        # type: (lex_mode_t, Id_t, bool) -> word_t
```
andychu: OK I see why  many of these were changed to return word_t instead of CompoundWord

With some of them, a redirect is impossible, e.g. for `${x//</y}`, the lexer switches modes

```
(C
  (w <Lit_Chars echo>)
  (w
    (BracedVarSub
      left:<Left_DollarBrace "${">
      name_tok:<VSub_Name x>
      var_name:x
      suffix_op:(suffix_op.PatSub pat:(w <Lit_Other "<">) replace:(w <Lit_Chars y>) replace_mode:Lit_Slash slash_tok:<VOp2_Slash />)
      right:<Right_DollarBrace "}">
    )
  )
)

```

we probably need some asserts and then type narrowing, let me see

andychu (commented):
Wow, this looks excellent!  I'm glad that it seems to fall relatively cleanly in to the type system and lexer structure

I am going to split up the spec tests a little, and then maybe try to narrow the `word_t` type in some cases ... that part is a bit tricky, and depends on whether the lexer can give you a Redir op in a given mode

andychu:
Actually I will do the rename to `left_tok` too, so we don't "collide" and cause merge conflicts

andychu:
Merged -- this is  amazing work, thank you!

I made some minor edits:

- Instead of the top level type RedirOp, I just used word::Redir
  - the `RedirOp %RedirOp` syntax is a special ASDL thing I added, in case you need to share top level types between sum types (which as I understand Rust can't do)
  - but in this case the type isn't shared, so I moved it into the sum type
- I narrowed the types by taking advantage of the fact that `lex_mode_e.ShCommand` is the only one that can return Kind.Redir.
  - So that involved introducing a couple tiny helper functions with casting

I'm glad it turned out to be a very clean change!  And yes even improved the redirect parsing

PR #2481, opened 2025-10-16
[osh] Check if `((` open an arith expr with a single-line lookahead

last-genius:
An expression is arithmetic if:
1) Opening and closing parens are balanced
2) And the closing parentheses are an uninterrupted pair - ))
3) And any other closing parentheses inbetween are closing a corresponding nested opening paren

If any of those conditions does not hold, then the expression is instead
treated as two nested subshells.

This is enough to handle the real world use cases (paren-ambiguity spec
tests), but does not follow bash and other shells completely, since
this look-ahead is single-line, biased towards nested subshells and so
can't handle multi-line arithmetic expressions.

Fixes: #2337

### DIFF
--- a/frontend/lexer.py
+++ b/frontend/lexer.py
@@ -248,22 +248,21 @@ class LineLexer(object):
         first_rparen = False  # Is the previous token an ')' ?
         parens_counter = 2
 
+        pos = self.line_pos
         while True:
             tok_type, end_pos = match.OneToken(lex_mode_e.Arith,
                                                self.src_line.content,
-                                               self.line_pos)
-            if self.line_pos == end_pos:
+                                               pos)
+            if end_pos == pos:  # end of line
                 break
-            self.line_pos = end_pos
+            pos = end_pos
 
-            # ( or $(
-            if tok_type in (Id.Arith_LParen, Id.Left_DollarParen):
+            if tok_type in (Id.Arith_LParen, Id.Left_DollarParen):  # ( or $(
                 parens_counter += 1
 
             if (tok_type == Id.Arith_RParen and first_rparen and
                     parens_counter == 1):
-                self.line_pos = original_pos
-                return True
+                return True  # saw closing ))
 
             if tok_type == Id.Arith_RParen:
                 parens_counter -= 1
@@ -279,8 +278,8 @@ class LineLexer(object):
             else:
                 first_rparen = False
 
-        # Recover after lookahead
-        self.line_pos = original_pos - num_bytes_back
+        # We may have to go back one byte for arithmetic
+        self.line_pos -= num_bytes_back
         return False
 
     def LookAheadFuncParens(self, num_bytes_back):


andychu:
Hm I think this limited approach is promising

I DON'T think we necessarily need to handle many of those things -- `$((` and multi-line lookahead and nested `(( ))`

As mentioned on Zulip, we are analyzing the `regtest/aports` results to see how often each thing is.   If the patch is simply adding a space, that can be acceptable

We want to handle some scripts unpatched though -- it's a judgement call


----

The only thing that looks a bit funny is the `ParsePipeline()` param

But in general I like the idea of just looking ahead on the current line, and using a heuristic !

I think it's good to keep the parser simple, rather than adding a ton of code for things that only happen in 1% of shell scripts, and are bad shell code anyway

andychu:
I merged the exact test cases into this branch, and pushed, to see how it behaves:

https://github.com/oils-for-unix/oils/commit/bf18bec21009c9a74985d0879ce77638900e0ff9

`test/spec.sh paren-ambiguity`

I have an idea for a possibly simpler lookahead rule, but let's see

andychu:
Hm it looks like it doesn't parse 2 of the 4 examples I copied into spec tests from #2337 

https://op.oilshell.org/uuu/github-jobs/10382/ovm-tarball.wwz/_tmp/spec/osh-py/paren-ambiguity.html

case `#5`

```
  echo $? >&4) 3>&- 5<&- </dev/null |
                       ^
[ stdin ]:5: Expected word type Id.Op_RParen, got Id.Redir_GreatAnd
```

 and case `#7`


```
 if ! ((test x"$i" = x-g) || (test x"$i" = x-O2)); then
              ^
[ stdin ]:1: Unexpected token after arithmetic expression (Id.Word_Compound != Id.Arith_RParen)
```

----

I have an idea which I will write up on Zulip


[Proposal for (( lookahead rule, that also fixes YSH if ((](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/Proposal.20for.20.28.28.20lookahead.20rule.2C.20that.20also.20fixes.20YSH.20if.20.28.28/with/545634286)

The basic idea is to

1. remove the token `((` and make it two `(` tokens
2. check for balanced `))` **within a line** - without spaces

And then if there happen to be any multi-line cases we don't handle, I think that

```
'(' '(' VarName ('=' | '[' )
```

will be another simple single line lookahead for arithmetic .  However I suspect we will NOT need that

frontend/lexer.py
```diff
@@ -226,6 +226,36 @@ def LookPastSpace(self, lex_mode):
 
         return tok_type
 
+    def LookAheadDParens(self):
+        # type: () -> bool
+        """For finding the closing arithmetic parens - ))
```
andychu: Hm actually I think my suggestion on Zulip pretty much boiled down to this

[Proposal for (( lookahead rule, that also fixes YSH if ((](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/Proposal.20for.20.28.28.20lookahead.20rule.2C.20that.20also.20fixes.20YSH.20if.20.28.28/with/545634286)

Except initially I was thinking we need to remove `((` as a token

(I think we do , but now I think it's a separate change)

I think it will actually be less error prone and cleaner to do this with **literal text**, not with the `Arith` lexer !

We start with `paren_counter = 2` for `((`, and then match balanced parens until the counter reaches 0 and we see two `))` in a row

---

I'm not sure why this PR doesn't parse cases `#5` and `#7` though -- I would have thought the rule does handle that

last-genius:
Updated.

Now the lookahead handles nested parens, `$((`, and more.

Non-synthetic paren-ambiguity tests are passing now. The only spec tests failing are for multi-line arithmetic (which, unless we change the bias of the lookahead, we will not be able to handle here)

One completion unit test is also failing - but the output there is really confusing.

Just checking for balanced parentheses wasn't enough - explained more in the comments and commits.

andychu:
Ah OK, some  spec tests failed in `spec/arith`

https://op.oilshell.org/uuu/github-jobs/10408/ovm-tarball.wwz/_tmp/spec/osh-py/arith.html

They are due to treating say

```
echo $((1
+ 2))
```

as commands now, rather than arithmetic, hm

----

And then same with `spec/whitespace`

https://op.oilshell.org/uuu/github-jobs/10408/ovm-tarball.wwz/_tmp/spec/osh-py/whitespace.html#details-1-osh

OK let me look at the rest

frontend/lexer.py
```diff
@@ -226,6 +226,60 @@ def LookPastSpace(self, lex_mode):
...
+        tok_type, end_pos = match.OneToken(lex_mode_e.Arith,
+                                           self.src_line.content,
+                                           self.line_pos)
+        first_match = False  # Is the previous token an ')'
+        parens_counter = 2
+        # It's not enough to just check if parens are balanced - we need to
+        # make sure that it's the starting parens that are closed at the end,
+        # and not some other parens, e.g. here, where parens are balanced
+        # but the closing pair still does not close an arithmetic expression:
+        # ((expr) || (expr))
```
andychu: OK this is interesting ...  this is a SUBSHELL/command, not arithmetic, right?  Please make that clearer in the comment

Actually wouldn't it be clearer as 

((command) || (command))

frontend/lexer.py
```diff
@@ -226,6 +226,60 @@ def LookPastSpace(self, lex_mode):
...
+        # (( (expr) ) || (expr))
```
andychu: likewise isn't this

(( (command) ) || (command))

Hm hm ... this is quite subtle ... the simpler rule is not sufficient to make the test cases pass?

frontend/lexer.py
```diff
@@ -340,6 +394,10 @@ def LookPastSpace(self, lex_mode):
         # type: (lex_mode_t) -> Id_t
         return self.line_lexer.LookPastSpace(lex_mode)
 
+    def LookAheadDParens(self, shift_back):
```
andychu, line 399: naming -- I think that `shift_back` is similar to `LookAheadFuncParens(unread)`

Although I think neither name is that good ... maybe it should be `num_bytes_back` or something ?   in both cases


```
        if self.cur_token.id == Id.Op_LParen:  # saw funcname(
            return self.lexer.LookAheadFuncParens(1)  # go back one char

        elif self.cur_token.id == Id.WS_Space:  # saw funcname WHITESPACE
            return self.lexer.LookAheadFuncParens(0)
```

frontend/lexer.py
```diff
@@ -226,6 +226,60 @@ def LookPastSpace(self, lex_mode):
...
+        # type: (int) -> bool
+        """For finding the closing arithmetic parens - ))
+        If none consecutive parens (not separated by any other token) occur,
+        returns False.
+
+        shift_back is the number of characters to go back for in case the
+        expression is not arithmetic.
+        """
+        original_pos = self.line_pos
+        tok_type, end_pos = match.OneToken(lex_mode_e.Arith,
```
andychu: Style: I personally prefer having a `while True ` loop with  single `match.OneToken()` call, with an `if ... break` for the condition

Since it is the same way to advance in both cases

frontend/lexer.py
```diff
@@ -226,6 +226,60 @@ def LookPastSpace(self, lex_mode):
 
         return tok_type
 
+    def LookAheadDParens(self, shift_back):
+        # type: (int) -> bool
+        """For finding the closing arithmetic parens - ))
+        If none consecutive parens (not separated by any other token) occur,
```
andychu: Maybe clarify that True means arithmetic, and False means command, with 2 short examples

I guess you could move the subtle examples up here too, the `((command); (command))` ones

andychu (commented):
Thanks for getting this working!  It looks pretty clean

I am wondering if we should commit the `$((` change separately from the `((` change


I wonder if that makes us differ from other shells in a way that matters -- the 2 cases are a bit different, in terms of how other shells behave

Let me look at your other PR

frontend/lexer.py
andychu: And I might choose an example that is more obviously a command, like `((command); (command))`   - since `||` is sometimes in arithmetic

osh/word_parse.py
```diff
@@ -826,7 +826,12 @@ def _ReadDoubleQuotedLeftParts(self):
             return self._ReadBracedVarSub(self.cur_token, d_quoted=True)
 
         if self.token_type == Id.Left_DollarDParen:
-            return self._ReadArithSub()
+            if (self.LookAheadDParens(shift_back=1)):
```
andychu: I think it's better to land the `$((` change separately ... we can just comment it out temporarily here, and then the other spec tests won't fail.  

And we can easily re-enable it later, after we make another run on `regtest/aports`

It's more likely that treating `$((` across lines as commands will break some scripts, compared with `((`


---

It's nice that the same lookahead works for both though!

frontend/lexer.py
last-genius: the simpler rule is not enough, yeah. in particular this real-world example in `paren-ambiguity` is failing:
```
if ! ((test x"$i" = x-g) || (test x"$i" = x-O2)); then
    CF="$CF $i"
fi
```

last-genius:
Commented out lookahead for `$((` (this means the spec and unit tests aren't failing anymore, it's all green!), addressed the review comments.

andychu:
Merged, thank you!  I made some minor cleanups, noted in the commit description

Woohoo time to run `regtest/aports` again

Issue #1472, opened 2023-01-29
@[strs] should work in expression mode, convert list to array of strings


andychu:
Noticed this after reading code sample from Kel

https://oilshell.zulipchat.com/#narrow/stream/325160-oil-discuss-public/topic/People.20still.20using.20JSON.20Template.3F

andychu:
In other words,  both of these should work:


```
$ bin/ysh -c 'var x = [1,2,3]; echo @[x]'
1 2 3


$ bin/ysh -c 'var x = [1,2,3]; = @[x]'
  var x = [1,2,3]; = @[x]
                     ^~
[ -c flag ]:1: Unexpected token in expression mode (Id.Expr_Reserved)

```

andychu:
This is useful because it lets people refactor code more easily

daveads:
@andychu  is this the expected behavior ? 

<img width="1052" height="72" alt="Image" src="https://github.com/user-attachments/assets/1751aa6b-6c1c-4e78-b835-f7678e56f097" />

andychu:
Hm yes this looks good!

I'd like to see a PR :-)

daveads:
> Hm yes this looks good!
> 
> I'd like to see a PR :-)

yea i will send it right away...

daveads:
@andychu i guess you might have to add me... since forking the project isn't needed here...

andychu:
Oh yes, I sent you an invite!

thank you

If you already forked I think you can patch it `git remote rm origin` and `git remote add origin` to point back to the `oils-for-unix/oils` repo

---

After you accept the invite, I think you can assign the `printf` bug to yourself!  Thank you!

Also please join https://oilshell.zulipchat.com/ if you haven't already, there is a lot  of context there :-)

PR #2482, opened 2025-10-17
Enable @[...] array splice in expression mode

daveads:
address this issue...  #1472

### DIFF
--- a/spec/ysh-expr-sub.test.sh
+++ b/spec/ysh-expr-sub.test.sh
@@ -41,7 +41,14 @@ key 42
 dq 42
 ## END
 
-#### @[expr splice]
+#### @[expr_splice] in expression mode
+
+# In addition to:
+#   write -- @[x]
+#
+# We can also do:
+#   var y = @[x]
+#   write -- @y
 
 var x = [1, 2, 3]
 


daveads:
Screenshot 2025-10-17 at 3 59 30 PM

daveads:
@andychu

andychu:
This looks promising !  Thank you for working on it

I notice some spec test failures:

Can you try `devtools/types.sh check-oils` ?  I think that might find some type errors that can catch some of these spec test failures:

https://op.oilshell.org/uuu/github-jobs/10383/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

Some how it affected OSH in a couple places

https://op.oilshell.org/uuu/github-jobs/10383/ovm-tarball.wwz/_tmp/spec/osh-py/osh-bugs.html

https://op.oilshell.org/uuu/github-jobs/10383/ovm-tarball.wwz/_tmp/spec/osh-py/strict-options.html#details-15-osh

---

Hm and thank you for debugging this NT_OFFSET thing ... I remember running into that awhile ago, and I think it was probably a code smell!

daveads:
Screenshot 2025-10-17 at 7 00 13 PM

@andychu

andychu:
Looks great!  

OSH spec tests pass now (`ovm-tarball`): 

https://op.oilshell.org/uuu/github-jobs/10384/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

YSH spec tests have 1 failure:

https://op.oilshell.org/uuu/github-jobs/10384/cpp-spec.wwz/_tmp/spec/ysh-py/index.html

It looks like that is related to the 256 -> 512 issue.  Maybe you can change that to 512 and it should pass?

Sorry you hit that issue

----


As for the C++ failures - https://op.oilshell.org/uuu/github-jobs/10384/

I see

https://op.oilshell.org/uuu/github-jobs/10384/cpp-coverage.wwz/_tmp/soil/logs/compare-gcc-clang.txt


```
_gen/bin/oils_for_unix.mycpp.cc:53634:36: error: static_cast from 'syntax_asdl::Token *' to 'syntax_asdl::ExprSub *', which are not related by inheritance, is not allowed
                            return static_cast<ExprSub*>(pnode->GetChild(1)->tok);
                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I should be able to help fix that later tonight ...

Typically I say that  the goal is to get it working in Python, and then once that happens, C++ sometimes "just works".  In this case I'm not sure of the difference, but it should be easy to fix

daveads:
> Looks great!
> 
> OSH spec tests pass now (`ovm-tarball`):
> 
> https://op.oilshell.org/uuu/github-jobs/10384/ovm-tarball.wwz/_tmp/spec/osh-py/index.html
> 
> YSH spec tests have 1 failure:
> 
> https://op.oilshell.org/uuu/github-jobs/10384/cpp-spec.wwz/_tmp/spec/ysh-py/index.html
> 
> It looks like that is related to the 256 -> 512 issue. Maybe you can change that to 512 and it should pass?
> 
> Sorry you hit that issue
> 
> As for the C++ failures - https://op.oilshell.org/uuu/github-jobs/10384/
> 
> I see
> 
> https://op.oilshell.org/uuu/github-jobs/10384/cpp-coverage.wwz/_tmp/soil/logs/compare-gcc-clang.txt
> 
> ```
> _gen/bin/oils_for_unix.mycpp.cc:53634:36: error: static_cast from 'syntax_asdl::Token *' to 'syntax_asdl::ExprSub *', which are not related by inheritance, is not allowed
>                             return static_cast<ExprSub*>(pnode->GetChild(1)->tok);
>                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> I should be able to help fix that later tonight ...
> 
> Typically I say that the goal is to get it working in Python, and then once that happens, C++ sometimes "just works". In this case I'm not sure of the difference, but it should be easy to fix

alright checking it out..

daveads:
@andychu have you been able to checked the C++ aspect

daveads:
@andychu i guess this pr is done right ?

ysh/expr_parse.py
```diff
@@ -260,6 +260,32 @@ def _PushYshTokens(parse_ctx, gr, p, lex):
 
             continue
 
+        # @[  for array splice - convert expression to list of strings
+        if 0: #tok.id == Id.Left_AtLBracket:
```
andychu: I removed this part because we only need Id.Expr_CastedDummy as a "hack" for things that are not expressions

Example:

```
var x = $(echo hi > out)    
```

There, the thing inside $() is not an expression.  So the parser generator `pgen2` can't handle it.  So we add a **hack** for the type system / homogeneous tree -- the entire command AST for the CommandSub is considered  **token** of type `Id.Expr_CastedDummy`, and then we cast it back to a `command_t` later

I'm just explaining why we have the Expr_CastedDummy, and why we don't need it in this case

---

Here, everything is an expression:

```
var x = @[42 + a[i]]
```

So we don't need to "hack up" the parse tree output from pgen2 (before turning it into an aST)

ysh/expr_to_ast.py
```diff
@@ -701,6 +701,12 @@ def Expr(self, pnode):
         elif typ == grammar_nt.braced_var_sub:
             return cast(BracedVarSub, pnode.GetChild(1).tok)
 
+        elif typ == grammar_nt.expr_sub:
+            left = pnode.GetChild(0).tok
+            e2 = self.Expr(pnode.GetChild(1))
+            right = pnode.GetChild(2).tok
+            return ExprSub(left, e2, right)
```
andychu: After removing the Expr_CastedDummy, this part was also simplified:

```
            left = pnode.GetChild(0).tok
            e2 = self.Expr(pnode.GetChild(1))
            right = pnode.GetChild(2).tok
            return ExprSub(left, e2, right)
```

This is what was causing the type error.  Now everything fits within the type system, since we don't need the casting hacks in this case.

andychu:
Yes, nothing more to do here!  I made some comments so you can see what I did, and then I will merge it

Thank you

---

Oh yes I also added spec tests for this change - in `spec/ysh-expr-sub`

ysh/expr_to_ast.py
daveads: oh yea i get... Thanks for the fix : )

daveads:
> Oh yes I also added spec tests for this change - in `spec/ysh-expr-sub`

cool

andychu:
Merged, thank you!

I'm impressed you dove into the code, changing the lexer / grammar / expr_to_ast / and evaluator

I will add you to the `#grant-planning` channel 

---

I also realized we don't have `var x = $[42 + 1]` either -- that is #2487 

That should be an easy change after this one -- it should be mostly writing a test in `spec/ysh-expr-sub`, and then adapting the grammar to accept `$[` as well as `@[`

daveads:
> I also realized we don't have `var x = $[42 + 1]` either -- that is #2487
> 
> That should be an easy change after this one -- it should be mostly writing a test in `spec/ysh-expr-sub`, and then adapting the grammar to accept `$[` as well as `@[`

i will assign myself to that issue.

Issue #2477, opened 2025-10-14
printf %d %X don't support hex 0x55 and octal 0055 - test/signal-report.sh


andychu:
Ran into this while running test/signal-report.sh

```
$ test/spec.sh builtin-printf -r 59
builtin-printf.test.sh
case    line    dash    bash    mksh    zsh     ash     osh
 59     1407    pass    pass    pass    BUG     pass    FAIL    printf %d %X support hex 0x5 and octal 055

                dash    bash    mksh    zsh     ash     osh
        pass    1       1       1       0       1       0
        BUG     0       0       0       1       0       0
        FAIL    0       0       0       0       0       1
        total   1       1       1       1       1       1
```

daveads:
assign me to this... @andychu

PR #2483, opened 2025-10-17
printf fix

daveads:
fix this issue #2477 

Support hex (0x) and octal (0) integer literals

### DIFF
--- a/builtin/printf_osh.py
+++ b/builtin/printf_osh.py
@@ -61,12 +61,15 @@ def _ParsePrintfInteger(s, blame_loc):
       042     octal
       42      decimal
     """
-    # Handle negative sign separately since the shell number lexer doesn't
-    # recognize negative numbers as a single token
+    # Handle +/- sign separately since the shell number lexer doesn't
+    # recognize signed numbers as a single token
     negative = False
     s_to_parse = s
     if len(s) > 0 and s[0] == '-':
         negative = True
+        s_to_parse = s[1:]
+    elif len(s) > 0 and s[0] == '+':
+        # Positive sign is optional but allowed
         s_to_parse = s[1:]
     
     # Normalize 0X to 0x for the lexer
--- a/spec/builtin-printf.test.sh
+++ b/spec/builtin-printf.test.sh
@@ -1462,6 +1462,42 @@ octal 5
 37
 ## END
 
+#### printf %d with + prefix (positive sign)
+
+echo decimal
+printf '%d\n' +42
+
+echo octal
+printf '%d\n' +077
+
+echo hex lowercase
+printf '%d\n' +0xab
+
+echo hex uppercase
+printf '%d\n' +0XAB
+
+## STDOUT:
+decimal
+42
+octal
+63
+hex lowercase
+171
+hex uppercase
+171
+## END
+
+## BUG zsh STDOUT:
+decimal
+42
+octal
+77
+hex lowercase
+171
+hex uppercase
+171
+## END
+
 #### printf %d with arbitrary base
 # bash, dash, and mksh print 64 and return status 1
 # zsh and ash print 0 and return status 1


daveads:
Screenshot 2025-10-17 at 9 55 39 PM

daveads:
@andychu Check it out..

andychu:
The tests pass, that is great!

https://op.oilshell.org/uuu/github-jobs/10387/ovm-tarball.wwz/_tmp/spec/osh-py/index.html

You can change `oils_failured_allowed:` at the top of the file to `0` now.  The tests assert the exact number of failures, and then it will pass

When I push a PR, I go look at  the results at `op.oilshell.org`:

https://op.oilshell.org/uuu/github-jobs/10387/

builtin/printf_osh.py
```diff
@@ -310,7 +310,8 @@ def _Percent(
 
             if match.LooksLikeInteger(s):
```
andychu: Hm changing the meaning of LooksLikeInteger may have some unintended consequences, since it is used in at least 2 other places

```
$ fgrep -n 'LooksLikeInteger(' */*.py
builtin/printf_osh.py:311:            if match.LooksLikeInteger(s):
builtin/process_osh.py:609:            if match.LooksLikeInteger(s):
...

osh/sh_expr_eval.py:1116:            if match.LooksLikeInteger(s):

```

andychu:
BTW that function supports the following behavior for shell - so we can adapt it for `printf` - EXCEPT for the last case.

Let's also add a test to `spec/builtin-printf` to make sure the `64#a` case does NOT work (or behaves like other shells).  As mentioned I think rejecting it completely is OK, I guess with  `status 1` from `printf`



```
$ osh -c 'echo $(( 055 ))'
45

$ osh -c 'echo $(( 0x55 ))'
85

$ osh -c 'echo $(( 55 ))'
55

$ osh -c 'echo $(( 64#a ))'
10

```

daveads:
attending to this now...

daveads:
@andychu can you check it now ?

andychu:
Hm looks like some other `printf` spec tests started failing?

https://op.oilshell.org/uuu/github-jobs/10418/

https://op.oilshell.org/uuu/github-jobs/10418/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-printf.html


Looks like negative numbers stopped working

```
  printf '[%6.4d]\n' -42
                     ^~~
[ stdin ]:5: printf expected an integer, got '-42'
```

builtin/printf_osh.py
```diff
@@ -49,6 +49,53 @@
...
+    Returns:
+      (True, value) when the string looks like an integer
+      (False, ...)  when it doesn't
+
+    Integer formats that are recognized:
+      0xAB    hex
+      042     octal
+      42      decimal
+    """
+    # Normalize 0X to 0x for the lexer
```
andychu, line 75: Oh actually this is a bug in OSH!

```
$ bash -c 'echo $(( 0XAA ))'
170

$ osh -c 'echo $(( 0XAA ))'
  echo $(( 0XAA ))
           ^
[ -c flag ]:1: fatal: Invalid integer constant '0XAA'
```

Let's change the lexer instead - search for `ShNumber_Hex`

---

And let's add a test in `spec/arith` to cover this case


Near these cases

```
#### Dynamic octal constant
zero=0
echo $(( ${zero}11 ))
## stdout: 9
## N-I mksh/zsh stdout: 11

#### Dynamic hex constants
zero=0
echo $(( ${zero}xAB ))
## stdout: 171

```

We should accept capital X, but we don't

frontend/lexer_def.py
```diff
@@ -960,7 +960,7 @@ def R(pat, tok_type):
 # Note: if you call match.LooksLikeInteger(s), mops.FromStr(s) may still
 # fail.  However you should call BOTH, because we don't rely want to rely on
 # the underlying stroll() to define the language accepted.
-LOOKS_LIKE_INTEGER = _WHITESPACE + '-?[0-9]+' + _WHITESPACE
+LOOKS_LIKE_INTEGER = _WHITESPACE + r'-?(0[xX][0-9a-fA-F]+|[0-9]+)' + _WHITESPACE
```
andychu: this changed can be reverted

mycpp/mops.py
```diff
@@ -117,10 +117,28 @@ def ToHexLower(b):
...
-    Simulate C++
+    Simulate C++ strtol() behavior for base=0:
+    - Supports hex: 0x prefix
+    - Supports old-style octal: 0 prefix (bash/C style, not Python 0o)
+    - Supports decimal
     """
     try:
-        big_int = BigInt(int(s, base))
+        # When base=0, handle old-style octal like bash/C (055 not 0o55)
+        if base == 0:
```
andychu: This can also be reverted, no?

builtin/printf_osh.py
daveads: > Oh actually this is a bug in OSH!
> 
> Let's change the lexer instead - search for `ShNumber_Hex`
> 

checking it out...

daveads:
@andychu check it out when free...

andychu:
Ah looks like spec tests are still failing - https://op.oilshell.org/uuu/github-jobs/10432/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-printf.html

The last case needs assertions for the other shells.  You can add `## OK` assertions in that case

andychu:
After you push, you can visit https://op.oilshell.org/uuu/github-jobs/10432/ for the failures

spec/builtin-printf.test.sh
```diff
@@ -1461,3 +1461,25 @@ octal 5
...
+
+# Test that 64#a is rejected (unlike in shell arithmetic where it's valid)
+printf '%d\n' '64#a' 2>/dev/null
+echo "status=$?"
+
+## STDOUT:
+status=1
+## END
+
+## OK dash STDOUT:
```
andychu: Oh this should mention all the other shells, like `OK dash/bash/mksh STDOUT:`

then you can write one assertion

builtin/printf_osh.py
```diff
@@ -49,6 +49,67 @@
 _ = log
 
 
+def _ParsePrintfInteger(s, blame_loc):
```
andychu, line 52: I think this function can be simplified

I realize it is different than the `_ParseOshInteger` function because of space and the negative sign

But maybe it can be reused a bit

We could do that in this PR, or in a separate PR

There is a step labeled "optional code cleanup" on the Kanban board, because sometimes it's better to do it separately

The first step is to get all the tests passing

andychu: I think it might be cleaner to have another little lexer, on top of ShNumber perhaps

Let me think about that

---

Weirdly, it also accepts +

```
$ dash -c "printf '%d\n' '+35'"
35

$ dash -c "printf '%d\n' '++35'"
dash: 1: printf: ++35: expected numeric value
0

```

daveads:
> After you push, you can visit https://op.oilshell.org/uuu/github-jobs/10432/ for the failures

noted

builtin/printf_osh.py
daveads: > I think it might be cleaner to have another little lexer, on top of ShNumber perhaps
> 
> Let me think about that
> 
> Weirdly, it also accepts +
> 
> ```
> $ dash -c "printf '%d\n' '+35'"
> 35
> 
> $ dash -c "printf '%d\n' '++35'"
> dash: 1: printf: ++35: expected numeric value
> 0
> ```

interesting

daveads:
@andychu

builtin/printf_osh.py
daveads: > 
> I realize it is different than the `_ParseOshInteger` function because of space and the negative sign
> 
> We could do that in this PR, or in a separate PR
> 


on a separate pr is better..

spec/builtin-printf.test.sh
```diff
@@ -1461,3 +1461,26 @@ octal 5
 55
 37
 ## END
+
```
andychu, line 1464: Can you add a test case for +42 +077 and +0xAB  ?  And then recognize +

Now that we know it's there

daveads: Screenshot 2025-10-24 at 8 41 24 PM
@andychu

andychu:
Thank you for fixing this!

builtin/printf_osh.py
daveads: > I think this function can be simplified
> 
> I realize it is different than the `_ParseOshInteger` function because of space and the negative sign
> 
> But maybe it can be reused a bit
> 
> We could do that in this PR, or in a separate PR
> 
> There is a step labeled "optional code cleanup" on the Kanban board, because sometimes it's better to do it separately
> 
> The first step is to get all the tests passing

are you going to create a separate issue for this ? @andychu

andychu: Yes I think we should change the style a bit, and maybe consolidate the functions

I want to use more of a strict lexing/parsing style, consistent with the rest of Oils

In this case, it may be a matter of style only (?).  But often I find bugs when I refactor into this style

---

You can see examples here from the `trap` builtin

[#blog-ideas > Lessons From Improving the Trap Builtin @ 💬](https://oilshell.zulipchat.com/#narrow/channel/266575-blog-ideas/topic/Lessons.20From.20Improving.20the.20Trap.20Builtin/near/544345450)

In that case, it was **essential** for upgrading the `trap` builtin from OSH to YSH

---

Actually in this case, we **do** have to make a decision about YSH

In YSH, we won't have `printf`

Here is a silly bash way to assign a **formatted string** to a variable - `printf -v`

Let me follow up on Zulip about this -- I like Zulip because it's easier to cross-link the threads

andychu:
Hm actually I found this code could be simplified a lot, without writing a new lexer:

https://github.com/oils-for-unix/oils/commit/f8b5059536688a94fba2a1c7886573ac1ace4af2

https://github.com/oils-for-unix/oils/commit/59ac1aff9c71701b5c82c331ec56bd9cd6af6810

There was some dead code, like normalizing `0x` to `0X`, which you changed the lexer to handle

---

The new function looks like this, which is pretty simple:

```
def _ParsePrintfInteger(s, blame_loc):

    # shells ignore space on the left, but not on the right!
    s = s.lstrip()

    # Handle +/- sign separately since the shell number lexer doesn't
    # recognize signed numbers as a single token
    negative = False
    if s.startswith('-'):
        negative = True
        s = s[1:]
    elif s.startswith('+'):
        # Positive sign is optional but allowed
        s = s[1:]

    # Borrow the lexer for $(( )), but handle it a bit differently.
    id_, pos = match.MatchShNumberToken(s, 0)
    if pos != len(s):  # no trailing data
        return (False, mops.BigInt(0))

    big_int = mops.ZERO
    if id_ == Id.ShNumber_Dec:
        ok, big_int = mops.FromStr2(s)

    elif id_ == Id.ShNumber_Oct:
        ok, big_int = mops.FromStr2(s[1:], 8)

    elif id_ == Id.ShNumber_Hex:
        ok, big_int = mops.FromStr2(s[2:], 16)

    else:  # Id.ShNumber_BaseN or Id.Unknown_Tok
        # Unlike $(( )), printf doesn't support 64#a
        return (False, mops.BigInt(0))

    if not ok:
        return (False, mops.BigInt(0))
    if negative:
        big_int = mops.Negate(big_int)
    return (True, big_int)
```

andychu:
Actually now that I look again, `blame_loc` is also unused!

daveads:
> Actually now that I look again, `blame_loc` is also unused!

oh okay ?

andychu:
We want to avoid having dead / unused code and the like

The code should be minimal :-)

daveads:
> We want to avoid having dead / unused code and the like
> 
> The code should be minimal :-)

sure

PR #2484, opened 2025-10-18
Add support for umask to parse symbolic input

### LINKED ISSUE #2460: yash, LuaRocks - umask with symbolic input isn't implemented
```
oils warning: umask with symbolic input isn't implemented

Error: LuaRocks 3.11.1 bug (please report at https://github.com/luarocks/luarocks/issues).
```

Comments here

[#distros > 2025-10-08-comm - running with aports/community @ 💬](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/2025-10-08-comm.20-.20running.20with.20aports.2Fcommunity/near/544324842)

Example: https://op.oils.pub/aports-build/2025-10-08-comm.wwz/_tmp/aports-report/2025-10-08-comm/diff_merged.html

EarthenSky:
close #2460

TODO:
- [x] add more tests touching on corner case behaviour
- [x] clean up code a bit (remove the double negative)
- [x] remove double umask calls
- [x] figure out why CI is failing

### DIFF
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1165,7 +1165,7 @@ Sets the bit mask that determines the permissions for new files and
 directories.  The mask is subtracted from 666 for files and 777 for
 directories.
 
-If no MODE, show the current mask.
+If no argument is provided, show the current mask.
 
 Flags:
 


EarthenSky:
@andychu I added two tests that are supposed to fail under osh, since we haven't implemented them yet (and probably don't need to atm). Is there any way I can let the CI know? https://op.oilshell.org/uuu/github-jobs/10389/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-process.html

It looks like `ovm-tarball` should be passing otherwise.

andychu:
Yes, you can change `oils_failured_allowed` at the beginning of the file

the last line of the run tells you whether the number of failures was expected

```

                dash    bash    mksh    zsh     osh
        pass    17      21      19      20      25
        ok      2       1       1       0       1
        N-I     2       1       1       1       0
        BUG     7       5       7       7       0
        FAIL    0       0       0       0       2
        total   28      28      28      28      28

builtin-process: note: Got 2 allowed Oils failures

```

EarthenSky:
Whew, lots of fighting with mycpp but looks good now. I'm open to any feedback you have about how the parsing is structured here @andychu!

builtin/process_osh.py
```diff
@@ -450,28 +458,116 @@ def Run(self, cmd_val):
...
+            print_stderr(
+                "oils warning: `%s` is an invalid symbolic mode operator" %
+                first_arg[0])
+            return 1
 
-        e_usage('umask: unexpected arguments', loc.Missing)
+        e_usage("unexpected number of arguments", loc.Missing)
+
+    def _SymbolicToOctal(self, initial_mask, arg_component):
+        # type: (int, str) -> Optional[Int]
```
andychu: We often use regular `int` and then something like `BAD_MASK = -1`  

Or you can also use `ok, octal = _ParseOctal()`, sort of like Go

I think one of those will work here

spec/builtin-process.test.sh
```diff
@@ -142,6 +143,135 @@ stat -c '%a' $TMP/umask-one $TMP/umask-two
...
+
+#### umask bad symbolic input
+umask b=rwx
+## status: 1
+## OK dash status: 2
+
+#### umask octal number out of range
+umask 0022
+umask 1234567
+# osh currently treats 0o1234567 as 0o0567
```
andychu: Can we handle this correctly in OSH?

If there is something TODO, then we just let OSH fail with oils_allowed_failures.

We don't mark OSH as a BUG -- that is only for non-OSH shells!

builtin/process_osh.py
```diff
@@ -450,28 +458,116 @@ def Run(self, cmd_val):
...
+                # the initial value (ex: umask u=rwx,a=rwx) although it's non-trivial to determine
+                # when, so it's probably not worth it
+                initial_mask = posix.umask(0)
+                new_mask = initial_mask
+
+                for arg_component in first_arg.split(","):
+                    maybe_new_mask = self._SymbolicToOctal(
+                        new_mask, arg_component)
+                    if maybe_new_mask is None:
+                        posix.umask(initial_mask)
```
andychu: Hm is it necessary to call umask() for each component?   Some high level description of the algorithm would be nice

builtin/process_osh.py
```diff
@@ -450,28 +458,116 @@ def Run(self, cmd_val):
...
+        elif len(arg_component) == 1:
+            print_stderr(
+                "oils warning: expected `=+-` after `%s` in symbolic mode operator"
+                % arg_component[0])
+            return None
+        elif arg_component[1] not in "=+-":
+            print_stderr(
+                "oils warning: `%s` is an invalid symbolic mode operator" %
+                arg_component[1])
+            return None
```
andychu: Hm I wonder if we can validate each part with a regex?  That might be simpler

We have some regex-based lexers in `frontend/lexer_def.py`, though you may need some help to create one, since there is some metaprogramming

builtin/process_osh.py
```diff
@@ -450,28 +458,116 @@ def Run(self, cmd_val):
...
+
+        mask_digit = int("111", 2)
```
andychu: Instead of `int()`, we should use octal literals like `0o123` or binary literals like `0b0101`

It looks like python2 supports them

```
Python 2.7.18 (default, Jan 23 2024, 02:54:29) 
[GCC 12.2.0] on linux2
>>> 0o0123
83
>>> 0b1001
9
```

builtin/process_osh.py
andychu: Yeah actually I looked at dash `miscbltin.c`, and it only calls umask once at the end of the loop

```
         return 1;
       }
       new_mask = ~new_mask;
     }
     umask(new_mask);

```

https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/miscbltin.c#n214

builtin/process_osh.py
```diff
@@ -450,28 +458,116 @@ def Run(self, cmd_val):
...
+                    new_mask = int(first_arg, 8)
+                except ValueError:
+                    # NOTE: This also happens when we have '8' or '9' in the input.
+                    print_stderr("oils warning: `%s` is not an octal number" %
+                                 first_arg)
+                    return 1
+
+                posix.umask(new_mask)
+                return 0
+            elif first_arg[0] in "ugoa":
```
andychu: I think this validation is possibly redundant

`help umask` says:


>    If MODE begins with a digit, it is interpreted as an octal number;   otherwise it is a symbolic mode string like that accepted by chmod(1).


So I think the existing error handling may suffice

spec/builtin-process.test.sh
```diff
@@ -142,6 +143,135 @@ stat -c '%a' $TMP/umask-one $TMP/umask-two
...
+## N-I dash/mksh/zsh status: 1
+
+#### umask -S
+# current mask as symbolic
+umask -S | grep 'u=[rwx]*,g=[rwx]*,o=[rwx]*' 
+## status: 0
+
+#### umask symbolic parsing
+
+umask 0000
```
andychu: It would be nice to have some comments here explaining the syntax

Actually I think we should update `doc/ref/chap-builtin-cmd.md` with the syntax

Now it says it only supports octal, but we are adding symbolic input

builtin/process_osh.py
andychu: Hm on second thought, I think most of these can go in "else" statements

andychu (commented):
Thanks for working on this!

The main thing I noticed is that it seems to make more umask calls than necessary

I think the whole string results in one umask call

And also I think some of the parsing can be consolidated

builtin/process_osh.py
```diff
@@ -450,28 +458,116 @@ def Run(self, cmd_val):
...
+        elif arg_component[1] not in "=+-":
+            print_stderr(
+                "oils warning: `%s` is an invalid symbolic mode operator" %
+                arg_component[1])
+            return None
+
+        mask_digit = int("111", 2)
+        for ch in arg_component[2:]:
+            if ch == 'r':
+                mask_digit &= int("011", 2)
```
andychu: Also I looked at dash source code - it seems to start with `positions = 0`, and `new_val = 0`, and then build up values with `|=`

Although there is other stuff going on

That could be more natural than using `&=`



```
      mask = ~mask;
       new_mask = mask;
       positions = 0;
       while (*ap) {
         while (*ap && strchr("augo", *ap))
           switch (*ap++) {
           case 'a': positions |= 0111; break;
           case 'u': positions |= 0100; break;
           case 'g': positions |= 0010; break;
           case 'o': positions |= 0001; break;
           }
         if (!positions)
           positions = 0111; /* default is a */
         if (!strchr("=+-", op = *ap))
           break;
         ap++;
         new_val = 0;
         while (*ap && strchr("rwxugoXs", *ap))
           switch (*ap++) {
           case 'r': new_val |= 04; break;
           case 'w': new_val |= 02; break;
           case 'x': new_val |= 01; break;
           case 'u': new_val |= mask >> 6;
               break;
           case 'g': new_val |= mask >> 3;
               break;
           case 'o': new_val |= mask >> 0;
               break;
           case 'X': if (mask & 0111)
               new_val |= 01;
               break;
           case 's': /* ignored */
               break;
           }
         new_val = (new_val & 07) * positions;
         switch (op) {
         case '-':
           new_mask &= ~new_val;
           break;
         case '=':
           new_mask = new_val
               | (new_mask & ~(positions * 07));
           break;
         case '+':
           new_mask |= new_val;
         }
```

andychu:
BTW these are the docs for the kill builtin:

https://github.com/oils-for-unix/oils/commit/e487951a29430515c4af28a0aff3c099e0ae43e8

https://op.oilshell.org/uuu/github-jobs/10567/ovm-tarball.wwz/_release/VERSION/doc/ref/chap-builtin-cmd.html#kill

The umask builtin is simpler

But I do think it helps to write out the docs, for testing and code structure

----

Compare with bash `help umask`
 
```
umask: umask [-p] [-S] [mode]
    Display or set file mode mask.
    
    Sets the user file-creation mask to MODE.  If MODE is omitted, prints
    the current value of the mask.
    
    If MODE begins with a digit, it is interpreted as an octal number;
    otherwise it is a symbolic mode string like that accepted by chmod(1).
    
    Options:
      -p        if MODE is omitted, output in a form that may be reused as input
      -S        makes the output symbolic; otherwise an octal number is output
    
    Exit Status:
    Returns success unless MODE is invalid or an invalid option is given.

```

andychu:
My first reading is something like

```
umask (OCTAL | SYMBOLIC)  # set the mask

umask FLAG*  # print the mask

Flags:
  -p   print in a form that may be reused as input
  -S  print in symbolic form

Mask symbols

  who operator permission*  # are all of these required?

  who :: u | g | o | a  
  operator :: = | + | -
  permission :: r | w | x
```

So yeah that makes more sense -- there are 2 forms, either get or set

---

Oh actually there is already a grammar, because it has the same syntax as POSIX chmod:

https://pubs.opengroup.org/onlinepubs/009696899/utilities/chmod.html


And yeah I was missing the `,` there

Well maybe we can just say it's the same as chmod, and link to that document

I wanted to understand the syntax a bit better, which will help me read the code

andychu:
But actually it seems more complex than I thought, and I wonder if we have all of it ... I will take a closer look

builtin/process_osh.py
EarthenSky: Since the function returns immediately after calling umask, the algorithm only ever calls `posix.umask` twice (to get the mask & in this case to reset it due to an error). However you make a good point that it's misleading to have the umask call inside the loop!

I'll move it outside when I clean up the logic & the high level description.

> Another interesting thought is what happens if an error is raised halfway through the function? My error handling is pretty brittle, so umask might not get unset! I'll double check that it's not possible, or otherwise add an except block.

spec/builtin-process.test.sh
```diff
@@ -142,6 +143,135 @@ stat -c '%a' $TMP/umask-one $TMP/umask-two
...
+umask a=rwx
+umask | tail -c 4
+
+umask 0124
+umask a+r
+umask | tail -c 4
+
+umask 0124
+umask a-r
+umask | tail -c 4
```
andychu: Why add `tail -c 4`  everywhere?  

Ah OK I see now bash and dash print 0022, but mksh and zsh print 022

That is annoying ...

builtin/process_osh.py
```diff
@@ -450,28 +458,116 @@ def Run(self, cmd_val):
...
+                mask_digit &= int("101", 2)
+            elif ch == 'x':
+                mask_digit &= int("110", 2)
+            else:
+                print_stderr(
+                    "oils warning: `%s` is an invalid symbolic mode character"
+                    % ch)
+                return None
+
+        if arg_component[0] == "u":
```
andychu: For the var names, let's use the same terminology as the POSIX doc:

clause

who

perm

permcopy

actionlist

----

By using that terminology, it looks like we are missing the perm symbol X

and the permcopy symbols `ugo`

andychu:
Hm  zsh doesn't support this either

```
$ zsh -c 'umask; umask uo=g; umask'
022
zsh:umask:1: bad symbolic mode permission: g
022
```


This is a rare case -- usually bash and zsh support much MORE than the other shells!

---

So actually I'd say it's good to start simple ... I like to start with the minimum, as long as it's documented

spec/builtin-process.test.sh
EarthenSky: ahh, that makes a lot of sense. Updated to the following

```sh
17     230     BUG     pass    BUG     BUG     FAIL    umask octal number out of range
```

https://github.com/oils-for-unix/oils/pull/2484/commits/cc1a1c337f7cd3850bd81c675cebe008edc4c5a0

builtin/process_osh.py
EarthenSky: It's been a while since I've used Go, but I do prefer that syntax. Good idea!
https://github.com/oils-for-unix/oils/pull/2484/commits/c8972d8a14aad8bc7698f616689f0ad40acd9c10

builtin/process_osh.py
EarthenSky: Refactored the umask call out of the loop, so it's clearer that it can never be called more than twice https://github.com/oils-for-unix/oils/pull/2484/commits/c8972d8a14aad8bc7698f616689f0ad40acd9c10

builtin/process_osh.py
EarthenSky: Sadly binary literals don't work with mycpp (I had them originally), but I can switch to octal since I use them later in the same function https://github.com/oils-for-unix/oils/commit/c8972d8a14aad8bc7698f616689f0ad40acd9c10

builtin/process_osh.py
EarthenSky: Good point!
https://github.com/oils-for-unix/oils/pull/2484/commits/49ad77bb307c4ed046811b93f5268200dc476918

builtin/process_osh.py
EarthenSky: I took a look at the POSIX doc and it's great! Next time I'll look for a spec before just trying to match `bash`'s behaviour. I see the cases you mention above that we don't support. I'll reimplement the function that evaluates a single clause in order to deal with these cases (except for the += one, probably), then add tests for them.

EarthenSky:
I've added support for everything the POSIX spec mentioned, except for one key issue: It seems that bash doesn't support permcopy (ugo), while dash/mksh merge permcopy and permlist into a single list, while POSIX says that only one permcopy character should be allowed for each `op`.

For the sake of compatibility, I have it set to mimic dash/mksh for now, since this behaviour is a superset of POSIX umask, but I'm interested in what you think is best? I'll wait to get your feedback until after I add a few more tests so we know what the behaviours are.

TODO:
- [x] update tests & add a bunch more to show the differences between shells

builtin/process_osh.py
```diff
@@ -446,32 +627,79 @@ def __init__(self):
...
+        wholist = parser.ParseWholist()
+        if parser.DoneParsing():
+            print_stderr("oils warning: actionlist is required")
+            return False, 0
+
+        while True:
+            ok, mask = parser.ParseNextAction(wholist, mask, initial_mask)
+            if not ok:
+                return False, 0
+            elif parser.DoneParsing():
```
andychu: I find that using a sentinel symbol can avoid extra guards like this

i.e. I discovered that `NUL-termination` is  actually a good thing from C that Python doesn't have idiomaticaly :-)

It makes parsers resemble grammars more --  there is a more straightforward translation

builtin/process_osh.py
```diff
@@ -446,32 +627,79 @@ def __init__(self):
...
+                posix.umask(new_mask)
+                return 0
+
+        e_usage("unexpected number of arguments", loc.Missing)
+
+    def _ParseClauseList(self, initial_mask, clause_list):
+        # type: (int, List[str]) -> Tuple[bool, int]
+        mask = initial_mask
+        for clause in clause_list:
+            ok, mask = self._ParseClause(mask, initial_mask, clause)
```
andychu: does this end up doing an |= , or it more complex than that ?

andychu (commented):
I took a quick look ...  this looks like it's easier to see what the language we accept is, thanks!

mycpp won't like the `@staticmethod` -- we don't use that anywhere else in the code

---

I would just make those free functions, and if you like, this could be moved into `builtin/umask_osh.py` , if you want to use the file as a namespace instead



https://op.oilshell.org/uuu/github-jobs/10715/cpp-coverage.wwz/_tmp/soil/logs/compare-gcc-clang.txt

builtin/process_osh.py
EarthenSky: Sadly not, but it was close! It mixes |= to add bits to the mask (`-` case), and &= to clear bits from the mask (`+=` case). We could possibly return two masks (an add mask and a clear mask), then apply them at the end. I hadn't considered that before. I'll take a look

builtin/process_osh.py
andychu: Actually I realized that our arg parser does use the same style -- `arg_r.AtEnd()`

So this is not very important, or it may or may  not be actionable

The termination conditions in parsers sometimes annoy me :-) 

For awhile I thought it may be better to parse tokens like `['-r', 'foo', None]` instead of `['-r', 'foo']`, but we are not doing that

EarthenSky:
> mycpp won't like the `@staticmethod` -- we don't use that anywhere else in the code

ahh, I figured haha

> I would just make those free functions, and if you like, this could be moved into `builtin/umask_osh.py` , if you want to use the file as a namespace instead

oh, this is a good idea. I'll give it a shot

builtin/process_osh.py
EarthenSky: I hear what you're saying, although I think it would be more work in this context since each token is a string. If we were working with string tokens I see how it would help avoid accidentally reading past the end though!

spec/builtin-process.test.sh
```diff
@@ -142,6 +143,402 @@ stat -c '%a' $TMP/umask-one $TMP/umask-two
...
+246
+## END
+## BUG dash/mksh STDOUT:
+ret0 = 0
+666
+ret1 = 0
+666
+ret2 = 0
+357
+## END
```
EarthenSky: > The perm symbol X shall represent the execute/search portion of the file mode bits if the file is a directory or if the current (unmodified) file mode bits have at least one of the execute bits set. **It shall be ignored if the file is not a directory and none of the execute bits are set in the current file mode bits.**

dash & mksh treat `X` as equivalent to `x`. This is surprising considering the code snippet from dash you provided seems to handle `X` properly. Perhaps it's a version issue.

spec/builtin-process.test.sh
```diff
@@ -142,6 +143,402 @@ stat -c '%a' $TMP/umask-one $TMP/umask-two
...
+124
+124
+## END
+## BUG zsh STDOUT:
+777
+ret = 1
+124
+124
+124
+## END
```
EarthenSky: zsh supports `umask =`, but not `umask =rwx`

spec/builtin-process.test.sh
```diff
@@ -142,6 +143,402 @@ stat -c '%a' $TMP/umask-one $TMP/umask-two
...
+000
+## END
+## N-I bash STDOUT: 
+124
+124
+124
+## END
+## BUG zsh status: 1
+## BUG zsh STDOUT:
+## END
```
EarthenSky: I'm not sure what this error is. I only get the same results if I run the following as a single command, rather than as 2 separate commands.

```sh
umask =+=
umask
```

EarthenSky: This same kind of IO-blocking failure also popped up with `umask =rw`, but only on the CI machine. Locally it didn't block any IO.

andychu:
Ah for the `zsh` failure, you can now add `compare_shells: zsh-5.9`

@btertoolen said that fixed things for him I think?

To test locally, you will have to install new wedges, as described here:

[Instructions for new setup ../oils.DEPS](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Instructions.20for.20new.20setup.20.2E.2E.2Foils.2EDEPS/with/548214479)

`zsh-5.9` is already in the CI image

If this doesn't help get past the error, please start a thread on Zulip

andychu:
I took a look, and it looks pretty nice now

One thing that would help me review is you can update the help in `doc/ref/chap-builtin-cmd.md` with a description of the subset we support

There seems to be significant differences among shells!  Thanks for finding that

---

I pushed some changes:

- moving the tests into their own file
- adding a flag parser, which will let us support `-p` and `-S`
- move methods that don't depend on `self` into  free functions

I also wonder if it makes sense to use exceptions for the error cases ... that is a judgement call, but in this case it seems like we parse a umask, or there is an error

And sometimes the errors can happen 2 or 3 functions deep?

EarthenSky:
Ty for the changes! I was just about to move the spec tests then write the docs, so nice timing.

> One thing that would help me review is you can update the help in `doc/ref/chap-builtin-cmd.md` with a description of the subset we support

Sure thing!

> I also wonder if it makes sense to use exceptions for the error cases ... that is a judgement call, but in this case it seems like we parse a umask, or there is an error

I added exception handling as a "just in case" kind of thing. If there was a bug where we accidentally read past the end of an array, that would raise an error (presumably in the mycpp version of osh too?) resulting in the value of umask being lost. Does this sound like overkill? I was hoping to get your opinion on this.

I also added some assertions for cases that should be unreachable, but those are essentially glorified comments and we shouldn't expect them to be triggered.

andychu:
Yeah I was wondering about the `except Exception`.  It is a funny issue that you have to restore the umask, so it makes sense

Sometimes we do that with Python context managers / C++ destructors

But I'd say it's OK for now.  That's a minor detail

EarthenSky:
> One thing that would help me review is you can update the help in `doc/ref/chap-builtin-cmd.md` with a description of the subset we support

Hey @andychu I updated umask's description in `chap-builtin-cmd.md`. Let me know if there's anything else left to do for this PR!

builtin/umask_osh.py
```diff
@@ -0,0 +1,284 @@
...
+
+        first_arg, first_loc = arg_r.ReadRequired2('expected an argument')
+        arg_r.Done()  # only one arg
+
+        if first_arg[0].isdigit():
+            try:
+                new_mask = int(first_arg, 8)
+            except ValueError:
+                # NOTE: This also happens when we have '8' or '9' in the input.
+                print_stderr("oils warning: `%s` is not an octal number" %
```
andychu, line 261: I think many of these "oils warning" can be a UsageError -- that is the style we use for invalid input

builtin/umask_osh.py
```diff
@@ -0,0 +1,284 @@
...
+
+            posix.umask(new_mask)
+            return 0
+
+        # NOTE: it's possible to avoid this extra syscall in cases where we don't care about
+        # the initial value (ex: umask ...,a=rwx) although it's non-trivial to determine
+        # when, so it's probably not worth it
+        initial_mask = posix.umask(0)
+        try:
+            ok, new_mask = _ParseClauseList(initial_mask, first_arg.split(","))
```
andychu, line 273: I wonder if we can parse it first, and avoid needing to restore it to `initial_mask` if the parse fails

PR #2485, opened 2025-10-18
Implement hash -p to set a command path

eswarseeram:
Adds support for the -p option in the hash builtin in Oil Shell.
- Allows manually mapping a command to a specific path.
- Updates the internal hash table for shell commands.

### DIFF
--- a/frontend/flag_def.py
+++ b/frontend/flag_def.py
@@ -91,6 +91,7 @@ SHOPT_SPEC.ShortFlag('-q')  # query option settings
 
 HASH_SPEC = FlagSpec('hash')
 HASH_SPEC.ShortFlag('-r')
+HASH_SPEC.ShortFlag('-p', args.String)
 
 ECHO_SPEC = FlagSpec('echo')
 ECHO_SPEC.ShortFlag('-e')  # no backslash escapes


andychu:
Thank you for looking into this!  We like to start with spec tests -- te.g. there are related ones  in `spec/command_.test.sh`

```
$ test/spec.sh command_ -r 8-15
command_.test.sh
  8      57     pass    pass    pass    pass    pass    $PATH lookup
  9      70     pass    pass    pass    ok      pass    filling $PATH cache, then insert the same command earlier in cache
 10     101     pass    pass    ok      ok      pass    filling $PATH cache, then deleting command
 11     133     pass    pass    pass    pass    pass    Non-executable on $PATH
 12     154     pass    ok      ok      ok      pass    hash without args prints the cache
 13     171     pass    pass    BUG     pass    pass    hash with args
 14     190     BUG     BUG     pass    pass    ok      hash -r doesn't allow additional args
 15     197     pass    pass    pass    pass    pass    Executing command with same name as directory in PATH (#2429)

                dash    bash    mksh    zsh     osh
        pass    7       6       5       5       7
        ok      0       1       2       3       1
        BUG     1       1       1       0       0
        total   8       8       8       8       8

```

builtin/pure_osh.py
```diff
@@ -382,6 +382,16 @@ def Run(self, cmd_val):
         arg = arg_types.hash(attrs.attrs)
 
         rest = arg_r.Rest()
+        if '-p' in attrs.attrs:
```
andychu, line 385: To add a flag, see

```
HASH_SPEC = FlagSpec('hash')
HASH_SPEC.ShortFlag('-r')
```

in `frontend/flag_def.py`  .  And then regenerated code with `build/py.sh all`

And then you can use `arg.p`

According to `help hash`, it is a flag that takes an argument, not 2 arguments.    `-p pathname`

So that's `ShortFlag('-p', args.String)`

eswarseeram:
Updated hash -p flag implementation as per maintainer feedback. Only single argument is taken now.

andychu:
As mentioned, this would still need spec tests

Issue #2416, opened 2025-09-12
abuild, ifupdown-ng, ... -  "cannot provide test case names with -l" - OSH disagrees on exit code of unquoted $(false)


andychu:
This is a common bug that we need to reduce -- I'm not sure what's going on

This is from `regtest/aports`

(This is currently cause 4)

melvinw:
OK finally got to the bottom of this one! It looks like an issue with the backtick eval syntax.

`ifupdown-ng`, and a few other projects, use a test framework called [ATF](https://github.com/freebsd/atf/tree/a7b2c4c89fb990c4ef34eb0e01fbd7b883ab49c4). Projects that use ATF provide tests as shell scripts that eventually get fed through this [harness script](https://github.com/freebsd/atf/tree/a7b2c4c89fb990c4ef34eb0e01fbd7b883ab49c4). In it there's a standard `getopts` loop. If the `-l` flag is present, its presence is recorded by setting a local called `_lflag` to `true`. While validating the args, the script checks to see if `_lflag` is set (by evaluating it with backticks) and errors out if there's anything left in argv, because it's meant to be used by itself. If you instrument the script with some prints, you'll see that whenever the script is called with `-l` there is nothing else in argv. I thought this was suspicious, so I put together the following script.

```
foo=false
if `${foo}`;
then
        echo TRUE
else
        echo FALSE
fi
```

lo and behold! bash and osh handle this differently.
```
$ bin/osh _tmp/foo.sh
TRUE
$ bash _tmp/foo.sh
FALSE
```

I'll send a spec test.

melvinw:
This gets more interesting! I wanted to see if `eval` had the same problem, so I changed the script to the following:
```
foo=$1
if eval $foo;
then
        echo TRUE
else
        echo FALSE
fi
if `$foo`;
then
        echo TRUE
else
        echo FALSE
fi
```

Apparently osh only breaks when using back ticks AND when the expression evaluates to something falsey
```
$ bash _tmp/foo.sh true
TRUE
TRUE
$ bash _tmp/foo.sh false
FALSE
FALSE


$ bin/osh _tmp/foo.sh true
TRUE
TRUE
$ bin/osh _tmp/foo.sh false
FALSE
TRUE
```

melvinw:
Parse tree for the original script, in case I don't get to the bottom of this tonight
```
(command.CommandList
  children:[
[...]

PR #2490, opened 2025-10-20
[core] When running a command where argv=[], keep the previous status code

last-genius:
Fixes: #2416

### DIFF
--- a/osh/cmd_eval.py
+++ b/osh/cmd_eval.py
@@ -1265,17 +1265,9 @@ class CommandEvaluator(object):
 
             self.tracer.OnShAssignment(lval, pair.op, rhs, flags, which_scopes)
 
-        # PATCH to be compatible with existing shells: If the assignment had a
-        # command sub like:
-        #
-        # s=$(echo one; false)
-        #
-        # then its status will be in mem.last_status, and we can check it here.
-        # If there was NOT a command sub in the assignment, then we don't want to
-        # check it.
-
-        # Only do this if there was a command sub?  How?  Look at node?
-        # Set a flag in mem?   self.mem.last_status or
+        # For the cases like
+        #   a=$(false)
+        #   $(false) $(exit 42)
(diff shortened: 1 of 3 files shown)


osh/cmd_eval.py
```diff
@@ -763,7 +763,8 @@ def _RunSimpleCommand(self, cmd_val, cmd_st, run_flags):
                 cmd_val = cast(cmd_value.Argv, UP_cmd_val)
                 self.tracer.OnSimpleCommand(cmd_val.argv)
                 return self.shell_ex.RunSimpleCommand(cmd_val, cmd_st,
-                                                      run_flags)
+                                                      run_flags,
+                                                      self.mem.LastStatus())
```
andychu: Hm does this have to be a parameter?  Could we return `self.mem.LastStatus()`  in RunSimpleCommand?

It has access to `self.mem` too right?




----

TBH I am not sure  why `$(false)`  sets the last status ... I remember there is a little hack in `_DoShAssignment` like this ... which seems very related

```
       if self.check_command_sub_status:
            last_status = self.mem.LastStatus()
            self._CheckStatus(last_status, cmd_st, node, loc.Missing)
            return last_status  # A global assignment shouldn't clear $?.
        else:
            return 0
```

andychu:
Note to myself: in `core/executor.py RunCommandSub`, it does call `SetLastStatus()`

I guess all shells do that - they don't just set `$?` on a simple command

---

I guess this to say that there was already what I see as a special case here, and this is a slight tweak on the same special case

Luckily in YSH we disallow empty argv, so this quirk doesn't come up!

---

But yeah I still wonder if we can avoid the param

last-genius:
Got rid of the parameter, it's a one-line fix now! Still passes all the tests and doesn't break anything else.

andychu:
Excellent, thank you!

I'm looking forward to doing a `regtest/aports` run with this fix, and the paren fix

Issue #2487, opened 2025-10-18
expr sub $[42] should also work in expression mode


andychu:
We did expr splice with @[ [1, 2, 3] ] with issue  #1472 , PR #2482

PR #2491, opened 2025-10-20
expr sub works in expression mode

daveads:
Handles this issue #2487

### DIFF
--- a/spec/ysh-expr-sub.test.sh
+++ b/spec/ysh-expr-sub.test.sh
@@ -81,16 +81,16 @@ pp test_ (@[ bad ])  # cannot be stringified
 
 # Basic integer expression
 var a = $[42]
-echo "a = $a"
+pp test_ (a)
 
 # Arithmetic expression
 var b = $[1 + 2]
-echo "b = $b"
+pp test_ (b)
 
 # Function call
 var arr = [1, 2, 3]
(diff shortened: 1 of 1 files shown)


daveads:
https://github.com/user-attachments/assets/1db44a4a-519e-4e05-bc3c-32d0a9a0a566

daveads:
@andychu

spec/ysh-expr-sub.test.sh
```diff
@@ -73,3 +73,38 @@ pp test_ (@[ bad ])  # cannot be stringified
...
 ## END
+
+
+#### $[expr_sub] in expression mode
+
+# Test that $[...] works in expression mode, not just in command/word mode
+
+# Basic integer expression
+var a = $[42]
+echo "a = $a"
```
andychu: Please change this to

pp test_ (a)
pp test_ (b)
pp test_ (c)

So then we can see that it's a `Str` value

Because `echo a = $a` will print the same thing for an `Int` value

`pp test_` is a special builtin we use for testing

----

Also, please test an error case at the end, like `var x = $[ [42] ]`   -- this can't be "stringified"

daveads:
@andychu done with this...

andychu:
Great, thank you!

Issue #2492, opened 2025-10-21
regtest/yash issue: osh does not respect line continuation "${x-}"


btertoolen:
This issue occured while building yash using osh in the regtest. The issue is the difference between bash and osh here:
```
[bram@bram-dell oils]$ value="${EMPTY_VALUE-a \
b}"
[bram@bram-dell oils]$ echo $value
a b
[bram@bram-dell oils]$ bin/osh
osh-0.36$ value="${EMPTY_VALUE-a \
b}"
osh-0.36$ echo $value
a \ b
```
Bash respects the line continuation, while osh doesnt (it adds the '\' to the actual value).

Note that without the `\` everything works as intended:
```
[bram@bram-dell oils]$ v="${EMPTY_VALUE-a
b}"
[bram@bram-dell oils]$ echo $v
a b
[bram@bram-dell oils]$ bin/osh
osh-0.36$ v="${EMPTY_VALUE-a
b}"
osh-0.36$ echo $v
a b
osh-0.36$
```

For more info see [this zullip thread](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/regtest.2Faports.20yash.20build.20failure/with/546126829)

PR #2493, opened 2025-10-21
add spec test for #2492 -- line continuation in "${x-}"

btertoolen:
Add a failing spec test for #2492


andychu:
Ah looks like `oils_allowed_failures` has to be adjusted, but maybe you are also working on the code fix too

btertoolen:
> Ah looks like `oils_allowed_failures` has to be adjusted, but maybe you are also working on the code fix too

@andychu No, I was thinking of maybe adding more tests if I come up with them, but I'll do that in a separate branch where I'm also working on the fix

btertoolen:
This PR will be replaced by #2498

Issue #2380, opened 2025-07-30
crash printing eggex


bar-g:
```
ysh-0.34$ echo $[/[\\0]/]

ysh: _gen/bin/oils_for_unix.mycpp.cc:58306: syntax_asdl::class_literal_term_t* 
expr_to_ast::Transformer::_NonRangeChars(pnode::PNode*): Assertion `0' failed.
```

andychu:
This looks fixed now, thanks for the report

```
ysh-0.35$ echo $[/\\0/]
  echo $[/\\0/]
            ^
[ interactive ]:3: Syntax error in expression (near Id.Expr_DecInt)

```

bar-g:
That is something else, though, and does fail cleanly here, too:
```
ysh-0.35$ echo $[/\\0/]
 echo $[/\\0/]
           ^
[ interactive ]:1: Syntax error in expression (near Id.Expr_DecInt)
```

But the actual issue reported is still the same:
```
ysh-0.35$ echo $[/[\\0]/]
ysh: _gen/bin/oils_for_unix.mycpp.cc:58629: syntax_asdl::class_literal_term_t* expr_to_ast::Transformer::_NonRangeChars(pnode::PNode*): Assertion `0' failed.
```

daveads:
Checking this out

PR #2494, opened 2025-10-21
regex

daveads:
fixes this issue... #2380

### DIFF
--- a/ysh/expr_to_ast.py
+++ b/ysh/expr_to_ast.py
@@ -1408,9 +1408,15 @@ class Transformer(object):
         if typ0 == grammar_nt.char_literal:  # /[ \n \u{3bc} \yff ]/
             return word_compile.EvalCharLiteralForRegex(child0.tok)
 
-        if typ0 in (Id.Expr_Name, Id.Expr_DecInt):  # /[ d 9 ]/
+        if typ0 == Id.Expr_DecInt:  # /[5]/
+            tok = child0.tok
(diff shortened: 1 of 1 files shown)


daveads:
Screenshot 2025-10-22 at 12 30 43 AM

andychu:
Ah thank you, this looks reasonable

I'm not sure what the original code was trying to do

andychu:
Merged, thank you!

I improved the error message a bit, when getting expressions like `/[25]/` , which are not valid

The idea here is that usually you write

```
/['0'-'9']/
/['0']/
```

Because in Eggex, you're supposed to quote strings.    This means that `-` and `!` and `^` are not ambiguous -- are the regex operators, or characters?

But as a special case, we allow you to write

```
/[0-9]/
/[0]/
```

like traditional regex.  So the last case was broken!

daveads:
> Merged, thank you!
> 
> I improved the error message a bit, when getting expressions like `/[25]/` , which are not valid
> 
> The idea here is that usually you write
> 
> ```
> /['0'-'9']/
> /['0']/
> ```
> 
> Because in Eggex, you're supposed to quote strings. This means that `-` and `!` and `^` are not ambiguous -- are the regex operators, or characters?
> 
> But as a special case, we allow you to write
> 
> ```
> /[0-9]/
> /[0]/
> ```
> 
> like traditional regex. So the last case was broken!

okay

Issue #2470, opened 2025-10-13
''' Multi-line strings don't enforce the r prefix, like non-multi-line strings


danyspin97:
`var y = '\u \\ '` is not valid due to backslashes, but `var y = r'\u \\'` is. 

```
var y = '''
\u \\
'''
```
The code above works as expected.

andychu:
not accepted:


```
ysh-0.36$ var y = '\u'
  var y = '\u'
           ^~
[ interactive ]:5: Strings with backslashes should look like r'\n' or u'\n' or b'\n'

```


accepted:


```
$ var y = ''' \u '''
```

andychu:
Related

- #2458

daveads:
> not accepted:
> 
> ```
> ysh-0.36$ var y = '\u'
>   var y = '\u'
>            ^~
> [ interactive ]:5: Strings with backslashes should look like r'\n' or u'\n' or b'\n'
> ```
> 
> accepted:
> 
> ```
> $ var y = ''' \u '''
> ```

i need some more clarification on the expected behavior of the multi-line strings...  @andychu

andychu:
Oh yes, the `'''` strings should REJECT the same strings as the `'` string

Namely when there are backslashes

We want the user to be clear.      In Python, note that `r'\n'` means something very different than `'\n'`

That's the same different in YSH

So `'\n'` in YSH is ambiguous -- we want it to be either `r'\n'` (raw)   or `u'\n'` (newline)

Likewise `'''\n'''` is ambiguous for the same reason

It should be either `r'''\n'''` or  `u'''\n'''`

andychu:
On the other hand, there is no problem with `'''foo'''` -- it doesn't contain any backslashes, so `r'''foo'''` and `u'''foo'''` mean exactly the same thing

Since they mean the same thing, you can write it with no prefix -- `'''foo'''`

PR #2495, opened 2025-10-22
multiline behavior

daveads:
fix this issue #2470


daveads:
@andychu check it out when free

andychu:
Thanks!  I also added it to the error catalog, like this:

https://op.oilshell.org/uuu/github-jobs/10456/ovm-tarball.wwz/_release/VERSION/doc/error-catalog.html#oils-err-20

Issue #2441, opened 2025-09-28
zfs install fails with a large token


melvinw:
From: https://op.oils.pub/aports-build/2025-09-27-disagree.wwz/_tmp/aports-report/2025-09-27-disagree/disagree-2025-09-18-bash/osh-as-sh/log/zfs.log.txt

```
Making install in tests/zfs-tests/tests
make[1]: Entering directory '/home/udu/aports/main/zfs/src/zfs-2.3.2/tests/zfs-tests/tests'
make[2]: Entering directory '/home/udu/aports/main/zfs/src/zfs-2.3.2/tests/zfs-tests/tests'
make[2]: Nothing to be done for 'install-exec-am'.
 ../../.././config/install-sh -c -d '/home/udu/aports/main/zfs/pkg/zfs/usr/share/zfs/zfs-tests/tests'
 ../../.././config/install-sh -c -d '/home/udu/aports/main/zfs/pkg/zfs/usr/share/zfs/zfs-tests/tests/functional/cli_root/zpool_upgrade/blockfiles'
 /usr/bin/install -c -m 644  functional/cli_root/zpool_upgrade/blockfiles/zfs-broken-mirror1.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-broken-mirror2.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v10.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v11.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v12.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v13.
dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v14.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v15.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v1.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v1mirror1.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v1mirror2.dat.bz2 functional/cli_root/zpool_upgrade/blockfiles/zfs-pool-v1mirror3.dat.bz2 function
[...]

PR #2497, opened 2025-10-22
[osh] Store token length in 32 bits instead of 16

last-genius:
This means that maximum length of a token increases from 2^16 to 2^32 accordingly.

Adjust the test expecting osh's failure in handling large tokens.

Fixes: #2441

### DIFF
--- a/core/alloc.py
+++ b/core/alloc.py
@@ -143,7 +143,8 @@ class Arena(object):
     def NewToken(self, id_, col, length, src_line):
         # type: (int, int, int, SourceLine) -> Token
 
-        if length >= 65536:
+        # 2**32 - token length is stored in an int (32 bits)
+        if length >= 4294967296:
             raise error.Parse(
                 '',  # ignored message
                 loc.TokenTooLong(src_line, id_, length, col))
--- a/frontend/syntax.asdl
+++ b/frontend/syntax.asdl
@@ -84,10 +84,10 @@ module syntax
   # Note that ASDL generates:
   #    typedef uint16_t Id_t;
   # So Token is
-  #    8 bytes GC header + 2 + 2 + 4 + 8 + 8 = 32 bytes on 64-bit machines
+  #    8 bytes GC header + 2 + 4 + 4 + 8 + 8 = 34 bytes on 64-bit machines
   #
   # We transpose (id, col, length) -> (id, length, col) for C struct packing.
-  Token = (id id, uint16 length, int col, SourceLine? line, str? tval)
+  Token = (id id, int length, int col, SourceLine? line, str? tval)
 
   # I wanted to get rid of Token.tval with this separate WideToken type, but it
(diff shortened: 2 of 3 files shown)


core/alloc.py
```diff
@@ -143,7 +143,8 @@ def SnipCodeString(self, left, right, inclusive=True):
     def NewToken(self, id_, col, length, src_line):
         # type: (int, int, int, SourceLine) -> Token
 
-        if length >= 65536:
+        # 2**32 - token length is stored in an int (32 bits)
+        if length >= 4294967296:
```
andychu, line 147: Hmm I was thinking we just get rid of this check altogether, and the associated TokenTooLong code

It is almost philosophical :-P  What happens if you feed a shell a file with 4 GiB of 'a' characters??

I am not sure I want to find out :-)

andychu:
Hm here is one example ...

```
9665197651986519965200652016520265203652046520565206652076520865209652106521165212652136521465215652166521765218652196522065221652226522365224652256522665227652286522965230652316523265233652346523565236652376523865239652406524165242652436524465245652466524765248652496525065251652526525365254652556525665257652586525965260652616526265263652646526565266652676526865269652706527165272652736527465275652766527765278652796528065281652826528365284652856528665287652886528965290652916529265293652946529565296652976529865299653006530165302653036530465305653066530765308653096531065311653126531365314653156531665317653186531965320653216532265323653246532565326653276532865329653306533165332653336533465335653366533765338653396534065341653426534365344653456534665347653486534965350653516535265353653546535565356653576535865359653606536165362653636536465365653666536765368653696537065371653726537365374653756537665377653786537965380653816538265383653846538565386653876538865389653906539165392653936539465395653966539765398653996540065401654026540365404654056540665407654086540965410654116541265413654146541565416654176541865419654206542165422654236542465425654266542765428654296543065431654326543365434654356543665437654386543965440654416544265443654446544565446654476544865449654506545165452654536545465455654566545765458654596546065461654626546365464654656546665467654686546965470654716547265473654746547565476654776547865479654806548165482654836548465485654866548765488654896549065491654926549365494654956549665497654986549965500655016550265503655046550565506655076550865509655106551165512655136551465515655166551765518655196552065521655226552365524655256552665527655286552965530655316553265533655346553565536: File name too long


andy@hoover:~/git/oils-for-unix/oils$ for i in {0..65536}; do echo -n $i; done | sh
```




```
36: File name too long
```

actually I think that is the MAX_ARG check

And then there are also comment tokens

andychu:
comments

```
$ { echo -n '#'; for i in {0..100000}; do echo -n $i; done; } | sh

(this is fine)

$ { echo -n '#'; for i in {0..100000}; do echo -n $i; done; } | osh
  #0123456789
   ^
[ stdin ]:1: Token starting at column 1 is too long: 488896 bytes (Id.Ignored_Comment)

```

(with existing OSH)

PR #2498, opened 2025-10-22
Line continuation

btertoolen:
Addresses issue #2492.
Fixes the following snippet
```
echo ${var-a \
b}
```
Before the output would be:
```
a \b
```

While it should be:
```
a b
```

### DIFF
--- a/frontend/lexer_def.py
+++ b/frontend/lexer_def.py
@@ -76,10 +76,12 @@ _BACKSLASH = [
     C('\\\n', Id.Ignored_LineCont),
 ]
 
+_DQ_ESCAPED_CHAR = R(r'\\[$`"\\]', Id.Lit_EscapedChar)
+
 # Only 4 characters are backslash escaped inside "".
 # https://www.gnu.org/software/bash/manual/bash.html#Double-Quotes
 _DQ_BACKSLASH = [
-    R(r'\\[$`"\\]', Id.Lit_EscapedChar),
+    _DQ_ESCAPED_CHAR,
     C('\\', Id.Lit_BadBackslash),  # syntax error in YSH, but NOT in OSH
 ]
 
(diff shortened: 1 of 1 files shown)


btertoolen:
@andychu This one is ready for review

frontend/lexer_def.py
```diff
@@ -442,12 +442,16 @@ def R(pat, tok_type):
     R(r'[^\0]', Id.Lit_Other),  # e.g. "$", must be last
 ]
 
-# Kind.{Lit,Ignored,VSub,Left,Right,Eof}
-LEXER_DEF[lex_mode_e.VSub_ArgDQ] = \
-  _DQ_BACKSLASH +  _VS_ARG_COMMON + _LEFT_SUBS + _VARS + [
-
+_DQ_VSUB_ARG_BACKSLASH = [
+    R(r'\\[$`"\\]', Id.Lit_EscapedChar),
```
andychu: This looks like the same as the _DQ_BACKSLASH rule right?

If so please extract it into a common var, similar to _U_BRACED_CHAR

could be _DQ_ESCAPED_CHAR

That makes it clearer that they are the same

spec/var-sub-quote.test.sh
```diff
@@ -311,3 +311,71 @@ echo "${var-"}"}"
 }
 ## END
 
+#### Var substitution with newlines (#2492)
+echo "${var-a \
```
andychu, line 315: It looks like all 4 shells agree on this first case, so let's separate that into its own test case

It is good to see GREEN across a row

Right now it looks like just bash and OSH agree, which is a bit misleading

spec/var-sub-quote.test.sh
```diff
@@ -311,3 +311,71 @@ echo "${var-"}"}"
...
+b}"
+echo "${var-a\nb}"
+echo "${var-a
+b}"
+
+echo "${var:-c \
```
andychu, line 320: this is also agreed on by all shells

It seems the main disagreement is \n, so that can go in  separate case

andychu (commented):
This looks good, thanks!

btertoolen:
@andychu Thanks! Fixed the issues you commented

andychu:
Merged, thank you!

Issue #2438, opened 2025-09-28
shorewall packages - Assoc array keys must be strings


andychu:
https://op.oils.pub/aports-build/2025-09-27-disagree.wwz/_tmp/aports-report/2025-09-27-disagree/diff_merged.html

https://op.oils.pub/aports-build/2025-09-27-disagree.wwz/_tmp/aports-report/2025-09-27-disagree/disagree-2025-09-18-bash/osh-as-sh/log/shorewall6.log.txt

```
>>> shorewall6: Unpacking /var/cache/distfiles/shorewall6-5.2.8.tar.bz2...
  vendor=${params[HOST]}
                  ^~~~
./configure:96: fatal: Assoc array keys must be strings: $x 'x' "$x" etc. (OILS-ERR-101)
```

PR #2499, opened 2025-10-22
Add patch handling for `regtest/aports/main`

last-genius:
A simple proof-of-concept. We haven't had to patch packages like this yet - but there's only one package (with its three subpackages) hitting #2438, and changing `osh` to accept this code is not the trade-off we are willing to make - see https://www.oilshell.org/blog/2016/10/20.html for more context. Instead, just add some quotes to assoc array keys.

### DIFF
--- a/regtest/patches/shorewall-core/shorewall-assoc-quote.patch
+++ b/regtest/patches/shorewall-core/shorewall-assoc-quote.patch
@@ -1,0 +1,148 @@
+From 39d124cf439a25bf28d58a39e9ff95d5059349be Mon Sep 17 00:00:00 2001
+From: Andrii Sultanov <sultanovandriy@gmail.com>
+Date: Wed, 22 Oct 2025 20:20:52 +0100
+Subject: [PATCH] configure: Quote assoc array keys as strings
+
+This prevents the following error in osh:
+
+    >>> shorewall6: Unpacking /var/cache/distfiles/shorewall6-5.2.8.tar.bz2...
+      vendor=${params[HOST]}
+                      ^~~~
+    ./configure:96: fatal: Assoc array keys must be strings: $x 'x' "$x" etc. (OILS-ERR-101)
+
+For more details, see the following:
+
+https://github.com/oils-for-unix/oils/issues/2438
+
+https://www.oilshell.org/blog/2016/10/20.html
+
+Signed-off-by: Andrii Sultanov <sultanovandriy@gmail.com>
+---
+ Shorewall-core/configure | 50 ++++++++++++++++++++--------------------
+ 1 file changed, 25 insertions(+), 25 deletions(-)
+
+diff --git a/Shorewall-core/configure b/Shorewall-core/configure
+index 741f74475..09b8f2bb7 100755
+--- a/Shorewall-core/configure
++++ b/Shorewall-core/configure
+@@ -93,7 +93,7 @@ done
+ 
+ cd $(dirname $0)
+ 
+-vendor=${params[HOST]}
++vendor=${params['HOST']}
+ 
+ if [ -z "$vendor" ]; then
+     if [ -f /etc/os-release ]; then
+@@ -117,48 +117,48 @@ if [ -z "$vendor" ]; then
+ 		;;
+ 	esac
+ 
+-	params[HOST]="$vendor"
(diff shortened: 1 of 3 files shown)


regtest/aports-setup.sh
```diff
@@ -450,6 +470,8 @@ fetch-all() {
   clone-aports
   clone-aci
   checkout-stable
+  patch-aports
```
andychu, line 477: Oh there is actually a function `patch-yash-to-disable-tests`, which is done AFTER we copy `../alpinelinux/aports` INTO the chroot

We should probably settle on one or the other

I guess I didn't want to mess up the local copy, so I patched the one in the chroot

regtest/aports-setup.sh
```diff
@@ -68,6 +68,26 @@ checkout-stable() {
...
 
+patch-aports() {
+  cur_dir=`pwd`
+  pushd ../../alpinelinux/aports
+
+  # Add our patches alongside Alpine's own patches
+  for patch_dir in $cur_dir/regtest/patches/*; do
+      patch_dir=$(basename $patch_dir)
+      for patch in $cur_dir/regtest/patches/$patch_dir/*; do
+          cp $patch main/$patch_dir/
```
andychu, line 81: Can you add some comments about how this works?  Does `abuild` apply the patches ?

regtest/aports-setup.sh
```diff
@@ -68,6 +68,26 @@ checkout-stable() {
...
+          patch_name=$(basename $patch)
+          shasum=$(sha512sum $patch | cut -d " " -f1)
+          sed -i "/source='*/ a         $patch_name" main/$patch_dir/APKBUILD
+          sed -i "/sha512sums='*/ a $shasum  $patch_name" main/$patch_dir/APKBUILD
```
andychu: Oh I see, is this modifying the line

```
sha512sums="
4dc63f2935e66d708dbecfdb77c6e5d697f4bc79af51ab860ed87499336d66d44f16abc7edf414c56f30d46de3a1a676c56c240557140c91aa388136e652cbdc  git-2.49.1.tar.xz
89528cdd14c51fd568aa61cf6c5eae08ea0844e59f9af9292da5fc6c268261f4166017d002d494400945e248df6b844e2f9f9cd2d9345d516983f5a110e4c42a  git-daemon.initd
fbf1f425206a76e2a8f82342537ed939ff7e623d644c086ca2ced5f69b36734695f9f80ebda1728f75a94d6cd2fcb71bf845b64239368caab418e4d368c141ec  git-daemon.confd
be5d568fc5b8b84c9afb97b31e471e41f32ccfe188eba0588ea0ef98b2d96c2ce4b2c1a3d70e88205aa4f6667f850b3f32c13bbb149ecddbf670344c162a4e25  fix-t4219-with-sticky-bit.patch
"
```

Note that in the `patch-yash` function, I took care to make the mutation idempotent

So if you run it twice, you get the same file back

I think that is important in practice.  Is that true here?

regtest/aports-setup.sh
last-genius: re-running `fetch-all` on changes to this branch works for me, but `prepare-all` is not idempotent: `ERROR: Directory oils-for-linux/oils/_chroot/aports-build should be empty` and I didn't want to remove the chroot every time.

One can always restore the local copy imo - it's a git repo, just `git restore .` works

regtest/aports-setup.sh
last-genius: Added a `git restore $apkbuild` here

regtest/aports-setup.sh
last-genius: Added a few more comments

regtest/aports-setup.sh
```diff
@@ -68,6 +68,30 @@ checkout-stable() {
   popd > /dev/null
 }
 
+patch-aports() {
+  cur_dir=`pwd`
```
andychu: please use `local cur_dir=$PWD` here

(backticks are gone in YSH :-) )

regtest/aports-setup.sh
andychu: OK, yeah I guess it's better to patch without root access, which `patch-yash` needs

andychu:
Thanks!

Issue #2475, opened 2025-10-14
noclobber is incorrectly applied to >> and &>>


PossiblyAShrub:
I found this while adding spec tests to https://github.com/oils-for-unix/oils/pull/2474.

`noclobber` (`set -C`) does not apply to `>>` and `&>>`. The [bash manual omits this](https://www.gnu.org/software/bash/manual/bash.html#Appending-Redirected-Output), but all shells in our spec suite agree.

Here's the test case:

```sh
set -C

rm -f $TMP/no-clobber
echo foo >> $TMP/no-clobber
echo stdout=$?
echo bar >> $TMP/no-clobber
echo again=$?
cat $TMP/no-clobber

rm -f $TMP/no-clobber
echo baz &>> $TMP/no-clobber
echo both=$?
echo foo &>> $TMP/no-clobber
echo again=$?
cat $TMP/no-clobber
```

The standard output should be:

```
stdout=0
again=0
foo
bar
both=0
again=0
baz
foo
```

OSH incorrectly raises an error because it currently applies the rule to `>>` and `&>>` redirects.

PR #2503, opened 2025-10-23
Fix >/dev/null and >>somefile when noclobber is set

PossiblyAShrub:
This is https://github.com/oils-for-unix/oils/issues/2191 and https://github.com/oils-for-unix/oils/issues/2475.

~~I ran into [a wart when trying to raise a noclobber error early](https://github.com/oils-for-unix/oils/pull/2503/files#diff-bc13b85d319057214d8ed9af7d5b50d39a6be28cf4a22dda71fbe54b51069dcfR417-R419). I've shared this on Zulip: [#oils-dev > We cannot raise IOError](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/We.20cannot.20raise.20IOError/with/546582250).~~

### DIFF
--- a/demo/noclobber-race.sh
+++ b/demo/noclobber-race.sh
@@ -1,6 +1,7 @@
 #!bin/osh
 #
 # Usage:
+#  demo/noclobber-race.sh setup-adversary-fast
 #  demo/noclobber-race.sh demo bin/osh
 #  demo/noclobber-race.sh demo _bin/cxx-opt/osh   # Will take longer, but will eventually get a race
 #                                                 # Always under ~150 times on my machine.


andychu:
This is great, thank you!   

I ran the race demo, and updated the help text at the top.  That does look like a bad bug, which is worth fixing

And yash fixed it too.  

I guess it depends on a better `stat()` API?  I can help with that, maybe we can talk about how to do it on Zulip

---

I'm always surprised that people use features like `set -C`, but it has come up before

Similarly with `set -a`, which we just implemented ... I guess everything under the sun is used by someone :-/

PR #2504, opened 2025-10-23
Update default PATH to match ash

### LINKED ISSUE #2502: lua-aports - failing many tests
[lua-aports is failing](https://op.oils.pub/aports-build/2025-10-22.wwz/_tmp/aports-report/2025-10-22/disagree-2025-10-15-main/osh-as-sh/log/lua-aports.log.txt) many tests, although it's not clear why? Perhaps this has 2 root causes.

## possible issues:

The following appears a lot:
```sh
Unable to deduce build architecture. Install apk-tools, or set CBUILD.
```

This also happens many times for different strings:
```sh
spec/abuild_spec.lua:37: Expected objects to be equal.
Passed in:
(string) ''
Expected:
(string) 'myvalue'

stack traceback:
	spec/abuild_spec.lua:37: in function <spec/abuild_spec.lua:36>

[  FAILED  ] spec/abuild_spec.lua:36: abuild get_conf should return the value of a configuration variable from the user config (9.90 ms)
```

also getting several unexpected nils that seem related to the following:
```sh
spec/db_spec.lua:86: Expected objects to not be the same.
Passed in:
(nil)
Did not expect:
type nil

stack traceback:
	spec/db_spec.lua:86: in function <spec/db_spec.lua:81>
```

This is a weird failure. Perhaps also based on the environment?
```sh
spec/pkg_spec.lua:121: Expected objects to be equal.
Passed in:
(string) '/tmp/lua_LljhEI/aports/myrepo/myapp/myapp-doc-1.0-r0.apk'
Expected:
(string) '/tmp/lua_LljhEI/pkgs/myrepo/aarch64/myapp-doc-1.0-r0.apk'

stack traceback:
	spec/pkg_spec.lua:121: in function <spec/pkg_spec.lua:120>
```

EarthenSky:
close #2502

### DIFF
--- a/core/sh_init.py
+++ b/core/sh_init.py
@@ -212,9 +212,13 @@ def InitVarsAfterEnv(mem, mutable_opts):
     # If PATH SHELLOPTS PWD are not in environ, then initialize them.
     s = mem.env_config.Get('PATH')
     if s is None:
-        # Setting PATH to these two dirs match what zsh and mksh do.  bash and
-        # dash add {,/usr/,/usr/local}/{bin,sbin}
-        mem.env_config.SetDefault('PATH', '/bin:/usr/bin')
+        # Setting PATH to these four dirs match busybox ash. zsh and mksh only
+        # do /bin:/usr/bin while bash and dash add {,/usr/,/usr/local}/{bin,sbin}
+        # The default PATH in busybox ash is defined here:
+        # busybox https://github.com/mirror/busybox/blob/371fe9f71d445d18be28c82a2a6d82115c8af19d/include/libbb.h#L2303
+        # The default PATH in bash is defined here:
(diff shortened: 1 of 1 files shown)


EarthenSky:
@andychu I'm not sure whether this is the right design choice to make (I noticed that the default path was targeting zsh and mksh instead of bash; there were probably reasons for that), but I am confident it'll at least fix the lua-aports issue.

andychu:
Thanks!  Given what you found in #2502,  this makes sense

Please add these links in the comment for reference

https://github.com/mirror/busybox/blob/371fe9f71d445d18be28c82a2a6d82115c8af19d/include/libbb.h#L2303

https://github.com/bminor/bash/blob/a8a1c2fac029404d3f42cd39f5a20f24b6e4fe4b/config-top.h#L61

andychu:
Thanks for tracking this down and fixing it !!

PR #2505, opened 2025-10-23
globignore impl

daveads:
solves this issue **Implement GLOBIGNORE #609**

### DIFF
--- a/doc/ref/chap-special-var.md
+++ b/doc/ref/chap-special-var.md
@@ -375,17 +375,16 @@ Examples:
     GLOBIGNORE='.:..'        # Match dotfiles, but filter . and ..
     echo .*                   # Shows .env, .gitignore, etc. (not . or ..)
 
-Pattern syntax supports standard glob metacharacters:
-
-- `*` - matches any string
-- `?` - matches any single character
-- `[...]` - matches character classes, including `[[:alnum:]]`, `[[:space:]]`, etc.
-
-Note: Colons inside bracket expressions like `[[:alnum:]]` are treated as part
-of the pattern, not as separators.
+The patterns use standard glob syntax (see [glob-pat][]). Note that colons
+inside bracket expressions like `[[:alnum:]]` are treated as part of the
+pattern, not as separators.
 
 When `GLOBIGNORE` is unset or set to an empty string, glob expansion returns to
 its default behavior.
+
+This is a bash-compatible feature.
+
+[glob-pat]: chap-mini-lang.html#glob-pat
 
 ## Shell Options
 
--- a/osh/glob_.py
+++ b/osh/glob_.py
@@ -449,6 +449,25 @@ def GlobToERE(pat):
 # - See 2 calls in osh/word_eval.py
 
 
+def _StringMatchesAnyPattern(s, patterns):
+    # type: (str, List[str]) -> bool
+    """Check if string matches any pattern in the list.
+
+    Returns True if s matches any pattern, or if s is . or ..
+    (which are always filtered when GLOBIGNORE is set).
+    """
+    basename = os_path.basename(s)
+    if basename in ('.', '..'):
+        return True
+
+    flags = 0
+    for pattern in patterns:
+        if libc.fnmatch(pattern, s, flags):
+            return True
+
+    return False
+
+
 class Globber(object):
 
     def __init__(self, exec_opts, mem):
@@ -508,24 +527,6 @@ class Globber(object):
 
         return patterns
 
-    def _MatchesGlobIgnore(self, filename, patterns):
-        # type: (str, List[str]) -> bool
-        """Check if filename matches any GLOBIGNORE pattern.
-
-        Filenames . and .. are always ignored when GLOBIGNORE is set.
-        """
-        basename = os_path.basename(filename)
-        if basename in ('.', '..'):
-            return True
-
-        flags = 0
-
-        for pattern in patterns:
-            if libc.fnmatch(pattern, filename, flags):
-                return True
-
-        return False
-
     def _Glob(self, arg, out):
         # type: (str, List[str]) -> int
         globignore_patterns = self._GetGlobIgnorePatterns()
@@ -560,7 +561,7 @@ class Globber(object):
             if globignore_patterns is not None:
                 filtered = []  # type: List[str]
                 for s in results:
-                    if not self._MatchesGlobIgnore(s, globignore_patterns):
+                    if not _StringMatchesAnyPattern(s, globignore_patterns):
                         filtered.append(s)
                 results = filtered
                 n = len(results)


daveads:
https://github.com/user-attachments/assets/ae12299e-30dd-4187-af9d-d3ec5a59c578

andychu:
OK nice, it passes the tests!

I will take a look

---

Please fix these type errors in the meantime:

https://op.oilshell.org/uuu/github-jobs/10465/dev-minimal.wwz/_tmp/soil/logs/oil-types.txt

```
osh/glob_.py:478: error: "value_t" has no attribute "s"
osh/glob_.py:559: error: Name 'tmp' already defined on line 554
osh/glob_.py:572: error: Name 'tmp' already defined on line 554
```

----

Sometimes mycpp does not generate good code, and leads to C++ compile errors, which I can help fix

But in this case, the problem isn't mycpp.  MyPy flags the errors

We have sort of a "filter"

1. must pass MyPy type checking
2. must pass mycpp translation
3. must pass C++ compilation

daveads:
> OK nice, it passes the tests!
> 
> I will take a look
> 
> Please fix these type errors in the meantime:
> 
> https://op.oilshell.org/uuu/github-jobs/10465/dev-minimal.wwz/_tmp/soil/logs/oil-types.txt
> 
> ```
> osh/glob_.py:478: error: "value_t" has no attribute "s"
> osh/glob_.py:559: error: Name 'tmp' already defined on line 554
> osh/glob_.py:572: error: Name 'tmp' already defined on line 554
> ```
> 
> Sometimes mycpp does not generate good code, and leads to C++ compile errors, which I can help fix
> 
> But in this case, the problem isn't mycpp. MyPy flags the errors
> 
> We have sort of a "filter"
> 
> 1. must pass MyPy type checking
> 2. must pass mycpp translation
> 3. must pass C++ compilation

Yea sure... I was taking a break 

Will send an update soon

andychu:
Yeah actually I meant  to say that I can help with `#2` and `#3`, because they are specific to Oils

But `#1` is not specific to Oils -- we just use MyPy

Thank you!

andychu:
On Zulip we are talking about how to make mycpp more friendly

Sometimes it gives bad error messages

daveads:
> Yeah actually I meant to say that I can help with `#2` and `#3`, because they are specific to Oils
> 
> But `#1` is not specific to Oils -- we just use MyPy
> 
> Thank you!

oh okayy

andychu:
This one still looks like it has type errors?

https://op.oilshell.org/uuu/github-jobs/10465/

I can help with the mycpp errors, after the type errors are fixed

daveads:
> This one still looks like it has type errors?
> 
> https://op.oilshell.org/uuu/github-jobs/10465/
> 
> I can help with the mycpp errors, after the type errors are fixed

yea sure i will tag you here when done.

daveads:
@andychu type errors fixed.

andychu:
Ah OK, there are some mycpp errors that should be easy to fix:

```
osh/glob_.py:480 Use explicit len(obj) or 'obj is not None' for mystr, mylist, mydict
osh/glob_.py:498 Use explicit len(obj) or 'obj is not None' for mystr, mylist, mydict
osh/glob_.py:505 Use explicit len(obj) or 'obj is not None' for mystr, mylist, mydict

mycpp/mycpp_main.py: 3 translation errors (after type checking)
```

https://op.oilshell.org/uuu/github-jobs/10556/cpp-coverage.wwz/_tmp/soil/logs/compare-gcc-clang.txt

Sometimes mycpp crashes, but these errors are good

It's basically because we need to be write explicit Python to translate to C++

osh/glob_.py
```diff
@@ -446,28 +449,90 @@ def GlobToERE(pat):
...
         #                  not glob().
-        #
-        # NOTE: Bash also respects the GLOBIGNORE variable, but no other shells
-        # do.  Could a default GLOBIGNORE to ignore flags on the file system be
-        # part of the security solution?  It doesn't seem totally sound.
+
+    def _GetGlobIgnorePatterns(self):
+        # type: () -> Optional[List[str]]
+        """Get GLOBIGNORE patterns as a list, or None if not set."""
+        from _devbuild.gen.value_asdl import value_e
```
andychu: these imports should be at the top

osh/glob_.py
```diff
@@ -446,28 +449,90 @@ def GlobToERE(pat):
...
+        from _devbuild.gen.runtime_asdl import scope_e
+        
+        cell = self.mem.GetCell('GLOBIGNORE', scope_e.GlobalOnly)
```
andychu: I think you should be able to use GetValue() instead of GetCell()

GetCell() is only if you need the flags or to mutate

osh/glob_.py
```diff
@@ -446,28 +449,90 @@ def GlobToERE(pat):
...
+        if cell is None:
+            return None
+        
+        val = cell.val
+        if val.tag() != value_e.Str:
+            return None
+        
+        globignore = cast(value.Str, val).s  # type: str
+        if not globignore:  # Empty string
```
andychu: the mycpp error is because this should be if len(glob_ignore) == 0:   

more like C++

osh/glob_.py
```diff
@@ -446,28 +449,90 @@ def GlobToERE(pat):
...
+            return None
+        
+        # Split by colon to get individual patterns, but don't split colons
```
andychu, line 503: Hm interesting that you have to parse this

Sometimes we cache the parsing

I think that might be a good idea here

Because GLOBIGNORE could be `'*.py:*.sh'`  for the entire program, and we don't want to parse this every time we do a glob

The cache could be a `Dict[str, List[str]]` or something similar

osh/glob_.py
```diff
@@ -446,28 +449,90 @@ def GlobToERE(pat):
...
+        if current:
+            patterns.append(''.join(current))
+        
+        return patterns
+
+    def _MatchesGlobIgnore(self, filename, patterns):
+        # type: (str, List[str]) -> bool
+        """Check if filename matches any GLOBIGNORE pattern.
+        
+        Filenames . and .. are always ignored when GLOBIGNORE is set.
```
andychu: OK hm that's an interesting behavior

It doesn't appear to be covered in spec/globignore?  I see a test with `GLOBIGNORE=.:..` but I don't think that covers it, and I don't see other instances of `..`

osh/glob_.py
```diff
@@ -446,28 +449,90 @@ def GlobToERE(pat):
...
+                in_bracket = True
+                current.append(c)
+            elif c == ']':
+                in_bracket = False
+                current.append(c)
+            elif c == ':' and not in_bracket:
+                # Colon outside brackets is a separator
+                if current:
+                    patterns.append(''.join(current))
+                    current = []
```
andychu: I think del current[:] produces less garbage, and I think mycpp supports it

We learned the hard way that idiomatic Python creates a lot of garbage!

andychu:
Hm weird, there is a failure only in C++

https://op.oilshell.org/uuu/github-jobs/10612/cpp-spec.wwz/_tmp/spec/osh-cpp/globignore.html

Usually that should not happen

I can look into it

daveads:
> Hm weird, there is a failure only in C++
> 
> https://op.oilshell.org/uuu/github-jobs/10612/cpp-spec.wwz/_tmp/spec/osh-cpp/globignore.html
> 
> Usually that should not happen
> 
> I can look into it

okayy

andychu:
If you can address the comments in Python first, then I will look into the C++ failure

Similar to the Segfault message PR -- as long as spec tests pass in Python, that is good

The C++ can be separate, and occasionally there are issues.

andychu:
Thanks!  Can you update doc/ref ?

```
$ grep -A 3 GLOBIGNORE  doc/ref/*
doc/ref/chap-special-var.md:### GLOBIGNORE
doc/ref/chap-special-var.md-
doc/ref/chap-special-var.md-TODO: bash compat
doc/ref/chap-special-var.md-
--
doc/ref/toc-osh.md:  [Shell Vars]    IFS             X LANG       X GLOBIGNORE
doc/ref/toc-osh.md-  [Shell Options] SHELLOPTS       X BASHOPTS
doc/ref/toc-osh.md-  [Other Env]     HOME              PATH
doc/ref/toc-osh.md-  [Other Special] BASH_REMATCH     @PIPESTATUS

```


Use `build/doc.sh all-ref` to preview the HTML, and to make sure it looks nice

As an example, you can see what I added in `doc/ref` for the `trap` PR:

https://github.com/oils-for-unix/oils/commit/9c6e2013ee9c3905a046ce78dce95f88fd01e70f

daveads:
> Thanks! Can you update doc/ref ?
> 
> ```
> $ grep -A 3 GLOBIGNORE  doc/ref/*
> doc/ref/chap-special-var.md:### GLOBIGNORE
> doc/ref/chap-special-var.md-
> doc/ref/chap-special-var.md-TODO: bash compat
> doc/ref/chap-special-var.md-
> --
> doc/ref/toc-osh.md:  [Shell Vars]    IFS             X LANG       X GLOBIGNORE
> doc/ref/toc-osh.md-  [Shell Options] SHELLOPTS       X BASHOPTS
> doc/ref/toc-osh.md-  [Other Env]     HOME              PATH
> doc/ref/toc-osh.md-  [Other Special] BASH_REMATCH     @PIPESTATUS
> ```
> 
> Use `build/doc.sh all-ref` to preview the HTML, and to make sure it looks nice
> 
> As an example, you can see what I added in `doc/ref` for the `trap` PR:
> 
> [9c6e201](https://github.com/oils-for-unix/oils/commit/9c6e2013ee9c3905a046ce78dce95f88fd01e70f)

@andychu  done.

spec/globignore.test.sh
```diff
@@ -162,3 +162,14 @@ echo *
 *
 *
 ## END
+
+#### . and .. always filtered when GLOBIGNORE is set
+# When GLOBIGNORE is set to any non-null value, . and .. are always filtered
```
andychu, line 165: that is a quirky behavior .  thanks for adding a test!

osh/glob_.py
```diff
@@ -446,28 +451,89 @@ def GlobToERE(pat):
...
+                current.append(c)
+
+        if len(current):
+            patterns.append(''.join(current))
+
+        self._globignore_cache[globignore] = patterns
+
+        return patterns
+
+    def _MatchesGlobIgnore(self, filename, patterns):
```
andychu: Let's make this a free function _StringMatchesAnyPattern(s, patterns) -- because it doesn't depend on `self` or a GLOBIGNORE value

osh/glob_.py
```diff
@@ -491,17 +557,28 @@ def _Glob(self, arg, out):
                 results = tmp  # idiom to work around mycpp limitation
                 n = len(results)
```
andychu, line 559: Hm OK I see we respect

- dotglob and GLOBIGNORE
- no_dash_glob
- then there's GLOBIGNORE
- then we don't fully support shopt -s globskipdots

But I see why you did that!  This code is complex, but that's just how bash is  ...

doc/ref/chap-special-var.md
```diff
@@ -350,7 +350,42 @@ TODO: bash compat
...
+    echo */*
+
+    GLOBIGNORE='.:..'        # Match dotfiles, but filter . and ..
+    echo .*                   # Shows .env, .gitignore, etc. (not . or ..)
+
+Pattern syntax supports standard glob metacharacters:
+
+- `*` - matches any string
+- `?` - matches any single character
+- `[...]` - matches character classes, including `[[:alnum:]]`, `[[:space:]]`, etc.
```
andychu: Thanks, let's add a link to the `glob-pat` topic

```
$ grep glob-pat doc/ref/*
doc/ref/chap-builtin-func.md:See [glob-pat][] for syntax.
doc/ref/chap-builtin-func.md:[glob-pat]: chap-mini-lang.html#glob-pat
```

Actually, now I see that is not quite complete ... and this one isn't quite complete, because it doesn't have say `echo *.[ch]`

Well I guess we need something like https://www.man7.org/linux/man-pages/man7/glob.7.html

Maybe do the best you can do consolidate it into one topic, and then I will make a follow-up pass

We shoiuldn't have 2 slightly different descriptions of glob syntax -- we should have one place that is complete

And it depends on libc, so actually it's hard to document!  GNU libc and musl libc could be slightly different

andychu (commented):
This looks good, thank you!

Just a small stylistic thing with the method -> function, and then less duplication in doc/ref

I will follow up about YSH glob work on Zulip

andychu:
OK also I am looking into this -- this is weird

```
$ test/spec-cpp.sh run-file globignore  -r 0 -v
        compare_shells      : bash
        legacy_tmp_dir      : true
        oils_failures_allowed: 1

globignore.test.sh
  0       5     pass    FAIL    Don't glob flags on file system with GLOBIGNORE
case: 0

[osh-cpp stdout]
Expected '-* hello zzzz?\n'
Got      ''


[osh-cpp status]
Expected 0
Got      99

```

I can reproduce the issue on my machine

andychu:
I fixed the C++ test

https://github.com/oils-for-unix/oils/pull/2505/commits/2e6c3a21432ed276660b4f2cdb4a32fa37972339

Not sure why that was there in the first place!

andychu:
Merged, thank you!

PR #2507, opened 2025-10-23
WIP: version of aidan/noclobber-fix with List[int]

PossiblyAShrub:
Just to run CI. Will merge into `aidan/noclobber-fix` (https://github.com/oils-for-unix/oils/pull/2503)

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -442,8 +442,7 @@ class FdState(object):
                         "Can't open %r: %s%s" %
                         (arg.filename, pyutil.strerror(e), extra),
                         blame_loc=r.op_loc)
-                    err_out.append(e.errno)
-                    return
+                    raise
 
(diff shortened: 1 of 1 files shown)


core/process.py
```diff
@@ -548,9 +545,11 @@ def Push(self, redirects, err_out):
             #log('apply %s', r)
             with ui.ctx_Location(self.errfmt, r.op_loc):
                 try:
-                    self._ApplyRedirect(r)
+                    self._ApplyRedirect(r, err_out)
                 except (IOError, OSError) as e:
-                    err_out.append(e)
+                    err_out.append(e.errno)
```
PossiblyAShrub, line 552: I still have to keep the try/except block as some calls inside `_ApplyRedirect` still raise `IOError` and `OSError`.

TODO: Figure out what these are and catch inside `_ApplyRedirect`?

Issue #2445, opened 2025-10-03
PATH cache is not cleared when mutating PATH


PossiblyAShrub:
I found this in https://github.com/oils-for-unix/oils/pull/2444 while creating test for https://github.com/oils-for-unix/oils/issue/2429:

```sh
mkdir -p _tmp/bin
mkdir -p _tmp/bin2
printf '#!/usr/bin/env sh\necho hi\n' >_tmp/bin/hello
printf '#!/usr/bin/env sh\necho hey\n' >_tmp/bin2/hello
chmod +x _tmp/bin/hello
chmod +x _tmp/bin2/hello

PATH="$PWD/_tmp/bin:$PATH" hello
PATH="$PWD/_tmp/bin2:$PATH" hello
```

The output of this under OSH is

```
hi
hi
```

Meanwhile, all the other shells output:

```
hi
hey
```

---

The issue is that `search_path.ClearCache` is only called when running `hash -r` or `shvar PATH=...` in YSH. It should also be called whenever `PATH` is overwritten.

A spec test for this bug has been added in https://github.com/oils-for-unix/oils/pull/2444.

andychu:
Oh thank you!  Great find

andychu:
This one might need callbacks, NOT sure

if so, then we discussed regarding the `ulimit` / file descriptor PR - https://github.com/oils-for-unix/oils/pull/2448

that callbacks are often done with classes in mycpp, similar to early Java

PR #2508, opened 2025-10-23
Clear PATH cache when mutating PATH

last-genius:
Fixes: https://github.com/oils-for-unix/oils/issues/2445

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -1993,6 +1993,11 @@ class Mem(object):
         assert flags & SetNameref == 0, flags
         assert flags & ClearNameref == 0, flags
 
+        # Hack: let SearchPath know it needs to reset the path
+        # cache before looking up a command
+        if lval.name == 'PATH':
+            self.clear_path_cache = True
+
         cell, var_frame = self._ResolveNameForYshMutation(
             lval.name, which_scopes, bool(flags & YshDecl))
 


andychu (commented):
This looks nice and simple!  Thank you

One issue is that YSH uses ENV.PATH

```
$ ysh -c 'echo $[ENV.PATH]; setvar ENV.PATH = "z"; ls'
/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games:/home/andy/bin:/wedge/oils-for-unix.org/pkg/python2/2.7.18/bin/
  echo $[ENV.PATH]; setvar ENV.PATH = "z"; ls
                                           ^~
[ -c flag ]:1: Command 'ls' not found (OILS-ERR-100)
[ -c flag ]:1: errexit PID 2636547: Command failed with status 127
```

Actually I think a first step would be to add a failing test ...  the hook may be a bit trickier (?)

----

I think `spec/path-cache` can be folded into `spec/command_`, where other cases for the PATH cache are

And then maybe add an analogous file `spec/ysh-command` for the YSH path cache test

last-genius:
Added a test for YSH (and made it pass), and in the process discovered there's another way of doing PATH mutation in osh, so added another test and fixed that as well.

osh/cmd_eval.py
```diff
@@ -1030,6 +1030,11 @@ def _DoMutation(self, node):
                             key = val_ops.ToStr(lval.index,
                                                 'Obj index should be Str',
                                                 loc.Missing)
+                            # Hack: let SearchPath know it needs to reset the path
+                            # cache before looking up a command
+                            if key == 'PATH':
```
andychu: Hm let me think about this  ... it seems a bit overzealous,  since it also clears the cache on `setvar other.PATH = 42`, I think

We could define YSH so you do `hash -r` aka `hash --clear` or something ... although maybe that's too unfriendly

andychu:
I think it would be OK to land a fix for OSH first, and then have the failing spec test for YSH as a TODO

So we're not blocked on a design issue

Although I guess this one doesn't necessarily make any regtest/aports pass -- although it might!

last-genius:
Dropped the `ysh` patch

core/state.py
```diff
@@ -1991,6 +1993,11 @@ def SetNamedYsh(self, lval, val, which_scopes, flags=0):
         assert flags & SetNameref == 0, flags
         assert flags & ClearNameref == 0, flags
 
+        # Hack: let SearchPath know it needs to reset the path
+        # cache before looking up a command
+        if lval.name == 'PATH':
```
andychu, line 1998: I think we should leave this out too -- this is in SetNamedYsh, so I think it only happens in YSH



In YSH, `setvar PATH =' foo'` is just a normal variable

`setvar ENV.PATH` would be the thing to clear the cache -- we have to think about that a bit

---

Although it's complicated because you can `shopt --set ysh:all` and so forth ... gah

core/state.py
```diff
@@ -2023,6 +2030,12 @@ def SetNamed(self, lval, val, which_scopes, flags=0):
         This can be used to FLIP flags, while NOT changing the variable.  Or it
         can also be used to initialize?
         """
+
+        # Hack: let SearchPath know it needs to reset the path
```
andychu, line 2034: I think the assumption here is that this is the ONLY way to change PATH

We also have an issue with `LANG=C` -- in bash, that doesn't just mutate a variable, it also makes a function call like `setlocale()` !!

andychu (commented):
Hm now I'm thinking about what we need for LANG=C ... hm

I guess for most of these, we tend to CACHE the parsing

https://github.com/oils-for-unix/oils/issues/527

e.g. GLOBIGNORE and PATH

We don't do anything special when you mutate them ... 

But LANG=C and PATH= are different!

core/state.py
last-genius: You can do `setglobal PATH` in `osh`

last-genius:
Rebased and fixed conflicts

core/state.py
andychu: OK good point, that is why I'm thinking about the invariants here

It is a fuzzy line between OSH and YSH

And I even wonder if we need something besides ENV.PATH in YSH ...

PR #2509, opened 2025-10-23
Describe what the syntax of the kill builtin is

andychu:


doc/ref/chap-builtin-cmd.md
```diff
@@ -1303,9 +1303,36 @@ If no JOB is specified, use the latest job.
 
 ### kill
 
-UNIMPLEMENTED
+The `kill` builtin sends a signal to one or more processes.  Usage:
+
+    kill (-s SIG | -SIG) WHAT+  # send SIG to the given processes
```
btertoolen: this is missing the -n option

doc/ref/chap-builtin-cmd.md
```diff
@@ -1303,9 +1303,36 @@ If no JOB is specified, use the latest job.
...
+
+where
+
+    SIG  = NAME | NUMBER   # e.g. USR1 or 10
+    WHAT = PID  | JOBSPEC  # e.g. 123 or %%
+
+Examples:
+
+    kill -s USR1 123 %%    # send SIGUSR1 to PID 123 and the current job
```
btertoolen: This is a nitpick, but I would maybe let most examples show only one pid or jobspec, and then put a separate example to show you can put multiple pids or jobspecs

doc/ref/chap-builtin-cmd.md
andychu: Yeah I decided to "keep it simple" for the user, and recommend `-s` first.   Showing `-n` as an example

Similarly I made `kill -L` the recommended one, and showed `-l` as an example

Issue #2501, opened 2025-10-22
[osh as bash] mc, graphicsmagick - 'set' got invalid flag '-a'


EarthenSky:
[mc is failing](https://op.oils.pub/aports-build/2025-10-22.wwz/_tmp/aports-report/2025-10-22/disagree-2025-10-15-main/osh-as-sh/log/mc.log.txt) one test as follows:

```sh
FAIL: run
============================================================================
Testsuite summary for "/src/vfs/extfs/helpers-list"
============================================================================
# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See tests/src/vfs/extfs/helpers-list/test-suite.log
Please report to https://www.midnight-commander.org/wiki/NewTicket
============================================================================
```

with the error in `tests/src/vfs/extfs/helpers-list/test-suite.log` being the following:

```sh
Testing /home/udu/aports/main/mc/src/mc-4.8.33/tests/src/vfs/extfs/helpers-list/data/hp48+.input
PASSED.
Testing /home/udu/aports/main/mc/src/mc-4.8.33/tests/src/vfs/extfs/helpers-list/data/iso9660.xorriso.input
        set -a  # "allexport: Export all variables assigned to."
            ^~
/home/udu/aports/main/mc/src/mc-4.8.33/tests/src/vfs/extfs/helpers-list/test_all:308: 'set' got invalid flag '-a'
        set +a
            ^
/home/udu/aports/main/mc/src/mc-4.8.33/tests/src/vfs/extfs/helpers-list/test_all:310: 'set' got invalid flag '-a'
```

Root cause:

```sh
bash$ set -a
bash$ bin/osh
osh-0.36$ set -a
  set -a
      ^~
[ interactive ]:1: 'set' got invalid flag '-a'
```

Same error as:
- graphicsmagick [#distros > 2025-10-08-comm - running with aports/community @ 💬](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/2025-10-08-comm.20-.20running.20with.20aports.2Fcommunity/near/544568293)

daveads:
@EarthenSky are you attending to this ?

EarthenSky:
> @EarthenSky are you attending to this ? 

I'm not, it's open to assign yourself and work on this if you'd like.

daveads:
[...]

PR #2511, opened 2025-10-23
'set' got invalid flag '-a'

daveads:
solves this issue #2501

Screenshot 2025-10-24 at 12 12 17 AM

### DIFF
--- a/spec/builtin-set.test.sh
+++ b/spec/builtin-set.test.sh
@@ -323,21 +323,19 @@ after_value
 set -a
 f() {
   local ZZZ=zzz
-  printenv ZZZ
+  printenv.py ZZZ
 }
 f
 ## STDOUT:
 zzz
 ## END
-## BUG mksh status: 1
-## BUG mksh stdout-json: ""
+## BUG mksh stdout: None
 
 #### set -a exports declare variables
 set -a
 declare ZZZ=zzz
-printenv ZZZ
+printenv.py ZZZ
 ## STDOUT:
 zzz
 ## END
-## N-I dash/mksh status: 1
-## N-I dash/mksh stdout-json: ""
+## N-I dash/mksh stdout: None


daveads:
https://github.com/user-attachments/assets/84e158f9-c0e0-43ae-ad97-e139e807b750

spec/builtin-set.test.sh
```diff
@@ -276,3 +276,45 @@ echo "$@"
...
+
+#### variables set before set -a are not exported
+BEFORE=before_value
+set -a
+AFTER=after_value
+printenv.py BEFORE AFTER
+## STDOUT:
+None
+after_value
+## END
```
andychu, line 320: Hm how about a test for local variables?

```
$ bash -c 'set -a; f() { local ZZZ=zzz; env|grep ZZZ; }; f'
ZZZ=zzz
BASH_FUNC_f%%=() {  local ZZZ=zzz;
 env | grep ZZZ
```

and `declare`

```
$ bash -c 'set -a; declare ZZZ=zzz; env|grep ZZZ;'
ZZZ=zzz
```

That is handled outside of `ShAssignment`, so I think it may not work

`ShAssignment` is for `a=b`

```
$ bin/osh -n -c 'a=b'
(command.ShAssignment left:<Lit_VarLike "a="> pairs:[(AssignPair left:...0x7fb3f44d8670 lhs:(sh_lhs.Name left:...0x7fb3f44d8670 name:a) op:assign_op.Equal rhs:(w <Lit_Chars b>))])
```

but then assignment builtins are command.Simple

```
$ bin/osh -n -c 'a=b'
(command.ShAssignment left:<Lit_VarLike "a="> pairs:[(AssignPair left:...0x7fb3f44d8670 lhs:(sh_lhs.Name left:...0x7fb3f44d8670 name:a) op:assign_op.Equal rhs:(w <Lit_Chars b>))])

```

daveads: @andychu

spec/builtin-set.test.sh
```diff
@@ -276,3 +276,68 @@ echo "$@"
...
+
+#### set -a exports local variables
+set -a
+f() {
+  local ZZZ=zzz
+  printenv ZZZ
```
andychu: Hm I was confused by how this works ... I didn't know that `/usr/bin/printenv` exists!

It's part of `coreutils` apparently

It is basically the same as our tool, except in the case of unexported vars


```
$ printenv ZZ YY bad
zz
yy
$ echo $?
1

$ spec/bin/printenv.py ZZ YY bad
zz
yy
None
```

I think we should use one or the other consistently

So far we use printenv.py everywhere

Issue #2506, opened 2025-10-23
reject bad code in mycpp: a,b = b, a


andychu:
[ASAN / mycpp problems on fc builtin](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/ASAN.20.2F.20mycpp.20problems.20on.20fc.20builtin/with/526311011)

```
a, b = b, a
```

```
builtin/readline_osh.py:436: KeyError: <mypy.nodes.TupleExpr object at 0x7f18335881d0>
```

andychu:
similar to #2480 

we should have a test in `mycpp/examples/invalid_*` for this

Issue #2480, opened 2025-10-15
mycpp gives confusing errors when initializing int variable as None - PartialType


andychu:
@btertoolen hit this in a recent PR - https://github.com/oils-for-unix/oils/pull/2478

the issue was

```
signal_to_send = None
signal_to_send = FuncReturningInt()
```

Root cause: mycpp is like "writing C++  in Python syntax" -- so an `int` var cannot be `None`

andychu:
This belongs on the 4th grant, bceause one of our goals is to make mycpp give **good** error messages

andychu:
Related - 
[Idea: Feed mycpp arbitrary Python, and ensure proper errors](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Idea.3A.20Feed.20mycpp.20arbitrary.20Python.2C.20and.20ensure.20proper.20errors/with/529584924)

PR #2513, opened 2025-10-24
Improve two mycpp errors

last-genius:
Addresses #2480 and #2506.

See [#oils-dev > Better mycpp errors](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Better.20mycpp.20errors/with/546942458) for more discussion on this (and the justification for such a large diff).

Intended as a starting point to explore improvements to mycpp - so these errors don't necessarily have the same style/structure yet. It might be worth pointing to the corresponding `examples/` in the error explcitly? Attach a unique code to an error like osh errors do?

### DIFF
--- a/spec/sh-options.test.sh
+++ b/spec/sh-options.test.sh
@@ -334,6 +334,8 @@ foo
 ## END
 
 #### noclobber on &> >
+case $SH in dash) exit ;; esac
+
 set -C
 
 rm -f $TMP/no-clobber
@@ -359,13 +361,6 @@ again=1
 baz
 ## END
 ## BUG dash STDOUT:
-stdout=0
-again=2
-foo
-both=0
-baz
-again=2
-foo
 ## END
 
 #### noclobber on &>> >>


andychu:
Thanks for looking at this!

I looked, and the `report_errors` method is inherited from `mycpp/visitor.py`

I wonder if we should just put `self.GetCType()` on the visitor too

---

I think it makes the typing a bit easier

BTW I think we should make mycpp into a typed program.  I moved toward that in December.  It used to have hundreds of type errors, now it has 26

I think this change lacks some type annotations, which will cause more errors.  We should probably have something in the CI to check that it doesn't regress



```
$ devtools/types.sh check-mycpp 
mycpp/visitor.py:421: error: "Expression" has no attribute "generator"
mycpp/pass_state.py:573: error: Function is missing a type annotation for one or more arguments
mycpp/cppgen_pass.py:391: error: "Type" has no attribute "args"
mycpp/cppgen_pass.py:392: error: "Type" has no attribute "args"
mycpp/cppgen_pass.py:505: error: Item "ProperType" of "Optional[ProperType]" has no attribute "arg_types"
mycpp/cppgen_pass.py:582: error: Item "ProperType" of "Optional[ProperType]" has no attribute "ret_type"
mycpp/cppgen_pass.py:821: error: Item "ProperType" of "Optional[ProperType]" has no attribute "ret_type"
mycpp/cppgen_pass.py:909: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:952: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:978: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:1100: error: "Expression" has no attribute "items"
mycpp/cppgen_pass.py:1710: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:1792: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:2365: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:2478: error: Function is missing a type annotation
mycpp/cppgen_pass.py:2506: error: Call to untyped function "_WritePopRoots" in typed context
mycpp/cppgen_pass.py:2516: error: Call to untyped function "_WritePopRoots" in typed context
mycpp/cppgen_pass.py:2707: error: "ProperType" has no attribute "ret_type"
mycpp/conversion_pass.py:139: error: "Type" has no attribute "ret_type"
mycpp/conversion_pass.py:241: error: Item "ProperType" of "Optional[ProperType]" has no attribute "arg_types"
mycpp/conversion_pass.py:255: error: Item "ProperType" of "Optional[ProperType]" has no attribute "ret_type"
mycpp/conversion_pass.py:261: error: Incompatible types in assignment (expression has type "ClassDef", variable has type "None")
mycpp/control_flow_pass.py:127: error: "Type" has no attribute "type"
mycpp/control_flow_pass.py:146: error: "Type" has no attribute "type"
mycpp/control_flow_pass.py:151: error: "Expression" has no attribute "fullname"
mycpp/control_flow_pass.py:312: error: "Expression" has no attribute "name"
Found 26 errors in 5 files (checked 10 source files)

```

mycpp/README.md
```diff
@@ -94,24 +94,29 @@ wiki.
...
+Passes that analyze:
 
-Turn turn the constant in `myfunc("foo")` into top-level `GLOBAL_STR(str1,
-"foo")`.
-  
-(2) Three passes in `cppgen_pass.py`.
+(1) `conversion_pass.py`
+    - compute virtual functions, locals, class members, yield, etc.
+    - this also computes forward_decls, and we write it in translate.py
+    - we determine which methods are virtual here
```
andychu: This comment is now redundant with "compute virtual functions" above

It could be "compute which methods are virtual"


(Thanks for updating this based on the Zulip thread!)

mycpp/README.md
```diff
@@ -120,20 +125,10 @@ declarations.  The `virtual` keyword is written in the next pass.
       void method();
     };
 
-More work in this pass:
-
-- Collect member variables and write them at the end of the definition
-- Collect locals for "hoisting".  Written in the next pass.
-
-(c) Definition Pass.
+    - collect member variables and write them at the end of the definition
```
andychu: This is now obsolete, because we "compute class members" in the first conversion_pass

That is, we look at all the dynamic `self.foo = ` statements approximately, and compute a static set of memebers

mycpp/README.md
```diff
@@ -120,20 +125,10 @@ declarations.  The `virtual` keyword is written in the next pass.
...
+    - collect locals for "hoisting".  Written in the next pass.
```
andychu: ditto, can remove this

mycpp/cppgen_pass.py
```diff
@@ -196,10 +196,21 @@ def GetCType(t: Type) -> str:
         c_type = 'void'
 
     elif isinstance(t, PartialType):
-        # I removed the last instance of this!  It was dead code in comp_ui.py.
-        raise AssertionError()
-        #c_type = 'void'
-        #is_pointer = True
+        # Only raise an error here when we are visiting an AssignmentStmt node,
```
andychu: this technically doesn't "raise" an error -- it appends an error to a list and keeps going

mycpp/cppgen_pass.py
```diff
@@ -196,10 +196,21 @@ def GetCType(t: Type) -> str:
...
+                node,
+                "Mismatched types: trying to assign expression of type '%s' to "
+                "a PartialType variable '%s' (was likely assigned None before).\n"
+                "Tip: If your type translates to a heap-allocated type (e.g. str "
+                "or class), you can annotate it with '# type: Optional[T]'. "
+                "If your type is allocated on the stack (int), then you can use "
+                "-1 or similar as an in-band null value" %
+                (t.var.type.type.name, t.var.name))
+
+        c_type = t.var.type.type.name
```
andychu: Hm so we append an error to a list

But then what's this `c_type`?   Does it generate invalid C++?  

It used to crash with AssertionError.

I suggested this style

```
c_type = self.GetCType(t)  # may append to errors
if c_type is None:  # bail early since we don't have a c_type
  return
```

Though I'm not sure exactly of all the consequences of that

mycpp/conversion_pass.py
```diff
@@ -356,20 +358,27 @@ def oils_visit_assignment_stmt(self, o: 'mypy.nodes.AssignmentStmt',
...
-                if isinstance(lval_item, NameExpr):
-                    if util.SkipAssignment(lval_item.name):
-                        continue
-                    self.current_local_vars.append((lval_item.name, item_type))
-
-                # self.a, self.b = foo()
-                if isinstance(lval_item, MemberExpr):
-                    self._MaybeAddMember(lval_item)
+            if isinstance(rval, TupleExpr):
+                self.report_error(
```
andychu, line 373: I think we usually `return` after `self.report_error`

Otherwise this will call `oils_visit_assignment_stmt`

I think it's better not to call it?    It may produce cascading nonsense errors

andychu:
Another thought: `conversion_pass` is now supposed to be where all the validation happens.  

We are converting from "arbitrary typed Python" to our **subset** of mycpp Typed Python -- ideally all in one place / in one pass

---

Unfortunately, mycpp grew to have the "shotgun parser" anti-pattern, where you pass around **potentially invalid** data like PartialType -- it is propagated between the **five** passes

And you just check it **right before** you write C++, or you don't check it at all.  This is bad for correctness / safety, and leads an explosion of if statements

(We explicitly avoided shotgun parsing in Oils; unfortunately the same was not true of mycpp.  Lesson learned twice!)

---

So GetCType() and disallowing `PartialType` and `Optional[int]` actually belongs in the conversion_pass!   According to the new architecture

So yeah I think it makes sense to put it on the visitor -- we can do `self.GetCType()`


---

So this change could **either** be:

- a minimal change to improve those 2 errors (they can still bail early)
- or a bigger change that moves us away from shotgun parsing -- i.e. `PartialType` should never be propagated between the 5 passes
  -  it should stop at the first pass
  -  this relates to giving mycpp our own IR -- the thing I started in `mycpp/mycpp.asdl`, but is NOT used

andychu:
BTW in `mycpp/translate.py`, here is the part that bails early on conversion pass:

```
         p_convert = ...

         p_convert.visit_mypy_file(module)
         MaybeExitWithErrors(p_convert)
```

So yeah because we don't accept `Optional[int]`, we should definitely do it in that stage (eventually)

---

So the problem becomes clear: `GetCType()` is in `cppgen_pass.py`  !!   But that's the stage that outputs source code -- it's already too late 

We should be able to look at types in `conversion_pass.py`, and probably the easiest way is to put it on the visitor

(I don't like all the inheritance, but we got that structure from MyPy)

andychu:
Also, right now "conversion_pass" doesn't actually convert to a new IR --- It simply bails if the MyPy AST isn't valid

The 4 subsequent passes can now assume a few things about the AST, so it is better in a sense


I think it would be even better to have our own IR, but that's a long term thing

---

And yeah this relates to the whole mycpp program being typed - it does help the structure / maintenance if we can type check

mycpp/cppgen_pass.py
last-genius: > But then what's this c_type? Does it generate invalid C++?

I've sort of assumed `report_error` is equivalent to a "recoverable" AssertionError - the command exits with `1`, reporting multiple (translation) errors - the C++ code generated thus far is assumed unusable/incorrect, right? Anyway, looking at `_gen/bin/oils_for_unix.mycpp.cc` - nothing but namespace+class declarations (240 lines or so) get generated.

`c_type = None` (or `c_type = 'void'`) do not work in this case, way too many asserts and assumptions by callers. the above C-type `if null` check would need to be added to quite a lot of places.

last-genius:
I've addressed the most immediate comments, but I think moving this to an earlier pass is a much better solution - I'll refactor this tomorrow

last-genius:
Moved everything to `conversion_pass.py` - it's much, much cleaner now :)

mycpp/conversion_pass.py
```diff
@@ -356,20 +368,28 @@ def oils_visit_assignment_stmt(self, o: 'mypy.nodes.AssignmentStmt',
...
+                    o, "mycpp currently does not handle code like "
+                    "'a, b = x, y'. You can move each assignment to its own line "
+                    "or return (x, y) from a function instead.")
+                return
+            else:
```
andychu: Now that we have return above, I think we don't need the else

Personally I tend to use the "early return" style a lot, since it reduces indentation

mycpp/conversion_pass.py
```diff
@@ -318,6 +318,18 @@ def oils_visit_assignment_stmt(self, o: 'mypy.nodes.AssignmentStmt',
...
+            t = self.types[lval]
+            self.report_error(
+                o,
+                "Mismatched types: trying to assign expression of type '%s' to "
+                "a PartialType variable '%s' (was likely assigned None before).\n"
+                "Tip: If your type translates to a heap-allocated type (e.g. str "
+                "or class), you can annotate it with '# type: Optional[T]'. "
+                "If your type is allocated on the stack (int), then you can use "
+                "-1 or similar as an in-band null value" %
+                (t.var.type.type.name, t.var.name))
```
andychu, line 331: I think we might as well `return` early here too?

We want to prevent the subsequent code from potentially using `self.types[lval]`

For other passes, that's handled by `report_error`

But in theory, this pass could grow some stuff that tries to use it, and crash ?

I think the default should be to `return` after `report_error`, unless it is clear we want to report multiple errors at once for the SAME node / in the same place

---

Even if we return early, we will still report one error per AST node -- that means we report **many** errors at once, which is user-friendly

andychu (commented):
Oh yes, looks much cleaner!

mycpp/conversion_pass.py
last-genius: Oh right, of course! Fixed now

last-genius:
Hmm, `dash` seems to have failed in a spec test:
```
sh-options: note: Got 2 allowed Oils failures
sh-options: FATAL: Got 3 total failures, but 2 are allowed
```

https://op.oilshell.org/uuu/github-jobs/10518/ovm-tarball.wwz/_tmp/spec/osh-py/sh-options.html#details-22-dash

andychu:
Oh I just saw that `noclobber` test failure in `dash` spec tests a couple days ago

The funny thing is that @PossiblyAShrub is working on that in OSH

I think dash just has a race condition -- even though I thought we read C code that said they protected against a race?

Anyway, we should disable that spec test with `case $SH in dash) exit ;; esac`, since it is known to be flaky

andychu:
Oh actually that was the failing test cases added last week

https://github.com/oils-for-unix/oils/commit/4665a12cd7

Well the spec tests are finding bugs in other shells, that's good :-)

andychu:
Thanks!

I checked manually that there is no big regression in type errors for mycpp itself

Over time we should enforce that it's typed.  Which is complicated by having an old copy of MyPy I think


```
~/git/oils-for-unix/oils (asv/mycpp-type-error)$ devtools/types.sh check-mycpp 
mycpp/visitor.py:421: error: "Expression" has no attribute "generator"
mycpp/pass_state.py:573: error: Function is missing a type annotation for one or more arguments
mycpp/cppgen_pass.py:389: error: "Type" has no attribute "args"
mycpp/cppgen_pass.py:390: error: "Type" has no attribute "args"
mycpp/cppgen_pass.py:503: error: Item "ProperType" of "Optional[ProperType]" has no attribute "arg_types"
mycpp/cppgen_pass.py:580: error: Item "ProperType" of "Optional[ProperType]" has no attribute "ret_type"
mycpp/cppgen_pass.py:819: error: Item "ProperType" of "Optional[ProperType]" has no attribute "ret_type"
mycpp/cppgen_pass.py:907: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:950: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:976: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:1098: error: "Expression" has no attribute "items"
mycpp/cppgen_pass.py:1708: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:1790: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:2363: error: "Expression" has no attribute "name"
mycpp/cppgen_pass.py:2476: error: Function is missing a type annotation
mycpp/cppgen_pass.py:2504: error: Call to untyped function "_WritePopRoots" in typed context
mycpp/cppgen_pass.py:2514: error: Call to untyped function "_WritePopRoots" in typed context
mycpp/cppgen_pass.py:2705: error: "ProperType" has no attribute "ret_type"
mycpp/conversion_pass.py:139: error: "Type" has no attribute "ret_type"
mycpp/conversion_pass.py:241: error: Item "ProperType" of "Optional[ProperType]" has no attribute "arg_types"
mycpp/conversion_pass.py:255: error: Item "ProperType" of "Optional[ProperType]" has no attribute "ret_type"
mycpp/conversion_pass.py:261: error: Incompatible types in assignment (expression has type "ClassDef", variable has type "None")
mycpp/conversion_pass.py:331: error: "Type" has no attribute "var"
mycpp/control_flow_pass.py:127: error: "Type" has no attribute "type"
mycpp/control_flow_pass.py:146: error: "Type" has no attribute "type"
mycpp/control_flow_pass.py:151: error: "Expression" has no attribute "fullname"
mycpp/control_flow_pass.py:312: error: "Expression" has no attribute "name"
Found 27 errors in 5 files (checked 10 source files)

```

PR #2516, opened 2025-10-25
Segmentation fault message

### LINKED ISSUE #722: Print message on segfault
bash:

```
$ build/mycpp.sh osh-eval-demo 

...

types/oil-slice.sh: line 15: 20263 Done                    echo 'echo hi'
     20264 Segmentation fault      (core dumped) | $osh_eval
```

oil:

```
[osh] lisa ~/git/oilshell/oil$ build/mycpp.sh osh-eval-demo

arg 'echo'
arg 'hi'
    echo 'echo hi' | $osh_eval
                     ^~~~~~~~~
types/oil-slice.sh:24: fatal: Exiting with status 139 (command in PID 20150)
    echo 'echo hi' | $osh_eval
                   ^
types/oil-slice.sh:24: fatal: Exiting with status 139 (pipeline invoked from PID 20150)
    types/oil-slice.sh demo _bin/osh_eval.dbg
    ^~~~~~~~~~~~~~~~~~
build/mycpp.sh:614: fatal: Exiting with status 139 (command in PID 20149)

```

daveads:
Attends to this issue https://github.com/oils-for-unix/oils/issues/722

close #722

### DIFF
--- a/spec/background.test.sh
+++ b/spec/background.test.sh
@@ -1,4 +1,5 @@
 ## oils_failures_allowed: 3
+## oils_cpp_failures_allowed: 4
 ## compare_shells: dash bash mksh
 
 # Job control constructs:


core/process.py
```diff
@@ -2019,6 +2025,18 @@ def NumRunning(self):
 NO_ARG = -20
 
 
+_SIGNAL_MESSAGES = {
+    SIGSEGV: 'Segmentation fault',
```
andychu: Hm do other shells hard-code these strings?

Or is there something like `strerror()` for `errno` with these strings?

Sometimes we have to look at the C source of other shells, as for the default path

https://github.com/oils-for-unix/oils/commit/f1483837667dc8d28e90c84ee68f02381b807e63

andychu (commented):
This looks like a good start, thanks

It is possible to add a spec test for it?  We could kill the child process like in `spec/builtin-kill`

Does this message happen in both batch mode and interactive mode?

andychu:
Some testing about what other shells do

```
$ bash -c 'sleep 1 & kill -HUP %%; wait; echo status=$?'
bash: line 1: 2674279 Hangup                  sleep 1
status=0

```

hm only bash does it?  interesting

```
$ zsh -c 'sleep 1 & kill -HUP %%; wait; echo status=$?'
status=0

andy@hoover:~/git/oils-for-unix/oils$ mksh -c 'sleep 1 & kill -HUP %%; wait; echo status=$?'
status=0

```

And it's not just interactive

andychu:
This is interactive bash ...

```
andy@hoover:~/git/oils-for-unix/oils$ sleep 1 & kill -USR1 %%; wait
[1] 2674297
[1]+  User defined signal 1   sleep 1

```

andychu:
Interactive zsh

```
~/git/oils-for-unix/oils% sleep 1 & kill -USR1 %%; wait
[1] 2674303
[1]  + user-defined signal 1  sleep 1

```

So zsh only does it interactively ?  But bash does it in both?

hm same with `mksh`, seems interactive only

```
$ sleep 1 & kill -USR1 %%; wait
[1] 2674315
[1] + User defined signal 1 \sleep 1

```

andychu:
I guess we can do whatever is simpler - it seems simpler to do the same thing in interactive and batch

core/process.py
daveads: @andychu  

https://github.com/bminor/bash/blob/a8a1c2fac029404d3f42cd39f5a20f24b6e4fe4b/siglist.c#L95

andychu: Ah thanks for looking it up

Although I asked Claude and it shows that Python 3 has strsignal(), which is a POSIX 2008 API

Bash predates this API -- so it has this comment:

> /* siglist.c -- signal list for those machines that don't have one. */


But unforunately our Python 2 build doesn't have a binding for it

https://docs.python.org/3/library/signal.html#signal.strsignal

https://claude.ai/share/5c6e5003-67c6-46fb-989b-0db63fc8d4ce


----

Could you try adding a binding in `libc.c` and `libc.pyi` , and you can test it with `libc_test.py` ?

```
build/py.sh pylibc
```

builds it and run tests

I think the binding could be similar to the one for `strerror()` in `pyext/posixmodule.c`

i.e. it just does a bit of arg parsing and so forth


```
 PyDoc_STRVAR_remove(posix_strerror__doc__,
 "strerror(code) -> string\n\n\
 Translate an error code to a message string.");
                                         
 static PyObject *
 posix_strerror(PyObject *self, PyObject *args) 
 {               
     int code;
     char *message; 
     if (!PyArg_ParseTuple(args, "i:strerror", &code))
         return NULL;
     message = strerror(code);
     if (message == NULL) {
         PyErr_SetString(PyExc_ValueError,
                         "strerror() argument out of range");
         return NULL;
     }   
     return PyString_FromString(message);
 }
     

```

andychu:
Also here is the binding from Python 3 in `signalmodule.c`


```
static PyObject *
 signal_strsignal_impl(PyObject *module, int signalnum)
 /*[clinic end generated code: output=44e12e1e3b666261 input=b77914b03f856c74]*/
 {
     char *res;
 
     if (signalnum < 1 || signalnum >= NSIG) {
         PyErr_SetString(PyExc_ValueError,
                 "signal number out of range");
         return NULL;
     }

...

     errno = 0;
     res = strsignal(signalnum);
 
     if (errno || res == NULL || strstr(res, "Unknown signal") != NULL)
         Py_RETURN_NONE;
 #endif
 
     return Py_BuildValue("s", res);
 }
```
 


I usually try to adapt a similar example, so between those 2 examples, we should be able to make a Python 2 binding for `strsignal()`

core/process.py
daveads: i can try that out

daveads:
@andychu  check now...

pyext/libc_test.py
```diff
@@ -367,6 +367,15 @@ def testSleepUntilError(self):
 
     # Not testing errno case
 
+  def testStrsignal(self):
+    self.assertEqual('Segmentation fault', libc.strsignal(11))
```
andychu, line 372: Could do `signal.SIGTERM` to be more readable

core/process.py
```diff
@@ -2019,6 +2022,31 @@ def NumRunning(self):
...
+    SIGABRT: 'Aborted',
+    SIGILL: 'Illegal instruction',
+    SIGFPE: 'Floating point exception',
+    SIGBUS: 'Bus error',
```
andychu: We should remove this dict now, and use libc.strsignal()!

daveads: > We should remove this dict now, and use libc.strsignal()!

Done

andychu:
This looks good, thanks!  How about adding a spec test for it?

That will make it easier to port to C++, so we can see that it does the same thing

I think something like this should work

```
$ bash -c 'sleep 1 & kill -HUP %%; wait; echo status=$?'
bash: line 1: 129997 Hangup                  sleep 1
status=0

```

Maybe in `spec/exit-status` or something

daveads:
> This looks good, thanks! How about adding a spec test for it?
> 
> That will make it easier to port to C++, so we can see that it does the same thing
> 
> I think something like this should work
> 
> ```
> $ bash -c 'sleep 1 & kill -HUP %%; wait; echo status=$?'
> bash: line 1: 129997 Hangup                  sleep 1
> status=0
> ```
> 
> Maybe in `spec/exit-status` or something

Okay

andychu:
The test  is tickling the right behavior:

https://op.oilshell.org/uuu/github-jobs/10636/cpp-spec.wwz/_tmp/spec/osh-cpp/background.html#details-26-osh

But it needs a better assertion

It is failing with OSH in Python, and passing with OSH in C++

It should be the other way around

```
[osh stderr] Expected '' Got 'Hangup (core dumped)\n'
```

Sometimes it may be useful to redirect `stderr` to a file, and then `grep` it for a string

Using `grep -o` perhaps

It would be nice if the test passes with bash too

andychu:
Hm yeah other shells don't print `status=129` ?  I wonder why that is

https://op.oilshell.org/uuu/github-jobs/10636/ovm-tarball.wwz/_tmp/spec/osh-py/background.html

andychu:
Oh and if you haven't already seen it, the commands to reproduce the spec test failures are here:

```
test/spec.sh background FLAG*
test/spec-cpp.sh run-file background FLAG*
```

andychu:
Hm the spec tests should still be failing locally for you ?

e.g. `test/spec.sh background` should be exiting **non-zero** now, which is a sign it is not ready

And I think Github sends you a message when the CI fails?  

---

Right now it is failing --  bash and OSH  now pass , but the other 2 shells need to be marked `N-I` 

https://op.oilshell.org/uuu/github-jobs/10649/ovm-tarball.wwz/_tmp/spec/osh-py/background.html


---


And then also we will need  to allow the C++ failure for now, like this

```
$ grep oils_cpp_failures */*.sh
spec/array-sparse.test.sh:## oils_cpp_failures_allowed: 2
spec/builtin-bind.test.sh:## oils_cpp_failures_allowed: 7
```



Sometimes the spec tests are half the problem, but they are important!  

Thanks for working on this

daveads:
@andychu what's the update on this...

core/process.py
```diff
@@ -2192,6 +2205,18 @@ def WaitForOne(self, waitpid_options=0):
             # Print newline after Ctrl-C.
             if term_sig == SIGINT:
                 print('')
+            else:
+                msg = GetSignalMessage(term_sig)
+                if msg is not None:
+                    if mylib.PYTHON:
+                        # WCOREDUMP is only available on some systems
+                        try:
+                            from os import WCOREDUMP  # type: ignore
```
andychu, line 2214: Hmmm now I see that WCOREDUMP is not POSIX

https://www.man7.org/linux/man-pages/man2/waitpid.2.html

I wonder what bash does?  Does bash use `#ifdef WCOREDUMP` ?



```
WCOREDUMP(wstatus)
              returns true if the child produced a core dump (see
              [core(5)](https://www.man7.org/linux/man-pages/man5/core.5.html)).  This macro should be employed only if
              WIFSIGNALED returned true.

              This macro is not specified in POSIX.1-2001 and is not
              available on some UNIX implementations (e.g., AIX, SunOS).
              Therefore, enclose its use inside #ifdef WCOREDUMP ...
              #endif.
```

andychu (commented):
I fixed the tests -- some assertions needed to be adjusted

But now I see this WCOREDUMP issue ...

we try to target POSIX; I didn't realize that wasn't POSIX

core/process.py
daveads: Taking a look at it.

daveads: @andychu  https://github.com/bminor/bash/blob/637f5c8696a6adc9b4519f1cd74aa78492266b7f/include/posixwait.h#L79-L85

i think there is some sort of Fallback when WCOREDUMP not available.

daveads: i guess it's time for me to play with the cpp side of this codebase

andychu: OK thanks for looking up the `#ifdef`!  That helps

And it seems relatively simple ... what is `s & 0200` I wonder?  Is that just one signal?

---

Hm yeah I would give it a try, but if you get stuck, it's not a big deal

@PossiblyAShrub has ported things to C++

However I guess this involves `./configure` as well ...

I wonder if we can **always** use `s & 0200` ?

andychu:
Hm I asked Claude for some code to test the theory that we can always use `s & 0200` ?

https://claude.ai/share/c62df9d4-1204-4b0c-b520-1836b6dbeb12


Seems worth trying / investigating

andychu:
Oh yes in glibc, it is always 0x80

https://github.com/bminor/glibc/blob/7cee1b4b40c6b0d6498a92decbfb397ef6ffe96a/bits/waitstatus.h#L59

andychu:
Same with musl libc: https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/include/sys/wait.h#L51

OK so yeah let me just merge this, since it passes tests

And then we can port it to C++ separately

Now we know there is a portable way to do it

daveads:
> Same with musl libc: https://github.com/bminor/musl/blob/1b76ff0767d01df72f692806ee5adee13c67ef88/include/sys/wait.h#L51
> 
> OK so yeah let me just merge this, since it passes tests
> 
> And then we can port it to C++ separately
> 
> Now we know there is a portable way to do it

okay please create an issue for that so i can attend to it.

andychu:
Let's discuss it on Zulip -- I will start a thread


(you can also create issues yourself if you like -- you can edit the Kanban board, since we have talked about this)

daveads:
> Let's discuss it on Zulip -- I will start a thread
> 
> (you can also create issues yourself if you like -- you can edit the Kanban board, since we have talked about this)

okayy

PR #2517, opened 2025-10-26
[mycpp] Fix issue with conflicting method and free function names

last-genius:
Add an example of such a conflict (which crashes before the fix)

Add a mycpp fix.

### DIFF
--- a/mycpp/TEST.sh
+++ b/mycpp/TEST.sh
@@ -244,9 +244,6 @@ test-invalid-examples() {
       */invalid_switch.py)
         expected_status=5
         ;;
-      */invalid_ctx_raise.py)
-        expected_status=2
-        ;;
       */invalid_partialtype.py)
         expected_status=2
         ;;
--- a/mycpp/examples/invalid_ctx_name.py
+++ b/mycpp/examples/invalid_ctx_name.py
@@ -1,0 +1,12 @@
+#!/usr/bin/env python2
+"""
+invalid_ctx_name.py
+"""
+from typing import Any
+
+
+class ShouldStartWithCtx(object):
+
+    def __exit__(self, type, value, traceback):
+        # type: (Any, Any, Any) -> None
+        pass
--- a/mycpp/examples/invalid_ctx_raise.py
+++ b/mycpp/examples/invalid_ctx_raise.py
@@ -9,13 +9,6 @@ from typing import Any
 from mycpp import mylib
 
 
-class ShouldStartWithCtx(object):
-
-    def __exit__(self, type, value, traceback):
-        # type: (Any, Any, Any) -> None
-        pass
-
-
 class ctx_MaybePure(object):
     """Regression for early return."""
 


mycpp/cppgen_pass.py
```diff
@@ -1010,6 +1010,18 @@ def oils_visit_call_expr(self, o: 'mypy.nodes.CallExpr') -> None:
             self.write('to_float')
         elif callee_name == 'bool':
             self.write('to_bool')
+        elif isinstance(o.callee, NameExpr):
```
andychu, line 1032: Hm I have a similar thought on this as I did for the previous change

I think `if callee_name == 'int'` (above)    and `if instance(o.callee_name, NameExpr)` in `cppgen_pass.py` are basically ad hoc parsing / checks right before we write C++

---

I think an interesting question is - what if we were to translate Oils to Rust (which I think is feasible, for at least "liboils") ?

I think if we did that, maybe the conversion_pass has to understand the difference between function and method calls too

Or maybe it's only a C++ thing, not sure


But in general it does seem nicer to put this info earlier

---

The `to_int()` and `to_float()` changes are because `int`  and `float` are keywords in C++

mycpp/visitor.py
```diff
@@ -235,12 +239,17 @@ def oils_visit_class_def(
             if isinstance(stmt, FuncDef):
                 method_name = stmt.name
 
+                if self.current_class_name in self.methods:
+                    self.methods[self.current_class_name].append(method_name)
+                else:
+                    self.methods[self.current_class_name] = [method_name]
```
andychu: Hm what if b() is defined before a() ?  I think we might be accumulating them in order here

andychu (commented):
Hm I am not sure about this, it might be simpler to disallow it ...

Or to understand it in an earlier pass

---

According to Claude, Rust does not have the ambiguity with f() that  C++ does ... it can only be a free function

Though still this requires us to keep more state in the visitor.

Generally I prefer having state passed between passes like in `translate.Run()`.  Each state has specific inputs and outputs

mycpp/visitor.py
```diff
@@ -31,6 +31,8 @@ def __init__(self) -> None:
 
         self.current_class_name: Optional[util.SymbolPath] = None
         self.current_method_name: Optional[str] = None
```
andychu: Yeah right now there are only 2 mutable members of the visitor

And I don't think we should add more, because they are **inherited** by all passes

The 2 mutable members are basically because visitors "lose the stack"

e.g. see

[oils-dev](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev) > [I dislike visitors](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/I.20dislike.20visitors/with/177846649)


----

I think we may want a simple hack to disallow it

and then later we can build up a full understanding of the namespaces, if we translate to Rust

Right now mycpp doesn't understand either Python or C++ namespacing; it is just a text printer ...

andychu:
Hm thinking about this more

I suspect there is the ordering bug where we build up methods and also visit their bodies at the same time, so we may miss some

----

Regardless of whether we choose the "disallow" or "write correct C++ code" solution, I think we might want to accumulate the methods in a pass BEFORE conversion_pass

And I think that the const pass can be moved before the conversionn pass

Right now it visits all string literals, and outputs global_strings

So we can also make it output `MethodDefs` or `Namespaces` or something.  For Class/Module namespaces. 

We could start with something simple for now, but eventually it would evolve into a full understanding of namespaces.  (That will presumably help with a translation to Rust)


```
     # [PASS]
     timer.Section('mycpp pass: CONST')
 
     global_strings = const_pass.GlobalStrings()
     p_const = const_pass.Collect(types, global_strings)
 
     for name, module in to_compile:
         p_const.visit_mypy_file(module)
         MaybeExitWithErrors(p_const)
```

And then either

- the conversion pass reads the namespaces, and disallows conflicting names
- or the cppgen pass reads the namespaces, and writes the correct code

----

Let me know what you think ... thanks for working on this!



The general idea is that we should structure it more like a pipeline in `translate.py Run()`, with explicit inputs and outputs to the visitor stages

And ALL the passes inherit from the visitor, so we don't want to add mutable state to ALL passes


And actually now that I wrote that, I see a place where we break that rule, and can improve.  I believe `self.current_class_name` can be passed in **explicitly** to our `oils_visit_*` functions, and it shouldn't be mutable member state

(which looks like mutable global state when it's a visitor)

mycpp/cppgen_pass.py
last-genius: Rust doesn't have the underlying issue - to call an associated function or a method on a struct you **need** to prefix them with `Struct::associated_func()` or `self.method()`. `func()` unambiguously calls the free function.

That said, the ordering issue is there and I think you're right about having to move this to an earlier pass - I'll refactor it

last-genius:
Refactored, moved the collection of method names and namespaces to the const pass (moving it before the conversion pass in turn)

last-genius:
The error seems to come from this function no longer being visited in `conversion_pass.py`:
```
    def visit_raise_stmt(self, o: 'mypy.nodes.RaiseStmt') -> None:
        if self.inside_dunder_exit:
            # Note: this doesn't check function calls that raise, but it's
            # better than nothing
            self.report_error(
                o, "raise not allowed within __exit__ (C++ doesn't allow it)")
            return
```

It's unclear to me why this is happening....

last-genius:
I've finally figured out what was causing the test failure - should be ready for review @andychu

mycpp/const_pass.py
```diff
@@ -123,6 +164,23 @@ def oils_visit_log_call(self, fmt: StrExpr,
...
+
+        for stmt in o.defs.body:
+            if isinstance(stmt, FuncDef):
+                self.method_defs.Add(current_class_name, stmt.name)
+                self.class_namespaces.Set(current_class_name,
+                                          self.current_file_name)
+        super().oils_visit_class_def(o, base_class_sym, current_class_name)
+
+    def oils_visit_mypy_file(self, o: MypyFile) -> None:
+        if util.ShouldSkipPyFile(o):
```
andychu: I think we should not need this ShouldSkip()?

In visitor, `visit_mypy_file()` calls it, and then it doesn't calls `oils_visit_mypy()` if it's true

The `oils_visit_*` methods are supposed to already be "narrowed"

mycpp/cppgen_pass.py
```diff
@@ -1024,6 +1029,19 @@ def oils_visit_call_expr(
...
+            # CallExpr means a free function is called here, not a method
+
+            # If there are methods on the current class with the same name,
+            # prefix the free function with the current namespace to ensure
+            # it's not the method (shadowing the free function) that's called
+            if (current_class_name and self.method_defs.ClassHasMethod(
```
andychu, line 1038: Comment for later: it feels like we should have visit_method_call() and  visit_func_call(), not a single visit_call_expr()

But I guess we can move there gradually

mycpp/examples/test_func_method_name_conflict.py
```diff
@@ -0,0 +1,45 @@
...
+    # type: () -> int
+    return 1
+
+
+class A:
+
+    def __init__(self):
+        # type: () -> None
+        pass
+
```
andychu, line 19: To test the previous bug, we could add another method function that calls a() before the method is defined

And maybe a free function that calls the free function a(), etc.

andychu (commented):
Thanks, this looks good

I would have also been fine with disallowing shadowing - a method name can't have the same name as a free function.  That would be simple 

Although I think this is good progress  in terms of mycpp understanding symbols better, and not just relying on the C++ compiler to create errors

PR #2518, opened 2025-10-26
Expose a more flexible stat API to reduce the number of stat calls in noclobber

PossiblyAShrub:
The [previous noclobber PR](2503) called stat twice because our stat API lacked the flexibility to check if a file exists and then check if that stated file was a regular file.

This improves the API by:
- Returning a `StatResult` object (or `nullptr`) from a stat function
- Further inspections of the `StatResult` can be done using methods on that object.

TODO:
- [x] Unit tests

### DIFF
--- a/mycpp/gc_mylib.cc
+++ b/mycpp/gc_mylib.cc
@@ -320,19 +320,16 @@ BigStr* BufWriter::getvalue() {
   }
 }
 
-StatResult::StatResult(struct stat stat_result)
-  : stat_result_(stat_result) {}
-
 bool StatResult::isreg() {
   return S_ISREG(stat_result_.st_mode);
 }
 
 StatResult* stat(BigStr* filename) {
-  struct stat stat_result;
-  if (::stat(filename->data_, &stat_result) < 0) {
+  auto* st = Alloc<StatResult>();
+  if (::stat(filename->data_, &st->stat_result_) < 0) {
     return nullptr;
   }
-  return Alloc<StatResult>(stat_result);
+  return st;
 }
 
 }  // namespace mylib
--- a/mycpp/gc_mylib.h
+++ b/mycpp/gc_mylib.h
@@ -378,7 +378,6 @@ class UniqueObjects {
 
 class StatResult {
 public:
-  StatResult(struct stat stat_result);
   bool isreg();
 
   static constexpr ObjHeader obj_header() {
@@ -389,7 +388,6 @@ public:
     return kZeroMask;
   }
 
-private:
   struct stat stat_result_;
 };
 


PossiblyAShrub:
This is ready for a review.

mycpp/gc_mylib.cc
```diff
@@ -320,4 +320,19 @@ BigStr* BufWriter::getvalue() {
...
+bool StatResult::isreg() {
+  return S_ISREG(stat_result.st_mode);
+}
+
+StatResult* stat(BigStr* filename) {
+  struct stat stat_result;
+  if (::stat(filename->data_, &stat_result) < 0) {
+    return nullptr;
+  }
+  return Alloc<StatResult>(stat_result);
```
andychu: Technically it is possible to Alloc<StatResult> first, and then pass `&(p.stat_result_)` to the `::stat()` call

That avoids copying the `stat_result` type

mycpp/gc_mylib.h
```diff
@@ -373,6 +375,26 @@ class UniqueObjects {
...
+  static constexpr ObjHeader obj_header() {
+    return ObjHeader::ClassFixed(field_mask(), sizeof(StatResult));
+  }
+
+  static constexpr uint32_t field_mask() {
+    return kZeroMask;
+  }
+
+private:
+  struct stat stat_result;
```
andychu: We generally use a trailing underscore for member vars, i.e. `stat_result_`

mycpp/mylib.py
```diff
@@ -508,3 +509,35 @@ class File:
...
+    except posix.error:
+        return None
+
+    return StatResult(stat_result)
+
+
+class StatResult:
+    def __init__(self, stat_result):
+        # type: (posix.stat_result) -> None
+        self.n_fields = stat_result.n_fields
```
andychu: Could this just save `self.stat_result` ?  And then `isreg()` can operate on `self.stat_result.st_mode` 

Less code :-)

andychu (commented):
Thanks looks great!  Thanks

I would try avoiding the copy of stat_result in C++, which may be kinda big sometimes

But if it doesn't work, it's not essential

---

I think we can Alloc<T>(StatResult) can construct a zero'd object, as long as StatResult has the right constructor

And then we can pass a pointer to that member directly (so it's probably not private anymore)

mycpp/gc_mylib.cc
PossiblyAShrub: One downside is that then we'd have to allocate even when we return a `nullptr`. But the copy has a cost too.

PossiblyAShrub:

andychu:
Yeah good point, I guess there could be some cases where most stats fail, and then you allocate a lot.  I guess with `$PATH` lookup, most will fail

It's probably not a big deal either way though - this seems good :-)

Issue #2426, opened 2025-09-19
[osh as bash] chrony - $[ ] used as synonym for $(( ))  rather than YSH


andychu:
```
  shift $[$OPTIND - 1]
          ^~~~~~~
./run:40: In expressions, remove $ and use `OPTIND`, or sometimes "$OPTIND"
```

andychu:
### happens in x42-plugins in aports/community

[x42-plugins build failure](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/x42-plugins.20build.20failure/with/554155648)

```
./genhead.sh 8 8 > x42/matrixmixer.h
        i=$[$i + 1]
            ^~
[ interactive ]:24: In expressions, remove $ and use `i`, or sometimes "$i"

```

found by @btertoolen

andychu:
[Change $[] to shell arithmetic like $(( ))](https://oilshell.zulipchat.com/#narrow/channel/503208-bug-post-mortem/topic/Change.20.24.5B.5D.20to.20shell.20arithmetic.20like.20.24.28.28.20.29.29/with/555461496)

PR #2519, opened 2025-10-26
Turn $[] into shell arithmetic in OSH, for compatibility

last-genius:
`chrony` is the only package using the backwards-compatible `$[$var]` so far, but it's tricky to patch since it's used quite a lot. in bash, `$[]` is `$(( ))`, but in osh, it's a ysh expression sub. Add an option allowing `$var` in expression subs from osh.

Probably could come up with a better name for the `shopt` option....

Fixes: #2426

### DIFF
--- a/ysh/run.sh
+++ b/ysh/run.sh
@@ -18,18 +18,16 @@ OSH_CPP=_bin/cxx-asan/osh
 
 parse-one() {
   echo ---
-  # TODO: these tests can be removed once the rest of ysh/ is translated
-  if test "$OSH" = "$OSH_CPP"; then
-    local prog="$1"
-    local skip=''
-    case $prog in
-      (*/assign.osh) skip=T ;;
-    esac
 
-    if test -n "$skip"; then
-      echo "skipping $prog"
-      return
-    fi
+  local prog="$1"
+  local skip=''
+  case $prog in
+    */assign.osh) skip=T ;;
+  esac
+
+  if test -n "$skip"; then
+    echo "skipping $prog"
+    return
   fi
 
   set +o errexit
@@ -101,12 +99,8 @@ soil-run() {
 }
 
 soil-run-cpp() {
-  local osh=$OSH_CPP
-
-  ninja $osh
-
-  # TODO: replace with run-test-funcs once the rest of ysh is translated
-  OILS_GC_ON_EXIT=1 OSH=$osh test-parse-osh
+  ninja $OSH_CPP
+  OSH=$OSH_CPP run-test-funcs
 }
 
 run-for-release() {
--- a/ysh/testdata/array-splice-demo.osh
+++ b/ysh/testdata/array-splice-demo.osh
@@ -1,22 +1,14 @@
 
-shopt -s oil-parse-at static-word-eval
-
 myglobal=global
 
 var myarray = :| $myglobal $'line\n' ${z:-default} |
+
 spec/bin/argv.py @myarray
-shopt -s parse_subexpr_shorthand
-echo len=$[len(myarray)]
-echo
 
 append _ '1 2' ${myglobal:-default} (myarray)
 spec/bin/argv.py @myarray
-echo len=$[len(myarray)]
 echo
 
 for item in (myarray) {
   echo "  $item"
 }
-
-# NOTE: This doesn't work for arrays
-#setvar myarray += %(more stuff)
--- a/ysh/testdata/sigil-pairs.sh
+++ b/ysh/testdata/sigil-pairs.sh
@@ -39,5 +39,5 @@ var cmd_sub = $(
 
 show $cmd_sub
 
-show $['quoted ' ++ 'words ' ++ "dq ${x:-default}"]
+#show $['quoted ' ++ 'words ' ++ "dq ${x:-default}"]
 


andychu:
Hm I'm not sure if adding an option is the best solution

#2426 is labeled "may-required-design" and in the "needs design" column

It can also be a 4 byte patch to this one package - `$[] -> $(( ))`

Depending on how often it happens

last-genius:
> Hm I'm not sure if adding an option is the best solution
> 
> #2426 is labeled "may-required-design" and in the "needs design" column
> 
> It can also be a 4 byte patch to this one package - `$[] -> $(( ))`
> 
> Depending on how often it happens

As I say in the PR description:
> chrony is the only package using the backwards-compatible $[$var] so far, but it's tricky to patch since it's used quite a lot

`$[$var` is used 45 times in 9 files!

andychu:
> $[$var is used 45 times in 9 files!

Oh that's all in the chrony package itself?  did you just grep for it?

---

The more conservative option is that `$[]` can be a synoym for `$(())` like bash

But then `shopt --set ysh:upgrade` turns `$[]` into the YSH behavior

That would have been the typical strategy, but I did not see bash `$[]` very much .  I am curious how many packages it happens in -- out of say 8,000 or 80,000

If it happens only 45 times in 1 package, that is different than happening 1 time in 45 packages ...

andychu:
In general YSH `$[]` is quite diffferent -- it has Python-like expressions

```
$ osh -c 'echo $[len([1,2,3]) === 3]'
true
```

So it's not clear that allowing `$var` is enough

Conversely, bash expressions are:

```
$ bash -c 'x=ff; echo $[ 0x$x == 255 ? 42 : 99 ]'
42
```

YSH basically deprecates shell arithmetic because it's too dynamic, and `$[]` is shell arithmetic, but ONLY in bash

last-genius:
Added tests for #2566 as well. Moved to treating `$[]` as an arithmetic expression by default (with an option to make it an YSH expression sub).

Tests that are failing are using `$[]` as ysh expression sub without the corresponding `shopt`

last-genius:
No idea why `ysh/testdata/array-splice-demo.osh` still treats `$[]` as an arithmetic expression even with the `shopt` in the failing CI test:

```
shopt -s parse_subexpr_shorthand
echo len=$[len(myarray)]
```

---
```
  echo len=$[len(myarray)]
                ^
ysh/testdata/array-splice-demo.osh:9: Unexpected token after arithmetic expression (Id.Arith_LParen != Id.Arith_RBracket)
```

It works for me locally...

spec/arith.test.sh
```diff
@@ -1042,3 +1042,41 @@ status=0
 ## BUG bash STDOUT:
 status=1
 ## END
+
+#### $[ is $((
```
andychu: minor organization: let's put these in `spec/arith-context`, since `spec/arith` has over 70 cases

And also I see spec/arith-context case 9 is "expr sub $[] happens to be the same in simple case", which is related to this

These 2 cases could be redundant

doc/ref/chap-option.md
```diff
@@ -224,26 +224,27 @@ simply quote it like `'@literal'` to fix the problem.
...
-      verbose_errexit         Whether to print detailed errors
-      verbose_warn            Print various warnings to stderr
+      parse_at                   echo @array @[arrayfunc(x, y)]
+      parse_brace                if true { ... }; cd ~/src { ... }
+      parse_equals               x = 'val' in Caps { } config blocks
+      parse_paren                if (x > 0) ...
+      parse_proc                 proc p { ... }
+      parse_triple_quote         """$x"""  '''x''' (command mode)
+      parse_ysh_string           echo r'\' u'\\' b'\\' (command mode)
+      parse_subexpr_shorthand    Makes $[] a YSH expression sub instead of a backwards-compatible equivalent of an arithmetic expression
```
andychu: How about `parse_ysh_expr_sub` ?   Similar to `parse_ysh_string`

We call `$[]` "expression sub", like `$()` is "command sub" ins hell


Thanks for updating the help!

osh/word_parse.py
```diff
@@ -1544,8 +1551,8 @@ def _ReadArithExpr(self, end_id):
                 loc.Word(self.a_parser.cur_word))
         return anode
 
-    def _ReadArithSub(self):
-        # type: () -> word_part.ArithSub
+    def _ReadArithSub(self, backwards_compatible=False):
```
andychu: instead of the boolean, how about a param with default `end_id=Id.Arith_RParen`

And then in the other case, pass `end_id=Id.Arith_RBracket`

We use that pattern elsewhere in this file, because sometimes the terminator is different


---

(We could also `assert end_id in (Id.Arith_RParen, Id.Arith_RBracket)` if desired)

osh/word_parse.py
```diff
@@ -1568,14 +1575,18 @@ def _ReadArithSub(self):
...
+            endid = Id.Arith_RBracket if backwards_compatible else Id.Arith_RParen
+            anode = self._ReadArithExpr(endid)
 
         self._SetNext(lex_mode_e.ShCommand)
 
-        # Ensure we get closing )
-        self._GetToken()
-        if self.token_type != Id.Right_DollarDParen:
-            p_die('Expected second ) to end arith sub', self.cur_token)
+        if not backwards_compatible:
```
andychu: and this could be `if end_id == Id.Arith_RParen`

andychu (commented):
This looks great -- thank you !!   I'm glad the code diff was small

Let me try running that failure you pointed out ...

andychu:
Ah OK the tests in `ysh/run.sh` were old and not testing much useful, so I removed them / cleaned them up

( The reason it failed is because it parsed the WHOLE FILE with `osh -n`, which does not RUN the shopt that changes parsing ... this is a general issue with `shopt` that affects bash too )

Let's see if the tests pass now ...

doc/ref/chap-option.md
andychu: Oh also for the help, I try to fit it within 80-100 chars

So we could say

```
parse_ysh_expr_sub   $[] is YSH expression sub, not synonym for $(( ))
```

I think that will be short enough

spec/arith.test.sh
andychu: And yes actually I think test case 9 can be deleted in favor of these tests

It is now obsolete, since $[] is the same as bash, and does not "happen to be the same" !

andychu:
OK cool, tests pass now!  Let me look at the `test` builtin PR now ...

last-genius:
Addressed all the comments.

andychu:
Merged, thank you!

PR #2525, opened 2025-10-27
Fix docs: ref/chap-type-method io->evalToDict

PossiblyAShrub:
`io->evalToDict` has been deprecated, and the docs were confusing before because it referenced `to_dict=true` and then used `io->evalToDict` in the example.


andychu:
thanks!

PR #2526, opened 2025-10-28
add missing source build/dev-shell.sh to regtest/aports-html.sh

btertoolen:
Add missing source command to `regtest/aports-html.sh`


andychu:
Thank you!

PR #2527, opened 2025-10-28
replace " with \' in sql files

btertoolen:
Replace double quotes with single quotes in regtest sql files to fix errors like this:
```
Parse error near line 8: no such column: "baseline/" - should this be a string literal in single-quotes?
  status1,     b.elapsed_secs as baseline,     cast("baseline/" || b.pkg_HREF as
                                      error here ---^
Parse error near line 29: no such column: "baseline" - should this be a string literal in single-quotes?
  create table metrics as   select     cast("baseline" as text) as config,     *
                              error here ---^
Parse error near line 35: no such column: "osh-as-sh" - should this be a string literal in single-quotes?
  insert into metrics select   cast("osh-as-sh" as text) as config,   * from osh
```

See also [#distros > Updating Causes of regtest/aports Failures](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/Updating.20Causes.20of.20regtest.2Faports.20Failures/with/547362321)

### DIFF
--- a/regtest/aports/cause.sql
+++ b/regtest/aports/cause.sql
@@ -17,11 +17,11 @@ where causes.pkg = diff_baseline.pkg;
 
 -- Set causes for signals/timeouts
 update diff_baseline
-set cause = "signal-124"
+set cause = 'signal-124'
 where status1 = 124 or status2 = 124;
 
 update diff_baseline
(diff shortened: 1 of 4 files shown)


andychu:
Thank you!

Issue #2530, opened 2025-10-29
gzip - builtin cat crashes subshell


last-genius:
[#oils-dev > gzip - PIPE mishandling](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/gzip.20-.20PIPE.20mishandling/with/547736811)

osh is using a builtin `cat` as an optimization, which means when it crashes (due to a broken pipe, for example), it brings down the whole subshell with itself, and not just its own subprocess. `gzip` uses a construct like below to test how `cat` behaves in case of a broken pipe, and osh doesn't report the exit status for it.

Compare builtin cat:
```
$ bin/osh -c '((cat </dev/zero; echo $? >&7) | true) 7>&1'
<empty> - incorrect
```
vs. external:
```
$ bin/osh -c '((/usr/bin/cat </dev/zero; echo $? >&7) | true) 7>&1'
141 - correct
```

This patch fixes this issue:
```diff
diff --git a/core/vm.py b/core/vm.py
index 89c43574e..e627fde08 100644
--- a/core/vm.py
+++ b/core/vm.py
@@ -13,7 +13,10 @@ from core.error import e_die
 from core import pyos
 from core import pyutil
 from display import ui
+from errno import EPIPE
+from mycpp import iolib
 from mycpp.mylib import log, tagswitch
+from signal import SIG_IGN, SIGPIPE

 from typing import List, Dict, Tuple, Optional, Any, cast, TYPE_CHECKING
 if TYPE_CHECKING:
@@ -272,10 +275,13 @@ class _Executor(object):
         with ctx_FlushStdout(io_errors):
             # note: could be second word, like 'builtin read'
             with ui.ctx_Location(self.errfmt, cmd_val.arg_locs[0]):
+                iolib.sigaction(SIGPIPE, SIG_IGN)
                 try:
                     status = builtin_proc.Run(cmd_val)
                     assert isinstance(status, int)
                 except (IOError, OSError) as e:
+                    if e.errno == EPIPE:
+                        return 141
                     self.errfmt.PrintMessage(
                         '%s builtin I/O error: %s' %
                         (cmd_val.argv[0], pyutil.strerror(e)),
```

We want to fix this properly though - either revert the `cat` optimization, or just fork (no exec) when builtin `cat` is used.

andychu:
Hm yeah I wonder if there is a spot in the code where we can put

"if this is `cat`, then fork" ... I don't think we have something like that

---

[...]

PR #2531, opened 2025-10-29
[spec] Add divergence test for gzip aports issue

last-genius:
Spec test for #2530


andychu:
Thanks!

`((` still looks wrong to me, but I'm glad OSH handles it :-)  


Either of these also shows the bug:

```
$ bash -c '( { cat </dev/zero; echo $? >&7; } | true) 7>&1'
141
$ bash -c '{ { cat </dev/zero; echo $? >&7; } | true; } 7>&1'
141
```

Issue #2336, opened 2025-07-02
`${line##*([}` intepreted as extended glob - Alpine


andychu:
[#osh > alpine xz - "${line##*([}" interpreted as extended glob](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/alpine.20xz.20-.20.22.24.7Bline.23.23*.28.5B.7D.22.20interpreted.20as.20extended.20glob)

andychu:
NOTE: It's not clear how to fix this one, it may require parser changes

We might need a different status for it

andychu:
In the worst case, we could do the "flightsim" hacks ... i.e. application-specific hacks

Or of course upstream patches, but that only affects **one distro** ... we don't want to patch 10 distros, because dash/bash run this code

andychu:
`test/spec.sh divergence`

PR #2533, opened 2025-10-30
[regtest] Patch /usr/bin/autopoint for xz

last-genius:
Escaping is horrific, but we need to handle two layers of shell+sed here.

xz builds successfully with this patch for me locally

Fixes: #2336

### DIFF
--- a/regtest/aports-setup.sh
+++ b/regtest/aports-setup.sh
@@ -77,14 +77,20 @@ patch-aports() {
   # also need to be specified in the 'source=' and 'sha512sums=' lists in APKBUILD
   for patch_dir in $cur_dir/regtest/patches/*; do
       local patch_dir=$(basename $patch_dir)
+      local apkbuild=main/$patch_dir/APKBUILD
+      git restore $apkbuild
       for patch in $cur_dir/regtest/patches/$patch_dir/*; do
-          cp $patch main/$patch_dir/
-          local patch_name=$(basename $patch)
-          local shasum=$(sha512sum $patch | cut -d " " -f1)
-          local apkbuild=main/$patch_dir/APKBUILD
-          git restore $apkbuild
-          sed -i "/source='*/ a $patch_name" $apkbuild
-          sed -i "/sha512sums='*/ a $shasum  $patch_name" $apkbuild
+          if [[ $patch == *.patch ]]; then
+              cp $patch main/$patch_dir/
+              local patch_name=$(basename $patch)
+              local shasum=$(sha512sum $patch | cut -d " " -f1)
+              sed -i "/source='*/ a $patch_name" $apkbuild
+              sed -i "/sha512sums='*/ a $shasum  $patch_name" $apkbuild
+          elif [[ $patch == *.copy ]]; then
+              cp $patch main/$patch_dir/
+          elif [[ $patch == *.apkbuild ]]; then
+              git apply $patch
+          fi
       done
   done
 
--- a/regtest/patches/xz/extglob.apkbuild
+++ b/regtest/patches/xz/extglob.apkbuild
@@ -1,0 +1,12 @@
+diff --git a/main/xz/APKBUILD b/main/xz/APKBUILD
+index 7ffe05e7024..f69634541a7 100644
+--- a/main/xz/APKBUILD
++++ b/main/xz/APKBUILD
+@@ -22,6 +22,7 @@ source="https://github.com/tukaani-project/xz/archive/refs/tags/v$pkgver/xz-$pkg
+ 
+ prepare() {
+ 	default_prepare
++	doas patch /usr/bin/autopoint -i ../../autopoint.patch
+ 	autoreconf -fi
+ }
+ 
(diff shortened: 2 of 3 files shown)


regtest/aports-setup.sh
```diff
@@ -88,6 +88,10 @@ patch-aports() {
       done
   done
 
+  # Patch xz's build dependency in the prepare stage (#2336)
+  git restore main/xz/APKBUILD
+  sed -i "/default_prepare/ a doas sed -i \"s/##\\\\\\\\*(\\\\\\\\[/##*\'([\'/\" /usr/bin/autopoint" main/xz/APKBUILD
```
andychu: Hm is it possible to make this `patch < autopoint-quote-fix.patch` ?

Oh I guess it needs `doas` because `/usr/bin/autopoint` is already installed

So then `doas patch <` ?

This will be more readable for a blog post, and it will help people who want to merge this patch into their own trees

(We should avoid propagating "Unix sludge" :-) )

andychu:
I guess we also have to make sure that abuild does not pick up this patch in the build stage, which is the normal thing

I am not sure what  the convention is -- does it automatically find `*.patch`?   We can call it something else then

last-genius:
Dropped the "Unix sludge" :)

last-genius:
So `patches/*/*.patch` work as patches to be copied into the package subdir and applied by `abuild` during the build stage. `patches/*/*.apkbuild` are instead patches that need to be applied by our scripts to package's `APKBUILD` itself (potentially changing any other stage of the build, `prepare` in this case).

regtest/patches/xz/extglob.apkbuild
```diff
@@ -0,0 +1,12 @@
+diff --git a/main/xz/APKBUILD b/main/xz/APKBUILD
+index 7ffe05e7024..9ee5d511136 100644
+--- a/main/xz/APKBUILD
++++ b/main/xz/APKBUILD
+@@ -22,6 +22,7 @@ source="https://github.com/tukaani-project/xz/archive/refs/tags/v$pkgver/xz-$pkg
+ 
+ prepare() {
+ 	default_prepare
++	doas sed -i "s/##\\*(\\[/##*'(['/" /usr/bin/autopoint
```
andychu: I still think this is Unix sludge because it mixes languages

- `\*` to escape regex metachar
- then `\\` to escape `\` within shell double quotes
- and double quotes are natural because the replacement string has single quotes

This sent me on a diversion, because these kinds of problems are things I'd like to address in YSH

[sed and awk have problems with constant strings versus regex](https://oilshell.zulipchat.com/#narrow/channel/447377-api-design/topic/sed.20and.20awk.20have.20problems.20with.20constant.20strings.20versus.20regex/with/548155685)

That is, I'd like to be able express this with the **literal** strings

```
*([

'*(['
```

Here is my attempt with awk, which looks WORSE than sed!  I discovered that `-v` does extra `\\` unquoting, which is should **not** do

```
$ sq="'"; awk -v search='##\\*\\(\\[' -v replace='##'$sq'*(['$sq  '{ gsub(search, replace); print }' << 'EOF'
${line##*([}
EOF
${line##'*(['}

```

I have a solution with `patch` which I will post

andychu:
Here is my solution with `patch`

Note I edited this file to have `autozz.orig` and `autozz`

```
$ cat foo.patch 
--- autozz.orig 2025-10-31 09:10:10.866813634 -0400
+++ autozz      2025-10-31 09:09:46.670532879 -0400
@@ -6,3 +6,5 @@
 6
 7
 8
+9
+10
```

and then


```
$ sudo sh -c 'seq 8 > /usr/bin/autozz'

$ sudo sh -c 'cd /usr/bin; patch -p 0 -i /home/andy/git/oils-for-unix/oils/foo.patch'; cat /usr/bin/autozz
patching file autozz
1
2
3
4
5
6
7
8
9
10
```

So I think we can use this same style

I like having the patch, because `/usr/bin/autopoint` is a part of the GNU `gettext` package, so others may use it

Then they can just copy the patch.  I think the `sed` is still obscure

andychu:
I also tried to improve on sed, to do it without escaping ... keeping `*([` literal in both cases

It seems impossible, since the search pattern is the regex language, and the replacement pattern is also a non-trivial constant language, that supports `\1` I think

This is the closest I got, but it doens't work with `([`

```
$ x='\*'; sed "s,$x,'$x'," << 'EOF'
${line##*([}
EOF
${line##'*'([}

```

So I think the patch solution is pretty clean

It is a double patch, but each patch is readable then

last-genius:
Switched to double-`patch`ing

regtest/aports-setup.sh
```diff
@@ -77,14 +77,20 @@ patch-aports() {
...
-          sed -i "/source='*/ a $patch_name" $apkbuild
-          sed -i "/sha512sums='*/ a $shasum  $patch_name" $apkbuild
+          if [[ $patch == *.patch ]]; then
+              cp $patch main/$patch_dir/
+              local patch_name=$(basename $patch)
+              local shasum=$(sha512sum $patch | cut -d " " -f1)
+              sed -i "/source='*/ a $patch_name" $apkbuild
+              sed -i "/sha512sums='*/ a $shasum  $patch_name" $apkbuild
+          elif [[ $patch == *.copy ]]; then
+              cp $patch main/$patch_dir/
```
andychu, line 90: Hm I'm bit confused how this works, but we'll see ...

PR #2534, opened 2025-10-30
[core] Run builtin cat in a subprocess

last-genius:
Otherwise if it crashes, it brings down the whole subshell with it and we don't capture its exit status correctly.

Create a new Thunk class for this purpose - its Run() just invokes a builtin and returns (no exec() involved).

With this change, the divergence test for 'cat' is passing.

Fixes: #2530

### DIFF
--- a/core/process.py
+++ b/core/process.py
@@ -70,6 +70,7 @@ if TYPE_CHECKING:
     from _devbuild.gen.syntax_asdl import command_t
     from builtin import trap_osh
     from core import optview
+    from core import vm
     from core.util import _DebugFile
     from osh.cmd_eval import CommandEvaluator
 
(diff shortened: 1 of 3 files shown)


core/process.py
```diff
@@ -32,6 +32,7 @@
 from core import pyutil
 from core import pyos
 from core import state
+from core.vm import _Executor
```
andychu: We usually follow the style `from core import vm`, and then the type annotation uses `vm._Executor`

And then put in the `if TYPE_CHECKING` block

(At some point, I think mycpp will need to gain an understanding of modules.  And then this kind of thing can be checked.  As mentioned, right now it is ignorant, and makes one big C++ translation unit)

andychu (commented):
This is great, thank you!!

After this I want to try a separate commit to remove the `interactive()` check, because I remember I cheated there

I believe it's due to the same "lack of process" problem

And it's more consistent without that (but we should land this first)

PR #2536, opened 2025-10-30
[mycpp] Turn global mutable state into private variables

last-genius:
Motivated by https://github.com/oils-for-unix/oils/pull/2517#issuecomment-3448756702

Make `self.current_class_name`, `self.current_method_name` and `self.at_global_scope` private variables and pass them explicitly into functions requiring them.

This still leaves `self.indent` as global mutable state - it's used in a lot more places and would be even more intrusive to change.

### DIFF
--- a/mycpp/control_flow_pass.py
+++ b/mycpp/control_flow_pass.py
@@ -434,7 +434,8 @@ class Build(visitor.SimpleVisitor):
 
     def oils_visit_assignment_stmt(self, o: 'mypy.nodes.AssignmentStmt',
                                    lval: Expression, rval: Expression,
-                                   current_method_name: Optional[str]) -> None:
+                                   current_method_name: Optional[str],
+                                   at_global_scope: bool) -> None:
         cfg = self.current_cfg()
         if cfg:
             lval_names = []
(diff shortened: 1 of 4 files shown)


mycpp/control_flow_pass.py
```diff
@@ -54,7 +54,6 @@ def __init__(self, types: Dict[Expression,
             SymbolPath, pass_state.ControlFlowGraph] = collections.defaultdict(
                 pass_state.ControlFlowGraph)
         self.current_statement_id = INVALID_ID
-        self.current_class_name: Optional[SymbolPath] = None
```
andychu, line 57: Oh that's nice you removed this from the control flow pass!

andychu:
BTW we have this "golden output" repo -- so we can check if the generated files change at all

https://github.com/oils-for-unix/mycpp-output/commits/main/

I checked and this refactoring is a no-op (as expected)

---

Now I remember I also want to move mycpp towards being a typed program

So we can put something in the CI to check for that

andychu:
(It would also be  cool if mycpp could be compiled and thus sped up with mycpp, but there are some reasons that's hard :-) )

Issue #2528, opened 2025-10-28
zeitgeist: Unexpected argument to 'exit'


btertoolen:
Currently osh and bash have different behavior in this case:
```
[bram@bram-dell oils]$ bin/osh /tmp/test
  if [ "$1" = "--help" ]; then exit 0 else
                                      ^~~~
/tmp/test:1: Unexpected argument to 'exit'
```
Bash runs this line without an error (as long as you don't provide an argument). In this specific case this was a bug in the script (a missing `;`), but we seem to encounter this more often.

For now just adding an issue here so we can link it as a cause.

See also:
[#distros > zeitgeist build failure](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/zeitgeist.20build.20failure/with/547576761)
[#osh > break/continue can be dynamic in OSH, static in YSH](https://oilshell.zulipchat.com/#narrow/channel/502349-osh/topic/break.2Fcontinue.20can.20be.20dynamic.20in.20OSH.2C.20static.20in.20YSH/with/544417683)

PR #2537, opened 2025-10-30
add new causes for zeitgeist and dircproxy

btertoolen:
add zeitgeist and dircproxy causes
image

#2528 also got added as a cause to tilda loudmouth and lasem, but they indeed do seem like the same issue.

image


andychu:
Great thanks!  I will make another run

PR #2542, opened 2025-10-31
add cause patterns for jq and libidn

btertoolen:
add causes for jq and libidn


andychu:
Thanks

I guess sometimes we have to use a string specific to the package, and not a "cause" that may appear in other packages

That's fine, as long as it the table helps us organize the bugs

PR #2548, opened 2025-11-03
add known-differences spec tests

btertoolen:
Add spec tests for cases where multiple shells disagree with each other

### DIFF
--- a/spec/known-differences.test.sh
+++ b/spec/known-differences.test.sh
@@ -1,0 +1,28 @@
+## oils_failures_allowed: 0
+## compare_shells: ash dash bash mksh zsh
+#
+# Here we list tests where different shells disagree with each other. For example we
+# encountered some cases where osh and bash agree, but ash doesnt. For alpine / abuild
+# this can cause build failures. So even if we don't directly plan on fixing them (ever)
+# it can still be useful to keep track of these cases. This is also what separates these
+# cases from the cases in the divergence spec tests (we plan on fixing those).
+# The packages where these disagreements were encountered are mentioned after the dash
+
+#### `set` output format - ifupdown-ng
+export FOO=bar
+set | grep bar | head -n 1
+## STDOUT:
+FOO=bar
+## END
+## OK ash/dash STDOUT:
+FOO='bar'
+## END
+## OK zsh STDOUT:
+## END
+
+#### nested function declaration - xcb-util-renderutil
+f() g() { echo 'hi'; }
+## STDOUT:
+## status: 2
+## OK ash/dash/mksh/zsh status: 0
+
--- a/test/spec.sh
+++ b/test/spec.sh
@@ -350,6 +350,10 @@ introspect() {
   test/spec-py.sh run-file introspect "$@"
 }
 
+known-differences() {
+  test/spec-py.sh run-file known-differences "$@"
+}
+
 let() {
   test/spec-py.sh run-file let "$@"
 }


andychu:
Hm looks like zsh is failing for some reason?

https://op.oilshell.org/uuu/github-jobs/10721/ovm-tarball.wwz/_tmp/spec/osh-py/known-differences.html

Does it pass locally?

----

I created  a new `zsh-5.9` wedge which could help?  Although I think zsh 5.1.1 should also work

You can change `compare_shells` from `zsh` to `zsh-5.9` if that makes it easier

----

But this also requires the new wedges which I mentioned on Zulip:


[Instructions for new setup ../oils.DEPS](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Instructions.20for.20new.20setup.20.2E.2E.2Foils.2EDEPS/with/548214479)


```
build/deps.sh rm-oils-crap
build/deps.sh install-wedges
```

btertoolen:
@andychu Yeah the old `zsh` would constantly hang on my system so I wasnt able to test that, but the new wedges seem to make zsh work again so thats very nice! By the way, if the PR is failing feel free to ignore it, I usually check them the next evening

andychu:
OK great, I'm glad the zsh bug is resolved now!  Whenever it's hanging, you can simply change the spec test to zsh-5.9 instead

spec/known-differences.test.sh
```diff
@@ -0,0 +1,36 @@
...
+## OK zsh STDERR:
+grep: (standard input): binary file matches
+## END
+
+#### nested function declaration
+f() g() { echo 'hi'; }
+## STDOUT:
+## OK bash status: 2
+## OK bash STDERR:
+bash: line 1: syntax error near unexpected token `g'
```
andychu: I typically do not assert the exact stderr, because it makes the test specific to a shell

So I think these can simply be removed

It matters when we compare against other shells e.g. with  the compatibility reports:

https://pages.oils.pub/spec-compat/2025-11-02/renamed-tmp/spec/compat/TOP.html

We don't want to write assertions for every shell, so for parse errors, I leave the error text out.  Exit code `2` is generally enough

Sometimes I also redirect `2>err.txt` and then `wc -l err.txt` to see that there IS or is NOT some output

But here that's probably not necessary

---

Also, it is possible that `status: 2` will become the spec, because we agree with bash.  However that depends on how many scripts run into this problem

AND if there is a bug.

If the ONLY uses of the construct are bugs, then it's a strong indication that the spec can disallow it!  

(My long term goal is to argue that OSH is "Bashix" -- a better spec than POSIX, that other shells can follow)

andychu (commented):
This looks good, thank you!

Except let's remove the exact stderr message assertions

spec/known-differences.test.sh
btertoolen: Yes good point about the `status`, I swapped them around now so bash and osh pass, and the rest is set to ok

spec/known-differences.test.sh
```diff
@@ -0,0 +1,27 @@
+## oils_failures_allowed: 0
+## compare_shells: ash dash bash mksh zsh
+#
+# Here we list tests where different shells disagree with each other. For example we
+# encountered some cases where osh and bash agree, but ash doesnt. For alpine / abuild
+# this can cause build failures. So even if we don't directly plan on fixing them (ever)
+# it can still be useful to keep track of these cases. This is also what separates these
+# cases from the cases in the divergence spec tests (we plan on fixing those).
+
+#### `set` output format
```
andychu: actually can you add in the comments for both cases which packages they occur in?

This will help us find things later, when writing the docs

I find it hard to keep track already :-)

Issue #2522, opened 2025-10-27
mycpp incompatibility - python formatting %c only supports integers


EarthenSky:
I ran into a small issue with python formatting. Looks like `%c` only expects an integer, not a single length string.

```py
# works in python, but fails after mycpp
"a char %c" % "a"

# works in python & after mycpp
"a char %c %" ord("a")
```

I haven't dug far into the issue because it was a quick fix, although it would be nice to have a list of these kinds of common "gotchas" somewhere in the oils dev docs.

Example error https://op.oilshell.org/uuu/github-jobs/10523/cpp-spec.wwz/_tmp/spec/osh-cpp/builtin-process.html#details-16-osh-cpp fixed with https://github.com/oils-for-unix/oils/pull/2484/commits/ddebdb9b52d2228a99c5cb3eeab9c4d2dd6d122c

andychu:
Yes, we should define what subset of `%x` strings we support

There is actually a parser in `mycpp/format_strings.py` that used to do this statically

But the problem is we sometimes use dynamic format  strings

andychu:
I think it would make sense to do "best effort" -- most format strings are static, so we can use the `format_strings.py` parser to verify them

PR #2549, opened 2025-11-04
[mycpp] Check format strings are valid

last-genius:
gc_str.cc currently only handles `%%`, `%r`, `%s`, `%d`, `%o` - any other format strings are illegal.

Fixes: #2522

### DIFF
--- a/mycpp/format_strings.py
+++ b/mycpp/format_strings.py
@@ -55,7 +55,7 @@ PAT = re.compile(
     '''
 ([^%]*)
 (?:
-  %([0-9]*)(.)   # optional number, and then character code
+  %-?([0-9]*)(.)   # optional number, and then character code
 )?
 ''', re.VERBOSE)
 


andychu:
Oh I actually think we should disallow this rather than support it

I mentioned that in #2522 , and also if you look at the fix linked:

https://github.com/oils-for-unix/oils/commit/ddebdb9b52d2228a99c5cb3eeab9c4d2dd6d122c

I think that  is actually better

---

I want mycpp to be a relatively small and simple metalanguage, so it can be translated to Rust and such.  This would add more work for other implementers, without much benefit

(Right now there is only ~8K lines of hand-written code  in the whole mycpp runtime)

There are also a ton of other Python format string operations that we don't support -- I would want to disallow this too, e.g. follow the **allowlist** strategy rather than denying specific ones like `%c` etc.

I think we only allow `%r %s %d %o %%` right now

andychu:
And btw I mentioned there that we used to have a format string compiler -- i.e. format strings were static like Rust, not dynamic like C


So if you had

```
"foo %d bar %s" % (a, b)
```

we would translate that unto a function:

```
void fmt99(int a, Str* b) {
  BufWriter b;
  b.write_const("foo ")
  b.format_d(a);
  b.write_const(" bar ")
  b.format_s(b)
  return b.toString();
}

```


https://github.com/oils-for-unix/oils/commit/064e00869eed9ba4456e6a7743c1dab559577718#diff-0991880193e6939147c6a54b0e4f2b904b87d9c7b1c5194fdeeea3e827ada915


But there were 2 problems:

1. sometimes we use dynamic format strings
2. it bloats the code!    the printf interpreter has some wisdom, versus the compiler :-)


However I think we can statically analyze the format strings that are constant, and disallow `%c`

---

A complication is we have **explicit** format strings like:

```
"foo %d" % i
```

and **implicit** ones like:

```
log("foo %d", i)
```

Right now the log() function and others are special cased ... 

We may want to move to a world where the explicit `%` is more common, for the sake of a simple mycpp metalanguage.  And simple static analysis

`log()` may perhaps be the **only** exception

last-genius:
I've disallowed the unimplemented format strings instead.

mycpp/const_pass.py
```diff
@@ -25,6 +26,9 @@
 _ALPHABET = string.ascii_lowercase + string.ascii_uppercase
 _ALPHABET = _ALPHABET[:32]
 
+# Pattern for format strings (%X)
+FORMAT_STRING_PAT = re.compile('[^%]*(?:%-?[0-9]*(.))?', re.VERBOSE)
```
andychu: Hm this is a bit subtle, technically we could include `[dsrO%]` in the regex too

I think it is nicer to take advantage of the re.VERBOSE format and add some comments

I think that finditer() could simply be passed a regex starting with `%` ?

andychu (commented):
Hm I thought we would use the parser in `mycpp/format_strings.py`, which has unit tests in `mycpp/format_strings_test.py`

Does this do the same thing?

oh yeah now I noticed it uses the same regex

```
PAT = re.compile(
    '''
([^%]*)
(?:
  %([0-9]*)(.)   # optional number, and then character code
)?
''', re.VERBOSE)
```

andychu:
Yeah I think I kept that code in case we want to use it again

It doesn't looked used now

So we should decide if we want to keep it or do something else

So far, this looks similar to me

andychu:
Now that I look more, I think it is better to have

1. code with unit tests - which I often use for parsing code like this
2. the re.VERBOSE style

So then `format_strings.py` already has those things

It does some unnecessary stuff now, but checking `char_code` before creating SubstPart shoudl do it

----

It is also nice to say what the wrong code was, e.g. 

> mycpp expected a printf format char in [drso], got 'c'

last-genius:
Moved to using `format_strings.py`.

mycpp/format_strings.py
```diff
@@ -74,9 +74,11 @@ def Parse(fmt: str) -> List[_Part]:
         if char_code:
             if char_code == '%':
                 part: _Part = LiteralPart('%')
-            else:
+            elif char_code in 'rsdo':
                 part = SubstPart(width, char_code, arg_num)
                 arg_num += 1
+            else:
+                raise AssertionError("mycpp expects a format specifier in [rsdo%%], got '%s' instead" % char_code)
```
andychu: Let's use RuntimeError here (or some kind of return value)


I try to use AssertionError only for bad code, not bad data

(although mycpp itself might be an exception to that ... it does tend to sloppily use AssertionError instead of actually reporting errors, so I'm glad we are fixing this)

andychu (commented):
Thanks, one small thing

mycpp/format_strings.py
last-genius: Done

andychu:
Excellent, thank you!

PR #2550, opened 2025-11-04
Fix `test` builtin parsing

last-genius:
Treat unary operator as a word if followed by a binary operator

A construct like `'-a' != 'string'` (when part of a long expression) would
be treated eagerly by the parser: `-a` would be detected as a unary
operator, `!=` as its argument, and `string` would be an unexpected
trailing word.

Instead, look ahead if the next token after a unary operator is a binary
operator. If so, treat the unary operator as a simple word (it does not
need to be quoted like above, `-a != 'string'` still works correctly)
and continue parsing the binary operator's right hand side argument.

Fixes: https://github.com/oils-for-unix/oils/issues/2409

### DIFF
--- a/spec/divergence.test.sh
+++ b/spec/divergence.test.sh
@@ -1,5 +1,5 @@
 ## compare_shells: bash dash mksh zsh ash
-## oils_failures_allowed: 3
+## oils_failures_allowed: 5
 
 # This file relates to:
 #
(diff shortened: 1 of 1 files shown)


last-genius:
One test is [failing](https://op.oilshell.org/uuu/github-jobs/10726/ovm-tarball.wwz/_tmp/spec/osh-py/dbracket.html#details-25-osh):
```
$ bin/osh -c ' [[ -f == ]] || echo false '
   [[ -f == ]] || echo false
                       ^~~~~
[ -c flag ]:1: Expected ]]
```

This is a curious one - before we handled `-f ==` as a unary operator and its argument, now we treat `-f == ]]` as a binary operator and its arguments. Properly preventing this would require unlimited lookahead (to check if `]]` is the closing one of the expression). 

Or, we could hardcode a `if you see DRightBracket as the rhs of binary op AND lhs is unary op, treat it as a unary op` condition, but that's not particularly clean

last-genius:
> Or, we could hardcode a `if you see DRightBracket as the rhs of binary op AND lhs is unary op, treat it as a unary op` condition, but that's not particularly clean

Went with this option for now just as a hacky fix

andychu:
Thanks, I want to look at this PR as well as the issue in the Zulip thread more closely

[wrong parsing in bash 'test'](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/wrong.20parsing.20in.20bash.20'test'/with/553705565)

i.e. when all the shells agree on the unintuitive behavior, that is what OSH should do

---

Also I noticed that none of the newer shells have this bug #2409, so it is OSH with the problem

andychu:
Hm I looked closer at this, and I think the general approach is right:


[test parsing bug #2409 - sqsh, etc.](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/test.20parsing.20bug.20.232409.20-.20sqsh.2C.20etc.2E/with/555510269)


We need to handle binary expressions before unary expressions

And bash has a hack for that too

---

Let me look at the details tomorrow

last-genius:
Rebased on top of `master`

Issue #2547, opened 2025-11-03
esh: osh does not always set `$OPTARG`


btertoolen:
Considering this test script:
```
#/tmp/test
while getopts ':dho:s:V' OPT; do
        echo "Opt: $OPT"
        echo "Opt arg: $OPTARG"
        case "$OPT" in
                d) EVALUATE=no;;
                h) print_help; exit 0;;
                o) OUTPUT="$OPTARG";;
                s) ESH_SHELL="$OPTARG";;
                V) echo "$PROGNAME $VERSION"; exit 0;;
                '?') echo "$PROGNAME: unknown option: -$OPTARG" >&2; exit 10;;
        esac
done
```

You can see `osh` doesnt always set $OPTARG, while `ash` and `bash` do:
```
osh-0.36$ /tmp/test -x
Opt: ?
Opt arg:
: unknown option: -

# bash
[bram@bram-dell oils]$ chmod +x /tmp/test
[bram@bram-dell oils]$ /tmp/test -x
Opt: ?
Opt arg: x
: unknown option: -x

# ash
/tmp/test -x
Opt: ?
Opt arg: x
: unknown option: -x
```

See also [#distros > esh build failure](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/esh.20build.20failure/with/553475616)

PR #2551, opened 2025-11-05
osh optarg handling

daveads:
Addresses this issue... #2547

### DIFF
--- a/builtin/pure_osh.py
+++ b/builtin/pure_osh.py
@@ -484,18 +484,18 @@ class GetOptsState(object):
         """Set OPTARG."""
         state.BuiltinSetString(self.mem, 'OPTARG', optarg)
 
-    def Fail(self, flag_char='', error_msg=''):
+    def Fail(self, flag_char, error_msg=''):
         # type: (str, str) -> None
         """Handle getopts failures.
 
         Silent mode (leading : in optspec) sets OPTARG to flag_char.
         OPTERR=0 disables error messages but doesn't change OPTARG.
         """
-        if self.silent and flag_char:
+        if self.silent and len(flag_char):
             self.SetArg(flag_char)
         else:
             state.BuiltinSetString(self.mem, 'OPTARG', '')
-            if self.opterr != 0 and error_msg:
+            if self.opterr != 0 and len(error_msg):
                 self.errfmt.Print_(error_msg)
 
 
@@ -518,11 +518,11 @@ def _GetOpts(
     #log('current %s', current)
 
     if current is None:  # out of range, etc.
-        my_state.Fail()
+        my_state.Fail('')
         return 1, '?'
 
     if not current.startswith('-') or current == '-':
-        my_state.Fail()
+        my_state.Fail('')
         return 1, '?'
 
     if current == "--": # special case, stop processing remaining args


daveads:
@andychu this seems to be a mycpp issue.

andychu:
Thanks for looking at this!   Can you add a failing test first in `spec/builtin-getopts` ?

Oh yeah and I guess this is the bug from `help getopts`

>     getopts reports errors in one of two ways.  If the first character
>    of OPTSTRING is a colon, getopts uses silent error reporting.

We didn't support that

----

I will look at the C++ errors afterward ... this does appear to be a mycpp problem, since MyPy passes

Not sure exactly why

andychu:
(And btw, I find this leading `:` to be a very confusing special case, since `:` means something else usually!  It means that the flag takes an agument)

daveads:
> Thanks for looking at this! Can you add a failing test first in `spec/builtin-getopts` ?
> 
> Oh yeah and I guess this is the bug from `help getopts`
> 
> > ```
> > getopts reports errors in one of two ways.  If the first character
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > of OPTSTRING is a colon, getopts uses silent error reporting.
> 
> We didn't support that
> 
> I will look at the C++ errors afterward ... this does appear to be a mycpp problem, since MyPy passes
> 
> Not sure exactly why

done

andychu:
Hm there are 4 failures in OSH Python , but only 3 are  allowed

https://op.oilshell.org/uuu/github-jobs/10775/

https://op.oilshell.org/uuu/github-jobs/10775/ovm-tarball.wwz/_tmp/spec/osh-py/builtin-getopts.html

( The `ovm-tarball` task has the Python spec test failures ...  `cpp-spec` is for C++.  It would be nice to move them closer)

spec/builtin-getopts.test.sh
```diff
@@ -389,3 +389,9 @@ err:?
 err:?
 err:?
 ## END
+
+#### getopts silent error reporting
```
andychu: Hm can you describe what this is testing?   e.g. what is "silent error reporting" exactly?

Also should we assert STDERR does not contain an error message?  I think it could be

```
## STDERR:
## END
```

which is different than if you don't have a leading `:` ?

----

I would like to see the two modes compared

from `help getopts`

If the first character     of OPTSTRING is a colon, getopts uses silent error reporting. 

>  In  this mode, no error messages are printed.  

> If an invalid option is seen, getopts places the option character found into OPTARG.  

OK good, we are testing this -- OPTARG=Z  -- let's write a comment about this

> If a required argument is not found, getopts places a ':' into NAME and sets OPTARG to the option character found. 

Can we write a test for this too?

builtin/pure_osh.py
```diff
@@ -573,15 +587,16 @@ def Run(self, cmd_val):
         var_name, var_loc = arg_r.ReadRequired2(
             'requires the name of a variable to set')
 
-        spec = self.spec_cache.get(spec_str)
-        if spec is None:
-            spec = _ParseOptSpec(spec_str)
-            self.spec_cache[spec_str] = spec
+        # OPTERR defaults to 1
+        try:
+            opterr = state.GetInteger(self.mem, 'OPTERR')
```
andychu, line 602: Thanks for adding support for OPTERR

Can you also add a test for it?

builtin/pure_osh.py
```diff
@@ -525,14 +537,16 @@ def _GetOpts(
...
-                              current)
-                tmp = [j8_lite.MaybeShellEncode(a) for a in argv]
-                print_stderr('(getopts argv: %s)' % ' '.join(tmp))
+                if silent:
+                    my_state.SetArg(flag_char)
+                    return 0, ':'
 
-                # Hm doesn't cause status 1?
+                my_state.Fail()
+                if opterr != 0:
```
andychu: Let's add a comment that OPTERR=0 means that getopts disables error messages

(  And also this is different than `silent` mode?  That's weird ...)

andychu:
Hm I also think it makes sense to put `OPTERR` in the GetOptsState() class as well .  That is persisted across invocations

Then we can handle OPTERR in Fail() as well, rather than in `_GetOpts()` ?


(Or maybe we have to read OPTERR every time that Fail() is called?  That is worth a test)

builtin/pure_osh.py
```diff
@@ -489,12 +489,18 @@ def Fail(self):
...
+    argv,  # type: List[str]
+    my_state,  # type: GetOptsState
+    errfmt,  # type: ui.ErrorFormatter
+    opterr,  # type: int
 ):
     # type: (...) -> Tuple[int, str]
+    silent = spec_str.startswith(':')
+    if silent:
+        spec_str = spec_str[1:]
+    spec = _ParseOptSpec(spec_str)
```
andychu, line 513: Hm I also think we should preserve the spec_cache.  Because `getopts` is often called in a loop like this

```
while getopts ':a:b:c:' flag; do
  ...
done
```

that prevents us  from parsing the spec over and over again, on each iteration

So instead of `Dict[str, bool]`, we could have a class, or we could use Zephyr ASDL

----

e.g. in `core/runtime.asdl`, it could be something like

```
GetOptsSpec = (Dict[str, bool] flags, bool silent)
```

And then we can cache that parsing in a dictionary `self.spec_cache` , e.g. `Dict[str, GetOpsSpec]`

----

I'm not sure if you've used Zephyr ASDL yet, but, it is a fairly important part of the codebase.  It lets us define algebraic data types concisely

If you write out the same thing with a Python class, it is more verbose

andychu:
Ah OK I see the error is

```
builtin/pure_osh.py:487 Fail has 2 default arguments.  Only 1 is allowed

mycpp/mycpp_main.py: 1 translation errors (after type checking)
```

https://op.oilshell.org/uuu/github-jobs/10950/cpp-coverage.wwz/_tmp/soil/logs/compare-gcc-clang.txt

---

BTW you can reproduce this locally by building C++

```
./NINJA-config.sh
ninja
```

So you don't have to wait for the CI

andychu:
So then the question is: can either `flag_char` or `error_msg` be **required** ? probably

(This is a C++ translation limitation, because C++ doesn't have optional named args!  We allow just one optional positional arg)

builtin/pure_osh.py
```diff
@@ -482,28 +484,45 @@ def SetArg(self, optarg):
...
+        # type: (str, str) -> None
+        """Handle getopts failures.
+
+        Silent mode (leading : in optspec) sets OPTARG to flag_char.
+        OPTERR=0 disables error messages but doesn't change OPTARG.
+        """
+        if self.silent and len(flag_char):
+            self.SetArg(flag_char)
+        else:
+            state.BuiltinSetString(self.mem, 'OPTARG', '')
```
andychu, line 497: this can be `self.SetArg('')` for consistency

builtin/pure_osh.py
```diff
@@ -436,6 +436,8 @@ def __init__(self, mem, errfmt):
         self.errfmt = errfmt
         self._optind = -1
         self.flag_pos = 1  # position within the arg, public var
+        self.silent = False  # type: bool
+        self.opterr = 1      # type: int
```
andychu, line 440: it would be nice to have some comments next to `self.silent` and `self.opterr`

what do they do?

builtin/pure_osh.py
andychu: Hm this change lost the `self.spec_cache` ... I am going to restore that

Issue #1482, opened 2023-02-03
mycpp should detect break inside tagswitch / switch


andychu:
Because the naive translation will cause a bug in C++

C++ breaks out of the switch, not the loop that encloses the switch, like Python

andychu:
TODO: add this to "invariants not checked" ?  Maybe add a test for this

last-genius:
I wonder if we could make it work in C++ instead of disallowing it?

Rust doesn't have this issue, its switch (`match`) does not need `break`, so it works like in Python: [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=426736dbae0f14a9441950ae7ba9e539)

We could make it work in C++ by labeling the exit from the loop and then translating a `break` in a `switch` to a `goto label`...

last-genius:
Tested quickly - there are no current cases where a `break` is used like this.

Though it is used several times inside a `switch`, but there is a nested inner `for` loop that it's breaking out of, like here:
https://github.com/oils-for-unix/oils/blob/d18d6e397444a6ffdc6b961265f1ec98b5491c80/builtin/func_eggex.py#L101-L111

andychu:
I think the `goto label` may have consequences for the control flow pass (which affects GC)

And maybe exceptions/constructors/destructors.  There are some invariants there too

If there are no instances in the code, I would definitely lean toward disallowing it

PR #2554, opened 2025-11-06
[mycpp] Disallow `break` inside a `switch`

last-genius:
Fixes: #1482

### DIFF
--- a/mycpp/TEST.sh
+++ b/mycpp/TEST.sh
@@ -249,6 +249,9 @@ test-invalid-examples() {
         ;;
       */invalid_switch_break.py)
         expected_status=3
+        ;;      
+      */invalid_format_strings.py)
+        expected_status=3
         ;;
     esac
 
--- a/mycpp/const_pass.py
+++ b/mycpp/const_pass.py
@@ -136,9 +136,16 @@ class Collect(visitor.TypedVisitor):
 
         self.current_file_name: Optional[str] = None
 
+    def verify_format_string(self, fmt: StrExpr) -> None:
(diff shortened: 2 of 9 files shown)


andychu:
Hm I see this is marked "draft", but I took a look

I wonder if makes sense to do this in an earlier pass, and not in cppgen_pass ?  

I realize that this doesn't happen in Rust, so I guess a rustgen_pass doesn't need it

But it is unclear to me now if we should really understand the control flow better, or if this is really just a C++ thing ...

andychu:
Actually even if Rust doesn't have this problem, I think it makes sense to validate this before code generation

Because it concerns "what is the mycpp language"?   That is independent of the back end

We are saying that `break` can't occur within the `with *switch` constructs, regardless of which language it's translated to



---

It seems like the control_flow_pass is  a logical place to put it, since I think it already handles `tagswitch / break` / loops / etc.

But I didn't look at the details

last-genius:
Looking at the CI failures -  apparently I was wrong, there *are* places in the code where we use `break` inside a `switch`!

last-genius:
Moved the check to the control flow pass

andychu:
Thanks!   Now I realize how confusing C/C++ are ...  

I realized that this should probably go in the conversion_pass / "narrowing pass"

But we don't have hooks for it ... I think I want to add `oils_visit_switch`, because that knowledge is used in both the control flow pass and the cppgen_pass

And now the narrowing pass

So the visitor can use some improvement.  That's a step towards our own IR

PR #2567, opened 2025-11-10
Update causes.awk for aports/community

EarthenSky:
New causes
- bmake
- crosstool-ng

 I suspect that a pr modifying `‎regtest/aports-html.sh` requires files to be created on the CI server, so I've omitted it for the time being.

### DIFF
--- a/regtest/aports/cause.awk
+++ b/regtest/aports/cause.awk
@@ -53,6 +53,12 @@ BEGIN {
   # expects and exact code string?
   patterns["#2547"] = "Undefined variable 'OPTARG'"
 
+  # crosstool-ng - $[] is not being treated identical to $(())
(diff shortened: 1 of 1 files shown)


andychu:
Hm this is still marked Draft?  It looks ready

EarthenSky:
oh, okay. I wasn't sure about how the cause-updating process went when one wasn't intending to do a publish a new update. It sounds like just updating `causes.awk` is sufficient?

andychu:
Yeah we should work on that

[Allowing contributors to publish regtest/aports](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/Allowing.20contributors.20to.20publish.20regtest.2Faports/with/548151174)

But for now, as long as you tested locally, just updating `cause.awk` is OK

Issue #2565, opened 2025-11-09
mycpp visitor.py should have visit_oils_*switch() method - switch is part of our IR


andychu:
I realized this after  https://github.com/oils-for-unix/oils/commit/2253fb51cc6e10ea6507391494379af0342534c7

switch is now used for **three**  reasons

1. control_flow_pass.py - - for GC rooting optimization
2. cppgen_pass - to write different C++ for each kind of switch
3. to disallow `break` within switch
    - this is currently in the control flow pass, but now I think it should be in the conversion pass

But the **visitor** needs to know about `switch`, so that **three** passes can use the switch knowledge

This is clearly a part of our IR !

PR #2570, opened 2025-11-10
fix mycpp visitor add visit oils switch

daveads:
add visit_oils_switch() method to visitor base class...  switch statements are now part of the IR.

Attend to this issue #2565

### DIFF
--- a/mycpp/control_flow_pass.py
+++ b/mycpp/control_flow_pass.py
@@ -308,6 +308,20 @@ class Build(visitor.SimpleVisitor):
                 with branch_ctx.AddBranch():
                     self.accept(default_block)
 
+    def oils_visit_switch(self, expr: 'mypy.nodes.CallExpr',
+                          o: 'mypy.nodes.WithStmt',
+                          switch_type: str) -> None:
+        """Build control flow graph for switch statements."""
+        cfg = self.current_cfg()
+        was_inside_loop = self.inside_loop
+        self.inside_switch = True
+        self.inside_loop = False
+        self._handle_switch(expr, o, cfg)
+        self.inside_switch = False
+        # Restore if we were inside a loop
+        if was_inside_loop:
+            self.inside_loop = True
+
     def visit_with_stmt(self, o: 'mypy.nodes.WithStmt') -> None:
         cfg = self.current_cfg()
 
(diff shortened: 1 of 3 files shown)


mycpp/cppgen_pass.py
```diff
@@ -2394,34 +2404,33 @@ def visit_with_stmt(self, o: 'mypy.nodes.WithStmt') -> None:
...
-            self._WriteStrSwitch(expr, o)
-        elif callee_name == 'tagswitch':
-            self._WriteTagSwitch(expr, o)
-        else:
-            assert isinstance(expr, CallExpr), expr
-            self.write_ind('{  // with\n')
-            self.indent += 1
+        if isinstance(expr.callee, NameExpr):
+            callee_name = expr.callee.name
+            if callee_name in ('switch', 'str_switch', 'tagswitch'):
```
andychu, line 2409: Hm it seems like we shouldn't need to duplicate this logic in cppgen_pass?

It's already handled by the base class in `visitor.py` ?

mycpp/control_flow_pass.py
```diff
@@ -319,29 +333,14 @@ def visit_with_stmt(self, o: 'mypy.nodes.WithStmt') -> None:
 
         # Note: we have 'with alloc.ctx_SourceCode'
         #assert isinstance(expr.callee, NameExpr), expr.callee
-        callee_name = expr.callee.name
+        if isinstance(expr.callee, NameExpr):
+            callee_name = expr.callee.name
+            if callee_name in ('switch', 'str_switch', 'tagswitch'):
```
andychu, line 338: ditto here, we shouldn't duplicate the logic

The way I think of it is that `visitor.py` is  creating a **"compiler IR"** for the "mycpp language" !

The IR (intermediate representation) now has `switch`, which Python does **not** have

And then `control_flow_pass` and `cppgen_pass` are **consuming** the IR.  But they shouldn't need to know the details -- that can be hidden


(We might want to move away from the visitor pattern and even away from the MyPy AST -- there are some discussions on Zulip about this)

daveads: link to the discussion ?

daveads: https://oilshell.zulipchat.com/#narrow/channel/410142-yaks/topic/Threads.20on.20the.20direction.20of.20mycpp

Issue #1674, opened 2023-07-13
[mycpp] Crash on statements following a `raise Exception(...)`


PossiblyAShrub:
While working on #1669 I ran into this issue which caused mycpp to crash with a fairly cryptic error message:

```py
# example.py
from __future__ import print_function

def foo():
    # type: () -> None
    raise NotImplementedError()
    print("hello")
```

```
$ _bin/shwrap/mycpp_main ".:pyext" _tmp/mycpp-demo example.py
	mycpp: LOADING example.py
	mycpp pass: CONST
	mycpp pass: FORWARD DECL
	mycpp pass: PROTOTYPES
Traceback (most recent call last):
  File "/home/aolse/oils/oil/mycpp/mycpp_main.py", line 369, in <module>
    sys.exit(main(sys.argv))
  File "/home/aolse/oils/oil/mycpp/mycpp_main.py", line 334, in main
    p3.visit_mypy_file(module)
  File "/home/aolse/oils/oil/mycpp/cppgen_pass.py", line 540, in visit_mypy_file
    self.accept(node)
  File "/home/aolse/oils/oil/mycpp/cppgen_pass.py", line 493, in accept
    node.accept(self)
  File "/home/aolse/wedge/oils-for-unix.org/pkg/mypy/0.780/mypy/nodes.py", line 676, in accept
    return visitor.visit_func_def(self)
  File "/home/aolse/oils/oil/mycpp/cppgen_pass.py", line 2170, in visit_func_def
    self.accept(
  File "/home/aolse/oils/oil/mycpp/cppgen_pass.py", line 493, in accept
    node.accept(self)
  File "/home/aolse/wedge/oils-for-unix.org/pkg/mypy/0.780/mypy/nodes.py", line 1004, in accept
    return visitor.visit_block(self)
  File "/home/aolse/oils/oil/mycpp/cppgen_pass.py", line 2601, in visit_block
    self._write_body(block.body)
  File "/home/aolse/oils/oil/mycpp/cppgen_pass.py", line 1583, in _write_body
    self.accept(stmt)
  File "/home/aolse/oils/oil/mycpp/cppgen_pass.py", line 493, in accept
    node.accept(self)
  File "/home/aolse/wedge/oils-for-unix.org/pkg/mypy/0.780/mypy/nodes.py", line 1019, in accept
    return visitor.visit_expression_stmt(self)
[...]

PR #2571, opened 2025-11-10
[mycpp] Clarify error on unreachable code

last-genius:
Fixes: #1674

### DIFF
--- a/mycpp/TEST.sh
+++ b/mycpp/TEST.sh
@@ -253,6 +253,9 @@ test-invalid-examples() {
       */invalid_format_strings.py)
         expected_status=3
         ;;
+      */invalid_unreachable_code.py)
+        expected_status=4
+        ;;
     esac
 
     if test $status -ne $expected_status; then
--- a/mycpp/examples/invalid_unreachable_code.py
+++ b/mycpp/examples/invalid_unreachable_code.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python2
+"""
+Unreachable code
+"""
+
+def a():
+    # type: () -> None
+    return
+    c = 1
+
+
+def b():
+    # type: () -> None
+    raise AssertionError()
+    d = 1


mycpp/visitor.py
```diff
@@ -737,6 +737,18 @@ def __init__(self, types: Dict[Expression, Type]) -> None:
         SimpleVisitor.__init__(self)
         self.types = types
 
+    def _GetType(self, expr: Expression) -> Type:
+        if expr in self.types:
+            return self.types[expr]
+        else:
+            self.report_error(expr, "Statement is unreachable")
```
andychu: How about "Expression is missing a type: statement may be unreachable"

In case this comes up for some other reason?

---

Also it looks like we have  gone from 28 type errors to 29 ...  I thought it might have been less when I added this


https://op.oilshell.org/uuu/github-jobs/10773/dev-minimal.wwz/_tmp/soil/logs/oils-types.txt

https://op.oilshell.org/uuu/github-jobs/10779/dev-minimal.wwz/_tmp/soil/logs/oils-types.txt

I will add a CI check for this ... I added it to the CI, but it doesn't check right now

last-genius: Renamed, fixed the type error

mycpp/conversion_pass.py
```diff
@@ -311,7 +307,7 @@ def _MaybeAddMember(self, lval: MemberExpr,
 
         if isinstance(lval.expr, NameExpr) and lval.expr.name == 'self':
             #log('    lval.name %s', lval.name)
-            lval_type = self.types[lval]
+            lval_type = self._GetType(lval)
```
andychu, line 310: Hm if we get None, doesn't it just crash in a different place?

We used to get KeyError, but now we will have None, which seems like it will blow up with exceptions

I would have thought we do

```
if lval_type is None:
  return
```

everywhere?

andychu (commented):
Hm I would expect this to push some crash bugs around ... ?  But I guess there is a test, so if it fixes this concrete bug, it is good progress

In an ideal world, our "conversion pass / narrowing pass" would eliminate all the unreachable code

But I don't know how to do that with MyPy, and it might not be a problem if we somehow move away from MyPy

Issue #2569, opened 2025-11-10
[OSH] Bash incompatibility: osh does not set/update `$RANDOM`


kseistrup:
**Oils** 0.36.0 (commit 9440e2f292629bd8d796fdec24887e2a051873ec)

```sh
[osh] ~$ printf '=%s=\n' "$RANDOM"
==
[osh] ~$ RANDOM=123
[osh] ~$ printf '=%s=%s=\n' "$RANDOM" "$RANDOM"
=123=123=
```

----

**Bash** 5.3.3

```
$ printf '=%s=\n' "$RANDOM"
=285=
$ RANDOM=123
$ printf '=%s=%s=\n' "$RANDOM" "$RANDOM"
=2850=18229=
```

----

`bash(1)` reads:

```txt
      RANDOM  Each time this parameter is referenced, it expands  to  a  random
              integer  between  0  and 32767.  Assigning a value to RANDOM ini‐
              tializes (seeds) the sequence of  random  numbers.   Seeding  the
              random number generator with the same constant value produces the
              same  sequence  of values.  If RANDOM is unset, it loses its spe‐
              cial properties, even if it is subsequently reset.
```

last-genius:
This is not POSIX, as far as I can tell, it's a bash feature.

I don't think we've seen test failures in `aports` with this so far, but it seems to be because no package does any analysis of the output of `$RANDOM`....

It's used by three `main` packages and two `community` packages.

In `apache2` package, in particular, our lack of proper `$RANDOM` would mean that the serial number on the certificate generated would always be the same, I think? `rust` relies on `$RANDOM` for a random temp dir, `bitcoin` to get a random TCP port. 

So this might not be strictly a compatibility issue, but an important correctness issue nonetheless.

kseistrup:
> This is not POSIX, as far as I can tell, it's a bash feature.

Correct, that's why I put “**Bash** incompabity” in the title.

Once upon a time, Oils announced itself for its Bash compatibility[^1]. If that is no longer the case, this issue is less important, but it _will_ prevent some Bash scripts from running as intended.

[^1]: E.g., »AMA: I spent 3 years creating a new bash-compatible shell called Oil« (2019)
https://old.reddit.com/r/linux/comments/c1pigw/ama_i_spent_3_years_creating_a_new_bashcompatible/

----

Edit: typo

PR #2574, opened 2025-11-11
[core] Implement $RANDOM

last-genius:
Fixes: #2569

### DIFF
--- a/core/state.py
+++ b/core/state.py
@@ -31,8 +31,8 @@ from frontend import location
 from frontend import match
 from mycpp import mops
 from mycpp import mylib
-from mycpp.mylib import (log, print_stderr, str_switch, tagswitch, iteritems,
-                         NewDict)
+from mycpp.mylib import (log, rand, srand, print_stderr, str_switch, tagswitch,
+                         iteritems, NewDict)
 from pylib import os_path
 
 from libc import HAVE_GLOB_PERIOD
@@ -1385,6 +1385,10 @@ class Mem(object):
         from core import sh_init
         self.env_config = sh_init.EnvConfig(self, defaults)
 
+        self.random_is_special = True
+        # Provide a seed, otherwise libc's rand() is always seeded with 1
+        srand(int(time_.time()) ^ posix.getpid())
+
     def __repr__(self):
         # type: () -> str
         parts = []  # type: List[str]
@@ -1502,7 +1506,7 @@ class Mem(object):
     def SetLocationForExpr(self, blame_loc):
         # type: (loc_t) -> None
         """
-        A more specific fallback location, like the $[ in 
+        A more specific fallback location, like the $[ in
 
             echo $[len(42)]
         """
@@ -1668,7 +1672,7 @@ class Mem(object):
 
     def PushEnvObj(self, bindings):
         # type: (Dict[str, value_t]) -> None
-        """Push "bindings" as the MOST visible part of the ENV Obj 
+        """Push "bindings" as the MOST visible part of the ENV Obj
 
         i.e. first() / propView()
         """
@@ -2071,6 +2075,11 @@ class Mem(object):
                 # set -o nounset; local foo; echo $foo  # It's still undefined!
                 val = value.Undef  # export foo, readonly foo
 
+            # RANDOM=foo seeds the random number generator with foo
+            if lval.name == 'RANDOM' and val.tag() == value_e.Str:
+                str_val = cast(value.Str, val)
+                srand(hash(str_val.s))
+
             cell = Cell(bool(flags & SetExport), bool(flags & SetReadOnly),
                         bool(flags & SetNameref), val)
             var_frame[cell_name] = cell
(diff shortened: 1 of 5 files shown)


spec/builtin-random.test.sh
```diff
@@ -0,0 +1,53 @@
...
+a=$RANDOM
+b=$RANDOM
+[[ $a -ne $b ]] && echo random
+
+## STDOUT:
+set
+random
+## END
+
+#### RANDOM=seed seeds a sequence of random numbers
```
andychu, line 16: OK this is interesting.  It is a case where an assignment has a side effect

Similar to

- `PATH=` - clears the cache (I know you have a PR open for this)
- `LANG=` - calls setlocale()

andychu (commented):
Hm my first reaction is that this is "overspecified" ... we are doing something too much like bash ?

Why not use libc rand() ?   I generally like less code to maintain, and that API is extremely stable, and I believe sufficient for shell `$RANDOM`

---

Thanks for testing the other shells though, that is useful!



But let's talk about it after the `test` PR, since that one is in `aports/main`

last-genius:
The main appeal of our own short pseudo-random generator is that Python's and C++'s seeds result in the same sequences... This is not the case if we use libc rand() in C++ and `random` in Python

Is there a way to specify different expected output for `osh` and `osh-cpp` in spec tests?

core/state.py
```diff
@@ -2469,6 +2487,8 @@ def Unset(self, lval, which_scopes):
             if case(sh_lvalue_e.Var):  # unset x
                 lval = cast(LeftName, UP_lval)
                 var_name = lval.name
+                if var_name == 'RANDOM':
+                    self.random_gen_enabled = False
```
andychu: Hm why do we need this boolean?

spec/builtin-random.test.sh
```diff
@@ -0,0 +1,57 @@
...
+## END
+## OK bash STDOUT:
+846
+21582
+## END
+## OK osh STDOUT:
+6138
+31103
+## END
+## OK oils-cpp STDOUT:
```
andychu: Ah OK the spec test framework doesn't support this now ...

```
ERROR in 'spec/builtin-random.test.sh'
FATAL: Invalid ## line '## OK oils-cpp STDOUT:\n'

```

https://op.oilshell.org/uuu/github-jobs/10854/interactive.wwz/_tmp/soil/logs/interactive-osh.txt



I think we should avoid over-specifying different random values for each shell in the tests

I think it is sufficient to test that it's a number between the expected values?   e.g. 


```
i=$RANDOM
test i -gt 0
test i -lt 32000
```

(or maybe shells don't agree on the max ?)

core/state.py
```diff
@@ -1385,6 +1385,12 @@ def __init__(
         from core import sh_init
         self.env_config = sh_init.EnvConfig(self, defaults)
 
+        self.random_gen_enabled = True
+        # Provide a seed, otherwise libc's rand() is always seeded with 1
+        srand(
+            int(time_.time()) ^ posix.getpid()
```
andychu: how about seeding with just the time and pid ?  that sees good enough, e.g. if you start 100 shells in parallel at almost the same time

core/state.py
last-genius: after `unset RANDOM`, it loses its special properties and no longer resolves to random numbers - this is a reset check

spec/builtin-random.test.sh
last-genius: This way we wouldn't test that setting a seed results in the same sequence of random numbers, seed could just be ignored

andychu: How about a test like this:


```
$SH -c 'RANDOM=seed; echo $RANDOM $RANDOM' > one.txt
$SH -c 'RANDOM=seed; echo $RANDOM $RANDOM' > two.txt
diff -u one.txt two.txt
echo diff=$?  

## STDOUT:
diff=0
## END
```

Then we are testing this property, without overspecifying it

One shell will fail this test, but the other shells will pass

It is nice to see GREEN all across without different assertions for each

For example I started using the pattern

```
if test $? -ne 0; then
  echo ok
fi
```

for non-zero status.  Instead of checking for 1 or 2, because shells differ, but the difference generally doesn't matter

---

These kinds of tests are preferable because then we don't have to write assertions for other shells

We are testing them: https://pages.oils.pub/spec-compat/2025-11-02/renamed-tmp/spec/compat/TOP.html

i.e.  it's nicer if we can  automatically see if other shells support the $RANDOM seed, rather than having to do manual any kind of manual categorization or work

last-genius:
Addressed the comments, but a new `ovm-tarball` failure showed up...

andychu:
Hm weird, I haven't seen this error before ...

https://op.oilshell.org/uuu/github-jobs/10858/ovm-tarball.wwz/_tmp/soil/logs/make-tarball.txt

```
     /home/uke/tmp/Python-2.7.13/Lib/random.py
  File "/home/uke/oil/opy/opy_main.py", line 600, in <module>
    sys.exit(main(sys.argv))
  File "/home/uke/oil/opy/opy_main.py", line 589, in main
    sys.exit(OpyCommandMain(argv[1:]))
  File "/home/uke/oil/opy/opy_main.py", line 421, in OpyCommandMain
    co = compiler.Compile(f, opt, 'exec')
  File "/home/uke/oil/opy/skeleton.py", line 90, in Compile
    return Compile(f, opt, self.gr, *args, **kwargs)
  File "/home/uke/oil/opy/skeleton.py", line 117, in Compile
    parse_tree = driver.PushTokens(p, tokens, gr, start_symbol)
  File "/home/uke/oil/pgen2/driver.py", line 88, in PushTokens
    if p.addtoken(type_, opaque, ilabel):
  File "/home/uke/oil/pgen2/parse.py", line 198, in addtoken
    raise ParseError("bad input", typ, opaque)
pgen2.parse.ParseError
make: *** [build/ref/portable-rules.mk:72: _build/oils-ref/opy-app-deps.txt] Error 1
make: *** Deleting file '_build/oils-ref/opy-app-deps.txt'
```


This is an old compiler trying to parse  Oils code ... the thing before mycpp

Let me see

core/state.py
andychu: OK let's add a comment here than `unset RANDOM` disables the random number generator

we could perhaps call it `self.random_is_special`

core/state.py
```diff
@@ -2071,6 +2075,10 @@ def SetNamed(self, lval, val, which_scopes, flags=0):
                 # set -o nounset; local foo; echo $foo  # It's still undefined!
                 val = value.Undef  # export foo, readonly foo
 
+            if lval.name == 'RANDOM' and val.tag() == value_e.Str:
```
andychu, line 2079: let's add  a comment that `RANDOM=foo` seeds the generator

(I find this feature of shells very non-obvious! )

andychu:
OK I repro'd this problem locally

```
  File "/home/andy/git/oils-for-unix/oils/opy/skeleton.py", line 117, in Compile
    parse_tree = driver.PushTokens(p, tokens, gr, start_symbol)
  File "/home/andy/git/oils-for-unix/oils/pgen2/driver.py", line 88, in PushTokens
    if p.addtoken(type_, opaque, ilabel):
  File "/home/andy/git/oils-for-unix/oils/pgen2/parse.py", line 198, in addtoken
    raise ParseError("bad input", typ, opaque)
pgen2.parse.ParseError
make: *** [build/ref/portable-rules.mk:72: _build/oils-ref/opy-app-deps.txt] Error 1
make: *** Deleting file '_build/oils-ref/opy-app-deps.txt'

```

hm

andychu:
Ah OK, this is because our (old) bytecode compiler doesn't like the Python 2 print statements in `random.py`!

That is a new dependency

Let me see what the best way to handle it is ... one slight issue is that `random.py` pulls in other dependencies

However it's arguable that we don't care so much about the non-C++ tarball anymore ...

PR #2576, opened 2025-11-12
Exit from set -u failures under eval

### LINKED ISSUE #2561: shunit 2 build failure: `eval` with `set -u` exits later in osh than in bash
When referencing an undefined variable while `set -u` is set, osh and bash behave differently.
Not all shells agree on this behavior but I think we will want to agree with bash in this case:
```bash
[bram@bram-dell oils]$ cat /tmp/min
set -u
test_function() {
    x=$1
}

echo "Calling eval test_function"
eval test_function
echo "Eval done, we shouldnt reach here"
[bram@bram-dell oils]$ bash /tmp/min
Calling eval test_function
/tmp/min: line 3: $1: unbound variable
[bram@bram-dell oils]$ bin/osh /tmp/min
Calling eval test_function
      x=$1
        ^~
/tmp/min:3: fatal: Undefined variable '1'
Eval done, we shouldnt reach here
[bram@bram-dell oils]$ ash /tmp/min
Calling eval test_function
/tmp/min: eval: line 3: 1: parameter not set
[bram@bram-dell oils]$ zsh /tmp/min
Calling eval test_function
test_function:1: 1: parameter not set
Eval done, we shouldnt reach here
[bram@bram-dell oils]$ mksh /tmp/min
Calling eval test_function
/tmp/min: 1: parameter not set
Eval done, we shouldnt reach here
[bram@bram-dell oils]$ dash /tmp/min
Calling eval test_function
/tmp/min: 3: eval: 1: parameter not set
```

This behavior causes tests in `shunit2` to fail with the message: `shunit2:ERROR test_boom() returned non-zero return code.`. There are more tests that are failing, but they all could have this cause. But debugging `shunit2` takes quite a bit of effort, so I think the easiest way to go about this is fix this issue and see if that makes all the tests pass. As a nice bonus, if the `shunit2` build passes, I guess that means `shunit2` can be used to unit test osh scripts

EarthenSky:
Close #2561

bash docs for `set -u`: ["An error message will be written to the standard error, and a non-interactive shell will exit."](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)

TODO:
- [x] add spec tests for this issue
- [x] read into exit
  - exits in osh are simply exceptions that get raised (although I'm not sure where they get caught, presumably it's after main_loop.Batch())
- [x] read into eval
  - calls main_loop.Batch() with the provided string as a command.
- [x] ensure that set -u errors raise util.UserExit()
  - we hopefully should be able to fix it by just adding this? Huh, it looks like not all errors are equal...
- [x] ~~figure out how to store / return the exit cause, to decide whether the shell exits or not~~
- [x] clean up the error into its own exception & move the tests out of divergence
- [x] Look into other uses of util.UserExit() and decide whether it seems like the right tool for the job (I suspect yes)
  - Looks like it's mostly just used at the _main level, except for a few places like subshells. It seems to only ever get raised for exit statements, so I doubt any bad side effects will occur.

### DIFF
--- a/core/error.py
+++ b/core/error.py
@@ -174,9 +174,11 @@ class NoUnset(FatalRuntime):
     Is raised in WordEvaluator then caught in CommandEvaluator.
     """
 
-    def __init__(self, exit_status, msg, location):
-        # type: (int, str, loc_t) -> None
-        FatalRuntime.__init__(self, exit_status, msg, location)
+    def __init__(self, msg, location):
+        # type: (str, loc_t) -> None
+
+        # exit status is always 1
+        FatalRuntime.__init__(self, 1, msg, location)
 
 
 class Expr(FatalRuntime):
--- a/osh/word_eval.py
+++ b/osh/word_eval.py
@@ -1447,7 +1447,7 @@ class AbstractWordEvaluator(StringWordEvaluator):
             tok_str = lexer.TokenVal(name_tok)
             name = tok_str[1:] if tok_str.startswith('$') else tok_str
             # same as e_die, except when non-interactive it exits
-            raise error.NoUnset(1, 'Undefined variable %r' % name, name_tok)
+            raise error.NoUnset('Undefined variable %r' % name, name_tok)
 
         return value.Str('')
 


andychu:
Ah sorry for the delay ... just reminding myself, this is wrong in OSH


```
$ osh -c 'set -u; f() { echo $x; }; eval f; echo "should not get here"'
  set -u; f() { echo $x; }; eval f; echo "should not get here"
                     ^~
[ -c flag ]:1: fatal: Undefined variable 'x'
should not get here
```

versus

```
$ bash -c 'set -u; f() { echo $x; }; eval f; echo "should not get here"'
environment: line 1: x: unbound variable

$ dash -c 'set -u; f() { echo $x; }; eval f; echo "should not get here"'
dash: 1: eval: x: parameter not set
```

andychu:
Thank you!  Great fix

It took me awhile to understand this, and I adjusted the tests a bit.  I'll write some more on Zulip

[Notes on set -u / eval bug](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Notes.20on.20set.20-u.20.2F.20eval.20bug/with/558481264)

PR #2541, opened 2025-10-31
Add failing spec test for xtrace and redirect bug

PossiblyAShrub:
This is https://github.com/oils-for-unix/oils/issues/2540

I am currently 2 attempts in. Currently working on a cleaner solution I proposed on Zulip, open to other ideas though.

> I think a cleaner solution would be something a little more sophisticated than the second solution. It should be possible to track the current fd that points to stderr in FdState. We can then have xtrace write to that (or prefer a debug file if set).

---

**Attempt 1**

This was a very hacky fix. This change moved the application of redirects to after we make the xtrace call. The issue is that this split the redirect evaluation and application, which I found to be messy with our model.

<details>
<summary>Click to expand the diff</summary>

```diff
diff --git a/osh/cmd_eval.py b/osh/cmd_eval.py
index fb2496273..b36d7ecdc 100644
--- a/osh/cmd_eval.py
+++ b/osh/cmd_eval.py
@@ -754,7 +754,7 @@ class CommandEvaluator(object):

         raise AssertionError('for -Wreturn-type in C++')

-    def _RunSimpleCommand(self, cmd_val, cmd_st, run_flags):
+    def _RunSimpleCommand(self, cmd_val, cmd_st, run_flags, redirects):
         # type: (cmd_value_t, CommandStatus, int) -> int
         """Private interface to run a simple command (including assignment)."""
         UP_cmd_val = cmd_val
@@ -762,17 +762,44 @@ class CommandEvaluator(object):
             if case(cmd_value_e.Argv):
                 cmd_val = cast(cmd_value.Argv, UP_cmd_val)
                 self.tracer.OnSimpleCommand(cmd_val.argv)
-                return self.shell_ex.RunSimpleCommand(cmd_val, cmd_st,
-                                                      run_flags)

             elif case(cmd_value_e.Assign):
                 cmd_val = cast(cmd_value.Assign, UP_cmd_val)
                 self.tracer.OnAssignBuiltin(cmd_val)
-                return self._RunAssignBuiltin(cmd_val)

             else:
                 raise AssertionError()

+        status = self._EvalAndPushRedirects(redirects)
+        if status != 0:
+            return status
+
+        io_errors = []  # type: List[int]
+        num_redirects = 0
+        if redirects is not None:
+            num_redirects = len(redirects)
+
+        with vm.ctx_Redirect(self.shell_ex, num_redirects, io_errors):
+            UP_cmd_val = cmd_val
+            with tagswitch(UP_cmd_val) as case:
[...]

PR #2577, opened 2025-11-12
[mycpp] POSIX I/O experiment

last-genius:
opening to see how this runs in the CI.

this is a very naive implementation where we read the file byte-by-byte to see how this fares in the benchmarks as the worst possible case.

next step - implement our own buffering (most importantly, it needs to support native `dup`-ing, preserving the buffer and the current seek position)

### DIFF
--- a/mycpp/gc_mylib.cc
+++ b/mycpp/gc_mylib.cc
@@ -114,40 +114,89 @@ LineReader* open(BigStr* path) {
   return reinterpret_cast<LineReader*>(Alloc<CFile>(fd));
 }
 
+void CFile::read_into_buffer() {
+  ssize_t len = 0;
+
+  while (end_ < (BUF_SIZE-1)) {
+    len = ::read(fd_, buffer_+end_, BUF_SIZE-end_);
+
+    if (len < 0) {
+      if (errno == EINTR && iolib::gSignalSafe->PollUntrappedSigInt()) {
+        // Raise KeyboardInterrupt like mylib.Stdin().readline() does in Python!
+        // This affects _PlainPromptInput() in frontend/reader.py.
+        throw Alloc<KeyboardInterrupt>();
+      } else if (errno == EINTR) {
+        continue;
+      } else if (end_ > 0) {
+        eof_ = true;
+        break;
+      } else {
+        throw Alloc<IOError>(errno);
+      }
+    } else if (len == 0) {
+      eof_ = true;
+      break;
+    }
+    end_ += len;
+  }
+}
+
 BigStr* CFile::readline() {
   // Reset errno because we turn the EOF error into empty string (like Python).
   errno = 0;
 
-  size_t i = 0;
-  bool null_terminated = false;
-  for (; i < 131071; i++) {
-    ssize_t len = read(fd_, &(line_[i]), 1);
-    if (len == 0) {
-      line_[i] = '\0';
-      null_terminated = true;
-      break;
-    } else if (line_[i] == '\n') {
-      line_[i + 1] = '\0';
-      i++;
-      null_terminated = true;
-      break;
-    } else if (len < 0) {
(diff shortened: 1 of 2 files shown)


last-genius:
There are still a few correctness issues left that I can't quite figure out, but benchmarks show that without buffering I/O (and reading byte-by-byte) we are definitely slower than `bash`. Depending on the case, this change makes `osh_cpp` from 5% to 300% slower than before

andychu:
Hmm this is a very good experiment!  Thank you

And it highlights that the only thing we really use is `readline()`

Let me see what other shells do for reading a line

I will started a thread on Zulip

last-genius:
Implemented buffering - seems to be mostly functional.

There's this odd error in `benchmarks` though:
```
if test -n "$ac_unrecognized_opts" && test "$enable_option_checking" != no; then
  { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: unrecognized options: $ac_unrecognized_opts" >&5
$as_echo "$as_me: WARNING: unrecognized options: $ac_unrecognized_opts" >&2;}
fi
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            ^~
benchmarks/testdata/configure-coreutils:67955: Couldn't find terminator for here doc that starts here
```

Some of the `mycpp-examples` are failing as well, but the cause is not completely clear to me.

last-genius:
Figured out the `benchmarks` error, only the `examples` one is left now...

last-genius:
Buffering certainly improved performance from the one-byte read implementation, but there's some way to go to get the libc i/o perf. 

Table of bash ratios:
```
|        Case       | Baseline | One-Byte Read | Buffering |
|:-----------------:|:--------:|:-------------:|:---------:|
| abuild-print-help |   1.30   |      3.95     |    1.45   |
| configure.tcc     |   1.04   |      1.24     |    1.04   |
| hello-world       |   2.45   |      2.66     |    2.61   |
```
Buffering benchmarks: https://op.oilshell.org/uuu/github-jobs/10880/benchmarks3.wwz/_tmp/osh-runtime/index.html
One-byte read: https://op.oilshell.org/uuu/github-jobs/10810/benchmarks3.wwz/_tmp/osh-runtime/index.html
Baseline: https://op.oilshell.org/uuu/github-jobs/10895/benchmarks3.wwz/_tmp/osh-runtime/index.html

andychu:
Ah OK interesting ... I think looking at end-to-end benchmarks is good, but we could  also write  some more focused benchmarks

Although maybe first, I'm not sure why we can use our own buffering, but we can't use libc's buffering?

---

BTW buffering may relate to our `io.stdin` API, which we need to figure out what to do with

- #2356 
- #2376

andychu:
Hm, so at a high level, would this lead to a fix for both

- the `ulimit` bug with descriptor 100 versus 10
- the `set -x` and redirects

?  I may have paged the details out of memory ...

last-genius:
> Hm, so at a high level, would this lead to a fix for both
> 
>     * the `ulimit` bug with descriptor 100 versus 10
> 
>     * the `set -x` and redirects
> 
> 
> ? I may have paged the details out of memory ...

`ulimit` has a separate fix, but currently it leads to additional seeking (since on `dup`, we create a new `FILE` and lose the old buffer, need to seek back to our actual position in the file), which was raised as a concern. Having our own buffering implementation would allow us to implement a `CFile.dup()` (instead of `posix.dup()`), which would avoid additional seeking, preserving the buffer while changing the underlying file descriptor - this is not yet implemented but shouldn't be too hard once we agree on the foundations.

`set -x` bug will benefit from the current state of this PR since there's no buffering for writes, so Aidan's `/dev/stderr` fix works in C++ (but doesn't in Python, since it still has buffered writes, it would still require this change: https://github.com/oils-for-unix/oils/pull/2541#issuecomment-3586042155); and if `/dev/stderr` fix is not taken up, it could benefit from the "persistent" fds work for the ulimit bug.

PR #2578, opened 2025-11-13
wget --directory fix and typo

eMBee:
change all cases where `wget` is called with `--directory` as argument instead of `--directory-prefix`. 
as discussed in https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/wget.20options.20incompatibility/with/555228242

fixes also an unrelated typo in the docs that was to small for a PR on its own.


andychu:
Thanks!

Issue #2579, opened 2025-11-13
makepasswd: `osh` doesn't handle longform args in `optargs` properly


btertoolen:
Consider this case:
```
$ cat /tmp/test
while getopts "c:" name; do
        case "$name" in
                c)
                        clean=1
                ?)
                        echo "????"
                        exit $?
                        ;;
        esac
done
shift $((OPTIND - 1))
if [ $# -eq 0 ]; then
        echo "Missing longform argument"
        exit $?
fi
```
When called with a long form argument, for example -- "value", the argument is not supposed to be handled in getopts (since its only set up to handle -c). This way it can be accessed manually later on (this is what the `if $#` statement does).
All shells seem to agree on this behavior, except for osh:
```
[bram@bram-dell oils]$ bash /tmp/test -- "Hello"
[bram@bram-dell oils]$ dash /tmp/test -- "Hello"
[bram@bram-dell oils]$ zsh /tmp/test -- "Hello"
[bram@bram-dell oils]$ mksh /tmp/test -- "Hello"
[bram@bram-dell oils]$ bin/osh /tmp/test -- "Hello"
????
```

Issue #2540, opened 2025-10-31
jq: 2> disagreement with bash with set -x


btertoolen:
When you do `set -x`, in osh the output of `-x` is also written to a file when piping stderr to a file like this:
```
osh-0.36$ set -x
osh-0.36$ printf 'aaaa' > /dev/null 2> /tmp/test_osh
osh-0.36$ set +x
+ set '+x'
osh-0.36$ cat /tmp/test_osh
+ printf aaaa
osh-0.36$
```

While in bash, this isnt the case:
```
[bram@bram-dell oils]$ set -x
[bram@bram-dell oils]$ printf 'aaaa' > /dev/null 2> /tmp/test_bash
+ printf aaaa
[bram@bram-dell oils]$ set +x
+ set +x
[bram@bram-dell oils]$ cat /tmp/test_bash
[bram@bram-dell oils]$
```

Or an even shorter version:
```
$ bash -x -c 'printf 'aaaa' > /dev/null 2> stderr'
+ printf aaaa

$ osh -x -c 'printf 'aaaa' > /dev/null 2> stderr'
```

See also: [#distros > jq build failure](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/jq.20build.20failure/with/548235551)

andychu:
@EarthenSky if you are interested in this, please chat with @PossiblyAShrub (Aidan) about it

I think Aidan might be working on it, although he didn't assign himself this bug

In any case it's good to come to agreement on how the existing code behaves, and possible solutions for changing it.  Often there are many possible solutions

andychu:
And actually I think it makes sense to commit the failing spec test **first**, because that often helps see what the options are

It can give clues to other people

PR #2580, opened 2025-11-13
add cause for #2579 and gphoto2 (#2540)

btertoolen:


spec/divergence.test.sh
```diff
@@ -144,3 +144,22 @@ status=0
...
+                c)
+                        ;;
+                ?)
+                        echo "shouldnt reach here"
+                        ;;
+        esac
+done
+shift $((OPTIND - 1))
+if [ $# -eq 0 ]; then
+  echo "\$# should be 1"
```
andychu: how about  `echo  "\$# = $#` ?  Do any shells disagree on the exact value?  If not that is a simpler test

spec/divergence.test.sh
```diff
@@ -144,3 +144,22 @@ status=0
 141
 141
 ## END
+
+
+#### longform args should be ignored in optargs if not specified (#2579)
```
andychu: hm is this a about --longflags?  Or is it about -- ?

btertoolen: Turns out its about --, which is a special case that osh doesnt seem to handle:
"The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character."

btertoolen:
Looked into implementing a fix for this issue, and as a result i ended up replacing the divergence spec test to multiple smaller tests in builtin-getopts spec tests anyway, so i removed the spec test from this PR

btertoolen:
Also added the gphoto2 cause i found to this since this PR is open anyway

andychu:
Thanks!

PR #2581, opened 2025-11-13
fix xargs_slot with default value

btertoolen:
Without this fix it fails on the unset variable for me:
regtest/aports-run.sh: line 301: XARGS_SLOT: unbound variable


andychu:
Ah I saw your Zulip message before I saw this commit!  I just fixed it, the exact same way :-)

Sorry for the breakage

PR #2582, opened 2025-11-15
Don't escape double quote char in here document

### LINKED ISSUE #2500: mdev-conf - \" should not be escape sequence in here docs
[mdev-conf is failing one test](https://op.oils.pub/aports-build/2025-10-22.wwz/_tmp/aports-report/2025-10-22/disagree-2025-10-15-main/osh-as-sh/log/mdev-conf.log.txt) as follows:

```sh
===> tests/persistent-storage_test:persistent_storage_by_label

...

Standard output:
Executing command [ sh /home/udu/aports/main/mdev-conf/src/mdev-conf-4.8/tests/../persistent-storage ]
Executing command [ readlink disk/by-label/EFI\x20System ]

Standard error:
Fail: incorrect exit status: 1, expected: 0
```

It's not clear yet what the root cause of this is, but I wonder if it's similar to #2416.

EarthenSky:
Close #2500

In `_ReadLikeDQ`, we turn all `Id.Lit_EscapedChar` that are `"` into `Id.Lit_Chars` tokens, rather than escaped chars.

### DIFF
--- a/osh/word_parse.py
+++ b/osh/word_parse.py
@@ -1097,7 +1097,7 @@ class WordParser(WordEmitter):
                     part = word_part.EscapedLiteral(tok,
                                                     ch)  # type: word_part_t
 
-                elif self.token_type == Id.Lit_EscapedDoubleQuote:
+                elif self.token_type == Id.Lit_BackslashDoubleQuote:
                     if left_token:
                         ch = lexer.TokenSliceLeft(tok, 1)
                         part = word_part.EscapedLiteral(tok, ch)


osh/word_parse.py
```diff
@@ -1067,8 +1067,8 @@ def _ReadBashRegexGroup(self):
 
         p_die('Expected word after ( opening bash regex group', self.cur_token)
 
-    def _ReadLikeDQ(self, left_token, is_ysh_expr, out_parts):
-        # type: (Optional[Token], bool, List[word_part_t]) -> None
+    def _ReadLikeDQ(self, left_token, is_ysh_expr, out_parts, is_here_doc=False):
+        # type: (Optional[Token], bool, List[word_part_t], bool) -> None
```
EarthenSky: I notice both `ReadHereDocBody` and `ReadForPlugin` set left_token = None, which is supposed to be set only when a here doc is being parsed.

When is `ReadForPlugin` called? It says "For $PS1, $PS4, etc.", although my understanding is the `PS*` are related to the rhs of the interactive terminal's cursor. I'm not sure how to write a spec test for that (or any interactive feature).

If `ReadForPlugin` should have this same behaviour, I'll remove `is_here_doc` in place of `left_token is None`.

https://github.com/oils-for-unix/oils/blob/e023633e1d0c81722e98d1115701163bcfa12a90/osh/word_parse.py#L2332-L2349

andychu: Hm interesting point ... It is called when rendering the prompt, in `osh/prompt.py`.   The PS1 is a template string for the prompt that is printed.


I tested these 3 cases manually, and currrently OSH behaves like bash

So it appears correct that $PS1 plugins behave more like here docs

---

I'm not sure if we need the extra is_here_doc boolean

I would say we should **lean on the tests** ...  if the tests don't break without the extra boolean, then it's OK

If there's a bug discovered, we can add a test to cover it

---

I think the issue is more with the doc strings/comments than the code



```
$ bash -c 'PS1=\"hi\"; echo $PS1; echo ${PS1@P}'
"hi"
"hi"

$ osh -c 'PS1=\"hi\"; echo $PS1; echo ${PS1@P}'
"hi"
"hi"
```

```
$ bash -c 'PS1=\\"hi\\"; echo $PS1; echo ${PS1@P}'
\hi\
hooveri\
$ osh -c 'PS1=\\"hi\\"; echo $PS1; echo ${PS1@P}'
\hi\
hooveri\
```

```
$ bash -c 'PS1=\\\\"hi\\\\"; echo $PS1; echo ${PS1@P}'
\\hi\\
\hi\
$ osh -c 'PS1=\\\\"hi\\\\"; echo $PS1; echo ${PS1@P}'
\\hi\\
\hi\

```

osh/word_parse.py
```diff
@@ -1094,8 +1094,18 @@ def _ReadLikeDQ(self, left_token, is_ysh_expr, out_parts):
                 if self.token_type == Id.Lit_EscapedChar:
                     tok = self.cur_token
                     ch = lexer.TokenSliceLeft(tok, 1)
-                    part = word_part.EscapedLiteral(tok,
-                                                    ch)  # type: word_part_t
+                    if ch == "\"" and is_here_doc:
```
andychu: We try to follow the rule that the lexer ONLY looks at the raw text

i.e. it does the "differentiating" for the parser and evaluator

I think we should try that strategy in parallel, and se how it looks

EarthenSky: The line is fuzzy, so I'm not entirely sure what counts as "looking at raw text". I'm thinking of 3 approachs:

1. Say we make a new token ESCAPED_DQ equal to `\"` (instead of _DQ_BACKSLASH); now we can just check if the current token has the same ID as ESCAPED_DQ, with the rest of the logic the same. However, it sounds like you were not looking for such a simple change when you said "It will result in a bigger diff".

2. ~~Add a new token ESCAPED_DQ but rather than turn it immediately into Lit_Chars, leave it and convert it back to \" later when the tokens are converted to string.~~ This one seemed like a good idea at first, but now feels really backwards.

3. Make a new lexer rule, completely separate from `LEXER_DEF[lex_mode_e.DQ]` that doesn't treat `\"` as part of ESCAPED_DQ (nor does it have special cases like Id.Right_DoubleQuote). I know you mentioned that it feels odd to only change one part of the rule but keep it the same otherwise [#distros > mdev-conf - \" should not be escape sequence in here docs @ 💬](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/mdev-conf.20-.20.5C.22.20should.20not.20be.20escape.20sequence.20in.20here.20docs/near/547398360) but it seems like a form of decoupling since `_ReadLikeDQ` is getting a bit complex?

I'm going to work on number 3 since it makes the most sense! I have no issue using this branch, because the old commits are still there.

frontend/types.asdl
```diff
@@ -8,6 +8,7 @@ module types {
               generate [integers]
 
   # Fifteen lexer modes for OSH
+  # TODO(GABE): should this be 16 now?
```
EarthenSky: I didn't really understand what this meant, but I counted 15 lines (not 15 tokens) under lex_mode. Not sure if we want `HereDoc` to be on the same line as DQ (or if that even matters)?

andychu:
Hm yeah this isn't quite what I meant - this looks like it is copying a lot of code.

I don't think we need a new lexer mode -- instead, I think we can to  make `\"` its own token type, just like `"` is

That is, `lex_mode_e.DQ` is used for both double quoted strings, and here docs

And it already has a conditional rule for `"` in the PARSER -- I would like to add one for `\"` as well.  Those are the 2 cases where double quoted strings differ from here docs

----

Also I suggested using another branch so we could more easily revert back to the first change, in case this didn't work out

The original change was only changing the evaluator -- I think we should change the lexer, parser, and evaluator, but not copy code

But it was OK since it was small and made tests pass -- probably without the extra `is_here_doc` boolean

---

We could do it in two steps -- commit the first thing, and then see if we can improve it, passing the same tests

EarthenSky:
> Hm yeah this isn't quite what I meant - this looks like it is copying a lot of code.

I see. I'm personally happier with this than the original because separating the here-doc and DQ cases makes the logic a lot clearer to me, but no worries! I'm sure we can get to a better middle groud soon.

> I don't think we need a new lexer mode -- instead, I think we can to make `\"` its own token type, just like `"` is
> ...
> And it already has a conditional rule for `"` in the PARSER -- I would like to add one for `\"` as well. Those are the 2 cases where double quoted strings differ from here docs

I see what you mean. We can modify the DQ lexer rule to remove `\"` from escape characters, just like `"` is removed from lit tokens, then the rest of the logic for what it should be converted into should be similar to the first passing commit.

> The original change was only changing the evaluator -- I think we should change the lexer, parser, and evaluator, but not copy code

What is the difference between the parser and the evaluator in this context? I've been treating the transformation from tokens into `parts` as "parsing." I actually haven't heard about an evaluator before, except for as a synonym for interpreter (like `eval`).

The newest solution I was thinking about modifies `_ReadLikeDQ` and the lexer, but I'm not sure it touches on the parser?

andychu:
> 
> I see what you mean. We can modify the DQ lexer rule to remove " from escape characters, just like " is removed from lit tokens, then the rest of the logic for what it should be converted into should be similar to the first passing commit.

Yes that sounds roughly right

> 
> > The original change was only changing the evaluator -- I think we should change the lexer, parser, and evaluator, but not copy code
> 
> What is the difference between the parser and the evaluator in this context? I've been treating the transformation from tokens into `parts` as "parsing." I actually haven't heard about an evaluator before, except for as a synonym for interpreter (like `eval`).

We take a lot of care to do more work in the parsing stage

Because parsing cost doesn't affect the cost of the running the shell script.   Generally you want to move as many decisions into lexing and parsing as you can.

Because think about a simple loop:

```
for i in {1..10000}; do
  cat << EOF
\" hi \"
EOF
done

```

in this case, a here doc does I/O, so the small difference probably won't matter.

But it is true that any decision that gets made at parse time is done ONCE, while any decision that's done at runtime happens 100,000 times.

So as a general principle, it's very important to be clear about what work you're doing when.

(And although this particular case has a small difference, I bet you could write a synthetic test case to observe it! )


> The newest solution I was thinking about modifies `_ReadLikeDQ` and the lexer, but I'm not sure it touches on the parser?

The lexer creates tokens, so it should create one for `\"`

Then I think the parser will choose `word_part.Literal` or `word_part.EscapedLiteral` for that token, based on whether it's a here doc, or a double-quoted string

----

e.g. this is correct because in double quotes, `\"` is an EscapedLiteral

```
$ osh -n -c 'echo "a \" b"'
(C (w <Lit_Chars echo>) (w (DQ <Lit_Chars "a "> (word_part.EscapedLiteral token:<Lit_EscapedChar "\\\""> ch:"\"") <Lit_Chars " b">)))

```


But this is incorrect, because in her docs, `\"` is NOT an EscapedLiteral -- it's a Literal

```
$ osh --ast-format text -n -c 'cat <<EOF
a \" b
EOF

'
(command.Simple
  blame_tok:(Token id:Lit_Chars length:3 col:0 line:(SourceLine line_num:1 content:"cat <<EOF\n" src:(source.CFlag)))
  more_env:[]
  words:[(CompoundWord parts:[...0x11ce])]
  is_last_cmd:F
  redirects:[
    (Redir
      op:(Token id:Redir_DLess length:2 col:4 line:...0x11d6)
      loc:(redir_loc.Fd fd:0)
      arg:(redir_param.HereDoc
        here_begin:(CompoundWord parts:[(Token id:Lit_Chars length:3 col:6 line:...0x11d6)])
        here_end_tok:(Token id:Undefined_Tok length:4 col:0 line:(SourceLine line_num:3 content:"EOF\n" src:...0x3ffffffc))
        stdin_parts:[(Token id:Lit_Chars length:2 col:0 line:(SourceLine line_num:2 content:"a \\\" b\n" src:...0x3ffffffc)) (word_part.EscapedLiteral token:(Token id:Lit_EscapedChar length:2 col:2 line:...0x1192) ch:"\"") (Token id:Lit_Chars length:3 col:4 line:...0x1192)]
      )
    )
  ]
)

```


(I had to use --ast-format text because there is a bug in the tool when I don't ...)

andychu:
I fixed the debug tool here, so you can examine the syntax tree of here docs:

https://github.com/oils-for-unix/oils/commit/d9b198097f34946416206d1fffb64772dafcce66

Yeah really it should be changed to Literal rather than EscapedLiteral -- that is the intention behind those AST nodes, and they are evaluated correctly

Literals evaluate to themselves, and  EscapedLiteral takes off the `\`


The `word_part_t` type often wraps the `Token` type

EarthenSky:
Thanks for all the description! It helps a lot for understanding how the evaluator behaves, since I haven't looked into it yet. Hopefully this is what you were looking for, although I'm feeling a bit more confident now \:)

osh/word_parse.py
```diff
@@ -1089,13 +1089,23 @@ def _ReadLikeDQ(self, left_token, is_ysh_expr, out_parts):
...
                     part = word_part.EscapedLiteral(tok,
                                                     ch)  # type: word_part_t
+
+                elif self.token_type == Id.Lit_EscapedDoubleQuote:
+                    if left_token:
+                        ch = lexer.TokenSliceLeft(tok, 1)
+                        part = word_part.EscapedLiteral(tok, ch)
+                    else:
+                        # in here docs \" should not be escaped, staying as literal characters
+                        part = Token(Id.Lit_Chars, tok.length, tok.col,
```
andychu: I think it is probably OK to mutate the token here

```
tok.id = Id.Lit_Chars
part = tok    # (*)
```

?  We do that a couple other times in `osh/word_parse.py`

(*) explaining why you can assign a Token to a  word_part_t and have it be `word_part_e.Literal` is an interesting corner of Zephyr ASDL

frontend/lexer_def.py
```diff
@@ -398,7 +398,8 @@ def R(pat, tok_type):
 ] + _BACKSLASH  # These have to come after RegexMeta
 
 LEXER_DEF[lex_mode_e.DQ] = [
-    _DQ_ESCAPED_CHAR,
+    R(r'\\[$`\\]', Id.Lit_EscapedChar),
+    C('\\"', Id.Lit_EscapedDoubleQuote),
```
andychu: Let's call this `Lit_BackslashDoubleQuote` 

Because in the lexer, we don't know YET if the  backslash starts an escape, or if it's literal!

andychu (commented):
Thanks, this looks great, and is what I suspected would be a clean change

It's consistent with the Oils architecture

So the interesting thing is that we only had to change the lexer and the parser -- NOT the evaluator!  That's good!

We should take advantage of more  knowledge at parse time, which is a more static approach than shells sometimes use

e.g. *OSH Parses Shell Scripts Up Front in a Single Pass. Other Shells Don't.* - https://www.oilshell.org/blog/2016/10/13.html

osh/word_parse.py
EarthenSky: Okay, can do! (I also noticed that the function `word_part.Literal()` didn't exist and dug into it enough to realize something funky was implemented here, haha)

https://github.com/oils-for-unix/oils/commit/d628fce3e7b987bd5beefa6ff4515025e287b5d1

frontend/lexer_def.py
EarthenSky: Fair enough!
https://github.com/oils-for-unix/oils/pull/2582/commits/d628fce3e7b987bd5beefa6ff4515025e287b5d1

PR #2583, opened 2025-11-15
Optarg: handle "--" case

### LINKED ISSUE #2579: makepasswd: `osh` doesn't handle longform args in `optargs` properly

btertoolen:
Fixes #2579

POSIX specifies specific behavior when "--" is passed as an argument to getopts, but osh was not handling this case:
"The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character."
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

Furthermore getopts should return `?` when it is processing `"--"` according to

### DIFF
--- a/builtin/pure_osh.py
+++ b/builtin/pure_osh.py
@@ -506,6 +506,10 @@ def _GetOpts(
         my_state.Fail()
         return 1, '?'
 
+    if current == "--": # special case, stop processing remaining args
+        my_state.IncIndex()
+        return 1, '?'
+
     flag_char = current[my_state.flag_pos]
 
     if my_state.flag_pos < len(current) - 1:


spec/builtin-getopts.test.sh
```diff
@@ -389,3 +389,76 @@ err:?
...
+                ?)
+                        echo "?"
+                        ;;
+        esac
+done
+## STDOUT:
+a
+## END
+
+#### getopts sets name to '?' when encountering '--' #2579
```
andychu: I think we can combine these 3 test cases

- it processes the flag -a before --
- it sets name to `?`
- it sets OPTIND

and then the last test case tests with args `-c operand`

So it could be 2 cases: without args after `--`, and with args

And in each case it checks 3 things

andychu:
This looks great, thank you!  Just a minor comment about the tests

btertoolen:
@andychu I think the comment is missing

andychu:
Oh sorry!  I didn't "publish" my code review

spec/builtin-getopts.test.sh
btertoolen: I had one test initially, but then I found out zsh failed it. So then I split it up to make it more clear what exactly zsh does differently. But then I found out that these tests don't run zsh, so it doesn't show up anyway. I tried adding zsh but it was failing a lot of test cases so I ended up leaving it out. Typing it out I guess there is indeed no more reason for these to be split up, so I'll merge it back into one

btertoolen: I split it up into 2 tests, where the second test is more focussed on leaving the remaining args as operands

andychu:
Thanks!

PR #2586, opened 2025-11-19
Implement trap ' ' (SIG_IGN) and trap --ignore 

### LINKED ISSUE #2476: Implement trap '' in OSH and trap --ignore in YSH - should be SIG_IGN
Many details here

[trap '' means SIG_IGN](https://oilshell.zulipchat.com/#narrow/channel/384942-language-design/topic/trap.20''.20means.20SIG_IGN/with/544422056)

daveads:
Implements POSIX-compliant `trap ''` behavior that sets signals to SIG_IGN, and adds a YSH-specific `trap --ignore` flag.

closes #2476

### DIFF
--- a/builtin/trap_osh.py
+++ b/builtin/trap_osh.py
@@ -78,6 +78,14 @@ class TrapState(object):
         self.traps[sig_num] = handler
 
         if handler.tag() == command_e.NoOp:
+            # This is the case:
+            #     trap '' SIGINT SIGWINCH
+            # It's handled the same as removing a trap:
+            #     trap - SIGINT SIGWINCH
+            #
+            # That is, the signal_safe calls are the same.  This seems right
+            # because the shell interpreter itself cares about SIGINT and
+            # SIGWINCH too -- not just user traps.
             if sig_num == SIGINT:
                 self.signal_safe.SetSigIntTrapped(False)
             elif sig_num == SIGWINCH:
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -935,17 +935,24 @@ issues][].
 ### trap
 
 The `trap` builtin lets you run shell code when events happen.  Events are
-signals or interpreter hooks.
+signals or shell interpreter hooks.
+
+These forms print the current `trap` state:
 
     trap -l          # List all events and their number
     trap -p          # Print the current trap state: events and handlers
-    trap CMD EVENT*  # Register handlers
-    trap - EVENT*    # Remove handlers
+
+These forms modify the `trap` state:
+
+    trap CMD EVENT*  # Register handler for the given events
+    trap -   EVENT*  # Remove handler for the given events (SIG_DFL)
+    trap ''  EVENT*  # Do nothing for the given events (SIG_IGN)
 
 Examples:
 
     trap 'echo hi' EXIT INT   # Register
-    trap - EXIT INT           # Remove
+    trap -  EXIT INT          # Remove
+    trap '' EXIT INT          # Ignore
 
 OSH also support legacy syntax, which is not recommended:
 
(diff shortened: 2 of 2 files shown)


daveads:
Screenshot 2025-11-19 at 3 03 56 AM

daveads:
Screenshot 2025-11-27 at 1 19 08 PM

daveads:
@andychu review when free...

spec/builtin-trap.test.sh
```diff
@@ -561,3 +561,63 @@ ok07
 trap-exit
 failure
 ## END
+
+#### trap '' sets handler to empty string (SIG_IGN)
```
andychu, line 565: Hmm I think this test already passed before this change?

The problem is that `''` could mean either

- handle the signal, but run nothing
- don't handle with  SIG_IGN 

the second is the right behavior ...

spec/ysh-builtin-trap.test.sh
```diff
@@ -44,3 +44,32 @@ register
...
+trap -p
+
+## STDOUT:
+trap -- '' SIGINT
+trap -- '' SIGUSR1
+## END
+
+#### trap --ignore removes hooks (like trap -)
+
+trap --ignore EXIT
```
andychu, line 60: hm this test would pass even without this line

builtin/trap_osh.py
```diff
@@ -396,6 +448,10 @@ def Run(self, cmd_val):
 
         arg_r.Next()
 
+        # If first arg is empty string '', ignore the specified signals
+        if first_arg == '':
```
andychu: `len(first_arg) == 0` is slightly faster

builtin/trap_osh.py
```diff
@@ -110,6 +113,20 @@ def _RemoveUserTrap(self, sig_num):
             # shells do.
             iolib.sigaction(sig_num, SIG_DFL)
 
+    def _IgnoreSignal(self, sig_num):
```
andychu: Hm this is very similar to `_RemoveUserTrap()` .... I wonder if we should consolidate them

builtin/trap_osh.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, signal_safe):
         self.signal_safe = signal_safe
         self.hooks = {}  # type: Dict[str, command_t]
         self.traps = {}  # type: Dict[int, command_t]
+        self.ignored_sigs = {}  # type: Dict[int, bool]  # Signals set to SIG_IGN
```
andychu: Hm instead of a separate dict, I wonder if we should reuse `command.NoOp`

That is defined in `frontend/syntax.asdl`, and then right now we use it for empty commands like `>out.txt`

osh/cmd_parse.py:

```
         if len(words) == 0:  # e.g.  >out.txt  # redirect without words
             assert len(redirects) != 0
             if typed_loc is not None:
                 p_die("Unexpected typed args", typed_loc)
             return command.Redirect(command.NoOp, redirects)
```

builtin/trap_osh.py
```diff
@@ -295,6 +312,13 @@ def _PrintState(self):
         # Print in order of signal number
         n = signal_def.MaxSigNumber() + 1
         for sig_num in xrange(n):
+            # Check for explicitly ignored signals
+            if sig_num in self.trap_state.ignored_sigs:
```
andychu: I think we should change `_PrintTrapEntry()` instead

If we use `command.NoOp`, that will be easier I think

builtin/trap_osh.py
```diff
@@ -346,6 +370,31 @@ def _RemoveTheRest(self, arg_r, allow_legacy=True):
             self.trap_state.RemoveItem(parsed_id)
             arg_r.Next()
 
+    def _IgnoreTheRest(self, arg_r):
```
andychu: Hm I think we could use `_AddTheRest()` with `command.NoOp` ... that will simplify the code

andychu (commented):
Thanks for looking at this!

I think using a `command.NoOp` will simplify and shorten the code

e.g.

- `_AddTheRest` 
-  `trap_state.AddItem()`
- `_PrintTrapEntry()`

these can all be reused with that representation, rather than copy and paste

----

I think it also "makes illegal states unrepresentable in the type system"

andychu:
Thanks, is this ready now?  If so please click "re-request review" next to "andychu" on upper right

Then it will appear in my TODO  list

Github has a funny/subtle UI -- it shows that I have made comments, so then it doesn't appear in my TODO unless you "re-request"

builtin/trap_osh.py
```diff
@@ -77,17 +77,25 @@ def _AddUserTrap(self, sig_num, handler):
...
-            #    pyos::InitSignalSafe() calls RegisterSignalInterest(SIGINT),
-            #    then we PollSigInt() in the osh/cmd_eval.py main loop
-            self.signal_safe.SetSigIntTrapped(True)
-        elif sig_num == SIGWINCH:
-            self.signal_safe.SetSigWinchCode(SIGWINCH)
+        if handler.tag() == command_e.NoOp:
+            if sig_num == SIGINT:
+                self.signal_safe.SetSigIntTrapped(False)
+            elif sig_num == SIGWINCH:
+                self.signal_safe.SetSigWinchCode(iolib.UNTRAPPED_SIGWINCH)
```
andychu, line 92: OK the SIGINT and SIGWINCH case are the same as the RemoveTraps() case

I think that is right ... becuase the interpreter still needs to handle SIGINT SIGWINCH

we are not running anything

andychu:
Merged, thank you for going through the iterations ... I like the simpler code a lot better

I'm going to note that in a thread on Zulip -- it's normal to make multiple iterations

PR #2588, opened 2025-11-22
[demo] Add survey of str.find/lastFind/contains

last-genius:


andychu:
Thanks, this is nice!  At least personally I find it helpful to explore what the design space is, and then from the Zulip discussion I think we have a reasonably clear idea of the API

Writing out / fixing the 6 `doc/ref` entries would be a useful next step, and that can also help others give input

Issue #2473, opened 2025-10-14
Check that existing builtins don't take accept args, and use shopt --set strict_arg_parse when other shells accept extra args but shouldn't


EarthenSky:
`osh` may have builtins that can be **more strict** about the number of arguments. For example, the following behaviour is allowed:

```sh
osh-0.36$ shopt -s strict:all
osh-0.36$ dirs
~/github/oils
osh-0.36$ dirs /tmp # possible bug!
~/github/oils
osh-0.36$ cd / /
  cd / /
       ^
[ interactive ]:10: 'cd' got too many arguments
```

While mundane, this is likely still an example of a bug because `dirs` doesn't take any parameters, while the user may think it does. Showing an error like in the `cd` builtin could be beneficial.

andychu:
Yes, `dirs` is a good example!


Actually that is a straight up bug in OSH

```
$ osh -c 'dirs foo'
~/git/oils-for-unix/oils
```


```
$ bash -c 'dirs foo'
bash: line 1: dirs: foo: invalid option
dirs: usage: dirs [-clpv] [+N] [-N]
```

It should **not** be conditional on `strict_arg_parse`  !  It should just be stricter

---

I'll modify this issue a bit

andychu:
Hm I thought we only checked some of the builtins, not all of them?

I think `spec/too-many-args` is a good idea ... but we can also leave it for a new contributor, since it is sort of mechanical

i.e. we should teach other people how to use the spec tests , and how to use the arg parser (thanks for clarifying and ipmroving arg_r.Done() !)

... You showed a great example, and then we can find more people.  (I continue to get e-mails about the job ad)

PR #2589, opened 2025-11-22
[builtin] "too many args" cleanup

last-genius:
First round of work for #2473, cleaned up some of the existing code, fixed `dirs` - haven't audited any other builtins yet.

### DIFF
--- a/core/shell.py
+++ b/core/shell.py
@@ -786,7 +786,7 @@ def Main(
     b[builtin_i.bind] = readline_osh.Bind(readline, errfmt, bindx_cb)
     b[builtin_i.history] = readline_osh.History(readline, sh_files, errfmt,
                                                 mylib.Stdout())
-    b[builtin_i.fc] = readline_osh.Fc(mem, readline, mylib.Stdout())
+    b[builtin_i.fc] = readline_osh.Fc(exec_opts, readline, mylib.Stdout())
 
     # Completion
     spec_builder = completion_osh.SpecBuilder(cmd_ev, parse_ctx, word_ev,
--- a/builtin/readline_osh.py
+++ b/builtin/readline_osh.py
@@ -8,7 +8,7 @@ from _devbuild.gen import arg_types
 from _devbuild.gen.runtime_asdl import cmd_value, scope_e
 from _devbuild.gen.syntax_asdl import loc
 from _devbuild.gen.value_asdl import value, value_e, value_str
-from core import pyutil, state, vm
+from core import pyutil, state, vm, optview
 from core.error import e_usage
 from frontend import flag_util, location
 from mycpp import mops
@@ -382,12 +382,12 @@ class Fc(vm._Builtin):
 
     def __init__(
             self,
-            mem,  # type: Mem
+            exec_opts,  # type: optview.Exec
             readline,  # type: Optional[Readline]
             f,  # type: mylib.Writer
     ):
         # type: (...) -> None
-        self.mem = mem
+        self.exec_opts = exec_opts
         self.readline = readline
         self.f = f  # this hook is for unit testing only
 
@@ -433,7 +433,7 @@ class Fc(vm._Builtin):
             last_index += num_items
         arg_r.Next()
 
-        if self.mem.exec_opts.strict_arg_parse():
+        if self.exec_opts.strict_arg_parse():
             arg_r.Done()
 
         if arg.l:


last-genius:
Audited all the bash builtins

andychu:
Hm it's hard for me to tell which parts are refactoring and which parts are changes

I think we do need tests for this, e.g. it looks like  `pwd` was changed, but does this match other shells?   Which ones ignore the extra args, etc.?

This is definitely the right idea though

last-genius:
> Hm it's hard for me to tell which parts are refactoring and which parts are changes

first two commits are refactorings, rest are changes.

added tests for `pwd`, `pushd`, `fc`

there already were tests for too many arguments for `cd` and `history`

spec/builtin-dirs.test.sh
```diff
@@ -241,3 +241,21 @@ pushd . >/dev/null
 dirs
 ## stdout: /tmp /tmp
 ## status: 0
+
+#### pushd does not take more than one argument
+pushd . . >/dev/null || echo too many args!
+## N-I zsh STDOUT:
+## END
+## OK osh/bash STDOUT:
```
andychu: this should be the green behavior, but OK is yellow

So remove `OK osh/bash`

spec/builtin-cd.test.sh
```diff
@@ -507,3 +507,12 @@ pwd
 ## N-I bash STDOUT:
 fail
 ## END
+
+#### pwd doesn't accept any args
+shopt -s strict_arg_parse || true
+pwd / >/dev/null || echo 'too many args!'
+## N-I bash/dash/ash STDOUT:
+## END
+## OK osh/mksh/zsh STDOUT:
```
andychu: we should have a green behavior here

And I think we need two tests:

- how does OSH `pwd` compare with `pwd` in other shells?  without `shopt -s strict_arg_parse`
- and then test OSH with `strict_arg_parse`

That's the same philosophy we (@EarthenSky ) followed with `cd` in the `regtest/aports` bug

- bash `cd` doesn't accept extra args
- ash DOES, and we need to be compatible with that behavior in order to run `regtest/aports`
- but OSH also adds `strict_arg_parse`

So the tests should express what the situation is with `pwd`

We should aim for "green rows" as much as possible

spec/builtin-fc.test.sh
```diff
@@ -245,3 +245,11 @@ case $SH in bash) echo '^D' ;; esac
 2	 echo 1
 ^D
 ## END
+
+#### fc with too many args
+fc -l 0 1 2 || echo too many args!
+## STDOUT:
+too many args!
+## END
+## N-I bash STDOUT:
```
andychu, line 259: Why does bash differ here?

For `fc`, should we hide the check behind `strict_arg_parse` to have bash compatibility?

last-genius:
Addressed the comments and discovered a dormant null dereference in `fc` - I guess we weren't memory-safe before :)

took a while to figure out the right fix, added a test as well.

alternatively, we could just check for None:
```diff
diff --git a/builtin/readline_osh.py b/builtin/readline_osh.py
index a04945cad..a137a05aa 100644
--- a/builtin/readline_osh.py
+++ b/builtin/readline_osh.py
@@ -461,10 +461,11 @@ class Fc(vm._Builtin):

             while i != limit:
                 item = readline.get_history_item(i)
-                if arg.n:
-                    self.f.write('\t %s\n' % (item))
-                else:
-                    self.f.write('%d\t %s\n' % (i, item))
+                if item is not None:
+                    if arg.n:
+                        self.f.write('\t %s\n' % (item))
+                    else:
+                        self.f.write('%d\t %s\n' % (i, item))
                 i += step

             return 0
```

core/shell.py
```diff
@@ -786,7 +786,7 @@ def Main(
     b[builtin_i.bind] = readline_osh.Bind(readline, errfmt, bindx_cb)
     b[builtin_i.history] = readline_osh.History(readline, sh_files, errfmt,
                                                 mylib.Stdout())
-    b[builtin_i.fc] = readline_osh.Fc(readline, mylib.Stdout())
+    b[builtin_i.fc] = readline_osh.Fc(mem, readline, mylib.Stdout())
```
andychu: Ah please pass `exec_opts` instead of `mem`

we try to pass the "minimal deps" , i.e. I think `fc` will never use memory/variables, but it now uses `exec_opts`

andychu (commented):
Hm thanks for fixing the `fc` issue

Hm so yeah mycpp doesn't protect against dereferencing NULL, so our generated code is not memory safe for that reason

We could turn on strict optional checking in MyPy -- there is a thread about that somewhere.  

But we can also document it in `doc/portability.md` .  i.e. we assume that the hardware platform traps NULL dereference (C does not provide that assumption)

That seems to be universally true on platforms that run Unix.  It seems only embedded platforms without an OS let you dereference NULL, leading undefined behavior

https://claude.ai/share/a838ed7f-50a1-4bdd-b466-9b747d1124e1


I'll make a note about it on Zulip

core/shell.py
last-genius: Done!

andychu:
Merged, thanks!

andychu:
Actually I was thinking that tactically speaking, we could have `spec/too-many-args`

And then just test all the builtins from A to Z

then it would be easier to see which ones we've fixed and which we haven't

and compare the "green rows" against bash and other shells

I often use the spec tests  as TODO list, and that may make it more convenient

PR #2590, opened 2025-11-23
Implement str.find/findLast/contains/strcmp

last-genius:


doc/ref/chap-builtin-func.md
```diff
@@ -160,7 +160,7 @@ Given a List of integer byte values, return a string.
 
 ### strcmp()
 
-TODO
+Returns 0 if s1 and s2 are equal, -1 if s1 is less than s2, and 1 if s1 is greater than s2.
```
andychu, line 164: How about adding something like "Compares 2 strings, by byte order"  or "lexicographically"

or using "lexographic order on bytes"

https://en.wikipedia.org/wiki/Lexicographic_order

doc/ref/chap-type-method.md
```diff
@@ -141,19 +141,45 @@ An `Obj` instance representing the string type.  Let's use this definition:
...
+
+If the substring is not found, returns -1.
 
 ```raw
 var i = mystr.find('y')
+var i = mystr.find('y', start=0, end=2)
 ```
 
-This is similar to
+### lastFind()
```
andychu: should be findLast

spec/ysh-string.test.sh
```diff
@@ -540,3 +540,101 @@ double
 ## status: 2
 ## stdout-json: ""
 
+#### str.find()
```
andychu, line 543: Please add findLast() to the title

or it could be in a separate spec test, to make it easier to read

spec/ysh-string.test.sh
```diff
@@ -540,3 +540,101 @@ double
...
+
+var mystr = 'abcaz'
+
+echo $[mystr.find('b')]
```
andychu: It would be easier to read to use the assert builtin in YSH now

e.g.

assert [1 === mystr.find('b')]
...

And then assert status: 0

builtin/method_str.py
```diff
@@ -323,6 +323,55 @@ def Call(self, rd):
...
+
+    def __init__(self):
+        # type: () -> None
+        pass
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+        string = rd.PosStr()
+
+        substr = rd.PosValue()
```
andychu: Hm this can just be PosStr()

builtin/method_str.py
```diff
@@ -323,6 +323,55 @@ def Call(self, rd):
...
+
+    def __init__(self, direction):
+        # type: (int) -> None
+        self.direction = direction
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+        string = rd.PosStr()
+
+        substr = rd.PosValue()
```
andychu: ditto, I think PosStr() should be fine

core/shell.py
```diff
@@ -826,8 +826,12 @@ def Main(
         'split': method_str.Split(),
         'lines': method_str.Lines(),
 
-        # finds a substring, optional position to start at
-        'find': None,
+        # finds a substring, optional named parameters specifying the slice
+        # of the string to search in - [start:end]
+        'find': method_str.Find(method_str.START),
+        'lastFind': method_str.Find(method_str.END),
```
andychu: should be findLast()

last-genius:
Addressed the comments

andychu:
Looks great!   One last change: in `doc/ref/toc-ysh.md`, please remove the X before each of the 4 topic names (and preserve space)

That turns into a red X, which is basically our TODO list

This change could be a very nice example commit that can go in the dev guide, for new contributors!

andychu:
Great thank you!

----

It occured to me that we are using slightly more STL  functions now ...  there actually is a rationale for moving away from dynamically linking `libstdc++`

```
$ ldd $(which osh)
        linux-vdso.so.1 (0x00007ffc879a7000)
        libreadline.so.8 => /lib/x86_64-linux-gnu/libreadline.so.8 (0x00007f8c0c554000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f8c0be00000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f8c0c474000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f8c0c454000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8c0c01f000)
        libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007f8c0bdcd000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f8c0c5cd000)

```

Most shells don't dynamicallyi link `libstdc++`, and I believe it actually slows process creation time, and shell startup time

i.e. it took awhile to get `benchmarks/osh-runtime` matching other shells -- it's a surprisingly constrained problem

----

We don't need to do anything about this right now, but just a heads up that we may want to reduce STL usage

This is actually exactly what fish did before it was ported to Rust -- it was written in C++, **without** STL usage

I think we're only using a tiny subset of `vector deque` and then a few other things like `std::min,max`, which are possible to replace

----

I think a "bootstrap-friendly" Oils would have no more dynamic dependencies than traditional shells in C

(And then for people who want more dependencies, the hypothetical Rust version could be what they build on)

PR #2592, opened 2025-11-25
add megacmd cause

btertoolen:
The pattern is very specific to the megacmd package, but im afraid it will match false positives otherwise


andychu:
Hm yeah I wonder what the cause of this one is ... it would be worth reducing and seeing what other shells do

Issue #2594, opened 2025-11-26
libnbd 'kill' can't translate number '134' to a name


btertoolen:
In bash we can pass exit codes of processes to `kill -l` and they get resolved to the signal that was used to kill the process (the exit code will be 128 + the number of the signal used to kill the process).

For example in ash and bash you can do:
```
$ kill -l 134 # ash
ABRT
$ bash -c "kill -l 134"
ABRT
```
But in osh this fails. 
A test in the libnbd package relies on this behavior, causing a test to fail when running it with osh

PR #2595, opened 2025-11-26
Kill list

btertoolen:
Handle the cases for kill -l 0 and kill -l >128 (process exit codes).
This should resolve issue #2594

### DIFF
--- a/doc/ref/chap-builtin-cmd.md
+++ b/doc/ref/chap-builtin-cmd.md
@@ -1337,6 +1337,7 @@ Examples:
     kill -l                # List all signals; -l is a synonym for -L
     kill -L USR1 USR2      # prints '10 12'
     kill -L USR1 15        # prints '10 TERM'
+    kill -L 134            # you can also pass exit codes, this prints 'ABRT'
 
 ## External
 


builtin/process_osh.py
```diff
@@ -754,11 +754,17 @@ def _ParseSignal(self, sig_str, blame_loc):
...
-        # type: (args.Reader) -> int
+        # type: (args.Reader) -> None
         while not arg_r.AtEnd():
             arg, arg_loc = arg_r.Peek2()
             if arg.isdigit():
-                sig_name = signal_def.GetName(int(arg))
+                sig_num = int(arg)
+                if sig_num == 0:
+                    print("EXIT") # special case, this is not really a signal
+                    return
```
andychu: Hm I think there is a bug here - it should continue and not return

```
$ kill -l 0 INT
EXIT
2
```

Please update the spec test to fail first, and see if that's true

btertoolen: Good catch, it should be continue not return. Added a spec test as well

builtin/process_osh.py
```diff
@@ -754,11 +754,18 @@ def _ParseSignal(self, sig_str, blame_loc):
...
-        # type: (args.Reader) -> int
+        # type: (args.Reader) -> None
         while not arg_r.AtEnd():
             arg, arg_loc = arg_r.Peek2()
             if arg.isdigit():
-                sig_name = signal_def.GetName(int(arg))
+                sig_num = int(arg)
+                if sig_num == 0:
+                    print("EXIT") # special case, this is not really a signal
+                    arg_r.Next()
```
andychu: Hm now I see the arg_r.Next() is duplicated

I think we could refactor this into a wrapper `_NumberToName()`, called from the `isdigit()` case.  Then we can use `return`

There is actually a similar function in `builtin/trap_osh.py` - `_ParseSignalOrHook`.  But it does something slightly different, because it always returns a string.

----

Now I see it catches `ValueError` from `int()`, which we should do too.   That's what happens on overflow in mycpp

So yeah maybe we can copy the beginning of that function into a new function next to Kill()

```
$ osh -c 'kill -l 123456789012345123123123123'
terminate called after throwing an instance of 'ValueError*'
Aborted
```

There is also this behavior, which confuses me.   I'd say we shouldn't bother unless a real script needs it to run

```
$ kill -l DEBUG RETURN EXIT
65
67
0
```

builtin/process_osh.py
```diff
@@ -754,11 +754,18 @@ def _ParseSignal(self, sig_str, blame_loc):
         return sig_num
 
     def _TranslateSignals(self, arg_r):
-        # type: (args.Reader) -> int
+        # type: (args.Reader) -> None
         while not arg_r.AtEnd():
```
andychu, line 787: Actually maybe a better structure is to have a loop that does 

```
while ...
  print(_TranslateSignal(arg))
  arg_r.Next()
```

and then _TranslateSignal() can return a string, and it can also return early

and then it's easy to see that we are doing exactly one thing for each arg -- the loop will be ~3 lines or so

andychu:
Merged, thanks

PR #2597, opened 2025-11-28
add -o index=off to mount commands (regtest/aports-run.sh & regtest/aports-setup.sh)

btertoolen:
Fixes this error:
```
bram@bram-dell /p/o/oils (master)> INTERACTIVE=1 regtest/aports-run.sh build-package-overlayfs osh-as-sh megacmd community
mount: /prj/oils-for-unix/oils/_chroot/package-slot99.overlay/merged: fsconfig() failed: Stale file handle.
       dmesg(1) may have more information after failed mount system call.
```

Also discussed here (a while back already): [#distros > setting up regtest/aports-setup.sh](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/setting.20up.20regtest.2Faports-setup.2Esh/with/544318771)

### DIFF
--- a/regtest/aports-setup.sh
+++ b/regtest/aports-setup.sh
@@ -370,9 +370,12 @@ create-osh-overlay() {
 
   mkdir -v -p $osh_overlay/{merged,work,layer}
 
+  # -o index=off fixes this error: fsconfig() failed: Stale file handle
+  # See also https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/setting.20up.20regtest.2Faports-setup.2Esh/with/544318771
   sudo mount \
     -t overlay \
     osh-as-sh \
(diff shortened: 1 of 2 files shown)


regtest/aports-setup.sh
```diff
@@ -373,6 +373,7 @@ create-osh-overlay() {
   sudo mount \
     -t overlay \
     osh-as-sh \
+    -o index=off \
```
andychu, line 378: OK thanks, in both cases, please add a short comment about what -o index=off`  avoids ... and possibly a link to Zulip


(I wish we could use OSH multi-line commands, which allow comments on each line ... but we have to put it before `sudo mount` here ...)

andychu:
Merged, thanks!

Issue #2488, opened 2025-10-19
Help fill out YSH stdlib functions and methods


andychu:
The ones we need help with are labeled with a red X on the Oils reference

https://oils.pub/release/0.36.0/doc/ref/toc-ysh.html

- find, findLast
- contains
- Dict clear accum update
- time strftime 
- runes encodeRunes
- bytes encodeBytes
- strcmp
- ...

Some of them definitely require **design**, so we need to discuss on Zulip first

---

Taking a survey of Python and JavaScript like this, to help us decide what to do, is great work!

```
$ ls */survey*
demo/survey-arith.sh      demo/survey-closure.rb  demo/survey-list.sh  demo/survey-pretty-print.sh  demo/survey-str-api.sh
demo/survey-case-fold.sh  demo/survey-closure.sh  demo/survey-loop.sh  demo/survey-static-names.sh
```

i.e. we don't just value coding, we also value design

EarthenSky:
Is there any more work still to be done regarding this issue? I see a lot of functions with Xs.

It looks like on zulip that
- Andriy is working on `str.find`/`findLast`/`contains` with https://github.com/oils-for-unix/oils/pull/2588

If nobody is working on them, I could look into the 4 (5?) container functions
- X includes()
- X Dict/clear()
- X accum()
- X update()

andychu:
Yes thanks, I'd start with `clear()` -- that's the simplest one ...  and probably `update()`

- maybe `includes()` is up for grabs too
- we're discussing accum() on Zulip

I found the **first step** is to check the `doc/ref` entry for each one, and make sure it makes sense!   And make sure **error** cases are documented

----

If anything doesn't make sense, then please start a thread on Zulip, so we can have a design discussion

Generally surveying Python/JS helps move us in the right direction

last-genius:
I've got implementations for Dict.clear/accum/inc/update - was discussing these on Zulip, feel free to grab anything else

PR #2598, opened 2025-11-29
Implement `dict->{clear,add,append}()`

last-genius:
Least ambiguous part of methods on dict from #2488

### DIFF
--- a/builtin/method_dict.py
+++ b/builtin/method_dict.py
@@ -93,28 +93,24 @@ class Add(vm._Callable):
 
         dictionary = rd.PosDict()
         key = rd.PosStr()
-        inc_val = rd.OptionalValue()
+        inc_val = rd.PosValue()
         rd.Done()
 
-        # If inc_val is provided, we treat it as the definitive type of the
-        # value to be incremented, if it isn't, we rely on the existing
-        # value in the dictionary (if that doesn't exist, the default is 0)
-        if inc_val is None:
-            right = value.Int(mops.ONE)  # type: value_t
-        else:
-            right = inc_val
-
+        # default value for the dictionary entry is zero
         if key in dictionary:
             left = dictionary[key]
         else:
             left = value.Int(mops.ZERO)
 
-        c, i1, i2, f1, f2 = expr_eval.ConvertForBinaryOp(left, right)
+        c, i1, i2, f1, f2 = expr_eval.ConvertForBinaryOp(left, inc_val)
 
         if c == coerced_e.Int:
             res = value.Int(mops.Add(i1, i2))  # type: value_t
         elif c == coerced_e.Float:
             res = value.Float(f1 + f2)
+        else:
+            raise error.TypeErr(left, 'add() expected the key to point to Int or float, got',
+                                rd.BlamePos())
 
         dictionary[key] = res
 
--- a/doc/ref/chap-type-method.md
+++ b/doc/ref/chap-type-method.md
@@ -676,26 +676,11 @@ Adds an increment specified to the value:
     call mydict->add('b', 2.14)
     = mydict # => (Dict)   {a: 3, b: 5.28}
 
-If the increment is not specified, its default value is 1 of the type of
-the value with the provided key:
-
-    call mydict->add('a')
-    = mydict # => (Dict)   {a: 4, b: 5.28}
-
-    call mydict->add('b')
-    = mydict # => (Dict)   {a: 4, b: 6.28}
-
 If there is no value in the dictionary with the provided key, its default
 value is 0 of the type of the increment (int or float):
 
     call mydict->add('c', 2.14)
     = mydict # => (Dict)   {a: 4, b: 6.28, c: 2.14}
-
-If the increment was not provided and the value with the provided key does not
-exist, then the result of the operation is an integer 1:
-
-    call mydict->add('d')
-    = mydict # => (Dict)   {a: 4, b: 6.28, c: 2.14, d: 1}
 
 Similar:
 
--- a/spec/ysh-methods.test.sh
+++ b/spec/ysh-methods.test.sh
@@ -579,20 +579,17 @@ pp test_ (book => get("author", ""))
 
 #### Dict -> add() with ints
 var histogram = {a: 1979}
-call histogram->add('a')
-assert [{"a": 1980} === histogram]
 call histogram->add('a', 3)
-assert [{"a": 1983} === histogram]
-call histogram->add('b')
-assert [{"a": 1983, "b": 1} === histogram]
-call histogram->add('c', 3)
-assert [{"a": 1983, "b": 1, "c": 3} === histogram]
-## status: 0
+assert [{"a": 1982} === histogram]
+call histogram->add('b', 3)
+assert [{"a": 1982, "b": 3} === histogram]
+
+var d = {k: 'foo'}
+call d->add('k', 42)
+## status: 3
 
 #### Dict -> add() with floats
 var histogram = {a: 2.5}
-call histogram->add('a')
-pp test_ (histogram)
 call histogram->add('a', 2)
 pp test_ (histogram)
 call histogram->add('a', 3.5)
@@ -605,15 +602,17 @@ call histogram->add('c', 1)
 pp test_ (histogram)
 call histogram->add('c', 3.0)
 pp test_ (histogram)
-## status: 0
+
+var d = {k: 'foo'}
+call d->add('k', 42.14)
+## status: 3
 ## STDOUT:
-(Dict)   {"a":3.5}
-(Dict)   {"a":5.5}
-(Dict)   {"a":9.0}
-(Dict)   {"a":9.0,"b":3.5}
-(Dict)   {"a":9.0,"b":5.5}
-(Dict)   {"a":9.0,"b":5.5,"c":1}
-(Dict)   {"a":9.0,"b":5.5,"c":4.0}
+(Dict)   {"a":4.5}
+(Dict)   {"a":8.0}
+(Dict)   {"a":8.0,"b":3.5}
+(Dict)   {"a":8.0,"b":5.5}
+(Dict)   {"a":8.0,"b":5.5,"c":1}
+(Dict)   {"a":8.0,"b":5.5,"c":4.0}
 ## END
 
 #### Dict -> append()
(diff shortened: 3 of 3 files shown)


andychu:
Hm I peeked at this ... I think the `setvar` should be a separate change we should discuss

It would be inconsistent if `setvar d[k] += 1` works, but `setvar i += 1` doesn't

And then there's also `*=` -- does that have a default of 1?   

Oh yeah, actually I think this change is wrong because of that ... only the defaults for `+=` and `-=` would be zero.  For the other ops, it doesn't make sense.  

----

This is another case where writing the docs first is good.  i.e. writing docs for `setvar` may surface these types of design errors

last-genius:
Damn, you're right! Dropped the `setvar +=` part

andychu:
Hm I'm not sure if this is ready for review, I think I peeked a little early

I just updated this wiki page  - https://github.com/oils-for-unix/oils/wiki/Making-Pull-Requests

I want to use the Github "request review" UI, and then also confirm after pushing that it's ready for review again

last-genius:
It is ready for review - I've undrafted it yesterday

andychu:
OK, please assign me as "reviewer" in the upper right ... this is a new thing

doc/ref/chap-type-method.md
```diff
@@ -641,6 +641,20 @@ TODO:
 call mydict->accum('key', 'string to append')
 ```
 
+### inc()
+
+Increments the value by the increment specified. If there is no value in the
+dictionary with the provided key, its default value is 0 of the type of the
+increment (int or float). If the increment is not specified, its default value
+is 1 of the type of the value with the provided key. If neither the increment
```
andychu: I think this is better illustrated by examples that the user can run

I know we are not consistent with that yet, but we should move toward it

And actually I started it with

    build/doc.sh run-code-all

That actually runs the code examples in `chap-type-method` that are NOT `raw`

So we should change these to 4 space indents, which means they are runnable code blocks

And then illustrate the int and float cases separately, what happens when the arg is missing, when the key is missing, etc.

doc/ref/chap-type-method.md
```diff
@@ -649,8 +663,6 @@ setvar mydict['k'] += 3  # TODO: default value of 0
 
 ### Dict/clear()
 
-TODO:
-
 Remove all entries from the Dict:
```
andychu, line 694: Let's also improve this entry, with something like

```
var d = {key: 'val'}
= len(d)  # => 1
call d->clear()
= len(d)  # => 0
```

So the user has something to verify

It makes the examples more readable

builtin/method_dict.py
```diff
@@ -64,6 +66,75 @@ def Call(self, rd):
...
+
+    def Call(self, rd):
+        # type: (typed_args.Reader) -> value_t
+
+        dictionary = rd.PosDict()
+        key = rd.PosStr()
+        inc_val = rd.OptionalValue()
+        rd.Done()
+
+        # If inc_val is provided, we treat it as the definitive type of the
```
andychu: Hm interesting, I am thinking if we can somehow share code with the `+` operator ... e.g. with the int float dispatch

doc/ref/toc-ysh.md
```diff
@@ -58,7 +58,7 @@ error handling, and more.
                                indexOf()        lastIndexOf()   X includes()
                                insert()         remove()
                                reverse()        List/clear()
-                   Dict        erase()        X Dict/clear()    X accum()
+                   Dict        erase()          Dict/clear()    X accum()
```
andychu: We should also add a link to `inc()`

spec/ysh-methods.test.sh
```diff
@@ -552,6 +552,19 @@ pp test_ (book)
 (Dict)   {"title":"The Histories"}
 ## END
 
+#### Dict -> clear()
+var book = {title: "The Histories", author: "Herodotus"}
+call book->clear()
+pp test_ (book)
+# confirm method is idempotent
+call book->clear()
+pp test_ (book)
```
andychu, line 561: Let's start using `assert [expected === actual]` where possible, as in the last PR

I know the tests are not consistent about this!  But it does make it easier to read

andychu:
It would also be useful to add the runnable examples (via `run-code-all`) for the previous methods, `find findLast` etc.

Although now I remember `strcmp()` is in `chap-builtin-func`, which may not be validated yet .... let me see

andychu:
Also, that function is not perfect, because it doesn't actually check all the results

But it's better than nothing -- it at least parses and runs the code

The history there is that there were too many errors in the docs!  It is hard to write accurate and readable examples ... 

So then I added a bit of automation to improve their quality

spec/ysh-methods.test.sh
last-genius: sadly, "equality isn't defined on float values" so these dicts can't be compared with an assert

builtin/method_dict.py
last-genius: Agreed, but with `setvar`, the increment is always provided, which makes this dispatch much simpler

last-genius:
Expanded the docs with runnable examples - could probably drop the lengthy specification of `inc()` now that it's illustrated with code?

spec/ysh-methods.test.sh
andychu: Hm, how about separating it to tests for floats and integers then

And use assert where we can (when it's not floats)

doc/ref/chap-type-method.md
```diff
@@ -641,6 +641,40 @@ TODO:
...
+Increments the value by the increment specified. If there is no value in the
+dictionary with the provided key, its default value is 0 of the type of the
+increment (int or float). If the increment is not specified, its default value
+is 1 of the type of the value with the provided key. If neither the increment
+nor the value with the provided key exist, then the result of the operation
+is an integer 1.
+
+    var mydict = {
+      a: 1,
+      b: 3.0,
```
andychu: for floats, I think it reads nicer if you use an example like `3.14`, i.e. something that's not an integer

doc/ref/chap-type-method.md
```diff
@@ -641,6 +641,40 @@ TODO:
...
+    }
+
+    = mydict # => (Dict)   {a: 1, b: 3.0}
+
+    call mydict->inc('a')
```
andychu: Yes let's remove the long description, moving it to descriptive text above each example

And it can consider ints and floats separately, so it's less abstract

----

I added an example for strcmp() here: https://github.com/oils-for-unix/oils/commit/e7ede10ecd7da1e8f000265278fd3a216abede8b

So it has both text and code, and they are close to each other

doc/ref/chap-type-method.md
```diff
@@ -641,6 +641,40 @@ TODO:
 call mydict->accum('key', 'string to append')
 ```
 
+### inc()
```
andychu: Now I wonder if it should be add() and append()

Because there is no `dec()` -- you do `d->inc(key, -3)` 

arguably `add(key, -3)` is clearer

i.e. the names `add() append()` might be better than `inc() accum()`

It's more concrete, since `+` is the addition operator

builtin/method_dict.py
andychu: Hm I think it should call the actual addition code, because the behavior differs

```
ysh-0.36$ = d
(Dict)  {f: 1.2, i: 42}

ysh-0.36$ call d->inc('f', 1)
  call d->inc('f', 1)
  ^~~~
[ interactive ]:7: fatal: Inc() expected int, got Float
```


All of these should work:

Int + Int
Int + Float
Float + Int
Float + Float

`ysh/expr_eval.py` has a _ConvertForBinaryOp() function

Let's make that public (removing _), and use that

Then this function can be very SHORT, and match the semantics of `+`

----

And yes I think it should be called `add()` and not `inc()`, since  it is literally addition, and should be implemented with the same code/spec as addition

andychu:
Hm I was also thinking that `d->add()` might be confused as "add item to collection" not "add a number to the value corresponding to a key"

Also, I think that explicitly requiring  1 is not bad, e.g. `d->inc(key, 1)` rather than `d->inc(key)`   

It's more parallel to `+= 1`

We didn't finish our design discussion on Zulip :-)

andychu:
Well I guess we can leave it `inc()` for now, and then change it later if we want to

last-genius:
I've already renamed it to `add` haha...

also moved to using `ConvertForBinaryOp`, split tests to use `assert`, moved the examples around.

Also added an implementation of `append`.

builtin/method_dict.py
```diff
@@ -64,6 +66,89 @@ def Call(self, rd):
...
+            left = dictionary[key]
+        else:
+            left = value.Int(mops.ZERO)
+
+        c, i1, i2, f1, f2 = expr_eval.ConvertForBinaryOp(left, right)
+
+        if c == coerced_e.Int:
+            res = value.Int(mops.Add(i1, i2))  # type: value_t
+        elif c == coerced_e.Float:
+            res = value.Float(f1 + f2)
```
andychu, line 110: let's add `else: raise AssertionError()` for readability - so we know all cases are handled

builtin/method_dict.py
```diff
@@ -64,6 +66,89 @@ def Call(self, rd):
...
+        dictionary = rd.PosDict()
+        key = rd.PosStr()
+        inc_val = rd.OptionalValue()
+        rd.Done()
+
+        # If inc_val is provided, we treat it as the definitive type of the
+        # value to be incremented, if it isn't, we rely on the existing
+        # value in the dictionary (if that doesn't exist, the default is 0)
+        if inc_val is None:
+            right = value.Int(mops.ONE)  # type: value_t
```
andychu: hm I'm on the fence about default 1 ... I think writing out 1 could be OK

```
call d->inc(k, 1)
call d->add(k, 1)
```

this seems harder to read:

```
call d->add(k)  # what are you adding?
```

doc/ref/chap-type-method.md
```diff
@@ -633,13 +633,69 @@ Ensures that the given key does not exist in the dictionary.
...
 
-TODO:
+Appends the given value to the list pointed to by the key:
 
-```raw
-call mydict->accum('key', 'string to append')
-```
+    var mydict = {a: [1]}
+    = mydict
+    # => (Dict)   {a: [1]}
```
andychu, line 642: hm I usually write `# =>` on the right, but I guess having it on the next line can be clearer in some cases

spec/ysh-methods.test.sh
```diff
@@ -564,6 +577,58 @@ pp test_ (book => get("author", ""))
...
+var histogram = {a: 1979}
+call histogram->add('a')
+assert [{"a": 1980} === histogram]
+call histogram->add('a', 3)
+assert [{"a": 1983} === histogram]
+call histogram->add('b')
+assert [{"a": 1983, "b": 1} === histogram]
+call histogram->add('c', 3)
+assert [{"a": 1983, "b": 1, "c": 3} === histogram]
+## status: 0
```
andychu: How about test cases where the original key would lead to a type error?

e.g. both of these in different cases:

```
var d = {k: 'foo'}

call d->add(k, 42)

call d->append(k, 'bar')
```

last-genius:
Dropped the default argument for `add()`, added tests - ready for another review now

andychu:
Merged, thanks!

PR #2599, opened 2025-11-30
Add dev-guide build script

PossiblyAShrub:
This clones the github wiki and renders all the markdown pages using doctools.

Usage details are in the `build/dev-guide.sh` script header comment.

### DIFF
--- a/.github/workflows/update-wiki.yml
+++ b/.github/workflows/update-wiki.yml
@@ -8,7 +8,6 @@ jobs:
   build-deploy:
     runs-on: ubuntu-22.04
     env:
-      OILS_PAGES_GITHUB_USER: ${{ secrets.OILS_PAGES_GITHUB_USER }}
       OILS_PAGES_GITHUB_TOKEN: ${{ secrets.OILS_PAGES_GITHUB_TOKEN }}
     steps:
       - name: Check out repository code
--- a/build/wiki.ysh
+++ b/build/wiki.ysh
@@ -18,11 +18,9 @@
 # This will get the latest oils-for-unix/oils-for-unix.github.io, copy the wiki
 # build into the clone and then commit + push it.
 #
-# This requires 2 environment variables to be set (usually in GitHub Actions
-# secrets):
-#   1. OILS_PAGES_GITHUB_USER    # GitHub username for the token/PAT holder
-#   2. OILS_PAGES_GITHUB_TOKEN   # A PAT with content: write permissions to
-#                                # the oils-for-unix.github.io repo.
+# This requires an environment variable to be set (usually in GitHub Actions
+# secrets): `OILS_PAGES_GITHUB_TOKEN`. This is a PAT with `content: write`
+# permissions to the oils-for-unix.github.io repo.
 
 const WIKI_OUT_DIR = '_tmp/wiki/out'
 const WIKI_DIR = '_tmp/wiki/git'
@@ -105,9 +103,8 @@ proc render() {
 proc commit-push() {
   ## Commit and push the most recent build to oils-for-unix.github.io
 
-  var USER = ENV.OILS_PAGES_GITHUB_USER
   var PAT = ENV.OILS_PAGES_GITHUB_TOKEN
-  pull-latest $PAGES_DIR "https://$USER:$PAT@github.com/PossiblyAShrub/oils-for-unix.github.io.git"
+  pull-latest $PAGES_DIR "https://x-access-token:$PAT@github.com/PossiblyAShrub/oils-for-unix.github.io.git"
 
   # Stash and pending changes (there should be none)
   cd $PAGES_DIR {


andychu:
Hm putting it in _release/VERSION does make sense

But sometimes we don't have releases for 6 weeks or more ...

I wonder if we should put it in `pages.oils.pub`?  That's not perfect either, because it then you might have to:

1. edit wiki
2. re-run this script and push HTML

Ideally we run this script automatically, but we can start running it manually for now

Re-publishing the wiki serves these two goals:

- make it crawlable
- make it look slightly nicer perhaps - Github UI is a bit slow lately
- and retain our existing wiki pages ... gradually refactoring them into a dev guide

PossiblyAShrub:
We can add a Github actions job that will run whenever the wiki is updated using `on: gollum`. [For reasons](https://dev.to/kalkwst/workflow-triggering-events-and-event-actions-5cec#the-gollum-event), they call this event type "gollum": https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows?versionId=free-pro-team%40latest#gollum

(We could also listen for updates using the [gollum webhook event](https://docs.github.com/en/webhooks/webhook-events-and-payloads?search-overlay-input=actions+trigger#gollum) if we ever wanted off Github Actions.)

Putting it on `pages.oils.pub` makes more sense. I'll recreate a PR there. And I'll use a YSH script ;)

andychu:
Oh interesting!  That makes sense

Unfortunately there is no wiki on Codeberg it seems ?  https://codeberg.org/oils/oils

We are getting a bit more locked in because of the Github wiki, but it is useful ...

andychu:
Hm does the web hook need to be in the same repo as the wiki?  I think the wiki is attached to `oils-for-unix/oils`, not the `oils-for-unix/oils-for-unix.github.io` repo

Actually it's a **third** repo, `oils-for-unix/oils.wiki.git`

```
andy@hoover ~/git/oils-for-unix/oils.wiki (master)$ git remote show origin
* remote origin
  Fetch URL: git@github.com:oils-for-unix/oils.wiki.git
  Push  URL: git@github.com:oils-for-unix/oils.wiki.git
  HEAD branch: master
```

Not sure where the code should go in order to use Github Actions this way

PossiblyAShrub:
Here is the design that I have settled on:

We will have 2 GitHub actions jobs:
1. One on the `oils-for-unix/oils` repo
  - This listens to the `gollum` event (wiki updated) and then sends a [repository dispatch event](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-dispatch-event) to the `oils-for-unix/oils-for-unix.github.io` repo
2. One on the `oils-for-unix/oils-for-unix.github.io` repo which listens for the [repo dispatch event](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#repository_dispatch)
  - This will pull in the wiki and build/deploy

The job (2) on the `oils-for-unix/oils-for-unix.github.io` repo could just be the current job which builds and deploys the `pages.oils.pub` site (this is implicitly created by GitHub on all pages sites) but adds a step to build the dev guide from the latest version of the wiki.

It's pretty dependent on Github right now. We could use raw webhooks as I suggested [on zulip #distros > automating regtest/aports new causes @ 💬](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/automating.20regtest.2Faports.20new.20causes/near/561747987).

I'll update when it's implemented. I may need some help adding a secret to CI so we can actually send that dispatch event in job (1).

PossiblyAShrub:
Most of job (2) mentioned above is done in https://github.com/oils-for-unix/oils-for-unix.github.io/pull/1

I've sent a request for a personal access token so I can implement job (1). We should talk about this in the meeting tomorrow. I don't really like holding onto a PAT that I need to renew every year so the dev guide can keep updating...

PossiblyAShrub:
Almost there on getting the auto deploy to work. It is frustratingly hard to test the actions triggers though. For example, the `gollum` event will only work if it's on `master`...

> This event will only trigger a workflow run if the workflow file exists on the default branch.
[docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows?versionId=free-pro-team%40latest#gollum)

I'll try running this in a fork.

PossiblyAShrub:
Okay, good news!  I now have this working. Here is an example commit (on my fork) of what the wiki update looks like:

https://github.com/PossiblyAShrub/oils-for-unix.github.io/commit/50ca96411755915b89b4b32679c29a43b7f43296

The result is then displayed on GitHub Pages: https://possiblyashrub.github.io/oils-for-unix.github.io/wiki/Dev-Guide.html

This is ready to merge into `master`, pending 2 steps from your end @andychu,
1. Either:
  a) My account needs write permissions to the `oils-for-unix.github.io` page, or
  b) You can create a new PAT with write permissions to the `oils-for-unix.github.io` page and update the `OILS_PAGES_GITHUB_TOKEN` and `OILS_PAGES_GITHUB_USER` [CI secrets](https://github.com/oils-for-unix/oils/settings/secrets/actions)
2. This PR needs a review

build/wiki.ysh
```diff
@@ -0,0 +1,129 @@
...
+          <link rel="stylesheet" type="text/css" href="../web/base.css" />
+        </head>
+        <body class="width40">
+    """
+
+    pre-render-wikilinks <$path | cmark
+
+    echo """
+      <div id="build-timestamp">
+        <i>Generated on $DATE</i>
```
andychu: Hm does this update all wiki pages with a timestamp when just one wiki page changes?

If so, I think we could leave out the timestamp to avoid that?

build/wiki.ysh
```diff
@@ -0,0 +1,129 @@
...
+
+proc render() {
+  pull-latest $WIKI_DIR https://github.com/oils-for-unix/oils.wiki.git
+  find $WIKI_DIR -name '*.md' -print0 | xargs -I {} -0 -- $0 render-one {}
+}
+
+
+proc commit-push() {
+  var USER = get(ENV, 'OILS_PAGES_GITHUB_USER')
+  var PAT = get(ENV, 'OILS_PAGES_GITHUB_TOKEN')
```
andychu: could be `ENV.OILS_PAGES_GITHUB_USER` ?   That will fail a little earlier than `USER === null` ?

or stylistically `ENV => get('OILS_PAGES_GITHUB_USER')` ?

.github/workflows/update-wiki.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Check out repository code
+        uses: actions/checkout@v4
+
+      - name: Install CommonMark
+        run: |
+          sudo apt install cmark
+
+      - name: Install YSH
+        run: |
+          expected="f4d41d20a0523dbcfbd4ba231f82edf25b08d4965d65bc71fcb56666d6743000"
```
andychu, line 22: I think eventually we will need to have a common shell function to either compile YSH or install the pre-built YSH 

Later we can replace this

build/wiki.ysh
```diff
@@ -0,0 +1,136 @@
...
+
+  pull-latest $WIKI_DIR https://github.com/oils-for-unix/oils.wiki.git
+  find $WIKI_DIR -name '*.md' -print0 | xargs -I {} -0 -- $0 render-one {}
+}
+
+
+proc commit-push() {
+  ## Commit and push the most recent build to oils-for-unix.github.io
+
+  var USER = ENV.OILS_PAGES_GITHUB_USER
```
andychu: hm what is the actual value of "USER"? :-)  I am confused on how the access works

Maybe it shouldn't be a secret -- just the token can be a secret?

PossiblyAShrub: In this case, it's just my username: `PossiblyAShrub`. But if _you_ generated the PAT, then you'd have to change it to `andychu`. The value technically isn't a secret, but placing these two values in the same place makes it harder to miss when we update the PAT. (The PAT is only valid for a year, then we have to regen.)

andychu: Oh I see

Yeah there is also a pages-build-deployment user ?  (as you pointed out)

https://github.com/oils-for-unix/oils-for-unix.github.io/actions

I wonder if we can use that one, or create a new one similar 

Let me see

andychu: Hm well I think when @btertoolen  moves the `regtest/aports published.html` to `pages.oils.pub`, then using your name could be an issue?

I want to use the same setup for both, but I'm not sure exactly how to do it

I also don't understand why here the commits appear as "Oils wiki bot"

https://github.com/PossiblyAShrub/oils-for-unix.github.io/commits/main/

I guess the git user is disconnected from the github actions user ... ?

andychu (commented):
Since I added everyone to the `oils-for-unix.github.io` repo, will this started working once I merge it?

I tried updating the wiki and didn't see any actions trigger, but maybe that's because I didn't merge yet

I think we should change to some kind of "role user" -- similar to `pages-build-deployment`  created by Github -- but not actually the same user

But we can do it in multiple steps if it's easier to test that way


---

I think for now it's better to hard code the user ( PossiblyAShrub) in the script, because logically we have one secret, not 2

PossiblyAShrub:
Summary:

- You can avoid the username in the script. This makes updating/changing the PAT easier and now we only have one secret (I've pushed this change!)
- Git and GitHub identities are different. The "Oils Wiki Bot" is *my* account, just with a different `git config` name and email. Changing the PAT to be under *your account* would cause those commits to be made by *your account*.
- One option to avoid having a PAT under my account is to create a "machine user", which is just another GitHub account

Below goes into a bunch of details with links about how this all works and directly addresses all of your questions @andychu.

---

> Since I added everyone to the `oils-for-unix.github.io` repo, will this started working once I merge it?

Yes.

---

> I tried updating the wiki and didn't see any actions trigger, but maybe that's because I didn't merge yet

I did some testing on the linked fork, so it *should* work. The only way to get the action to trigger is if it is on `master`: https://github.com/oils-for-unix/oils/pull/2599#issuecomment-3635075417.

---

> Hm well I think when @btertoolen moves the regtest/aports published.html to pages.oils.pub, then using your name could be an issue?
> 
> I want to use the same setup for both, but I'm not sure exactly how to do it

Bram can use the same setup. Technically the commits will be made with my account, but the git user associated with those commits is configurable. (See below for the idea of a "machine account" and that GitHub + Git identities are different.) **I've pushed a commit removing the username altogether.**

I did some more digging, and it turns out [you can use `x-access-token` instead of a user name](https://docs.github.com/en/apps/creating-github-apps/writing-code-for-a-github-app/building-ci-checks-with-a-github-app#add-code-to-clone-a-repository) (found after seeing [this article](https://blog.infra.kiwi/pushing-new-commits-on-github-actions-with-a-personal-access-token-pat-8c9240fbc8d2)).

---

> I also don't understand why here the commits appear as "Oils wiki bot"
> 
> https://github.com/PossiblyAShrub/oils-for-unix.github.io/commits/main/
> 
> I guess the git user is disconnected from the github actions user ... ?

Yeah, the tricky part is GitHub is a very leaky abstraction. Git has decentralized identity management meanwhile GitHub centralizes it. This means that the git users and github users are completely detached. For example, [Why are my commits linked to the wrong user? (Github Docs)](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user).

Technically, in these actions, all the commits are made using my account. But the email and name associated with those commits can be any account. I figured it doesn't make sense to use my name+email, as this is a bot, so I made up a bogus name + email here.

There is also [what GitHub calls a "**Machine Account**"](https://docs.github.com/en/get-started/learning-about-github/types-of-github-accounts#:~:text=This%20type,%28CI%29%20workflows.). This account is just a normal GitHub account, but it's sole purpose is to do automated actions. [Rust's bors bot](https://github.com/bors) is an example.

---

> I think we should change to some kind of "role user" -- similar to `pages-build-deployment` created by Github -- but not actually the same user
> 
> But we can do it in multiple steps if it's easier to test that way

Interestingly, the `pages-build-deployment` actions are all done in the commiter's name.

As an example, there is an [action for automatically creating commits (but only in the same repository)](https://github.com/stefanzweifel/git-auto-commit-action). In [the article detailing how it works](https://michaelheap.com/git-auto-commit/),

> In order, the following steps are executed:
> 
>   -  [Check out the specified branch](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/entrypoint.sh#L40-L54) ([defaults](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/action.yml#L14) to github.head_ref)
>   -  [Add required files](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/entrypoint.sh#L56-L61). [By default](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/action.yml#L22) all files will be added with .. You can specify a pattern such as src/* or *.js with the file_pattern input
>   - [Create a commit](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/entrypoint.sh#L63-L79) containing the changes. The [default](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/action.yml#L27-L38) identity will commit the changes as GitHub Actions, **with the change author being [set to the person that triggered the workflow run](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/action.yml#L38).** You can also specify commit_options such as --signoff to add a trailer.
>   -  If you provide a [tagging_message](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/action.yml#L39-L42) a tag [will be created](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/entrypoint.sh#L81-L91) pointing to the commit that was just created.
>   -  Finally, the commit (and possibly a tag) will be [pushed to GitHub](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/entrypoint.sh#L93-L117). You [can specify push_options](https://github.com/stefanzweifel/git-auto-commit-action/blob/a6006229e0c9695d16a585acf619a3a9548b18f9/action.yml#L43-L46) to add additional flags to git push
(Emphasis on "with the change author being set to the person that triggered the workflow run." is mine.)

**We could create a new account which maintains the PAT instead of my personal account**. I don't mind maintaining a PAT for now. And, if I am unable to create a new PAT for whatever reason, fixing that is as simple as creating a new PAT with another account and updating the secret.

PR #2602, opened 2025-12-04
Upgrade to aports 3.23 for abuild fix

### LINKED ISSUE #2424: [osh as ash] heimdal, perl, tk - "Found uncompressed man pages:"
This bug only occurs with OSH as busybox ash

Not sure what the root cause is

https://op.oils.pub/aports-build/2025-09-17-ash2.wwz/_tmp/aports-report/2025-09-17-ash2/diff_merged.html

[2025-09-17-ash2 - regtest/aports run](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/2025-09-17-ash2.20-.20regtest.2Faports.20run/with/539924176)

EarthenSky:
Close #2424

I've tested that all of `heimdal`, `tcl`, `tk`, and `perl` succeed locally. I had to remove `../../alpinelinux` in order for the new alpine chroot to be setup correctly, although ymmv.

We want this change https://gitlab.alpinelinux.org/alpine/abuild/-/commit/73203da0e713bf1f0d022d91363c47d96301f4cc?__goaway_challenge=cookie&__goaway_id=03e89dcf81c655e8833b8f9940d96903 from abuild-3.16.0-r0 which first appears in aports-3.23. The change fixes a small bug that was accidentally unnoticed because of bash's lastpipe behaviour.

### DIFF
--- a/regtest/aports-setup.sh
+++ b/regtest/aports-setup.sh
@@ -54,7 +54,7 @@ checkout-stable() {
   # dl-cdn.alpinelinux.org URL
   # https://alpinelinux.org/releases/
 
-  local branch='3.22-stable'
+  local branch='3.23-stable'
   git checkout $branch
   git log -n 1
   popd > /dev/null
@@ -153,7 +153,7 @@ make-chroot() {
 
   # "branch" is one of these: https://dl-cdn.alpinelinux.org/
   # it's not the aports branch
-  local branch='v3.22'
+  local branch='v3.23'
   time sudo $aci -n -d $PWD/$CHROOT_DIR -b $branch
 }
 


EarthenSky:
I'm not sure if there's anything else that should be changed, but if so feel free to let me know (or modify it yourself if it's quicker)!

andychu:
Ha I could have sworn 3.22 was the latest version when I looked, and it turns out 3.23 was released yesterday :-)

https://www.alpinelinux.org/releases/

@btertoolen is working on reproducing `regtest/aports` on `he.oils.pub`

- I think our current number should be **11** on `aports/main`
- with this change, does it go down to **7** ?

So I guess I will leave it up to Bram to say whether this should be merged now, or if he wants to wait until we have a number with 3.22

---

We can also have multiple people running `regtest/aports`, but in that case it's better to coordinate on the machine usage

This kind of experience will be useful for `regtest/debian`

andychu:
But yes, I think upgrading to `3.23` is the right solution now, since you've tested that the packages work with OSH

This way we avoid any patches ... it's a clean solution (and we can describe the caveat with 3.22 if we like)

andychu (commented):
(oops I should have reviewed rather than commented, to keep my TODO list clean)

PR #2606, opened 2025-12-09
Update causes script

btertoolen:
I added a `regtest/aports-update-causes.sh` script that can be used to generate a new report using `causes.awk`, based on the latest report. It currently does it for either community reports or main reports since these are currently separate. But we could also modify the script to always regenerate both.

The main function of `regtest/aports-update-cause.sh` is `update-latest` (name should be improved). The high level steps are explained in that function:
```
  ### 1. Determine latest report name / epoch
  ### 2. Fetch latest published report
  ### 3. Generate new report using new causes.awk
  ### 4. Upload new report to op.oils.pub
```

Still on the first iteration here, there are some assumptions that need to be handled better:
  * always assumes main runs are full runs
  * always assumes community runs are partial runs
  * `regtest/aports-update-causes.sh` gets the most recent reports from the server, based on the date of the report but:
     * this assumes community reports have `comm` in their name
     * assumes main reports are all reports that dont have `comm` in their name
     * assumes all reports start with a date with the yyyy-mm-dd format (this is fine I guess, but we could also look at the last modified date of the file).
  * assumes the script is only executed once per day, else it fails because `$new_epoch` already exists
 

The intended use for `regtest/aports-update-causes.sh` is to run it like this:
```bash
regtest/aports-update-causes.sh update-latest main # generate new report for main
regtest/aports-update-causes.sh update-latest community # generate new report for community
```

I added a small python server to run `regtest/aports-update-causes.sh` whenever a request is given to its port. But this is just for a proof of concept, we can replace this with a proper server if needed.

And also of course the trigger for the github action should eventually be changed to pull requests that change `causes.awk`

### DIFF
--- a/soil/web-worker.sh
+++ b/soil/web-worker.sh
@@ -278,12 +278,7 @@ deploy-job-results() {
   local run_dir=$2  # e.g. 1234  # make this dir
   local job_name=$3  # e.g. cpp-small for example.com/github-jobs/1234/cpp-small.wwz
 
-  local payload_type=${prefix}jobs
-  if [ ${job_name} == 'aports-report' ]; then
-    payload_type='github-jobs'
-  fi
-
-  # rest of args are more env vars
+    # rest of args are more env vars
   shift 3
 
   # writes $job_name.wwz
@@ -309,7 +304,7 @@ deploy-job-results() {
     my-scp $job_name.{wwz,tsv,json} "$SOIL_USER_HOST:$remote_dest_dir"
   else
     curl --include --fail-with-body \
-      --form "payload-type=${payload_type}" \
+      --form "payload-type=${prefix}jobs" \
       --form "subdir=$run_dir" \
       --form "file1=@${job_name}.wwz" \
       --form "file2=@${job_name}.tsv" \
@@ -417,17 +412,13 @@ remote-event-job-done() {
   ### "Client side" handler: a job calls this when it's done
 
   local prefix=$1  # 'github-' or 'sourcehut-'
-  # TODO: remove this, temporary workaround
-  if [ prefix != "github-" ]; then
-    prefix="github-"
-  fi
   local run_id=$2  # $GITHUB_RUN_NUMBER or git-$hash
 
   log "remote-event-job-done $prefix $run_id"
 
   # Deployed code dir
   if false; then
-    sshq soil-web/soil/web.sh event-job-done $prefix $run_id
+    sshq soil-web/soil/web.sh event-job-done "$@"
   else
     # Note: I think curl does URL escaping of arg1= arg2= ?
     curl --include --fail-with-body \


.github/workflows/new-report.yml
```diff
@@ -0,0 +1,15 @@
...
+    branches:
+      - update-causes-script
+
+jobs:
+  trigger:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Trigger new report generation
+        run: |
+          curl "http://he.oils.pub:8080?token=super-secret&kind=main"
```
andychu: Hm I don't think we should have another server?

I guess we were not talking about the same thing -- when I mentioned `curl`, I was saying that Github Actions probably has hooks you can trigger on `curl` ?

I think we should try to make the entire thing run **within** Github Actions, not on `he.oils.pub`?

I think that should be possible?  It can sync a given set of results, and then update the causes

e.g. 

> (latest results .wwz + causes.awk in Oils repo) => new HTML `.wwz` published

btertoolen: Yes I did it like this for now as a test. I wasnt really clear on this. 

We can run `regtest/aports-update-causes.sh` directly in github actions, but for that we need to add an SSH key secret to github, so that the ci can download and upload to the op.oils.pub server. But I dont have permissions to do anything with github secrets, so that was blocking me a bit last night. And we will need (or atleast it would be nice) to have some sort of pre-built binary that can be downloaded like we discussed, but thats not a blocker because i can look into that myself.

And then eventually we could also turn this into a workflow_dispatch trigger, but you can only test a workflow_dispatch trigger once it is merged into main

andychu: Ah OK Aidan (@PossiblyAShrub )  tested it on his own fork -- which has its own `main` branch

And also requested some secrets

So we want to use the same approach for both

andychu: BTW I think the whole thing can be tested without secrets in Github actions up until the last step -- deploying `.wwz`

1. fetch old .wwz
2. run the report again
3. create new .wwz

Once that works, then we can hook up the same/similar auth that we're doing for the wiki

deps/Dockerfile.soil-aports-report
```diff
@@ -0,0 +1,46 @@
...
+
+# Copy again to prevent unsound caching
+COPY deps/from-apt.sh /home/uke/tmp/deps/from-apt.sh
+
+# layer-locales also has to install packages
+RUN --mount=type=cache,id=var-cache-apt-debian-12,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,id=var-lib-apt-debian-12,target=/var/lib/apt,sharing=locked \
+    du --si -s /var/cache/apt /var/lib/apt && \
+    deps/from-apt.sh layer-debian-12 && \ 
+    deps/from-apt.sh ovm-tarball && \
```
andychu, line 11: TODO: let's look at these two

regtest/aports-update-causes.ysh
```diff
@@ -0,0 +1,151 @@
+#!bin/ysh
```
andychu, line 1: Name this file `aports-update-causes.ysh` to identify the language

regtest/aports-update-causes.sh
```diff
@@ -0,0 +1,151 @@
...
+# Generate a new report, based on the latest report available.
+# Used to update causes in a report without rebuilding any packages.
+# 
+# Example usage:
+#   regtest/aports-update-causes.sh update-latest main # generate new report for main
+#   regtest/aports-update-causes.sh update-latest community # generate new report for community
+
+const REPORTS_PATH="_tmp/aports-report"
+
+proc stdout_to_stderr() {
```
andychu: let's remove this since you said it was no longer necesssary

(related to the "pattern 8: redirecting shell functions/blocks")

regtest/cause-server.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+import http.server
+import subprocess
+import urllib.parse
+import os
```
andychu: The cause server is no longer needed I believe

Since we can trigger through Github Actions

soil/apt_install.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+
+apt-get update
+apt-get install -y $1 --no-install-recommends
```
andychu: hm do we need this small wrapper?

We are often using `deps/from-apt.sh`

soil/host-shim.sh
```diff
@@ -23,8 +23,8 @@ live-image-tag() {
   local image_id=$1
 
   # podman build migration!
-  echo 'v-2025-11-02'
-  return
+  # echo 'v-2025-11-02'
+  # return
```
btertoolen, line 27: Need to check with @andychu what the proper fix is here. Probably just update the case switch from below?

soil/host-shim.sh
```diff
@@ -319,7 +323,15 @@ run-job-uke() {
       fi
   esac
 
-  local image="docker.io/oilshell/soil-$image_id"
+  if [ $job_name == aports-report ]; then
+    flags+=( -e "REGTEST_CAUSES_BASE_REPORT=$REGTEST_CAUSES_BASE_REPORT" )
```
btertoolen, line 327: TODO: can we pass these as arguments to the script update causes script in the run command? Then you dont need to set env variables if you want to run the script locally

soil/host-shim.sh
andychu: Yes we can have separate versions for each container

When I rebuilt all of the containers, I changed it to a single version

But we can go back to different versions

andychu:
This PR should be split up a bit, e.g. I want to merge the `ghcr.io` changes first

btertoolen:
I put a separate PR for the ghcr.io changes in #2635

PR #2608, opened 2025-12-10
[devtools fix] Fix git config

PossiblyAShrub:
Arghh... third times the charm! I don't think I can ever get actions to work reliably on the first attempt :P

Second fix for https://github.com/oils-for-unix/oils/pull/2599

I forgot to add git credentials this time: https://github.com/oils-for-unix/oils/actions/runs/20115714722/job/57724554927

### DIFF
--- a/.github/workflows/update-wiki.yml
+++ b/.github/workflows/update-wiki.yml
@@ -37,4 +37,6 @@ jobs:
 
       - name: Commit and Push
         run: |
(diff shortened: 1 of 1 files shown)


andychu:
In theory I want everything in containers where it can be tested locally ... but yeah for now doing some iterations with Github seems to be the state of the art ...

OK let's see what happens!

PR #2609, opened 2025-12-11
Implement dict->update and ++ for dicts

last-genius:


doc/ref/chap-type-method.md
```diff
@@ -684,6 +684,24 @@ If the key doesn't exist, then its default value is zero:
 In other words, it's like `setvar mydict.k += 3`, but you don't have to
 initialize each value to zero.
 
+### Dict/update()
```
andychu: I think the link from the TOC might be broken?  unless we call it update() not Dict/update()

spec/ysh-methods.test.sh
```diff
@@ -642,6 +642,24 @@ var d = {k: 'foo'}
 call d->append('k', 'bar')
 ## status: 3
 
+#### Dict -> update()
+var d = {a: 2, b: 3}
```
andychu, line 646: let's also add var alias = d, and then test that it is also modified

ysh/expr_eval.py
```diff
@@ -687,10 +687,18 @@ def _Concat(self, left, right, op):
             c.extend(right.items)
             return value.List(c)
 
+        elif left.tag() == value_e.Dict and right.tag() == value_e.Dict:
+            left = cast(value.Dict, UP_left)
+            right = cast(value.Dict, UP_right)
+
+            res = left.d.copy()
```
andychu, line 694: Hm technically if we wanted to keep the mycpp runtime small, and reduce total code

We could just inline the copy, and avoid the copy() method:

```
res = {}
for ...

```

It is more efficient to do the reserve() though, so I guess it's fine to keep it.  One more thing for a Rust translation :-)

---

Actually now I realized the C++ version can be optimized even more by copying the 4 structs/slabs (Dict, index, keys, values)

which avoids hashing

So I guess we can keep the copy() method  and maybe do it later

andychu (commented):
Looks great, thank you!

I like the the code continues to be short :-)

spec/ysh-methods.test.sh
last-genius: Done

doc/ref/chap-type-method.md
last-genius: Fixed

andychu:
Great thank you!  (I made minoir changes to the docs)

PR #2612, opened 2025-12-14
Add strsignal() translation and wcoredump support

### LINKED ISSUE #2610: Translating strsignal() and WCOREDUMP to C++
PR #2516 added signal messages for killed background processes (fixing #722), but only in Python mode. C++ mode needs the same functionality.

When a background job is killed by a signal, Python OSH displays the signal name, but C++ OSH shows nothing:

```bash
# Python OSH
$ sleep 1 & kill -HUP $!; wait $!
Hangup
status=129

# C++ OSH
$ sleep 1 & kill -HUP $!; wait $!
status=129
(no message)
```

**Failing test:** `test/spec-cpp.sh run-file background -r 26`

### References

- Original issue: #722
- Python implementation: #2516 (merged)
- https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Translating.20strsignal.28.29.20and.20WCOREDUMP.20to.20C.2B.2B

daveads:
C++ translation Add strsignal() and WCOREDUMP support

Translates the signal message functionality from Python to C++ mode.

The WCOREDUMP fallback uses (status & 0200) which is portable across all architectures.

closes #2610

### DIFF
--- a/spec/builtin-completion.test.sh
+++ b/spec/builtin-completion.test.sh
@@ -1,5 +1,4 @@
 ## oils_failures_allowed: 3
-## oils_cpp_failures_allowed: 3
 ## compare_shells: bash
 
 #### complete with no args and complete -p both print completion spec
--- a/spec/builtin-trap.test.sh
+++ b/spec/builtin-trap.test.sh
@@ -1,6 +1,5 @@
 ## compare_shells: dash bash mksh ash
 ## oils_failures_allowed: 2
-## oils_cpp_failures_allowed: 2
 
 #### traps are not active inside subshells $() ()  trap | cat
 
--- a/spec/sh-options.test.sh
+++ b/spec/sh-options.test.sh
@@ -1,6 +1,5 @@
 ## compare_shells: bash dash mksh
 ## oils_failures_allowed: 1
-## oils_cpp_failures_allowed: 1
 ## tags: interactive
 
 # Test options to set, shopt, $SH.


andychu:
Hm this is still failing tests?

Did you see the messages about the spec-cpp allowed failures?

[Why is this CI job failing?](https://oilshell.zulipchat.com/#narrow/channel/121539-oils-dev/topic/Why.20is.20this.20CI.20job.20failing.3F/with/563698548)


Also Andriy fixed the failure messages

cpp/libc.cc
```diff
@@ -250,6 +252,23 @@ int sleep_until_error(double seconds) {
...
+  // Validate signal number range
+  if (sig_num < 1 || sig_num >= NSIG) {
+    throw Alloc<ValueError>(StrFromC("signal number out of range"));
+  }
+
+  errno = 0;
+  char* res = ::strsignal(sig_num);
+
+  // Return nullptr if there's an error, NULL result, or "Unknown signal" message
+  if (errno != 0 || res == nullptr || strstr(res, "Unknown signal") != nullptr) {
```
andychu: I don't think we should check for `errno != 0` here, or in `pyext/libc.c`

The C idiom is to check the return value, and then only `errno`

I looked at `man strsignal` and it is a little odd

I am also not sure where "Unknown signal" comes from ... is that ever returned?

core/process.py
```diff
@@ -2208,14 +2207,10 @@ def WaitForOne(self, waitpid_options=0):
...
-                        try:
-                            from os import WCOREDUMP  # type: ignore
-                            if WCOREDUMP(status):
-                                msg = msg + ' (core dumped)'
-                        except (ImportError, AttributeError):
-                            pass
+                    # WCOREDUMP checks if process dumped core
+                    # On systems without WCOREDUMP, this will be False
+                    if WCOREDUMP(status):
+                        msg = msg + ' (core dumped)'
```
andychu, line 2213: I also learned that core dumps aren't always on ... so maybe we shouldn't say "core dumped"

I'll think about that

pyext/libc.c
```diff
@@ -395,14 +395,13 @@ func_strsignal(PyObject *self, PyObject *args) {
...
     return NULL;
   }
-  
-  errno = 0;
+
   res = strsignal(sig_num);
-  
-  if (errno || res == NULL || strstr(res, "Unknown signal") != NULL) {
+
+  if (res == NULL || strstr(res, "Unknown signal") != NULL) {
```
andychu: I feel like we don't need the "Unknown signal" check here either?  What do you think?

cpp/libc.cc
```diff
@@ -250,6 +252,22 @@ int sleep_until_error(double seconds) {
...
+BigStr* strsignal(int sig_num) {
+  // Validate signal number range
+  if (sig_num < 1 || sig_num >= NSIG) {
+    throw Alloc<ValueError>(StrFromC("signal number out of range"));
+  }
+
+  char* res = ::strsignal(sig_num);
+
+  // Return nullptr if there's a NULL result or "Unknown signal" message
+  if (res == nullptr || strstr(res, "Unknown signal") != nullptr) {
```
andychu: Ditto, I think if the string is "unknown signal", it's fine to print that

core/process.py
```diff
@@ -2208,14 +2207,10 @@ def WaitForOne(self, waitpid_options=0):
...
-                    if mylib.PYTHON:
-                        # WCOREDUMP is only available on some systems
-                        try:
-                            from os import WCOREDUMP  # type: ignore
-                            if WCOREDUMP(status):
-                                msg = msg + ' (core dumped)'
-                        except (ImportError, AttributeError):
-                            pass
+                    # WCOREDUMP checks if process dumped core
+                    # On systems without WCOREDUMP, this will be False
```
andychu: This comment doesn't seem to match the implementation ... if WCOREDUMP is not defined, we fill it in?

Issue #2605, opened 2025-12-08
YSH builtin cat should not fork, as OSH does


andychu:
OSH does this for compatibility

[gzip - PIPE mishandling](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/gzip.20-.20PIPE.20mishandling/with/546392296)

Because the optimization wasn't sound:

https://oils.pub/blog/2025/12/links.html#io

But it is a speed issue, so I think YSH should not

andychu:
PR by @last-genius to fix the soundness issue

https://github.com/oils-for-unix/oils/commit/5647c933a097847359cba756b4fdfc53f3485f2c

I think this could be `shopt --set ysh_rewrite_extern`, part of `shopt --set ysh:upgrade`

(In addition to  of `shopt --set rewrite_extern`, which enables `builtin cat rm` .  We should also document this in `doc/ref` under the `builtin cat` and `chap-options.md`)

PR #2613, opened 2025-12-15
[core] Do not fork builtin cat in YSH

last-genius:
Provides a new ysh_rewrite_extern option (enabled with ysh:upgrade) that does not fork cat when rewriting it to a builtin.

Adds a spec test to ysh-options verifying this behavior.

Fixes: #2605

### DIFF
--- a/doc/ref/toc-osh.md
+++ b/doc/ref/toc-osh.md
@@ -197,7 +197,7 @@ X [Unsupported]   enable
   [Interactive]    emacs           vi
   [Compat]         eval_unsafe_arith            ignore_flags_not_impl
                    ignore_shopt_not_impl
-  [Optimize]       rewrite_extern
+  [Optimize]       rewrite_extern               ysh_rewrite_extern
 ```
 
 <h2 id="special-var">
--- a/spec/ysh-options.test.sh
+++ b/spec/ysh-options.test.sh
@@ -198,6 +198,7 @@ shopt -s simple_word_eval
 shopt -s verbose_errexit
 shopt -s verbose_warn
 shopt -s xtrace_rich
+shopt -s ysh_rewrite_extern
 ## END
 
 #### osh -O ysh:upgrade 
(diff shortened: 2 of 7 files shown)


spec/ysh-options.test.sh
```diff
@@ -758,3 +759,14 @@ echo ${#assoc[@]}
 ## STDOUT:
 2
 ## END
+
+#### builtin cat crashes the main process with ysh_rewrite_extern
```
andychu: Hm thanks for testing this

Although "crashes" sounds bad ... the spec should be that YSH never crashes :-)

We can say that a forked YSH/Oils child process exits with code indicating SIGPIPE   ?


I see this now

```
stdout:
141
stderr:
Broken pipe (core dumped)
Broken pipe (core dumped)
```

(although now I want to get rid of the `core dumped` thing, because core dumps aren't always on!)


---

Also I think this test should PASS and not fail ... this is the YSH spec -- and we are deciding that YSH has ysh_rewrite_extern ON by default

the OSH spec matches bash, but the YSH spec doesn't have to, so it should pass 

(unless we see any problems with this behavior, which I don't think there are)

doc/ref/toc-osh.md
```diff
@@ -197,7 +197,7 @@ X [Unsupported]   enable
   [Interactive]    emacs           vi
   [Compat]         eval_unsafe_arith            ignore_flags_not_impl
                    ignore_shopt_not_impl
-  [Optimize]       rewrite_extern
+  [Optimize]       rewrite_extern  ysh_rewrite_extern
```
andychu: Maybe add spaces to align   ysh_rewrite_extern aligns with ignore_flags_not_impl

(It's a bit of an art, not an exact rule ... but I think that will look better :-) )

spec/ysh-options.test.sh
last-genius: Made it so the test passes instead.

PR #2617, opened 2025-12-16
Address dev friction with compare.html

last-genius:
Fixes compare.html not being generated for ysh-all on errors and makes it more informative so that it's clearer what caused the errors

Here's how osh-cpp's compare.html page looks like with this PR: 
image

See https://op.oilshell.org/uuu/github-jobs/11017/cpp-spec.wwz/_tmp/spec/osh-cpp/compare.html for an example of "before", where it's unclear what the issue is.

### DIFF
--- a/test/spec-cpp.sh
+++ b/test/spec-cpp.sh
@@ -193,15 +193,24 @@ summary-tsv-row() {
   local spec_subdir=$1
   shift
 
+  local status cpp_failures_allowed cpp_failures
+
   if test $# -eq 1; then
     local spec_name=$1
     local -a tsv_files=( _tmp/spec/$spec_subdir/$spec_name.result.tsv )
+    status=$( awk '{printf $1}' "_tmp/spec/$spec_subdir/$spec_name.task.txt" )
+    cpp_failures_allowed=$( awk '{printf $6}' "_tmp/spec/$spec_subdir/$spec_name.stats.txt" )
+    cpp_failures=$( awk '{printf $7}' "_tmp/spec/$spec_subdir/$spec_name.stats.txt" )
   else
     local spec_name='TOTAL'
     local -a tsv_files=( "$@" )
+    cpp_failures_allowed="-"
+    cpp_failures="-"
+    status=0
   fi
 
-  awk -v spec_name=$spec_name '
+  awk -v spec_name=$spec_name -v status=$status \
+      -v cpp_failures_allowed=$cpp_failures_allowed -v cpp_failures=$cpp_failures '
 # skip the first row
 FNR != 1 {
   case_num = $1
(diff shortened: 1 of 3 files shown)


test/spec-cpp.sh
```diff
@@ -190,12 +196,19 @@ summary-tsv-row() {
   if test $# -eq 1; then
     local spec_name=$1
     local -a tsv_files=( _tmp/spec/$spec_subdir/$spec_name.result.tsv )
+    status=`awk '{printf $1}' "_tmp/spec/$spec_subdir/$spec_name.task.txt"`
+    local -a cpp_failures_allowed=(`awk '{printf $6}' "_tmp/spec/$spec_subdir/$spec_name.stats.txt"`)
+    local -a cpp_failures=(`awk '{printf $7}' "_tmp/spec/$spec_subdir/$spec_name.stats.txt"`)
   else
     local spec_name='TOTAL'
     local -a tsv_files=( "$@" )
+    local -a cpp_failures_allowed=( "$@" )
```
andychu: This case is for the TOTAL row at the top, and it looks like this part is not right

Actually I think we don't need an array in either case

I think it can be

```
local status cpp_failures_allowed cpp_failures  # declare string variables first
if test ...
  status=$(...)
  cpp_failures_allowed=$(awk ...)
  cpp_failures=$(awk ...)
else
  status='-'  # or maybe 0 if - doesn't work
  cpp_failures_allowed='-'
  cpp_failures='-'
fi
```

I think then the TOTAL row will look better

andychu (commented):
This looks great, thank you!

Glad you were able to decipher my old awk scripts :-)  This was "research" for the YSH language design

which I still want to do :-)

regtest/aports and test/spec-cpp both use what I call the "parallel process table pattern" -- with xargs

But I want YSH to have something nicer than xargs !  i.e. the patterns in these scripts are basically "YSH in crappy shell syntax" :-)

  I'll mention this on Zulip

test/spec-cpp.sh
```diff
@@ -222,12 +235,18 @@ END {
     row_css_class = "cpp-good"  # green
   }
 
-  row = sprintf("%s %s %s %d %d %d",
+  if (status != 0) {
```
andychu, line 240: Oh yeah status is only used here, to set the color

Not in the table contents itself

So I guess it can default to 0

test/spec-cpp.sh
```diff
@@ -187,15 +193,24 @@ summary-tsv-row() {
...
   shift
 
+  local status cpp_failures_allowed cpp_failures
+
   if test $# -eq 1; then
     local spec_name=$1
     local -a tsv_files=( _tmp/spec/$spec_subdir/$spec_name.result.tsv )
+    status=`awk '{printf $1}' "_tmp/spec/$spec_subdir/$spec_name.task.txt"`
+    cpp_failures_allowed=`awk '{printf $6}' "_tmp/spec/$spec_subdir/$spec_name.stats.txt"`
+    cpp_failures=`awk '{printf $7}' "_tmp/spec/$spec_subdir/$spec_name.stats.txt"`
```
andychu: Please use $() for command sub everywhere rather than backticks

(this could be a good rule for a linter)

andychu:
Also it looks like the Docker registry is flaky now, so we can just restart the job

https://github.com/oils-for-unix/oils/actions/runs/20310083784/job/58338811562

test/spec-cpp.sh
last-genius: Done!

PR #2619, opened 2025-12-20
Fix GOROOT issue due to /proc mount order

### LINKED ISSUE #2489: Analyze / figure out the baseline failures in regest/aports / Fix GOROOT (?)
[Almost 1/3 of community packages fail at baseline?](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/Almost.201.2F3.20of.20community.20packages.20fail.20at.20baseline.3F/with/544209167)

This might be a good issue for someone who has experience with  Alpine Linux packaging

---

Our first priority is to fix bugs, but we also want to have accurate meausrements

EarthenSky:
Close #2489

when entering the overlayfs for building a package /proc, /sys, and /dev mounts were being bypassed, so /proc appeared empty. Go depends on `/proc/self/exe` to find its path, which it was failing to do (hence GOROOT as an issue).

I expect this to fix a large number of pacakge builds, notably many packages written in Rust.

There is still an outgoing issue regarding causing `distrobuilder`'s baseline to fail, but it appears more nuanced.

[more discussion on zulip](https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/reproducing.20Go.20package.20failures.20in.20aports.2Fcommunity/with/564371283)

### DIFF
--- a/regtest/aports-run.sh
+++ b/regtest/aports-run.sh
@@ -290,14 +290,17 @@ build-package-overlayfs() {
     $merged
 
   # mount over the overlay, onto the merged
+  # https://man7.org/linux/man-pages/man5/procfs.5.html
   sudo mount \
     -t proc \
     "_chroot_proc_${xargs_slot}" $merged/proc
 
+  # https://man7.org/linux/man-pages/man5/sysfs.5.html
   sudo mount \
     -t sysfs \
     "_chroot_sysfs_${xargs_slot}" $merged/sys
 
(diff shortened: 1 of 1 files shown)


EarthenSky:
@andychu any ideas about this error https://github.com/oils-for-unix/oils/actions/runs/20386457019/job/58588234851?pr=2619 ? Seems to be a storage issue on the CI or something like that?

PR #2620, opened 2025-12-21
[doc] fix incorrect example in ysh-vs-python

pltrz:
Output of '22' < '3' is changed from true to false.


andychu:
Merged, thank you!

PR #2629, opened 2025-12-27
regtest/debian experiment

btertoolen:
WIP

Right now it is possible to set up a package overlay based on baseline or osh-as-sh layer, fetch the source, install the build deps, and build the binary package

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,8 +33,7 @@ tags
 # Our own temp dirs
 _bin/
 _build/
-_debian_chroot/
-_debian_build/
+_regtest
 _chroot/
 _devbuild/
 _gen
--- a/regtest/sh_wrapper.sh
+++ b/regtest/sh_wrapper.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 OSH_BIN="/usr/local/bin/oils-for-unix"
 if [ "$1" = "-c" ] && [ "$2" = "--" ]; then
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -33,6 +33,8 @@ tags
 # Our own temp dirs
 _bin/
 _build/
+_debian_chroot/
+_debian_build/
```
andychu: Right now the chroot is in `_chroot/aports-build`

We could have `_chroot/debian-build` too ?

---

There are some things in `_chroot` that are not chroots, since we moved to overlay FS

We could also perhaps call it `_regtest/aports-build` and `_regtest/debian-build`, but probably not worth changing now

regtest/debian-build.ysh
```diff
@@ -0,0 +1,68 @@
+#!bin/ysh
+
```
andychu, line 12: I generally  like to put at least a couple examples at the top in the comments

To remind myself (and others) what the common usage is  (e.g.  I just updated `deps/image.sh` like this)

regtest/sh_wrapper.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+OSH_BIN="/usr/local/bin/oils-for-unix"
```
andychu, line 3: This wrapper is OK to make progress, but please file a bug for the underlying OSH issue (and a failing spec test is also appreciated)

.gitignore
btertoolen: I'll make the change for regtest/debian now, then we can see later if we want to move aports over as well

PR #2632, opened 2026-01-08
[frontend] Cleanup in lexer_gen.py

last-genius:


last-genius:
not so sure about the second commit - might have been useful in debugging while strictly speaking it was commented out

andychu:
Thanks, it's good to clean this up

Except I think we should keep PrintTree() because it helps remind us what the internal structure of Python is ... it is a debugging function

But yes the rest seems fine to delete.

last-genius:
> Except I think we should keep PrintTree() because it helps remind us what the internal structure of Python is ... it is a debugging function

Dropped this commit.

PR #2633, opened 2026-01-09
[wedge] Update re2c to 4.3.1

last-genius:
Version 4.0 introduced several improvements which make the Rust translation so much easier (since it's awkward to do pointer arithmetic, YYINPUT allows to use YYCURSOR as an index instead of a pointer).

Version 4.3 also improved compile-time performance.

Not upgrading to 4.4 (the latest version at the moment), since that is breaking backwards compatibility in a rare case (4.3 introduces a warning to track if we use the deprecated feature).

### DIFF
--- a/build/deps.sh
+++ b/build/deps.sh
@@ -44,7 +44,7 @@ mkdir -p $WEDGE_2025_DIR
 
 readonly DEPS_SOURCE_DIR=_build/deps-source
 
-readonly RE2C_VERSION=3.0
+readonly RE2C_VERSION=4.3.1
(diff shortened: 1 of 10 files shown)


andychu:
OK thanks for testing this out

To land this we will have to rebuild the CI containers, which are now hosted here: https://github.com/orgs/oils-for-unix/packages

I would like others to be able to do that, but let me see if I can run the  script tomorrow .... I thinK I automated everything a couple months ago

andychu:
thread: >
[Rebuild containers with new version of re2c](https://oilshell.zulipchat.com/#narrow/channel/308821-containers/topic/Rebuild.20containers.20with.20new.20version.20of.20re2c/with/567288905)

andychu:
Great thanks!  I am going to rebuild everything to test this out

PR #2634, opened 2026-01-10
Add exec -a flag to set argv[0]

### LINKED ISSUE #2558: `exec -a` not implemented
This is also a disagreement between bash and osh.
Bash:
```
[bram@bram-dell oils]$ exec -a FOOPROC /bin/bash -c 'echo $0'
FOOPROC
```
Osh:
```
osh-0.36$ exec -a FOOPROC /bin/bash -c 'echo $0'
  exec -a test123 /bin/bash -c 'echo $0'
       ^~
[ interactive ]:12: 'exec' doesn't accept flag -a
```

Encountered this while looking into `tclx` build failure (https://oilshell.zulipchat.com/#narrow/channel/522730-distros/topic/tclx.20build.20failure/with/554195525), but this was not the cause of that failure. So far I haven't encountered this as a cause for a build failure yet.

daveads:
The exec builtin now supports the -a flag to set argv[0], matching bash behavior.

close #2558

### DIFF
--- a/builtin/process_osh.py
+++ b/builtin/process_osh.py
@@ -246,11 +246,9 @@ class Exec(vm._Builtin):
             e_die_status(127, 'exec: %r not found' % cmd, cmd_val.arg_locs[1])
 
         # shift off 'exec', and remove typed args because they don't apply
+        c2_argv = cmd_val.argv[i:]
         if arg.a is not None:
-            c2_argv = [arg.a]
-            c2_argv.extend(cmd_val.argv[i+1:])
-        else:
-            c2_argv = cmd_val.argv[i:]
+            c2_argv[0] = arg.a
 
         c2 = cmd_value.Argv(c2_argv, cmd_val.arg_locs[i:],
                             cmd_val.is_last_cmd, cmd_val.self_obj, None)
--- a/spec/builtin-process.test.sh
+++ b/spec/builtin-process.test.sh
@@ -44,7 +44,7 @@ stdout
 ## BUG mksh stdout-json: ""
 
 #### exec -a sets argv[0]
-exec -a FOOPROC /bin/bash -c 'echo $0'
+exec -a FOOPROC sh -c 'echo $0'
 ## STDOUT:
 FOOPROC
 ## END


daveads:
Screenshot 2026-01-10 at 6 05 23 AM

issue

daveads:
Screenshot 2026-01-10 at 5 58 58 AM

daveads:
@andychu review.

spec/builtin-process.test.sh
```diff
@@ -43,6 +43,14 @@ stdout
 ## BUG mksh status: -11
 ## BUG mksh stdout-json: ""
 
+#### exec -a sets argv[0]
+exec -a FOOPROC /bin/bash -c 'echo $0'
```
andychu: I would make this `bash` or better yet just `sh` because `/bin/bash` may not exist on every system

builtin/process_osh.py
```diff
@@ -245,7 +246,13 @@ def Run(self, cmd_val):
             e_die_status(127, 'exec: %r not found' % cmd, cmd_val.arg_locs[1])
 
         # shift off 'exec', and remove typed args because they don't apply
-        c2 = cmd_value.Argv(cmd_val.argv[i:], cmd_val.arg_locs[i:],
+        if arg.a is not None:
+            c2_argv = [arg.a]
+            c2_argv.extend(cmd_val.argv[i+1:])
+        else:
+            c2_argv = cmd_val.argv[i:]
```
andychu: the arg_locs should match the argv

andychu: I think this would be simpler like:

```
argv = cmd_val.argv[i:]
if arg.a is not None:
  argv[0] = arg.a

```

OK and actually I guess there is no way to change the location now ... but that's OK

```
exec -a foo python -c
             ^^^^^ # location is still python, not foo
```

Maybe write a comment about the location -- that controls this error

```
$ osh -c 'exec foo bar'
  exec foo bar
       ^~~
[ -c flag ]:1: fatal: exec: 'foo' not found

```

andychu (commented):
Thanks for adding this!  

(sorry for the delayed review, usually I aim for 24 hours -- but I'm still busy with family stuff)

PR #2635, opened 2026-01-13
update image urls to ghcr.io registry

btertoolen:


andychu:
Great thanks!  I will test this out

Issue #2637, opened 2026-01-15
`osh -c -- echo hi`: Command '--' not found


btertoolen:
Ran into this while working on regtest/debian.

`apt` or one of its hooks seems to invoke the system shell with `-c --` before the actual commands that should be executed. But osh tries to execute `--` as a command instead of treating it like an argument delimiter.

Here is the full command that is executed: `
[pid 21762] execve("/bin/sh", ["sh", "-c", "--", "dpkg-source --no-check -x hello_"...], 0x7ffe5af89d28 /* 19 vars */) = 0`

Or a minimal reproducable example:
```
$ osh -c -- echo hi
Unexpected error in execvpe('--', ['--'], ...): [Errno 2] No such file or directory
```

PR #2638, opened 2026-01-15
add spec test for osh -c --

btertoolen:
Spec test for #2637


andychu:
Ah OK looks like the failure count has to be updated

`test/spec.sh divergence` should fail locally for you now

e.g.

```
test/spec.sh divergence; echo $?
```

but it should also print a message

btertoolen:
@andychu Yes of course, makes sense. Got it confused in my head thinking the divergence spec tests are supposed to fail

Issue #2640, opened 2026-01-16
Osh hangs when reading a device file and piping through multiple commands to "head"


asokoloski-m:
This is a weird one that I think we've boiled down to a minimal case:

This hangs:
`cat /dev/zero | cat | head -c 5`

This does not:
`cat /dev/zero | head -c 5`

(using /dev/urandom has the same issue)

It completes successfully without hanging in Bash.  Thanks in advance!  Oils rocks!

PossiblyAShrub:
Odd... I've added a failing spec test in https://github.com/oils-for-unix/oils/pull/2642 so we can track this issue. Thanks for the bug report!

Interestingly, the file also has to be a character device. A regular file works just fine on my machine:

```
ysh-0.37$ cat README.md | cat | head -c 5
Oils 
```

andychu:
Hm interesting, I wonder if this is the same bug as 

- #2365

PR #2642, opened 2026-01-19
Add failing spec test for #2640

PossiblyAShrub:
## Changes

- Adds a failing spec test for https://github.com/oils-for-unix/oils/issues/2640

### DIFF
--- a/spec/bugs.test.sh
+++ b/spec/bugs.test.sh
@@ -1,5 +1,5 @@
 ## compare_shells: bash dash mksh zsh ash
-## oils_failures_allowed: 0
+## oils_failures_allowed: 1
 
 #### echo keyword
 echo done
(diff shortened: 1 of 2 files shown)


andychu:
Thanks!  Let's move this to `spec/bugs` -- because it is a slower test that may rely on timing, and it may not be specifically related to pipelines (?)

Also, I'm pretty sure we have seen this bug before!  Oh yeah I think it related to the Kakoune editor or something  ...

PossiblyAShrub:
> Thanks! Let's move this to spec/bugs -- because it is a slower test that may rely on timing, and it may not be specifically related to pipelines (?)

Done. This is ready for another review.

andychu:
Thanks!

Issue #2643, opened 2026-01-19
AssertionError in evalHay after some syntax errors


PossiblyAShrub:
From  Alexey Shamrin on Zulip ([#oils-discuss > Segmentation fault in evalHay after some syntax errors](https://oilshell.zulipchat.com/#narrow/channel/121540-oils-discuss/topic/Segmentation.20fault.20in.20evalHay.20after.20some.20syntax.20errors/with/568655519)),

> Seems like a bug:
> 
> ```
> $ cat crash.hay 
> echo =
> $ cat crash.ysh 
> const h = parseHay('crash.hay')
> const result = evalHay(h)
> $ ysh --version
> Oils 0.37.0             https://oils.pub/
> 
> git commit = a9e1764bc62097e63e385a631a476c04bce534e8
> $ ysh crash.ysh 
>   echo =
>        ^
> crash.hay:1: Unexpected = (Hint: use var/setvar, or quote it)
> Segmentation fault
> ```

When run in a python build we see the AssertionError:

```
ysh-0.37$ = parseHay('crash.hay')
  echo =
       ^
crash.hay:1: Unexpected = (Hint: use var/setvar, or quote it)
Traceback (most recent call last):
  File "/home/aolse/oils/oil/bin/oils_for_unix.py", line 263, in <module>
    sys.exit(main(sys.argv))
  File "/home/aolse/oils/oil/bin/oils_for_unix.py", line 232, in main
    return AppBundleMain(argv)
  File "/home/aolse/oils/oil/bin/oils_for_unix.py", line 186, in AppBundleMain
    return shell.Main(lang, arg_r, environ, login_shell, loader, readline)
  File "/home/aolse/oils/oil/core/shell.py", line 1214, in Main
    prompt_plugin, waiter, errfmt)
  File "/home/aolse/oils/oil/core/main_loop.py", line 280, in Interactive
    is_return, _ = cmd_ev.ExecuteAndCatch(node, 0)
  File "/home/aolse/oils/oil/osh/cmd_eval.py", line 2366, in ExecuteAndCatch
    status = self._Execute(node)
  File "/home/aolse/oils/oil/osh/cmd_eval.py", line 2164, in _Execute
    status = self._Dispatch(node, cmd_st)
  File "/home/aolse/oils/oil/osh/cmd_eval.py", line 1975, in _Dispatch
    status = self._DoExpr(node)
[...]

PR #2644, opened 2026-01-19
Fix hay crash when parseHay encounters syntax error

PossiblyAShrub:
Details about the error given in https://github.com/oils-for-unix/oils/issues/2643.

Fixes: https://github.com/oils-for-unix/oils/issues/2643

## Changes

- Fix the crash when assigning/reading from `parseHay` after a syntax error
- Adds a spec test for the above


andychu:
Thank you!

BTW I still want to "lower" Hay into "just YSH" ... so in an ideal world `parseHay()` would be subsumed by `parseCommand()`

https://oils.pub/release/latest/doc/ref/chap-builtin-func.html

PR #2650, opened 2026-01-24
fix prompt disappearing in interactive shell

### LINKED ISSUE #2367: Prompt disappears in interactive shell after shopt --set ysh:upgrade
Also freeze using long option:

    osh-0.34$ shopt --set ysh:upgrade

    # nothing happening, also not after Ctrl-C or -Z

daveads:
When unsetting ysh:all, the shell switches from YSH mode (env_obj enabled) to OSH mode (env_obj disabled). 

This changes where PS1 is looked up:

   - YSH mode: PS1 in ENV dict or __defaults__
   - OSH mode: PS1 as regular shell variable

   Added MaybeMigrateEnvToVars() to migrate PS1/PS2/PS3/PS4 from ENV dict
   to regular variables when unsetting ysh:all or ysh:upgrade.

   Fixes #2367

### DIFF
--- a/core/sh_init.py
+++ b/core/sh_init.py
@@ -8,7 +8,6 @@ from core import pyutil
 from core import optview
 from core import state
 from frontend import location
-from mycpp import mylib
 from mycpp.mylib import iteritems, tagswitch, log
 from osh import split
 from pylib import os_path
@@ -21,9 +20,6 @@ if TYPE_CHECKING:
     from _devbuild.gen import arg_types
 
 _ = log
-
-if mylib.PYTHON:
-    _env_config_instance = None  # type: EnvConfig
 
 
 class EnvConfig(object):
@@ -48,17 +44,12 @@ class EnvConfig(object):
 
     def __init__(self, mem, defaults):
         # type: (state.Mem, Dict[str, value_t]) -> None
-        if mylib.PYTHON:
-            global _env_config_instance
-            assert _env_config_instance is None, "Only one EnvConfig instance"
-            _env_config_instance = self
-
         self.mem = mem
         self.exec_opts = mem.exec_opts
         self.defaults = defaults
-        # Lazy init because exec_opts is None when Mem.__init__ creates EnvConfig
-        self._did_capture_env_obj = False  # type: bool
-        self._env_obj_at_startup = False  # type: bool
+        # -1: not yet captured, 0: False (OSH), 1: True (YSH)
+        # Lazy because exec_opts is None when Mem.__init__ creates EnvConfig
+        self._env_obj_at_startup = -1  # type: int
 
     def GetVal(self, var_name):
         # type: (str) -> value_t
@@ -66,16 +57,14 @@ class EnvConfig(object):
         YSH: Look at ENV.PATH, and then __defaults__.PATH
         OSH: Look at $PATH
         """
-        # Capture env_obj on first use (can't do in __init__, exec_opts is None then)
-        if not self._did_capture_env_obj:
-            self._env_obj_at_startup = self.mem.exec_opts.env_obj()
-            self._did_capture_env_obj = True
+        if self._env_obj_at_startup == -1:
+            self._env_obj_at_startup = 1 if self.mem.exec_opts.env_obj() else 0
 
         use_env_obj = self.mem.exec_opts.env_obj()
         # Prompt variables use startup value to fix #2367: prompt shouldn't
         # disappear when 'shopt --unset ysh:all' or 'shopt --set ysh:all'
         if var_name in ('PS1', 'PS2', 'PS3', 'PS4'):
-            use_env_obj = self._env_obj_at_startup
+            use_env_obj = bool(self._env_obj_at_startup)
 
         if use_env_obj:  # e.g. $[ENV.PATH]
 


andychu:
Thanks for doing this!  I see the motivation here

I am wondering if there is a solution where we don't have to maintain migration code ...  although arguably PS1 PS2 PS3 PS4 will **never** change?

Hm hm

---

Or maybe we  can slightly change what `ysh:all` and `ysh:upgrade` do?  Maybe they shouldn't affect the prompt?

Maybe we can continue to use the prompt that the shell **started** with?

e.g. 

- if you start with `osh`, it always uses the OSH prompt, even if you do `shopt --set ysh:all` 
- if you start with `ysh`, it always uses the YSH prompt, even if you do `shopt --unset ysh:all`

???

Not sure if that is possible, hm hm hm

andychu:
OK actually I see the problem is that we get the value from `env_config.Get()`

This is `EvalFirstPrompt()`

```
         # Now try evaluating $PS1
         ps1_val = self.mem.env_config.GetVal('PS1')
         #log('ps1_val %s', ps1_val)
         UP_ps1_val = ps1_val
         if UP_ps1_val.tag() == value_e.Str:
             ps1_val = cast(value.Str, UP_ps1_val)
             return self.EvalPrompt(ps1_val.s)
```


and then `GetVal()` looks at `exec_opts.env_obj()`, which is `shopt --set env_obj`:

```
     def GetVal(self, var_name):
         # type: (str) -> value_t
         """
         YSH: Look at ENV.PATH, and then __defaults__.PATH
         OSH: Look at $PATH
         """
         if self.mem.exec_opts.env_obj():  # e.g. $[ENV.PATH]
     

```

andychu:
So considering that, how about we use the **startup** rule:

in `EnvConfig.__init__`

```
self.env_obj_at_startup = self.exec_opts.env_obj()
```

and then `env_config.GetPromptVal()` can look at `self.env_obj_at_startup` rather than `self.exec_opts.env_obj()` ?

Would that work?

daveads:
> So considering that, how about we use the **startup** rule:
> 
> in `EnvConfig.__init__`
> 
> ```
> self.env_obj_at_startup = self.exec_opts.env_obj()
> ```
> 
> and then `env_config.GetPromptVal()` can look at `self.env_obj_at_startup` rather than `self.exec_opts.env_obj()` ?
> 
> Would that work?

let me check that out.

core/sh_init.py
```diff
@@ -47,14 +47,24 @@ def __init__(self, mem, defaults):
...
 
     def GetVal(self, var_name):
         # type: (str) -> value_t
         """
         YSH: Look at ENV.PATH, and then __defaults__.PATH
         OSH: Look at $PATH
         """
-        if self.mem.exec_opts.env_obj():  # e.g. $[ENV.PATH]
+        if not self._did_capture_env_obj:
+            self._env_obj_at_startup = self.mem.exec_opts.env_obj()
```
andychu: Oh I see ... hm I think there is only one EnvConfig object, so it should be OK to put this in `__init__` ?

i.e. have one variable instead of two?

yeah it is initialized with `state.Mem` ... and there is only one of those, from `core/shell`

---

So we can just make it a global variable, or would you feel more comfortable with maintenance if we have some kind of global assertion or unit test saying there's only one `EnvConfig`?  

I think that is possible, hm

core/sh_init.py
```diff
@@ -47,14 +47,24 @@ def __init__(self, mem, defaults):
...
+            self._did_capture_env_obj = True
+
+        use_env_obj = self.mem.exec_opts.env_obj()
+        if var_name in ('PS1', 'PS2', 'PS3', 'PS4'):
```
andychu, line 66: This should also have a comment on why these variables are special ... i.e. to fix this prompt issue

andychu:
It is possible to add code under:

```
if mylib.PYTHON:
  pass
```

that does not get translated to C++

So if we want to assert there is only one `EnvConfig`, we can use `id(self)` or similar tricks ... hidden by `mylib.PYTHON`

I think that can make the code simpler, while also protecting invariants, which will make the code easier to "jump into".  I can see why it's non-obvious / non-local that there is one `EnvConfig`

spec/ysh-prompt.test.sh
```diff
@@ -158,3 +158,13 @@ hi
 ## END
 ## stderr-json: "<Runtime error: Func 'renderPrompt' takes no positional args, but got 1><Runtime error: Func 'renderPrompt' takes no positional args, but got 1>"
 
+
+#### PS1 prompt persists after unsetting ysh:all
+
+echo 'shopt --unset ysh:all
```
andychu, line 164: Hm are you sure this test fails before the change?

It looks like `shopt --unset ysh:all` is not going to be piped into the `$[ENV.SH]` ???

andychu: Also, how about a test for the opposite case?  Where you start with OSH and `shopt --set ysh:all` ?

That can go either in this file or in `spec/prompt.test.sh`, whichever is more convenient

I guess you can do `ENV.SH.replace('ysh', 'osh')` to get an osh path .... ?  or maybe there's a better way

core/sh_init.py
daveads: The two variable approach is necessary because exec_opts is None when `EnvConfig` is created during `Mem.__init__`. We can't initialize directly in `__init__` as suggested.

daveads: @andychu we could look at restructuring the initialization order but that would be a bigger change.

core/sh_init.py
```diff
@@ -44,17 +48,36 @@ class EnvConfig(object):
...
+        if mylib.PYTHON:
+            global _env_config_instance
+            assert _env_config_instance is None, "Only one EnvConfig instance"
+            _env_config_instance = self
+
         self.mem = mem
         self.exec_opts = mem.exec_opts
         self.defaults = defaults
+        # Lazy init because exec_opts is None when Mem.__init__ creates EnvConfig
+        self._did_capture_env_obj = False  # type: bool
```
andychu: We don't need `_did_capture_env_obj` if there is only one EnvConfig instance?  That was the idea behind the assertion

Issue #2646, opened 2026-01-24
User-Facing Docs Site Search


PossiblyAShrub:
We want to have a search feature like:
- [Python3 docs]()
- [Rust docs]()

Search was implemented in #2616 for the dev-guide using [pagefind](https://pagefind.app/).

PossiblyAShrub:
Unfortunately pagefind doesn't seem to work well for the reference style docs. It:
- Tends to surface side comments referencing symbols (eg. `parseArgs`) instead of the main doc entry
- Doesn't jump the to point on the page where the result was found
- Fails terribly on typos (eg. `parseArhs` returns nonsense)

For example, take a look at [their MDN demo](https://mdn.pagefind.app/). If you search for `"uint8array"`, the top link is [`TypedArray.prototype.fill()`](https://developer.mozilla.org/en-us/docs/web/javascript/reference/global_objects/typedarray/fill/). In fact, the root [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) docs don't even show in the first page of results!

---

The python docs search is very slow and pulls in ~MB of data for each query! Instead, I found that the rust docs are both low bandwidth and high signal. They detail [their search system in the dev guide](https://rustc-dev-guide.rust-lang.org/rustdoc-internals/search.html). I want to try implementing our own. I don't think it will be much work!

PR #2651, opened 2026-01-24
Docs-wide search

PossiblyAShrub:
This is https://github.com/oils-for-unix/oils/issues/2646

### DIFF
--- a/deps/Dockerfile.soil-ovm-tarball
+++ b/deps/Dockerfile.soil-ovm-tarball
@@ -72,8 +72,12 @@ COPY --chown=uke \
 
 COPY build/common.sh /home/uke/tmp/build/common.sh
 COPY deps/from-tar.sh /home/uke/tmp/deps/from-tar.sh
+COPY deps/from-binary.sh /home/uke/tmp/deps/from-binary.sh
 
 # For bootstrapping the OVM build
 RUN deps/from-tar.sh layer-cpython
+
+# For doc tests
+RUN deps/from-binary.sh layer-nodejs
 
 CMD ["sh", "-c", "echo 'hello from oilshell/ovm-tarball'"]
--- a/deps/from-binary.sh
+++ b/deps/from-binary.sh
@@ -60,4 +60,19 @@ test-clang() {
   $CLANGXX --version
 }
 
+NODEJS_URL='https://nodejs.org/dist/v24.13.1/node-v24.13.1-linux-x64.tar.xz'
+NODEJS_SHASUM='30215f90ea3cd04dfbc06e762c021393fa173a1d392974298bbc871a8e461089 node-v24.13.1-linux-x64.tar.xz'
+
+layer-nodejs() {
+  ### We need a newer version of Node.js than is available in Debian
+  ### Used in web/search.test.js
+
+  pushd $DEPS_DIR
+  wget --no-clobber $NODEJS_URL
+  sha256sum --check <(echo "$NODEJS_SHASUM")
+  time tar -x --xz <node-*-linux-x64.tar.xz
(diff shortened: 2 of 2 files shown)


doctools/search_index.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+
+# This tool reads in the headings on a doc/ref page and produces a list of all
+# the symbols (and their anchors) which can be used as a search index.
```
andychu: I think this docstring can be improved

It doesn't mention it's JSON

What format is the JSON in?

I think you pointed to the Rust compiler dev guide -- that link may help here?

andychu: We should have a `search_index_test.py` , so the next person to change it doesn't break the search!

web/search.js
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * Computes the DL distance between textA and textB. This is the number of edits between A and B. The possible edits are:
+ * - Insertion (adding a single char)
+ * - Deletion (removing a single char)
+ * - Substitution (changing a single char)
+ * - Transposition (swapping the locations of 2 chars)
+ *
+ * [0] https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance#Distance_with_adjacent_transpositions
+ */
+function damerauLevenshteinDistance(textA, textB) {
```
andychu, line 31: We should have some kind of unit tests for this JavaScript too

I think a bunch of it can be tested outside the browser 

I don't think we have any instances of that in the repo though

doc/old/errors.md
```diff
@@ -1,6 +1,6 @@
 ---
 in_progress: yes
-css_files: ../../web/base.css ../../web/manual.css ../../web/toc.css
+css_files: ../../web/base.css ../../web/manual.css ../../web/toc.css ../../web/search.css
```
andychu: I don't think it's worth changing anything in `doc/old` ?  I think this directory is purely historical and nobody should read these docs ...

build/doc.sh
```diff
@@ -952,6 +956,8 @@ run-for-release() {
 
   copy-web
 
+  build-search-index
```
andychu: could just inline this function since it calls a YSH script?

doctools/doc_html.py
```diff
@@ -105,12 +105,33 @@ def Header(meta, f, draft_warning=False):
...
+      if (docIndex === -1) {
+        window.basePath = '.'; // If not found, assume we are at the root
+      } else {
+        window.basePath = parts.slice(0, docIndex).join('/');
+      }
+
+      // Load scripts irrespective of our location in the doc-tree
+      const script = document.createElement('script');
+      script.type = 'module';
+      script.src = `${window.basePath}/web/search.js`;
```
andychu, line 132: Hm this seems like a bit of a roundabout way to have a `<script>` tag?  

What's the issue iwth making it more static?  I guess it's because we have `css_files:` metadata but not `js_files:` metadata?

PossiblyAShrub: This is partially a roundabout way to have a `<script>` tag, but also necessary to compute the base path for the search functionality. Without the base path, we don't know where each search result's anchor resolves to.

Computing this base path statically is possible, but will require some deep changes to how we build docs. It is better placed in another PR.

PossiblyAShrub:
I've added a bunch of tests. The `soil-ovm-tarball` container image will have to be rebuilt. [Sounds like you are the only one who can do that?](https://oilshell.zulipchat.com/#narrow/channel/308821-containers/topic/Rebuild.20containers.20with.20new.20version.20of.20re2c/with/567288905)

When that container is rebuilt, CI should be green. So this is ready for a review.

PR #2652, opened 2026-01-24
add xz-utils to Debian list in deps.sh

shamrin:
this fixes an error during `build/deps.sh fetch`:

```
tar (child): xz: Cannot exec: No such file or directory
```


andychu:
Thank you!

PR #2654, opened 2026-01-25
fix ARM64 build type detection for zsh 5.1.1

### LINKED ISSUE #2653: build/deps.sh install-wedges fails on arm64
zsh-5.1.1 step of `build/deps.sh install-wedges` fails on `aarch64`:

```
checking build system type... /home/shamrin/oils/_build/deps-source/zsh/zsh-5.1.1/config.guess: unable to guess system type
```

Same error on Alpine and Debian.

shamrin:
Fixes #2653

### DIFF
--- a/deps/source.medo/zsh/WEDGE
+++ b/deps/source.medo/zsh/WEDGE
@@ -18,6 +18,12 @@ wedge-make() {
   local build_dir=$2
   local install_dir=$3
 
+  # FIX for zsh 5.1.1: 2009-era config.guess fails on ARM64 with "cannot guess build type"
+  local arch_flags=""
+  if [[ "$(uname -m)" == "aarch64" && "$WEDGE_VERSION" == "5.1.1" ]]; then
+    arch_flags="--build=aarch64-unknown-linux-gnu"
+  fi
+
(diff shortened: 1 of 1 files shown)


deps/source.medo/zsh/WEDGE
```diff
@@ -18,6 +18,12 @@ wedge-make() {
   local build_dir=$2
   local install_dir=$3
 
+  # FIX for zsh 5.1.1: 2009-era config.guess fails on ARM64 with "cannot guess build type"
+  local arch_flags=""
+  if [[ "$(uname -m)" == "aarch64" && "$WEDGE_VERSION" == "5.1.1" ]]; then
```
andychu, line 23: Please use `test "$(uname -m)" = ... && test "$WEDGE_VERSION" = ...`

I think these scripts run with `/bin/sh`, which doesn't always have `[[`

shamrin: I've moved the change to `oils-for-unix` branch, with the fix you suggested: see PR #2655

shamrin:
Closed in favor of PR #2655

PR #2655, opened 2026-01-25
fix ARM64 build type detection for zsh 5.1.1

### LINKED ISSUE #2653: build/deps.sh install-wedges fails on arm64

shamrin:
Fixes #2653

### DIFF
--- a/deps/source.medo/zsh/WEDGE
+++ b/deps/source.medo/zsh/WEDGE
@@ -18,6 +18,12 @@ wedge-make() {
   local build_dir=$2
   local install_dir=$3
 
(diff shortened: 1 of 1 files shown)


shamrin:
This includes the fix [suggested](https://github.com/oils-for-unix/oils/pull/2654#discussion_r2725626032) in the previous version of this PR.

andychu:
Thank you!

PR #2656, opened 2026-01-26
Support readline emulation provided by libedit

### LINKED ISSUE #228: tab completion doesn't work on OS X
This is likely a combination of 2 factors

- I copied a bunch of readline #defines from `pyport.h` into `Modules/readline.c`.  Python's autotools ./configure actually detects some features, which may be different on OS X.
- OS X uses libedit.

TODO: Get a copy of Python's pyconfig.h on OS X?  Is it always the same?

### LINKED ISSUE #460: Build OSH with other line editing libraries (libedit, etc.)
Right now it's tied to GNU readline, and I would like to avoid that long term.  We use GNU readline is a slightly different way than bash does, because we want to be principled about interactive completion.

This is simpler than #185, and maybe a step on the way.

### LINKED ISSUE #551: Oil doesn't work interactively on OS X because of the native/line_input binding
I think OS X has (a non-GPL) libedit which tries to emulate the readline interface?  I'm not an expert in this area.

This is #460 but I didn't realize how broken Oil was until this thread:

https://oilshell.zulipchat.com/#narrow/stream/119655-new-members/topic/Testing.20osh

I don't have enough motivation to do this myself at the moment, so I'd need some help from an OS X user.

### LINKED ISSUE #1969: License compatability: Oils vs readline
While commenting on another issue, I noted that GNU readline is used in Oils, though I am not sure if this is 100% always true on all platforms as I only came across Oils a couple days ago.

This is an issue due to the license of GNU readline, though being a library, unusually it is under the full GPL not the LGPL.  GPL being a deliberate decision at it's inception many years ago.

As I am sure people are aware static linking a GPL binary or a binary object derived from GPL source code automatically makes the resulting object GPL, however Oils is under the more permissive Apache2 licence.  In fact some distributions omit Readline completely because of this, and use, e.g. libedit.

As suggested in the other issue, there are alternative libraries that can be used to circumvent this, e.g. [https://github.com/reeflective/readline](https://github.com/reeflective/readline)

shamrin:
Fixes #228, fixes #460, fixes #551, fixes #1969

## Why

* allow `./configure && _build/oils.sh` to work out-of-the-box on MacOS, with auto-completion and line editing
* prepare for distributing Oils binaries without stepping into GPL

## macOS

On my macOS readline library and headers all symlink to libedit. That's why there is no need to say `-l edit` or specify path to headers. Usual [installation][] instructions just work.

[installation]: https://oils.pub/release/latest/doc/INSTALL.html

```
$ cd $(xcrun --show-sdk-path)
$ ls -l usr/lib/libreadline.tbd 
lrwxr-xr-x  1 root  wheel  13 Aug 27  2024 usr/lib/libreadline.tbd -> libedit.3.tbd
$ ls -l usr/include/readline
total 0
lrwxr-xr-x  1 root  wheel  22 Aug 27  2024 history.h -> ../editline/readline.h
lrwxr-xr-x@ 1 root  wheel  22 Aug 27  2024 readline.h -> ../editline/readline.h
```

I don't have Homebrew installed. Only Xcode Command Line Tools.

## How

I've ported most of the hacks from Python [readline.c](https://github.com/python/cpython/blob/main/Modules/readline.c). Except for the ["mad world"](https://github.com/python/cpython/blob/639c1ad4f1ef5c2409a62fa8ed16e6aa3a6f9ab8/Modules/readline.c#L1370) workaround that is [no longer needed](https://github.com/python/cpython/issues/62658#issuecomment-1093621555) since 2019-era MacOS 10.9.

## Next steps

After this PR is merged:

* Support direct linking with `libedit`. While we could ask to install [libeditreadline-dev][] emulation (at least on Debian or Ubuntu), a better way is to link with "real" libedit, via `-l edit`. Some distros don't have readline emulation package. Also nice to completely avoid GPL consequences for Oils binaries.
* Document that `libedit` is also supported

(By the way, libedit is installed by default on Alpine, Arch, CentOS, Fedora, Debian, Ubuntu, ... And only Ubuntu has readline *in addition* to libedit.)

[libeditreadline-dev]: https://packages.debian.org/trixie/libeditreadline-dev

### DIFF
--- a/cpp/frontend_pyreadline.cc
+++ b/cpp/frontend_pyreadline.cc
@@ -20,7 +20,6 @@ namespace py_readline {
 
 static Readline* gReadline = nullptr;
 
-// Assuming readline 4.0+
 #if HAVE_READLINE
 
 static char* do_complete(const char* text, int state) {
@@ -68,7 +67,7 @@ static void display_matches_hook(char** matches, int num_matches,
 
 static void free_history_entry_portable(HIST_ENTRY* entry) {
   // See also: https://github.com/python/cpython/issues/53695
-  #if HAVE_READLINE && HAVE_READLINE_FREE_HISTORY_ENTRY
+  #if HAVE_READLINE_FREE_HISTORY_ENTRY
   // GNU Readline 5.0+
   histdata_t data = free_history_entry(entry);
   free(data);


build/common.sh
```diff
@@ -14,11 +14,16 @@ set -o nounset
...
 LLVM_VERSION=14.0.0
+ARCHITECTURE="$(uname -m)"
 
-# New version is slightly slower -- 13 seconds vs. 11.6 seconds on oils-for-unix
-readonly CLANG_DIR_RELATIVE="../oil_DEPS/clang+llvm-$LLVM_VERSION-x86_64-linux-gnu-ubuntu-18.04"
+if test "$ARCHITECTURE" = "aarch64"; then
+  readonly CLANG_DIR_RELATIVE="../oil_DEPS/clang+llvm-$LLVM_VERSION-aarch64-linux-gnu"
+else
+  readonly CLANG_DIR_RELATIVE="../oil_DEPS/clang+llvm-$LLVM_VERSION-x86_64-linux-gnu-ubuntu-18.04"
+fi
```
shamrin, line 26: This change is to help developers on arm64 run the same clang-format.

cpp/frontend_pyreadline.cc
```diff
@@ -274,7 +303,7 @@ int Readline::get_endidx() {
 
 void Readline::clear_history() {
 #if HAVE_READLINE
-  rl_clear_history();
+  ::clear_history();
```
shamrin, line 306: libedit does not have `rl_clear_history`, but `clear_history` exists in both libedit and readline.

deps/from-binary.sh
```diff
@@ -21,14 +21,23 @@ readonly DEPS_DIR=$REPO_ROOT/../oil_DEPS
...
+  fi
 else
   # This version was 4.7 GB
   LLVM_VERSION=14.0.0
-  CLANG_URL='https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.0/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz'
+  if test "$ARCHITECTURE" = "aarch64"; then
+    CLANG_URL='https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.0/clang+llvm-14.0.0-aarch64-linux-gnu.tar.xz'
+  else
+    CLANG_URL='https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.0/clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz'
+  fi
```
shamrin, line 40: Allow developers on ARM64 download same versions of LLVM (mostly for clang-format).

build/detect-readline.c
```diff
@@ -1,19 +1,13 @@
...
-  return 0;
-}
+/* clang-format on */
 
 int main(void) {
   char *line = readline("");
 
-  /* Ensure readline version is recent enough.
-     This line will break the build otherwise: https://git.io/vhZ3B */
-  rl_event_hook = test_event_hook;
```
andychu, line 16: Hm do we still need this?  I wonder why this was added?

andychu: Hm it was added way back in 2018 for the Python build ... maybe there is no point for C++ ?

Does the detection break `libedit` support?

configure
```diff
@@ -284,13 +291,97 @@ detect_readline() {
 
       readline_dir="$FLAG_readline"
       have_readline=1
+
+      if cc_quiet build/detect-readline-callback-sigcleanup.c \
+        -L "$FLAG_readline/lib" \
```
andychu, line 296: Since all the flags here and above are the same, I would make a shell function like

```
if cc_quiet_for_readline build/detect-foo.c; then
  have_foo=1
fi
```

configure
```diff
@@ -284,13 +291,97 @@ detect_readline() {
...
+        have_readline_resize_terminal=1
+      fi
     fi
     return
   fi
 
   # Detect in default location
   if cc_quiet build/detect-readline.c -l readline; then
     have_readline=1
+    if cc_quiet build/detect-readline-callback-sigcleanup.c -l readline; then
```
andychu, line 357: Hm now that we have multiple `detect-*` to run, I think it's better to consolidate these two branches ...

i.e. have a shell function that does all the checks, and takes a param for the user-specified location versus the default location

